PR #3, opened 2019-09-03
Image trigger fix

Gundersanne:


.tasks
```diff
@@ -13,4 +13,4 @@
 
 set -ex
 
-bots/image-trigger
+# bots/image-trigger
```
martinpitt: It seems weird to mix that change into this commit. I'd suggest to drop the other commit as well and leave .tasks not executable until we switch over. Alternatively, if the non-executable file causes trouble, please move this change into the other commit.

Gundersanne: Oh sorry, the executable commit should have been merged yesterday. Not sure why that pops up here.

PR #17, opened 2019-09-23
make-checkout: Supply bots/ from the bots project

Gundersanne:


make-checkout
```diff
@@ -71,15 +71,14 @@ def main():
...
-
-    # HACK
-    # We can drop this check once bots is no longer present in cockpit
-    if api.repo != "cockpit-project/bots" and (api.repo != "cockpit-project/cockpit" or (opts.base and opts.base != "master")):
-        execute("git", "remote", "add", "bots_origin", "https://github.com/cockpit-project/cockpit", cwd=TARGET_DIR)
-        execute("git", "fetch", "--depth=1", "bots_origin", opts.bots_ref or "master", cwd=TARGET_DIR)
-        bots_ref = execute("git", "rev-parse", "FETCH_HEAD", cwd=TARGET_DIR).strip()
-        execute("git", "checkout", "--force", bots_ref or "bots_origin/master", "--", "bots/", cwd=TARGET_DIR, error_on_fail=False)
-        execute("git", "reset", "bots", cwd=TARGET_DIR)
+    if api.repo != "cockpit-project/bots":
```
martinpitt: This should work fine for now, but I think this should rather be `if opts.bots_ref`. We do make a distinction in tests-scan between `bots_ref = None` and setting it to some HEAD, right? tests-scan should translate the former into not supplying `--bots-ref` at all, thus this should be None here.

tests-scan
```diff
@@ -390,8 +382,18 @@ def cockpit_tasks(api, update, branch_contexts, repo, pull_data, pull_number, sh
                 checkout_ref = ref
                 if project != repo:
                     checkout_ref = "master"
+                    bots_ref = ""
```
martinpitt: This looks like a leftover. We shouldn't set it to an empty string, but also it always gets set below.

PR #21, opened 2019-09-24
test: Add run script

Gundersanne:
Moved over the relevant bits of cockpit's test-static-code. And call
test-bots.

Fixed any errors.

### DIFF
--- a/test/run
+++ b/test/run
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+set -eu
+
+# run static code checks like pyflakes and pep8
+PYFILES=$(grep -lI '^#!.*python' `find -path ./make-checkout-workdir -prune -o -type f ! -name '.*'` 2>/dev/null || true)
+if python3 -m pyflakes $PYFILES >&2; then
+    echo "ok 1 pyflakes"
+else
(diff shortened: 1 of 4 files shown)


test/run
```diff
@@ -0,0 +1,31 @@
...
+
+# pycodestyle/pep8 python syntax check
+if python3 -m pycodestyle /dev/null >/dev/null 2>&1; then
+    PYTHON_STYLE_CHECKER="pycodestyle"
+elif python3 -m pep8 /dev/null >/dev/null 2>&1; then
+    PYTHON_STYLE_CHECKER="pep8"
+fi
+
+# FIXME: Fix code for the warnings and re-enable them
+out=$(python3 -m "$PYTHON_STYLE_CHECKER" --ignore E501,E265,E261,W504,W605 test/* --exclude=test/verify/nested-kvm,test/README.md,test/run) || true
```
martinpitt: let's drop the entire if/elif/fi from above, and just hardcode `$(python3 -m pycodestyle`) here, for a better error behaviour. I think pep8 was still from the time when this had to run in Ubuntu 14.04.

PR #28, opened 2019-09-25
vm-run: improve --quiet to exit on machine quit

allisonkarlitskaya:
vm-run would keep running even after the machine quit (from running
'shutdown' over ssh, for example).  Make it exit (and cleanup) when the
machine exits.


martinpitt:
Triggering the host test to get the basic python checks

martinpitt:
Fails pyflakes test:

    ./vm-run:23: 'signal' imported but unused

PR #31, opened 2019-09-26
task: Ensure that test-github talks to our mock GitHub API

martinpitt:
Otherwise it talks to the real GitHub API if it happens to run before
test-task.


martinpitt:
Example: https://logs-https-cockpit.apps.ci.centos.org/logs/pull-30-20190926-094545-c4321a12-cockpit-project-bots--host/log

PR #32, opened 2019-09-26
images: Shrink windows-10 image and install QRes

martinpitt:
Use sdelete to zero unused blocks, so that the image does not balloon as
much after installing updates.

Install QRes so that developers or our tests can switch to a higher
screen resolution programmatically. Set a default 1280x1024 resolution
which is a better start than the tiny 800x600 of the previous image.


martinpitt:
This seems to break the [cockpit-composer test](https://209.132.184.41:8493/logs/pull-32-20190926-134646-e2270e5e-weldr-cockpit-composer--fedora-30-edge/log), at some point the browser or selenium seems to crash.

martinpitt:
@henrywang : Do you have some time to check why the c-composer tests reliably fail with this new image? It's exactly the same windows and edge versions, the main difference is the different resolution (which should avoid a lot of these "scrolled out of sight" bugs). However, the tests fail with "Couldn't connect to selenium server", which seems unrelated to "not found" failures.

henrywang:
@martinpitt, correct! It's not caused by this PR but flaky issue.

martinpitt:
@henrywang : There's some change, though -- I retried this test about 5 times and it consistently fails, while on the current image it works much more often.

henrywang:
@martinpitt, `Error: Couldn't connect to selenium server` always stucks at the 3rd case of `imageBuilding.test.js`, which is building a tar image. The same issue can be found on [some of PR edge test](https://209.132.184.41:8493/logs/pull-793-20191001-134848-a085ca7b-weldr-cockpit-composer--fedora-30-edge/log).
But it passed locally with the latest Windows-10 image. So I suspect it's related with resource. The building image uses a lot of resource and it caused selenium hub in Win10 cannot response the request. That caused `Error: Couldn't connect to selenium server`. That is also able to explain why some of edge test got passed sometime.
So I think we can skip building image case on Edge browser.

martinpitt:
@henrywang Thanks for your investigations! I'm a bit confused, though -- it sounds as the resource contention would be in the fedora-30 VM, as that's the place that runs composer and the image build. But how is that specific to which browser (from an entirely different VM) talks to it? Or does the test try to download a big tar file, and fails because it runs out of memory? We can certainly try and bump the VM size even more (either for f-30 or windows-10). I'll make a test commit to test this theory.

martinpitt:
Argh, cockpit-composer's Makefile resets bots to master -- this is bogus and must be fixed.

martinpitt:
... and at the same time this means that the test here never actually *used* the refreshed image, so all my musing about "why this fails" already applies to the current image.

martinpitt:
I sent https://github.com/weldr/cockpit-composer/pull/794 to fix the bots clobbering, and I'll send similar PRs to starter-kit and friends.

martinpitt:
ooooh, [composer/edge passed](https://209.132.184.41:8493/logs/pull-32-20191002-051414-6b16ebbb-weldr-cockpit-composer--fedora-30-edge/log)!!! So let's land https://github.com/weldr/cockpit-composer/pull/795 and then we can drop the hack again.

martinpitt:
So composer works now, and cockpit Edge test is in a similar state as with the current image. I think this is ready to land now. \o/

PR #33, opened 2019-09-26
PEP-8 fixes

martinpitt:
Mostly done with autopep8.
    
Keep a few remaining ignores which are hard/impractical to fix.

### DIFF
--- a/tests-scan
+++ b/tests-scan
@@ -19,7 +19,7 @@
 
 # Random extra options for tests-invoke
 REPO_EXTRA_INVOKE_OPTIONS = {
-    'mvollmer/subscription-manager': [ "--html-logs" ]
+    'mvollmer/subscription-manager': ["--html-logs"]
 }
 
 import argparse
@@ -42,6 +42,7 @@ try:
 except ImportError:
     no_amqp = True
 
+
 def main():
     parser = argparse.ArgumentParser(description='Bot: scan and update status of pull requests on GitHub')
     parser.add_argument('-v', '--human-readable', action="store_true", default=False,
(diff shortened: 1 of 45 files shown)


martinpitt:
The remaining test failures are unrelated. The cockpit selenium tests are currently rather impossible to get green..

tests-score
```diff
@@ -36,13 +36,17 @@ BOTS = os.path.dirname(os.path.realpath(__file__))
 # https://github.com/cockpit-project/cockpituous/blob/master/learn/README.md
 
 # Here we're looking for a field in a record that only has one value
+
```
marusak: here it would make more sense to put those two newlines above the function description.

tests-score
```diff
@@ -36,13 +36,17 @@ BOTS = os.path.dirname(os.path.realpath(__file__))
...
 def value(record, field):
     values = record.get(field, [])
     if len(values) == 1:
         return values[0][0] or ""
     return None
 
 # Here we're looking for the count of a specific field/value in the record
+
+
```
marusak: detto

tests-scan
```diff
@@ -123,13 +126,16 @@ def tests_human(priority, name, number, revision, ref, context, base, original_b
...
     )
 
+
 def is_internal_context(context):
     for pattern in ["rhel", "edge", "vmware", "openstack"]:
         if pattern in context:
             return True
     return False
 
 # Prepare a test invocation command
```
marusak: I would keep function description right next to the function itself and put those two newlines above the description.

tests-scan
```diff
@@ -106,6 +107,8 @@ def main():
     return 0
 
 # Prepare a human readable output
+
```
marusak: detto

tests-score
martinpitt: Well spotted, thanks! Indeed autopep doesn't handle these well. I went through the entire diff and fixed all of these.

martinpitt:
@marusak: The interdiff is fairly trivial, so I triggered just two tests. Is that ok for you?

martinpitt:
Argh, [host test failure](https://209.132.184.41:8493/logs/pull-33-20190927-085757-890b2d5b-cockpit-project-bots--host/log) -- that was supposed to be fixed by PR #31, but is apparently still the case.

PR #35, opened 2019-09-27
WIP: 1-cockpit-8: [no-test] clean up BOTS vs. BASE directory

martinpitt:
* [ ] image-refresh centos-7
 * [x] naughty-prune

PR #37, opened 2019-09-27
Centralize definition of bots and base directory

martinpitt:
Use the constants from machine_core/constants.py everywhere.

 * [ ] naughty-prune
 * [ ] image-refresh rhel-7-7

----

Split out from PR #35 to test this separately.


martinpitt:
Nevermind, I fixed #35.

PR #39, opened 2019-09-30
Image refresh for rhel-atomic

cockpituous:
Image refresh for rhel-atomic


marusak:
Ech, I don't have permissions to land it. Could I get one? @Gundersanne :)

marusak:
Oh, nevermind the previous comment. I (probably) have permissions, but it needs `host` test to finish.
This is a bit annoying - image-refresh never triggers `host` but it is required in every PR.
So I think we should either:
1. Trigger `host` on image-refreshes (I know it it not needed to test, but...)
2. Drop the `no-test` label, since now by default it would only trigger host, which we need
3. Do not require `host` on image-refreshes (which may be tricky do to)

I prefer no. 2

martinpitt:
> Drop the no-test label, since now by default it would only trigger host, which we need

That's not true -- a normal PR to bots would trigger the entire glory of ~ 35 tests from all projects. Given how fast "host" is, I'd prefer just running it for image refreshes. Until then, I stop marking it as a required test.

PR #42, opened 2019-09-30
Image refresh for continuous-atomic

cockpituous:
Image refresh for continuous-atomic
 * [x] resurrect centos-7: PR #55 
 * [x] fix missing systemd-devel on rhel-7.7 image: PR #58


martinpitt:
Hm, recently built centos-7 image disappeared from public mirror.

martinpitt:
Ah, this needs a rebase now -- of course there's no bots-ref, as that already is a PR with an image refresh.

martinpitt:
Bah, same missing systemd-devel as on the rhel-7.7 refresh.

martinpitt:
Fixed in #58

martinpitt:
Restarting, this should work now.

martinpitt:
Rebased to pick up the fixed centos-7 image.

PR #41, opened 2019-09-30
Image refresh for rhel-7-7

cockpituous:
Image refresh for rhel-7-7
 * [x] image-refresh rhel-7-7


martinpitt:
Apparently [systemd-devel is now missing in the mock](https://209.132.184.41:8493/logs/pull-41-20190930-054646-a04b491e-cockpit-project-cockpit-rhel-7.7-rhel-7-7/log.html)

martinpitt:
Reason: our image build scripts take master's cockpit.spec, which has

    BuildRequires: systemd-devel >= 235

And yum being stupid it silently ignores missing dependencies:
```
DEBUG: No package gnutls-devel >= 3.4.3 available.
DEBUG: Package sed-4.2.2-5.el7.x86_64 already installed and latest version
DEBUG: No package systemd-devel >= 235 available.
DEBUG: Package gdb-7.6.1-115.el7.x86_64 already installed and latest version
```

martinpitt:
Fixed in #58

PR #56, opened 2019-09-30
image-refresh: Always trigger `host` as well

marusak:


martinpitt:
I re-enabled host as a required test.

PR #55, opened 2019-09-30
Image refresh for centos-7

cockpituous:
Image refresh for centos-7
 * [x] image-refresh centos-7


martinpitt:
Fished out/reuploaded the image from cockpit-7 and re-triggered tests. It also feels like we should give this some more exercise, on e. g. our rhel-7.7 branch.

marusak:
Why does it try to pull the old image?
`image-download: image-download: couldn't find file anywhere: centos-7-e7497bde5980429e29ed739d1d27f8db7d9a8cb4c5bc2a209ef563a8b4bca66f.qcow2`
Does it pick wrong branch? (which would be sad since we landed bunch of refreshes today)

martinpitt:
@marusak: Wondering about that myself.. centos-7/starter-kit test ought to use the bots-ref from this PR.

martinpitt:
I'd go out on a limb and land this now, as this is a bit of an emergency.

PR #86, opened 2019-10-03
testmap: add manual rhel-7-7/tar target for composer

jikortus:
The need for this arose from weldr/lorax [PR #823](https://github.com/weldr/lorax/pull/823).

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -102,6 +102,10 @@ REPO_BRANCH_CONTEXT = {
             'rhel-7-7/openstack',
             'rhel-7-7/vmware',
         ],
+        # These can be triggered manually with bots/tests-trigger
+        '_manual': [
+            'rhel-7-7/tar',
+        ],
     },
     'weldr/cockpit-composer': {
         'master': [


task/testmap.py
```diff
@@ -102,6 +102,10 @@
             'rhel-7-7/openstack',
             'rhel-7-7/vmware',
         ],
+        # These can be triggered manually with bots/tests-trigger
```
marusak: It is now only `tests-trigger` not `bots/tests-trigger` :) but that is the smallest nitpick and also it is used in the file on other places as well, so not something that need repush :)

jikortus: Oops, sorry for that, I missed this as I just copied the already existing lines together with the comment. As for your question -- yes, the goal is to enable manual triggering of `rhel-7-7/tar` in `weldr/lorax` repository`. Thank you :)

atodorov:
> Looks correct, just to make sure what is goal of this PR - you want to be able to manually (with test-trigger) trigger rhel-7-7/tar in weldr/lorax repository, but you don't want to run it automatically on PRs, correct?

Yes, this is correct. We need this to be triggered manually so we can verify that https://github.com/weldr/lorax/pull/823 works correctly in CI before triggering this scenario automatically. 

Can you please merge this PR.

PR #88, opened 2019-10-04
testmap: run rhel-7-7/tar automatically for weldr/lorax

jikortus:
Now that the `rhel-7-7/tar` test passed in CI (weldr/lorax#823), it should be executed automatically. One thing I'm not sure about is whether I can leave the `_manual` list empty just for reference, or should I rather remove it completely?


marusak:
> One thing I'm not sure about is whether I can leave the _manual list empty just for reference, or should I rather remove it completely?

It is fine to keep.

PR #102, opened 2019-10-07
Drop bots symlink

martinpitt:
Not necessary any more since commit 7233e2236.


martinpitt:
I lied -- tests-scan and issue-scan still create bots/tests-invoke commands, which works for everything except the "host" test. We first need to move that to ../tests-invoke (accepting that this then cannot be tested from PRs any more).

martinpitt:
#103 landed, so this is unblocked. Rebased, and "host" test succeeded now.

PR #104, opened 2019-10-07
Test centos on rhel-7.8 branch

marusak:
That means:
1. When centos-7 image refreshes, test rhel-7.8 branch
2. When rhel-7.8 branch refreshes, test centos-7 image


marusak:
> That context isn't known right now, so we need to hack that into the webhook pod to actually run the test.

Hmm, I was almost sure that only part before @ needs to be known. Well, I'll trigger it manually from centos.

martinpitt:
This [failure](https://209.132.184.41:8493/logs/pull-104-20191007-190505-111472a7-cockpit-project-cockpit-rhel-7.8-centos-7/log.html) looks like it's really haunting us -- the current centos-7 image is supposed to have a fixed systemd-devel.

martinpitt:
Ah, @marusak, can you please rebase this to current master, to include the latest centos-7 image?

marusak:
> can you please rebase this to current master, to include the latest centos-7 image?

sure, will do, but shouldn't it rebase automatically?

martinpitt:
> sure, will do, but shouldn't it rebase automatically?

The project's own tests ("host") will, but not a foreign test -- for these, we check out the bots PR as-is, without rebasing.

martinpitt:
Much better, but still some test failures. We need to fix them in the rhel-7.8 branch, or add naughties in bots.

martinpitt:
We just did a clean run in https://github.com/cockpit-project/cockpit/pull/12961, so let's land this.

PR #113, opened 2019-10-09
issue-scan: Drop --bots-ref

martinpitt:
Commit 5588e4d264 dropped make-checkout's --bots-ref option. Stop
generating it in issue-scan for non-bots projects.

As in this scenario the external projects don't clone bots/ for
themselves (as nothing calls their Makefiles), do that as part of the
generated command.

### DIFF
--- a/issue-scan
+++ b/issue-scan
@@ -167,20 +167,19 @@ def output_task(command, issue, repo, verbose):
     context = context.strip()
     checkout = "PRIORITY={priority:04d} "
 
-    # when working on bots, don't check out bots into itself, and run from project root
     if repo == "cockpit-project/bots":
-        bots_ref = ""
+        # when working on bots run from project root
         cmd = "./{name} --verbose --issue='{issue}' {context}"
     else:
-        bots_ref = "--bots-ref master"
-        cmd = "bots/{name} --verbose --issue='{issue}' {context}"
+        # for external projects, nothing checks out bots/ subdir for them, so do it here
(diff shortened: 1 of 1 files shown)


martinpitt:
This "works" in the sense that `GITHUB_BASE=cockpit-project/cockpit ./issue-scan` now drops the non-existing `--bots-ref` option, but it still doesn't actually work:
```
PRIORITY=0009 ./make-checkout --verbose --repo cockpit-project/cockpit master && cd make-checkout-workdir && bots/tests-data --verbose --issue='12935' tests-train-1.jsonl.gz ; cd ..
```

Because for issue-scan, there is now nothing that actually creates bots/ in the checkout, so that calling bots/tests-data simply doesn't exist. I didn't have this in mind when I proposed moving to the `$COCKPIT_BOTS_REF` approach.

@Gundersanne, @marusak: Do you have a gut feeling how to fix this? We could let issue-scan just emit another shell snipped that generates bots -- that could just be a symlink to the current directory (which is a master bots checkout).

martinpitt:
This works now:

```
❱❱❱ GITHUB_BASE=weldr/cockpit-composer ./issue-scan 
PRIORITY=0009 ./make-checkout --verbose --repo weldr/cockpit-composer master && cd make-checkout-workdir && git clone .. bots && bots/npm-update --verbose --issue='803'  ; cd ..
```
and running the command succeeds. (see https://github.com/weldr/cockpit-composer/pull/801, which happened when I let it run too far)

PR #111, opened 2019-10-09
Image refresh for rhel-7-7

cockpituous:
Image refresh for rhel-7-7
 * [x] image-refresh rhel-7-7


martinpitt:
Fixed in https://github.com/cockpit-project/cockpit/pull/12942

martinpitt:
Two failures, but good enough. We mostly need this for composer now.

PR #95, opened 2019-10-09
Image refresh for fedora-30

cockpituous:
Image refresh for fedora-30
 * [x] image-refresh fedora-30


marusak:
selenium is failing (which is almost rule not exception now) and welder to which I don't understand. @henrywang can you please take a look at it, thanks!

henrywang:
@marusak, The `fedora-30/edge@weldr/cockpit-composer` log looks a little bit strange.
1. the log should be a text format file, not a html file, like [fedora-30/firefox@weldr/cockpit-composer](https://209.132.184.41:8493/logs/pull-95-20191009-134646-6fcde155-weldr-cockpit-composer--fedora-30-firefox/log)
2. the edge log does not provide any useful info. I just wondering edge has the same log as [fedora-30/chrome@weldr/cockpit-composer](https://209.132.184.41:8493/logs/pull-95-20191009-134646-6fcde155-weldr-cockpit-composer--fedora-30-chrome/log), but the chrome one passed.
Could you please re-run the test? Thanks.

marusak:
> the log should be a text format file, not a html file, like fedora-30/firefox@weldr/cockpit-composer

It is a new feature. Now every log is html log. If you are used to the raw log, just delete `.html` from url :) 

> the edge log does not provide any useful info. I just wondering edge has the same log as fedora-30/chrome@weldr/cockpit-composer, but the chrome one passed.

Ah I see. It was retried since I commented so the logs are different. The two logs you mentioned - they show collision - which is situation when two bots start working on those tests, but then one notices that the other one is already working on it and aborts it. I am not sure why it is successful and collision- I suppose that log comes from one bot and the success state from the other one. I'll retry.

martinpitt:
@marusak, @henrywang : For that collision, go to the parent dir of that log, and look at backup.kD2SKt/ -- this is the other test run that "won" the collision race, and which looks successful.

henrywang:
@marusak, the new log looks reasonable, but still failed. I re-tried it, lets see.

henrywang:
@marusak, the edge test passed, but the log still looks strange, doesn't make any sense.
One more issue I found, the log URL might have issue, e.g.
`https://209.132.184.41:8493/logs/pull-95-20191012-181616-6fcde155-weldr-cockpit-composer--fedora-30-edge`, there's two `-` face to face in this URL. I think there should be a word between two `-`

martinpitt:
All tests succeeded, except for the expected cockpit/edge.

PR #117, opened 2019-10-10
Cleanup in tests-invoke no.2

marusak:
I am looking into how we can move the rebasing out of invoke and how we can make it simpler (and then split into logging and really tests invoking (which may not need to be even script only executing command on cmdline)).

- For commit `Do not parse HEAD `: I've checked `tests-scan` and it seem to be always set (from what github sends)

- For commit `Do not set 'base'`: I haven't found where this is used. @Gundersanne @martinpitt can you find such case?

### DIFF
--- a/tests-invoke
+++ b/tests-invoke
@@ -177,13 +177,6 @@ class PullTask(object):
         status["link"] = "log.html"
         status["extras"] = ["https://raw.githubusercontent.com/cockpit-project/bots/master/task/log.html"]
 
-        # Include information about which base we're testing against
-        if self.base:
-            subprocess.check_call(["git", "fetch", "origin", self.base])
-            commit = subprocess.check_output(["git", "rev-parse", "origin/" + self.base],
-                                             universal_newlines=True).strip()
-            status["base"] = commit
-
         # For other scripts to use
         os.environ["TEST_DESCRIPTION"] = description
         self.sink = sink.Sink(host, identifier, status)
--- a/tests-invoke
+++ b/tests-invoke
@@ -55,6 +55,9 @@ def main():
     # TEST_BRANCH would be `rhel-8.0`
     test_branch = os.environ.get("TEST_BRANCH")
 
+    if opts.publish and not revision:
+        parser.error("TEST_REVISION is required for proper publishing")
+
     try:
         task = PullTask(name, revision, opts.context, opts.rebase,
                         test_project=test_project, test_branch=test_branch)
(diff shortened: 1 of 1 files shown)


tests-invoke
```diff
@@ -177,13 +177,6 @@ class PullTask(object):
         status["link"] = "log.html"
         status["extras"] = ["https://raw.githubusercontent.com/cockpit-project/bots/master/task/log.html"]
 
-        # Include information about which base we're testing against
-        if self.base:
-            subprocess.check_call(["git", "fetch", "origin", self.base])
-            commit = subprocess.check_output(["git", "rev-parse", "origin/" + self.base],
-                                             universal_newlines=True).strip()
-            status["base"] = commit
```
martinpitt: Ack, seems the [sink](https://github.com/cockpit-project/cockpituous/blob/master/sink/sink) doesn't look at that field at all.

tests-invoke
```diff
@@ -242,10 +242,6 @@ class PullTask(object):
             self.start_publishing(opts.publish)
             os.environ["TEST_ATTACHMENTS"] = self.sink.attachments
 
-        head = subprocess.check_output(["git", "rev-parse", "HEAD"], universal_newlines=True).strip()
-        if not self.revision:
-            self.revision = head
```
martinpitt: From tests-invoke's perspective, it's not guaranteed to be defined. It does `revision = os.environ.get("TEST_REVISION")`, so it could happen that this env var is simply not set. I agree that it *should* be always set to have a meaningful collision detection and publishing, so if we use a sink it's mandatory.

So if publishing is on (`$TEST_PUBLISH` or `--publish`), please change the `os.environ.get()` to a `os.environ[]` and give a meaningful error message on `KeyError` -- i. e. let's enforce/assert that it's always set.

marusak: Put there check as you suggested. And also noticed that we read revision from env variable and then set this variable back in, so I dropped it as well.

tests-invoke
```diff
@@ -55,6 +55,10 @@ def main():
     # TEST_BRANCH would be `rhel-8.0`
     test_branch = os.environ.get("TEST_BRANCH")
 
+    if opts.publish and not revision:
+        sys.stderr.write("TEST_REVISION is required for proper publishing")
```
martinpitt: Missing `\n`. However, it's IMHO better to use `parser.error("TEST_REVISION is required for proper publishing")` (without `\n`), that also exits the program.

marusak: `parser.error()` is cool and useful! Shame I didn't know about it. Used that :)

PR #121, opened 2019-10-11
Rebase in make-checkout and not in tests-invoke

marusak:
It makes much more sense to do this in make-checkout. The only problem
is that if rebasing fails, it won't be in logs. But since the goal is to
make from logging wrapper around all commands, then also make-checkout
would log what is happening.

Also it could be written as shell script - but that may be a follow-up.

### DIFF
--- a/make-checkout
+++ b/make-checkout
@@ -37,6 +37,7 @@ def main():
     parser = argparse.ArgumentParser(description="Fetch and checkout specific revision")
     parser.add_argument("-v", "--verbose", action="store_true", help="Verbose output")
     parser.add_argument("--repo", nargs='?', help="Repository to check out")
+    parser.add_argument("--rebase", metavar="branch", help="Branch to rebase to")
     parser.add_argument("ref", help="The git ref to fetch")
     parser.add_argument("revision", nargs='?', help="Actual commit to check out, defaults to ref")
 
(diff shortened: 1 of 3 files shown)


marusak:
I am testing this (therefore draft), for now I ran only one cockpit-project/cockpit PR in master, I need to check external projects, bots and non-master branches. And forgot no-test label, sorry :)

marusak:
Tests:
1. cockpit-project/cockpit PR in master
```
BASE_BRANCH=master PRIORITY=0005 ./make-checkout --verbose --repo=cockpit-project/cockpit pull/12948/head && cd make-checkout-workdir && COCKPIT_BOTS_REF=master TEST_PROJECT=cockpit-project/cockpit TEST_NAME=pull-12948-20191011-101616 TEST_REVISION=287cead1c74ed51367b5aadce33825e8149db846 ../tests-invoke --pull-number=12948  fedora-30 ; cd ..
```
bots from master, commits from #12948 rebased on current master - OK

2. cockpit-project/cockpit PR in stable branch
```
TODO
```
bots from master, commits from ... rebased on the stable branch - TODO

3. cockpit-project/bots external project
```
TODO
```
bots from ..., commits from master - TODO

4. cockpit-project/cockpit-podman
```
TODO
```
bots from master, commits from ... rebased on master - TODO

make-checkout
```diff
@@ -44,6 +44,8 @@ def main():
     if not opts.revision:
         opts.revision = "FETCH_HEAD"
 
+    base = os.environ.get("BASE_BRANCH")
```
martinpitt: TBH I'd rather move away from environment variables in favor of options. It seems weird to mix the two. Let's please also make this a bit more descriptive, perhaps `--rebase BRANCH`?

make-checkout
```diff
@@ -37,6 +37,7 @@ def main():
     parser = argparse.ArgumentParser(description="Fetch and checkout specific revision")
     parser.add_argument("-v", "--verbose", action="store_true", help="Verbose output")
     parser.add_argument("--repo", nargs='?', help="Repository to check out")
+    parser.add_argument("--rebase", dest="branch", help="Branch to rebase to")
```
martinpitt: To avoid confusion, please drop dest and leave it as `opts.rebase`.

make-checkout
```diff
@@ -67,6 +68,14 @@ def main():
     execute("git", "fetch", "origin", opts.ref, cwd=TARGET_DIR, error_on_fail=False)
     execute("git", "checkout", "--detach", opts.revision, cwd=TARGET_DIR, error_on_fail=False)
 
+    if opts.base:
```
martinpitt: And this needs to be `opts.branch` with the current PR, but it should better be `opts.rebase` wrt. my comment above.

make-checkout
marusak: I just wanted to have in help `--rebase BRANCH` and not `--rebase REBASE`. Is it possible to specify this without dest?

make-checkout
marusak: noticed when I ran it :) But thanks for double checking after me ;)

make-checkout
martinpitt: Yes, use `metavar=`

marusak:
@martinpitt variable named "the right way" ;) and I finished testing (see the big comment). Is there some usecase I should test as well?

PR #135, opened 2019-10-15
Split tests-invoke into publisher and invoker

marusak:
I am currently testing it, but I thought I first may get some options so I can include them in the final testing.

@Gundersanne and @martinpitt options?

### DIFF
--- a/publish-wrapper
+++ b/publish-wrapper
@@ -1,0 +1,199 @@
+#!/usr/bin/env python3
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import os
+import socket
+import subprocess
+import sys
+
+from task import github
+from task import sink
+
+sys.dont_write_bytecode = True
+
+HOSTNAME = socket.gethostname().split(".")[0]
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Log to sink and publish into github')
+    parser.add_argument('--publish', help='Publish results centrally to a sink',
+                        default=os.environ.get("TEST_PUBLISH"))
+    parser.add_argument('--test-name', help='Test name', required=True)
+    parser.add_argument('--repo', help="The repository in which the tested PR is opened", default=None)
+    parser.add_argument('--revision', help="Revision of the PR head", required=True)
+    parser.add_argument('--github-context', help="The full context as written in github",
+                        required=True)
+    parser.add_argument('command', help="Command to wrap")
+    opts = parser.parse_args()
+
+    if not opts.publish:
+        parser.error("Publishing address is required")
+
+    try:
+        task = PullTask(opts.repo)
+        ret = task.run(opts)
+    except RuntimeError as ex:
+        ret = str(ex)
+
+    if ret:
+        sys.stderr.write("log-wrapper: {0}\n".format(ret))
+        return 1
+    return 0
+
+
+class PullTask():
+    def __init__(self, repo):
+        self.sink = None
+        self.github_status_data = None
+        self.api = github.GitHub(repo=repo)
+
+    def start_publishing(self, opts):
+        # build a unique file name for this test run
+        id_context = opts.github_context.replace("@", "-").replace("#", "-")
+        identifier = "-".join([
+            opts.test_name.replace("/", "-"),
+            opts.revision[0:8],
+            id_context.replace("/", "-")
+        ])
+
+        description = "{0} [{1}]".format(github.TESTING, HOSTNAME)
+
+        self.github_status_data = {
+            "state": "pending",
+            "context": opts.github_context,
+            "description": description,
+            "target_url": ":link"
+        }
+
+        status = {
+            "github": {
+                "token": self.api.token,
+                "requests": [
+                    # Set status to pending
+                    {
+                        "method": "POST",
+                        "resource": self.api.qualify("commits/" + opts.revision + "/statuses"),
+                        "data": self.github_status_data
+                    }
+                ],
+                "watches": [{
+                    "resource": self.api.qualify("commits/" + opts.revision + "/status?per_page=100"),
+                    "result": {
+                        "statuses": [
+                            {
+                                "context": opts.github_context,
+                                "description": description,
+                                "target_url": ":link"
+                            }
+                        ]
+                    }
+                }]
+            },
+            "revision": opts.revision,
+            "onaborted": {
+                "github": {
+                    "token": self.api.token,
+                    "requests": [
+                        # Set status to error
+                        {
+                            "method": "POST",
+                            "resource": self.api.qualify("statuses/" + opts.revision),
+                            "data": {
+                                "state": "error",
+                                "context": opts.github_context,
+                                "description": "Aborted without status",
+                                "target_url": ":link"
+                            }
+                        }
+                    ]
+                },
+            }
+        }
+
+        status["link"] = "log.html"
+        status["extras"] = ["https://raw.githubusercontent.com/cockpit-project/bots/master/task/log.html"]
+
+        self.sink = sink.Sink(opts.publish, identifier, status)
+
+    def stop_publishing(self, ret):
+        def mark_failed():
+            if "github" in self.sink.status:
+                self.github_status_data["state"] = "failure"
+
+        def mark_passed():
+            if "github" in self.sink.status:
+                self.github_status_data["state"] = "success"
+
+        if isinstance(ret, str):
+            message = ret
+            mark_failed()
+            ret = 0
+        elif ret == 0:
+            message = "Tests passed"
+            mark_passed()
+        else:
+            message = "Tests failed with code {0}".format(ret)
+            mark_failed()
+            ret = 0  # A failure, but not for this script
+        self.sink.status["message"] = message
+        if "github" in self.sink.status:
+            self.github_status_data["description"] = message
+        try:
+            del self.sink.status["extras"]
+        except KeyError:
+            pass
+        self.sink.flush()
+
+        return ret
+
+    def detect_collisions(self, opts):
+        statuses = self.api.get("commits/{0}/statuses".format(opts.revision))
+        for status in statuses:
+            if status.get("context") == opts.github_context:
+                if status.get("state") != "pending":
+                    return "Status is not pending"
+                if status.get("description") not in [github.NOT_TESTED, github.NOT_TESTED_DIRECT]:
+                    return "Status description is not in [{0}, {1}]".format(
+                        github.NOT_TESTED, github.NOT_TESTED_DIRECT)
+                return None  # only check the newest status of the supplied context
+        return None
+
+    def run(self, opts):
+        ret = self.detect_collisions(opts)
+        if ret:
+            sys.stderr.write("Collision detected: {0}".format(ret))
+            sys.stderr.flush()
+            return 0
+
+        self.start_publishing(opts)
+        os.environ["TEST_ATTACHMENTS"] = self.sink.attachments
+
+        msg = "Running on: `{0}`\nCommand: `{1}`\n".format(HOSTNAME, opts.command)
+        sys.stderr.write(msg)
+        sys.stderr.flush()
+
+        ret = subprocess.call(opts.command, shell=True)
+        ret = self.stop_publishing(ret)
+
+        return ret
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/tests-invoke
+++ b/tests-invoke
@@ -19,47 +19,57 @@
 
 import argparse
 import os
-import socket
 import subprocess
 import sys
 import time
 
-sys.dont_write_bytecode = True
-
 from task import github
-from task import sink
 from machine.machine_core.directories import BASE_DIR
 
-
-HOSTNAME = socket.gethostname().split(".")[0]
-DEVNULL = open("/dev/null", "r+")
+sys.dont_write_bytecode = True
 
 
 def main():
     parser = argparse.ArgumentParser(description='Run integration tests')
-    parser.add_argument('--publish', dest='publish', default=os.environ.get("TEST_PUBLISH", ""),
-                        action='store', help='Publish results centrally to a sink')
-    parser.add_argument('-v', '--verbose', action='store_true', help='Verbose output')
-    parser.add_argument('--pull-number', help="The number of the pull request to test")
-    parser.add_argument('context', help="The context or type of integration tests to run")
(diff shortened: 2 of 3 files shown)


publish-wrapper
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/env python3
```
marusak: Lets bikeshed about the name of this script :)

publish-wrapper
```diff
@@ -0,0 +1,205 @@
...
+    parser.add_argument('--publish', help='Publish results centrally to a sink',
+                        default=os.environ.get("TEST_PUBLISH"))
+    parser.add_argument('--test-name', help='Test name',
+                        default=os.environ.get("TEST_NAME", "tests"))
+    parser.add_argument('--test-project', help="Project which is being tested",
+                        default=os.environ.get("TEST_PROJECT", ""))
+    parser.add_argument('--base-project', help="Project in which the tested PR is opened",
+                        default=os.environ.get("BASE_PROJECT"))
+    parser.add_argument('--test-branch', help="Branch to be tested - if differs to PR base branch",
+                        default=os.environ.get("BASE_BRANCH", ""))
```
marusak: I still am not sure what would be the best approach, if to have only ENV or only positional arguments...

publish-wrapper
```diff
@@ -0,0 +1,205 @@
...
+                        default=os.environ.get("TEST_REVISION"))
+    parser.add_argument('--context', help="The context or type of integration tests to run",
+                        default=os.environ.get("TEST_CONTEXT"))
+    parser.add_argument('command', help="Command to wrap")
+    opts = parser.parse_args()
+
+    if not opts.test_project:
+        opts.test_project = opts.base_project
+
+    print(opts)
```
marusak: sorry, this is forgotten, needs to be removed

publish-wrapper
```diff
@@ -0,0 +1,205 @@
...
+    for opt in ["publish", "base_project", "revision", "context"]:
+        if getattr(opts, opt) is None:
+            parser.error("the following argument is required: {0}".format(opt))
+
+    try:
+        task = PullTask()
```
marusak: Kinda pointless to have class for this, would just the `main` function do?

tests-scan
```diff
@@ -142,6 +142,7 @@ def tests_invoke(priority, name, number, revision, ref, context, base, original_
 
     checkout = "PRIORITY={priority:04d} ./make-checkout --verbose"
     cmd = "TEST_PROJECT={repo} TEST_NAME={name}-{current} TEST_REVISION={revision} ../tests-invoke --pull-number={pull_number} "
+    wrapper = "TEST_PROJECT={repo} BASE_PROJECT={github_base} TEST_NAME={name}-{current} TEST_REVISION={revision} ./publish-wrapper "
```
marusak: Here is redundancy. I thought I could just drop those env variables that `cmd` and `wrapper` share - but then I would need to reconstruct it in the wrapper in order to log the command which is being executed.

tests-scan
```diff
@@ -153,9 +154,10 @@ def tests_invoke(priority, name, number, revision, ref, context, base, original_
 
     # Let tests-invoke know that we are triggering different branch - it needs to post correct status
     if base != original_base:
-        cmd = "TEST_BRANCH={base} " + cmd
+        wrapper = "TEST_BRANCH={base} " + wrapper
 
-    cmd += " {context}"
+    cmd += " --context={context}"
```
marusak: and yes, context is the one left `--type` argument :) consistency is the key!

publish-wrapper
martinpitt: It's just for not having a four-pages long main() function. You can of course turn the class into four simple functions. I still like being able to read main() in high-level terms, and separate functions do the details.

publish-wrapper
martinpitt: My gut feeling is that in order to be meaningful the sink needs a certain set of properties, like `TEST_NAME, `TEST_PROJECT`. These should be positional parameters or mandatory options. Things like `BASE_BRANCH` might be optional, and thus should become proper options.

`TEST_PUBLISH` now seems kind of meaningless to publish-wrapper itself; run-queue should look at this and call publish-wrapper around the command if it is set. However, I realize that this might be difficult as run-queue shouldn't interpret the command (and read TEST_PROJECT etc.) In tests-scan it's also too early, as that runs on a different environment (webhook) that doesn't have `$TEST_PUBLISH`. So maybe leave the `--publish` option as-is for now?

publish-wrapper
martinpitt: blue :bike: :house: !

I thought of "sink-logger", but that doesn't capture the GitHub statuses aspect. publish-wrapper is good enough for now *shrug*

publish-wrapper
```diff
@@ -0,0 +1,205 @@
...
+from task import github
+from task import sink
+
+sys.dont_write_bytecode = True
+
+HOSTNAME = socket.gethostname().split(".")[0]
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Log and publish into github')
```
martinpitt: "Log to sink and .."

publish-wrapper
```diff
@@ -0,0 +1,205 @@
...
+    parser.add_argument('--publish', help='Publish results centrally to a sink',
+                        default=os.environ.get("TEST_PUBLISH"))
+    parser.add_argument('--test-name', help='Test name',
+                        default=os.environ.get("TEST_NAME", "tests"))
+    parser.add_argument('--test-project', help="Project which is being tested",
+                        default=os.environ.get("TEST_PROJECT", ""))
+    parser.add_argument('--base-project', help="Project in which the tested PR is opened",
+                        default=os.environ.get("BASE_PROJECT"))
+    parser.add_argument('--test-branch', help="Branch to be tested - if differs to PR base branch",
```
martinpitt: When would that be used? For our foreign tests, like testing starter-kit in a bots PR? Wouldn't that just be TEST_PROJECT=cockpit-starter-kit?

publish-wrapper
```diff
@@ -0,0 +1,205 @@
...
+    parser.add_argument('command', help="Command to wrap")
+    opts = parser.parse_args()
+
+    if not opts.test_project:
+        opts.test_project = opts.base_project
+
+    print(opts)
+    for opt in ["publish", "base_project", "revision", "context"]:
+        if getattr(opts, opt) is None:
+            parser.error("the following argument is required: {0}".format(opt))
```
martinpitt: I haven't tried, but does the [required](https://docs.python.org/3/library/argparse.html#required) option work with `default` if it gets it from the environment? (Aside from the fact that I'd really like to drop these as much as possible)

Also, is it a thing to just use the GitHub status wrangling without log publishing? I. e. to what extent is `--publish` actually optional? (Also see my remark above)

publish-wrapper
```diff
@@ -0,0 +1,205 @@
...
+
+    try:
+        task = PullTask()
+        ret = task.run(opts)
+    except RuntimeError as ex:
+        ret = str(ex)
+
+    if ret:
+        sys.stderr.write("log-wrapper: {0}\n".format(ret))
```
martinpitt: This whole ret= thing could be simplified. Drop the `return 0`, and move this error handling right into the `except:` above.

tests-invoke
```diff
@@ -23,240 +23,81 @@ import socket
...
-                        action='store', help='Publish results centrally to a sink')
     parser.add_argument('-v', '--verbose', action='store_true', help='Verbose output')
     parser.add_argument('--pull-number', help="The number of the pull request to test")
-    parser.add_argument('context', help="The context or type of integration tests to run")
+    parser.add_argument('--test-name', help='Test name',
+                        default=os.environ.get("TEST_NAME", "tests"))
+    parser.add_argument('--revision', help="Revision of the PR head",
+                        default=os.environ.get("TEST_REVISION"))
+    parser.add_argument('--context', help="The context or type of integration tests to run",
+                        default=os.environ.get("TEST_CONTEXT"))
```
martinpitt: Oh, so was that the motivation for using env vars, so that they get shared between the logger and tests-invoke? Hmm. Does tests-invoke actually need all these? I would have thought it's mainly interested in setting `$TEST_SCENARIO`, which it parses from the context?

publish-wrapper
marusak: If there is PR in `cockpit-project/bots` (like image refresh) and you test `context@cockpit-project/cockpit-podman` then:
- `TEST_PROJECT=cockpit-project/cockpit-podman` - used to create correct github context
- `BASE_PROJECT=cockpit-project/bots` - used so we know where to publish those contexts

publish-wrapper
marusak: the `ret` is also set up in `try` as well as in `except`. I think it would make it just more complicated?
```
     try:                                                                        
          task = PullTask()                                                       
          ret = task.run(opts)                                                    
          if ret:                                                                 
              sys.stderr.write("log-wrapper: {0}\n".format(ret))                  
          return 0                                                                
      except RuntimeError as ex:                                                  
          ret = str(ex)                                                           
          if ret:                                                                 
              sys.stderr.write("log-wrapper: {0}\n".format(ret))                  
          return 1 
```

tests-invoke
marusak: well, it does need:
- `--pull-number` to detect collisions and thus needs also `--revision`. I am not sure where is the best place for this, but I guess here?
- `--test-name` seems indeed redundant as it is used only used to output name of the test
` `--context` is used as scenario

publish-wrapper
marusak: Since we are passing full github context, this indeed is not needed anymore.

marusak:
Tests scenarios:
1. starter-kit PR
```
```
NOT TESTED

2. bots PR testing bots (host)
```
```
NOT TESTED

3. bots PR testing starter-kit
```
```
NOT TESTED

4. starter-kit PR testing bots PR (`@bots#..`)
```
```
NOT TESTED

5. cockpit 8.2 PR
```
```
NOT TESTED

6. bots PR testing cockpit 8.2 
```
```
NOT TESTED

marusak:
OK, lets call this ready for review :)
I ran a few tests and posted here 2 results (with command and output log). I ran a few more tests, but since the command changed ever so slightly, I don't show them here. Once this gets lgtm, I'll run all the mentioned scenarios, before we land this.

tests-scan
```diff
@@ -139,34 +139,26 @@ def tests_invoke(priority, name, number, revision, ref, context, base,
...
-        cmd = "COCKPIT_BOTS_REF={bots_ref} " + cmd
+        checkout += " {ref}"
+        cmd += " --bots-ref {bots_ref}"
     else:
         # we are testing the repo itself, checkout revision from the PR
-        checkout += " {ref} {revision} && "
+        checkout += " {ref} {revision}"
 
-    return (checkout + "cd make-checkout-workdir && " + cmd + " ; cd ..").format(
+    return (wrapper + " \"" + checkout + " && cd make-checkout-workdir && " + cmd + "\"").format(
```
martinpitt: nitpick: using single quotes is easier here, avoids the \

tests-scan
```diff
@@ -318,15 +310,6 @@ def cockpit_tasks(api, update, contexts, repo, pull_data, pull_number, sha, amqp
             # end up with a bunch of half tested pull requests
             baseline += 1.0 - (min(100000, float(number)) / 100000)
 
-        def get_externals():
```
martinpitt: This is nice code cleanup, but should be a separate commit.

tests-invoke
```diff
@@ -19,257 +19,82 @@
...
 
 def main():
     parser = argparse.ArgumentParser(description='Run integration tests')
-    parser.add_argument('--publish', dest='publish', default=os.environ.get("TEST_PUBLISH", ""),
-                        action='store', help='Publish results centrally to a sink')
     parser.add_argument('-v', '--verbose', action='store_true', help='Verbose output')
-    parser.add_argument('--pull-number', help="The number of the pull request to test")
-    parser.add_argument('context', help="The context or type of integration tests to run")
+    parser.add_argument('--revision', help="Revision of the PR head", required=True)
+    parser.add_argument('--base-project', help="Project in which the tested PR is opened",
```
martinpitt: Why "base" project? Is there any other kind?

tests-invoke
```diff
@@ -19,257 +19,82 @@
...
+                        required=True)
+    parser.add_argument('--pull-number', help="The number of the pull request to test",
+                        required=True)
+    parser.add_argument('--context', help="Test context", required=True)
```
martinpitt: I have a gut feeling that this should be dropped, and setting `TEST_{OS,SCENARIO}` should be put into the command line by tests-scan directly. This avoids propagating the context syntax into too many different scripts.

But this is fine for a follow-up, of course.

tests-invoke
```diff
@@ -19,257 +19,82 @@
...
+    parser.add_argument('--bots-ref', help="Reference from which bots can be prepared")
```
martinpitt: In the same vein, this also feels a bit like busywork: This just translates an option into an env var, tests-scan could just set this directly. (Again, fine for follow-up cleanup, if we care enough)

tests-invoke
```diff
@@ -19,257 +19,82 @@
...
-        # All done
-        if self.sink:
-            ret = self.stop_publishing(ret)
+        entry_point = os.path.join(BASE_DIR, ".cockpit-ci", "run")
+        alt_entry_point = os.path.join(BASE_DIR, "test", "run")
+        if not os.path.exists(entry_point) and os.path.exists(alt_entry_point):
+            entry_point = alt_entry_point
+        cmd = ["timeout", "120m", entry_point]
+        if opts.verbose:
+            cmd.append("--verbose")
```
martinpitt: I know this isn't new code, but none of our projects actually supports this. So this and the whole `--verbose` option could be dropped.

publish-wrapper
```diff
@@ -0,0 +1,204 @@
...
+                    ]
+                },
+            }
+        }
+
+        status["link"] = "log.html"
+        status["extras"] = ["https://raw.githubusercontent.com/cockpit-project/bots/master/task/log.html"]
+
+        # For other scripts to use
+        os.environ["TEST_DESCRIPTION"] = description
```
martinpitt: I can't find any place which uses this env var, so this is another potential cleanup target.

publish-wrapper
```diff
@@ -0,0 +1,204 @@
...
+                if status.get("description") not in [github.NOT_TESTED, github.NOT_TESTED_DIRECT]:
+                    return "Status description is not in [{0}, {1}]".format(
+                        github.NOT_TESTED, github.NOT_TESTED_DIRECT)
+                return None  # only check the newest status of the supplied context
+        return None
+
+    def run(self, opts):
+        ret = self.detect_collisions(opts)
+        if ret:
+            sys.stderr.write(ret)
```
martinpitt: The previous version prefixed the error with "Collision detected:". Did you remove that on purpose?

publish-wrapper
```diff
@@ -0,0 +1,204 @@
...
+            sys.stderr.flush()
+            return 1
```
martinpitt: Current master treats this as success (`return None`), now this counts as a failure. The former seems more correct to me, as another bot is already handling the test -- i. e. we should avoid treating a non-zero exit as an infra failure that causes run-tests to requeue the test. Although I don't think that actually happens (run-queue seems to entirely ignore the exit code).

Did you change that on purpose?

marusak:
Thank you @martinpitt for the review. I updated all of your comments. Also updated the examples and logs.

tests-invoke
marusak: Technically there is also `test-project` (which we pass into `make-checkout`).
`test-project` is the one, we actually test - thus passing into `make-checkout`.
`base-project` is the one, in which the PR is opened - thus passing into `publish_wrapper` and `tests-invoke` (both use it for querying statuses and doing collision detection).

But I named them all `--repo`. Help explains what repo it expects.

marusak:
Ok, I am happy with tests I ran (github fails to update the message so no more updates).
This can land when safe to do so. I will then watch out tests to see if there are some regression and be ready for PPRs :)

PR #160, opened 2019-10-21
[no-test] [WIP] [Tests] Add new manual composer test 

jrusz:
I want to add a new manual action to try my test.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -107,6 +107,7 @@ REPO_BRANCH_CONTEXT = {
         ],
         # These can be triggered manually with bots/tests-trigger
         '_manual': [
+            'fedora-30/ci',
         ],
     },
     'weldr/cockpit-composer': {


task/testmap.py
```diff
@@ -107,6 +107,7 @@
         ],
         # These can be triggered manually with bots/tests-trigger
         '_manual': [
+            'fedora-30/ci',
```
martinpitt: "ci" is a rather sad scenario name -- each test is CI, no? What's that supposed to cover? Or is that meant to be a generic placeholder for "random stuff" you want to try in test/run?

jrusz: Yes that's true.. I will find a better name for it when I am done with the test. So yes it's a placeholder for stuff I want to try now manully. It's for this https://github.com/weldr/lorax/pull/882 I want to run pylint and some coverage with it.

martinpitt: Note that for such a temporary case you don't need to invent a place-holder -- you can just mutate any existing scenario in test/run to do something different. However, there's nothing specifically bad at this, as long as it doesn't move into the "automatic" section.

PR #157, opened 2019-10-20
Image refresh for rhel-8-1

cockpituous:
Image refresh for rhel-8-1
 * [x] image-refresh rhel-8-1

PR #163, opened 2019-10-23
Image refresh for openshift

cockpituous:
Image refresh for openshift
 * [x] image-refresh openshift


marusak:
`Please install intltool package` as well as in https://github.com/cockpit-project/bots/pull/157

PR #164, opened 2019-10-23
Image refresh for selenium (deprecated)

cockpituous:
Image refresh for selenium
 * [x] image-refresh selenium
 * [x] after https://github.com/cockpit-project/bots/pull/301 lands, close this


martinpitt:
I robustified the selenium tests a bit, let's retry this.

martinpitt:
This grew stale, I rebased it to current master.

martinpitt:
This [breaks composer tests](https://209.132.184.41:8493/logs/pull-164-20191106-062727-ccb630da-weldr-cockpit-composer--fedora-30-firefox/log.html), lots of "HTTP method not allowed" errors. @henrywang, does that ring a bell? Seems newer firefox breaks  something here?

henrywang:
@martinpitt, the root cause is [geckodirver v0.25.0](https://github.com/mozilla/geckodriver/releases/tag/v0.25.0) dropped support for legacy Selenium web element references. Since [selenium-hub image v3.141.59-vanadium](https://github.com/SeleniumHQ/docker-selenium/releases/tag/3.141.59-vanadium), the geckodirver has been updated to v0.25.0. And the new selenium image just updated selenium-hub image to 3.141.59-vanadium. That caused `HTTP method not allowed` error.
The solution is to update webdirverio to v5.x which supports WebDriver standard. I've tried before, the issue of the new version is the [`setValue` command does not work with Windows Edge](https://github.com/webdriverio/webdriverio/issues/3324) and I used a lot of `setValue` to fill in form or box, like name, password, etc. According to this issue comments, the root cause comes from Microsoft Edge browser, and their's a [bug](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/22429136/) filed already in developer.microsoft.com. But no any response and fix in that bug.
My current solution is try some work-arounds, but I do not know whether they work or not.

martinpitt:
@henrywang : Thanks for your investigations! I was just wondering why we don't see that error in cockpit's selenium tests. If this is hard to work around, I'm also okay with pinning down the firefox container to the current version; but that's of course a sad long-term fix as the point of these tests is also to cover current browsers.

henrywang:
@martinpitt, right! I'm working on some work around. I hope one of them works!

martinpitt:
As this has been blocked for a while, I created https://github.com/cockpit-project/bots/pull/281 so that we can use a current firefox/chrome version for cockpit testing.

henrywang:
Thanks @martinpitt! The webdriver.io update PR is https://github.com/weldr/cockpit-composer/pull/833.
Lots of code changed due to framework changed. Need some time make test stable again.

martinpitt:
This is deprecated, close this once https://github.com/cockpit-project/bots/pull/301 lands.

martinpitt:
#301 landed, closing.

PR #168, opened 2019-10-23
Image refresh for fedora-30

cockpituous:
Image refresh for fedora-30
 * [x] image-refresh fedora-30


martinpitt:
oh, weldr/lorax moved to fedora-31, that's why these tests don't get served. Good enough then. Retrying the cockpit selenium tests a bit, though.

PR #169, opened 2019-10-24
image-refresh cleanups

martinpitt:
In particular this fixes missing or obsolete test triggers. I tested this by adding a `print(triggers)` and `return`.


martinpitt:
FTR, with this we can also implement a tests-trigger for a manual image refresh PR. Right now, this requires manual test map lookup and copying the tests individually, which is a bit annoying.

martinpitt:
> Thanks! This is great since we don't need to maintain it on two places. I checked what differences it produces, here is the list: (I guess some are wanted)

Thanks for the review!
 
>     1. `continuous-atomic` is in `rhel-7.7` branch but not in `rhel-7.8` branch (but we were triggering `rhel-7.8`) - this seems like misconfiguration in testmap.py

It's technically okay, as continous-atomic is built from centos-7 (see `get_build_image()` in machine/machine_core/machine_virtual.py). We just want to test changes to the rhel-7.x branches on c-atomic. It doesn't hurt to add that, I did that in a separate commit now.

>     2. `fedora-29` was triggering `fedora-29@podman` and `fedora-atomic`. Now it only triggers F29.
>     5. rhel-7-7 does not trigger `rhel-atomic` anymore

Thanks for spotting! The atomics are indeed missing.

>     3. Fedora-30 triggers stable branches, like `fedora-30/selenium-firefox@cockpit-project/cockpit/rhel-8.1`
>     4. Fedora-31, rhel-7-8 and rhel-8-1 triggers all `weldr/lorax` contexts
>     7. rhel-atomic also triggers `rhel-atomic@cockpit-project/cockpit/rhel-7.7`

These are an example of "woefully incomplete", i. e. what this PR wants to fix in a generic way.

>     6. rhel-8-0 does not trigger anything.

That's a consequence from this morning's PR #170. So that's correct.

PR #170, opened 2019-10-24
testmap: Drop rhel 8.0 branches

martinpitt:
The 8.0 branches realistically won't change any more, so deprecate them.
Add missing cockpit-composer tests for the rhel-8.1 branch.


martinpitt:
@henrywang, @larskarlitski  : Is that ok from your (c-composer) perspective?

PR #197, opened 2019-11-03
Image refresh for fedora-29

cockpituous:
Image refresh for fedora-29
 * [x] image-refresh fedora-29


martinpitt:
@marusak: Yes, IPA should be fairly simple to move. I would actually like to centralize our "external services" like ipa, candlepin, selenium, and perhaps openshift into one "cockpit-services" VM image, to avoid the overhead of downloading and dragging around 4 images. I think as the first step we should try and move the IPA server into a docker deployment. I'm happy to look into this next week.

Fedora Atomic's successor is Fedora Core. We started the discussion about the sucessor of the cockpit/ws container a while ago in issue #11002, but never really finished it. IMHO it would be fine to at least try and keep cockpit/ws working on Fedora Core, and thus create a fedora-core image as successor of fedora-atomic. At this point this shouldn't be much work.

PR #201, opened 2019-11-03
images: install strace and ltrace

allisonkarlitskaya:
strace and ltrace are extremely useful tools, and quite small.  Let's
have them in our images.


martinpitt:
I suggest doing this in three separate PRs, as having one PR that essentially rebuilds *all* our images is  rather unwieldy.

allisonkarlitskaya:
Okay.

PR #206, opened 2019-11-04
Revert "testmap: Revert switch to Fedora-31 for weldr/lorax"

atodorov:
This reverts commit 1417905f839dfba8a221b0e9bb630b8d07717f8e.

Nested virt issues have been sorted out so lorax-composer test suite shouldn't have any problems with qcow2, live-iso or kickstart tar tests going forward.


martinpitt:
This didn't actually run the f31 tests, I'm triggering them now.

PR #208, opened 2019-11-06
Add new "services" image

martinpitt:
Let's reduce the number of auxiliary images that we carry around and have to download, and put the external services that tests talk to into a single "services" image.

 * [x] image-refresh services


cockpituous:
image-refresh in progress on 3-ci-srv-03.
Log: https://209.132.184.41:8493/logs/image-refresh-208-20191106-084848/

cockpituous:
Task aborted.

cockpituous:
image-refresh in progress on 2-cockpit-10.
Log: https://209.132.184.41:8493/logs/image-refresh-208-20191106-093131/

cockpituous:
Task failed: https://209.132.184.41:8493/logs/image-refresh-208-20191106-093131/

cockpituous:
image-refresh in progress on 4-ci-srv-01.
Log: https://209.132.184.41:8493/logs/image-refresh-208-20191106-095757/

cockpituous:
Task failed: https://209.132.184.41:8493/logs/image-refresh-208-20191106-095757/

martinpitt:
Added a workaround for broken candlepin Ansible scripts.

cockpituous:
Log: https://209.132.184.41:8493/logs/image-refresh-208-20191106-105151/

cockpituous:
Task failed: https://209.132.184.41:8493/logs/image-refresh-208-20191106-105151/

martinpitt:
[Failed](https://209.132.184.41:8493/logs/image-refresh-208-20191106-105151/log) with 

    barnabycourt.fedorapeople.org: Name or service not known

that seems to work for me locally, retrying.

cockpituous:
image-refresh in progress on 3-ci-srv-03.
Log: https://209.132.184.41:8493/logs/image-refresh-208-20191106-112828/

martinpitt:
Hmm, restarting the container after its creation works great, but not after a reboot:
```
Error response from daemon: error creating overlay mount to /var/lib/docker/overlay2/8183f0db5b6eca68ce8b6693d5cfaebfe868bc81416ae6091668cd464078ff33/merged: invalid argument
Error: failed to start containers: freeipa
```
I need to try something else then :-/

cockpituous:
image-refresh services done: https://github.com/cockpituous/bots/commits/image-refresh-services-20191106-122121

mvollmer:
This increases coupling, no?  If we need a newer version of FreeIPA we might have to battle problems with installing Candlepin, and vice versa.  This would be frustrating.

mvollmer:
> This increases coupling, no? If we need a newer version of FreeIPA we might have to battle problems with installing Candlepin, and vice versa. This would be frustrating.

Let me gently slap my own hand right away: We need to always care about all of them anyway, and it's not as if the old images would disappear.

martinpitt:
I found a better way to handle the freeipa container that doesn't rely on docker volumes. Locally I have check-realms working now.

cockpituous:
image-refresh in progress on 2-cockpit-9.
Log: https://209.132.184.41:8493/logs/image-refresh-208-20191106-131515/

cockpituous:
image-refresh services done: https://github.com/cockpituous/bots/commits/image-refresh-services-20191106-140909

martinpitt:
We don't have any tests yet that use this new image. I locally ran them on a local cockpit branch. I'll submit that once this lands.

For now, I just trigger a cockpit test to confirm that the old candlepin symlink still works. It's an orphan now, but it only needs to last for a day or two.

martinpitt:
testLibvirt is the flake du jour, and unrelated to ipa or candlepin images.

martinpitt:
Triggering rhel-7.8 as well, that uses candlepin more than master.

images/scripts/services.setup
```diff
@@ -0,0 +1,138 @@
...
+cat > playbook.yml <<- EOF
+- hosts: dev
+
+  environment:
+    JAVA_HOME: /usr/lib/jvm/java-1.8.0/
+
+  roles:
+     - role: candlepin
+       candlepin_git_pull: True
+       candlepin_deploy_args: "-g -a -f -t"
```
jirihnidek: '-r' option can create testing repositories, but it also requires some other packages.

images/scripts/services.setup
```diff
@@ -0,0 +1,138 @@
...
+ipa dnsconfig-mod --forwarder=8.8.8.8
+poweroff
+'
+
+
+#############################
+#
+# candlepin setup
+#
+#############################
```
jirihnidek: BTW: we provide docker images of candlepin server too, but I don't have any experience with candlepin in docker.

martinpitt: @jikortus : Thanks for the hint! Indeed I'm not a big fan of the ansible scripts -- in half of the cases when we run them (about once every 3 months) they fail. Most recent one was [this failure](https://209.132.184.41:8493/logs/image-refresh-208-20191106-095757/log) from [this PR](https://github.com/candlepin/ansible-role-candlepin/pull/7). Where do you release the docker images? I did not find any on [docker hub](https://hub.docker.com/search?q=candlepin&type=image).

jikortus: @martinpitt : It looks like the above comment should have been addressed to @jirihnidek :).

martinpitt: Whoops, indeed, sorry!

jirihnidek: @martinpitt I'm sorry, but internal building of candlepin docker images is broken ATM. I will provide more details on IRC chat.

PR #209, opened 2019-11-06
Change obsolete SSH ACCESS message

jrusz:
It is no longer possible to log in to fedora-31 using just ssh and password like in fedora-30.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -27,8 +27,7 @@ TTY LOGIN
   To quit use Ctrl+], Ctrl+5 (depending on locale)
 
 SSH ACCESS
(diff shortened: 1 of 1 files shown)


machine/machine_core/machine.py
```diff
@@ -27,8 +27,7 @@
   To quit use Ctrl+], Ctrl+5 (depending on locale)
 
 SSH ACCESS
-  $ ssh -p {ssh_port} {ssh_user}@{ssh_address}
-  Password: foobar
+  $ ssh -p {ssh_port} {ssh_user}@{ssh_address} -i bots/machine/identity
```
martinpitt: Hah, I didn't even notice, as pretty well everyone in the team has that in their ~/.ssh/config. Can you please move the -i after the -p, so that the options all come before the positional arguments?

PR #215, opened 2019-11-07
naughty: Consistent pattern for unexpected audit messages

### LINKED ISSUE #214: Update naughties for `testlib.Error`, or fix the format
PR cockpit-project/cockpit#9571 moves verify tests to Python 3, which changes the error output from `Error: ...` to `testlib.Error:`. tests-policy normalizes that to `Error:`. Either find a clever way to make the exception print out the old format, or change the naughties and normalize into the other direction for Python 2 test runs.

martinpitt:
The precise output format of unexpected messages has changed over time.
A long time ago with Pythoh 2 they were prefixed with `Error:`, then
with Python 3 it changed to `testlib.Error:`, and with [1] that prefix
moved to a separate line.

Change the naughties to accept any prefix, so that this works across all
branches.

[1] https://github.com/cockpit-project/cockpit/pull/13022

### DIFF
--- a/naughty/fedora-30/118-selinux-rpcbind-name_bind
+++ b/naughty/fedora-30/118-selinux-rpcbind-name_bind
@@ -1,1 +1,1 @@
-testlib.Error: audit: type=1400 audit(*): avc:  denied  { name_bind } * comm="rpcbind"
+* type=1400 audit(*): avc:  denied  { name_bind } * comm="rpcbind"
--- a/naughty/continuous-atomic/62-selinux-rhsmd-dbus
+++ b/naughty/continuous-atomic/62-selinux-rhsmd-dbus
@@ -1,1 +1,1 @@
-Error: type=1400 audit(*): avc:  denied  { write } for  pid=* comm="rhsmd" name="dbus"
+* type=1400 audit(*): avc:  denied  { write } for  pid=* comm="rhsmd" name="dbus"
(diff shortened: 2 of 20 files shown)


martinpitt:
Oh, thinko -- the testlib.Error rewriting doesn't just apply to unexpected messages, but to stack traces as well. My bad!

martinpitt:
Fixed, I'll send a separate PR for the testlib.Error cleanup.

marusak:
Hmm, this should be taken care of, right? https://209.132.184.41:8493/logs/pull-215-20191107-101010-b866bcfb-cockpit-project-cockpit--fedora-30/log.html#226-2

naughty/fedora-30/118-selinux-rpcbind-name_bind
```diff
@@ -1 +1 @@
-testlib.Error: audit: type=1400 audit(*): avc:  denied  { name_bind } * comm="rpcbind"
+* audit: type=1400 audit(*): avc:  denied  { name_bind } * comm="rpcbind"
```
marusak: heh, this cannot be like this, there is no space before audit. And also it is not "consistent pattern" :)

martinpitt: Argh, sorry -- `git grep audit:` is empty now. Wash, rinse, repeat! (Pushed)

PR #225, opened 2019-11-11
image-download: Add fallback RedHat-internal image server in AWS

martinpitt:
This is only accessible via internal VPN. This will be used as a
fallback when the primary image server in e2e does not work.


martinpitt:
Right, this is most probably not the final version anyway. But I'd like to add it as a fallback now, so that we can gain some experience with it.

PR #227, opened 2019-11-12
Image refresh for rhel-8-1

cockpituous:
Image refresh for rhel-8-1
 * [x] image-refresh rhel-8-1


marusak:
[Two](https://209.132.184.41:8493/logs/pull-227-20191112-073636-bd54d993-cockpit-project-cockpit-rhel-8-appstream-rhel-8-1/log.html) failed tests, retying to compare

marusak:
[testMultipleZones](https://209.132.184.41:8493/logs/pull-227-20191112-103939-bd54d993-cockpit-project-cockpit-rhel-8-appstream-rhel-8-1/backup.OY3pUy/log.html#31) failed again. @Gundersanne I know this test was one of those tests that it took longer to get green but I though it was fixed. This is in `rhel-8-appstream` branch - is it possible we just did not backport some fix there?

martinpitt:
These happen on the current image as well, e. g. [here](https://209.132.184.41:8493/logs/pull-13130-20191112-101818-2283dc9e-cockpit-project-cockpit--rhel-8-1-distropkg/log.html). So not a regression. I propose we ignore this for now to avoid needless work: we don't plan to do an AppStream update for 8.1.1 right now. If we do, we can still ignore the bits that apply to BaseOS, like firewall.

PR #230, opened 2019-11-12
Image refresh for rhel-7-7

cockpituous:
Image refresh for rhel-7-7
 * [x] image-refresh rhel-7-7


marusak:
So we retired rhel-7-7 branch but we kept this image. Do we plan to keep it for some time just in case?

PR #232, opened 2019-11-12
images: Add workaround for fedora-coreos bugs

martinpitt:
This blocks https://github.com/cockpit-project/cockpit/pull/13110.

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -687,5 +687,5 @@ class VirtMachine(Machine):
 
     def needs_writable_usr(self):
         # On atomic systems, we need a hack to change files in /usr/lib/systemd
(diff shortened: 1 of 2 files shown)


machine/machine_core/machine_virtual.py
```diff
@@ -687,5 +687,5 @@ def add_netiface(self, networking=None):
 
     def needs_writable_usr(self):
         # On atomic systems, we need a hack to change files in /usr/lib/systemd
-        if self.atomic_image:
+        if self.atomic_image or "coreos" in self.image:
```
marusak: this could just be `if self.ostree_image:`

martinpitt: Indeed, and I had that locally, but forgot to `--amend` it, bah! Pushed.

PR #237, opened 2019-11-13
images: Drop fedora-atomic

martinpitt:
Fedora 29, and with it, Fedora Atomic, is EOL now. We now have its
successor, Fedora CoreOS.


martinpitt:
[testJournal](https://209.132.184.41:8493/logs/pull-237-20191113-225454-d1e62a65-cockpit-project-cockpit--fedora-coreos/backup.nEhY4m/log.html#51-2) seems to be flaky, I'll try and reproduce locally. It seems we don't get screenshots with f-coreos.

PR #243, opened 2019-11-14
Fix pyflakes/pycodestyle

martinpitt:
Our test/run was missing all our python modules.


martinpitt:
Rebased after #242 landed, to fix conflicts.

PR #244, opened 2019-11-14
testmap: Remove manual composer test

jrusz:
We don't need this manual trigger anymore since we're dropping the test.


jrusz:
Right, makes sense, I'll try to remember next time :)

PR #272, opened 2019-11-22
testmap: Add centos-8-stream to starter-kit master tests

Gundersanne:
These just work out of the box :)


marusak:
What about cockpit-podman, have you tried running that one?

Gundersanne:
@marusak let's find out :p

PR #274, opened 2019-11-25
[no-test] testmap: update all rhel-8-1 related triggers to rhel-8-2 for cockpit-composer

henrywang:
Update all rhel-8-1 related triggers to rhel-8-2 for cockpit-composer

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -136,20 +136,14 @@ REPO_BRANCH_CONTEXT = {
     'weldr/cockpit-composer': {
         'master': [
             'fedora-30/chrome',
-            'fedora-30/firefox',
-            'fedora-30/edge',
+            'rhel-8-2/firefox',
+            'fedora-31/edge',
             'fedora-31/firefox',
             'rhel-7-8/firefox',
             'rhel-8-1/chrome',
         ],
-        'rhel-8.1': [
-            'rhel-8-1/chrome',
-            'rhel-8-1/firefox',
-            'rhel-8-1/edge',
-        ],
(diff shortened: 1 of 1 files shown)


task/testmap.py
```diff
@@ -140,16 +140,15 @@
...
+            'rhel-8-2/chrome',
         ],
-        'rhel-8.1': [
-            'rhel-8-1/chrome',
-            'rhel-8-1/firefox',
-            'rhel-8-1/edge',
+        'rhel-8.2': [
+            'rhel-8-2/chrome',
+            'rhel-8-2/firefox',
+            'rhel-8-2/edge',
```
martinpitt: There actually is no "rhel-8.2" composer branch, and "rhel-8.1" did not exist either. "rhel-8.0" still exists (but is largely neglected now, I think we can just delete it). I would recommend to just drop this paragraph completely. We can reintroduce it once we actually need and have a stable branch again, WDYT?

task/testmap.py
```diff
@@ -140,16 +140,15 @@
             'fedora-30/edge',
             'fedora-31/firefox',
             'rhel-7-8/firefox',
-            'rhel-8-1/chrome',
+            'rhel-8-2/chrome',
```
martinpitt: We would lose RHEL 8.1 tests on master with that. I would recommend to keep 8.1 until 8.2 gets released. Instead, fedora-30 is rather over-represented here, how about moving one f-30 test to f31, one to rhel-8-2, and keep the third as f-30?

PR #299, opened 2019-12-03
Image refresh for rhel-8-1

cockpituous:
Image refresh for rhel-8-1
 * [x] image-refresh rhel-8-1


martinpitt:
There is a new SELinux regression:
```
testlib.Error: audit: type=1400 audit(1575466842.614:9): avc:  denied  { dac_override } for  pid=2141 comm="rhsmcertd-worke" capability=1  scontext=system_u:system_r:rhsmcertd_t:s0 tcontext=system_u:system_r:rhsmcertd_t:s0 tclass=capability permissive=0
```

martinpitt:
The above SELinux violation is either the same as, or very similar to https://bugzilla.redhat.com/show_bug.cgi?id=1753991 . I reported it there for now. The others look like flakes. As this image was built on e2e, it's currently unavailable, so we cannot further test *this* PR. (We can of course just close this and let the AWS cloud build a new one. -- I'm trying that for PR #300 first)

PR #301, opened 2019-12-04
image-trigger: Drop selenium image refresh

martinpitt:
This image got deprecated by the services image (commit 4b17bf2), and
can't be refreshed any more until cockpit-composer gets along with
current Firefox (webdriver.io v5).

Refresh the services image every 30 days, so that we notice regressions
earlier.

Fix inconsistent quoting style in the map while we are at it.

### DIFF
--- a/image-trigger
+++ b/image-trigger
@@ -20,7 +20,6 @@
 DAYS = 7
 
 REFRESH = {
-    "services": {"refresh-days": 120},
     "centos-7": {},
     "centos-8-stream": {},
     "continuous-atomic": {},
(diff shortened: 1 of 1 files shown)


marusak:
Thanks. But are we keeping selenium image for now, just not update it? Is anyone still using this image?

martinpitt:
@marusak: Yes, see https://github.com/cockpit-project/bots/pull/164 . Once https://github.com/weldr/cockpit-composer/pull/833 lands, we can move cockpit-composer to the services image, and drop our selenium image.

PR #300, opened 2019-12-04
Image refresh for rhel-atomic

cockpituous:
Image refresh for rhel-atomic


martinpitt:
Seems this new image [breaks sosreport](https://209.132.184.41:8493/logs/pull-300-20191204-091616-20fe1999-cockpit-project-cockpit-rhel-7.8-rhel-atomic/log.html)

martinpitt:
This image was built on e2e, which went down shortly afterwards. Let's let the bots build a new image on AWS instead.

PR #306, opened 2019-12-05
Fixes for absent e2e cluster

martinpitt:
Our primary internal image server (https://cockpit-11.e2e.bos.redhat.com:8493/) is down right now, and trying to connect to it hangs eternally. This also causes the "am I in the RH network?" detection to fail, thus currently our AWS and PSI bots don't run any rhel* and windows tests.

### DIFF
--- a/image-download
+++ b/image-download
@@ -83,7 +83,8 @@ def find(name, stores, latest, quiet):
 
             def curl(*args):
                 try:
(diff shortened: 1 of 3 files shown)


martinpitt:
I manually applied these commits to the checkout in 1-aws-tasks container, and triggered a rhel test. However,  cockpit-tasks resets the checkout to origin/master all the time, so that doesn't help. I'm afraid we just have to land this without a CI run. I manually tested the changes here, and am quite confident in them.

PR #309, opened 2019-12-05
Image refresh for rhel-atomic

cockpituous:
Image refresh for rhel-atomic


martinpitt:
Nice, seems the previous error magically fixed  itself :)

PR #319, opened 2019-12-09
Fix tests explosions on tests-trigger

marusak:
This PR has two fixes that I found while tracing why we trigger so many tests on simple `tests-trigger`.

What is happening is that when we do `tests-trigger`, it creates status, which is picked by webhook, then placed into queue and then interpreted by `run-queue`. Then since it was status it hits [this](https://github.com/cockpit-project/bots/blob/master/run-queue#L59) branch and thus runs command like:
```
./tests-scan sha <sha> --amqp <amqp> --context <context> --repo <repo>
```
Normally we would run it with `-p` (pull request number) but status does not have this information. `tests-scan` when does not have pull request number specified, it scans all pulls in given repo. (well before this PR it was scanning in all repos all pulls even though it was given repo).

The main problem though was that we were ignoring context for external scans - which meant that we did for every triggered context one full scan. (so `tests-trigger c1 c2 c3` would do 3 scans and each would find all valid contexts and add them into queue.)

Now I need to run a whole bunch of tests to see if I broke some other use case.

### DIFF
--- a/tests-scan
+++ b/tests-scan
@@ -37,6 +37,24 @@ except ImportError:
     no_amqp = True
 
 
+def build_policy(repo, context):
+    policy = testmap.tests_for_project(repo)
+    if context:
+        short_contexts = []
+        for c in context:
+            short_contexts.append(c.split("@")[0])
+        new_policy = {}
+        for (branch, contexts) in policy.items():
+            branch_context = []
+            for c in short_contexts:
+                if c in contexts:
+                    branch_context.append(c)
+            if branch_context:
+                new_policy[branch] = branch_context
+        policy = new_policy
+    return policy
+
+
 def main():
     parser = argparse.ArgumentParser(description='Bot: scan and update status of pull requests on GitHub')
     parser.add_argument('-v', '--human-readable', action="store_true", default=False,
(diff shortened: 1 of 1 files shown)


marusak:
I realised that this would effectively disable us from being able to scan for external projects in any other repo than `bots`. But I don't think that there is any use case where we need it?

marusak:
so commit `tests-scan: Use correct repo when scanning externals` is correct, but it breaks the somewhat weird login on how we scan external contexts. I will look if it makes sense to have the old behaviour (if it does not cause  the explosion itself) or we need to rewrite the login of how we scan external contexts.

marusak:
Ok, so I played with this for some time and realized that doing it exactly the same we did before with just "fixing" it was impossible so I rewrote how we validate contexts. Generating of default contexts is the same, it only changes how we check if existing status is valid. Also it is much stricter when context is passed on cmdline. (for how the logic works see commit message).

Coudl you please take a look if you see something that I broke?

tests-scan
```diff
@@ -88,7 +89,7 @@ def main():
 
         # When run from c-p/bots checkout without PR number or PR data we want to scan also all
         # external projects. E.g. `./tests-scan` in c-p/b checkout will scan all projects
-        if opts.repo == "cockpit-project/bots" and not opts.pull_data and not opts.pull_number:
+        if not repo_specified and opts.repo == "cockpit-project/bots" and not opts.pull_data and not opts.pull_number:
```
martinpitt: This is another hack, though. At first sight it's not clear why  this should differ between an implied default and an explicit `--repo cockpit-project/bots`. I thought this mainly needs to differentiate between a tests-scan that got triggered in response to a "PR created" event (in which case we want external tests) and one in response to a "statuses" event (in which case we don't). So please add a `# HACK` at least, so that we can clean this up later on, and add a detailled commit message.

marusak: It indeed is - but it is due to another hack being there - the one where we set up `opts.repo`. If we would not do that, we could just check if `opts.repo` is defined and then check `api.repo`. I'll add there comment explaining it.

marusak: Done

martinpitt: That's not really what I meant. Of course the introduction of `repo_specified` is a shallow hack around changing `opts.repo`. But my concern is conceptual, not syntactical. I meant that running external tests or not should not depend on whether `--repo` is given on the command line or not, but for which GitHub event it is being called.

marusak: So as talked on IRC - this was misunderstanding. And we decided that we don't really need the concept of scanning of external projects and webhook can do it itself. So I changed that commit to drop this part all together.

martinpitt: Thanks, this is much better indeed! @Gundersanne, can you please review again?

PR #314, opened 2019-12-11
Image refresh for rhel-8-1

cockpituous:
Image refresh for rhel-8-1
 * [x] image-refresh rhel-8-1


martinpitt:
Several cockpit-appstream tests [fail](https://209.132.184.41:8493/logs/pull-314-20191212-133131-42a60acd-cockpit-project-cockpit-rhel-8-appstream-rhel-8-1-distropkg/log.html#142) with

    .terminal div:nth-child(1) is ambigous

Apparently the image now has a newer cockpit version, and we need to adjust the tests for that. But as we don't really care about the current version of rhel8-appstream any more (I'll rebase it soon), I ignore this.

lorax/aws is not an image problem either.

PR #323, opened 2019-12-11
Image refresh for rhel-7-7

cockpituous:
Image refresh for rhel-7-7
 * [x] image-refresh rhel-7-7
 * [x] get back rhel-atomic image: #328


martinpitt:
The current rhel-atomic image is not present anywhere any more, it got lost in between the e2e and aws juggling. I triggered a new one in #328

PR #334, opened 2019-12-13
naughty: Close 60: Debian: parted clobbers in-memory state of extended partitions

### LINKED ISSUE #60: Debian: parted clobbers in-memory state of extended partitions
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=842269

See also https://bugzilla.redhat.com/show_bug.cgi?id=1135493 for the original.

This has been fixed with a path in Fedora only, it wasn't fixed upstream.  Thus, debian-unstable suffers from it.

Basically, parted tells the kernel the wrong size for extended partitions.

Apparently, the bug was masked on debian-unstable by systemd-udevd independently re-reading the partition table.  However, recent versions of Storaged block systemd-udevd from re-reading since that causes other bugs.  Thus, this bug is now visible.

cockpituous:
Known issue which has not occurred in 24 days

Debian: parted clobbers in-memory state of extended partitions

Fixes #60

### DIFF
--- a/naughty/debian-stable/60-parted-clobbers-extended-partitions
+++ b/naughty/debian-stable/60-parted-clobbers-extended-partitions
@@ -1,1 +1,0 @@
-self.content_row_wait_in_col(3, 1, "xfs File System")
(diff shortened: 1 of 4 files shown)


cockpituous:
Please comment on the upstream distro bug manually before accepting this pull request.

If you wish to keep this known issue open, then update it

PR #340, opened 2019-12-17
testmap: test cockpit-composer master on rhel-8-2

larskarlitski:
The rhel-8.2 branch will be removed upstream.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -142,11 +142,6 @@ REPO_BRANCH_CONTEXT = {
             'rhel-7-8/firefox',
             'rhel-8-1/chrome',
         ],
-        'rhel-8.2': [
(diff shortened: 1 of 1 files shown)


task/testmap.py
```diff
@@ -141,8 +141,6 @@
             'fedora-31/firefox',
             'rhel-7-8/firefox',
             'rhel-8-1/chrome',
-        ],
-        'rhel-8.2': [
```
martinpitt: Please just remove the section. rhel-8-2 is already being tested on master. We don't need to test all browsers there, pairwise testing across browsers and OSes is enough IMHO.

larskarlitski: Of course. Updated.

Issue #71, opened 2019-04-25
rhel-atomic: SELinux is preventing chronyd from sendto access on the unix_dgram_socket


marusak:
SELinux violations like:
`type=1400 audit(1556197202.146:5): avc:  denied  { sendto } for  pid=2758 comm="chronyd" path="/host/run/chrony/chronyc.5500.sock" scontext=system_u:system_r:chronyd_t:s0 tcontext=system_u:system_r:spc_t:s0 tclass=unix_dgram_socket permissive=0`

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1703192

PR #349, opened 2019-12-18
Image refresh for rhel-atomic

cockpituous:
Image refresh for rhel-atomic


marusak:
New AVC: `type=1400 audit(1576708971.847:5): avc:  denied  { sendto } for  pid=771 comm="chronyd" path="/host/run/chrony/chronyc.3238.sock" scontext=system_u:system_r:chronyd_t:s0 tcontext=system_u:system_r:spc_t:s0 tclass=unix_dgram_socket permissive=`

Probably needs naughty.

martinpitt:
@marusak: We already have that, known issue #71. However, was that a literal copy&paste? Your's is missing a double space between `denied` and `{ send to }`.

However, I just saw that our pattern has a bug: There's a double space between "for" and "pid".

martinpitt:
The double space is indeed in the log, so the pattern is fine. I just piped the test output to `./tests-policy rhel-atomic` and it works.

martinpitt:
[Same issue again](https://209.132.184.41:8493/logs/pull-349-20191219-065050-7b1b7320-cockpit-project-cockpit-rhel-7.8-rhel-atomic/log.html#93-2)

marusak:
I put some logging here to see why this one naughty is not matching anything.
It confirms that we use correct context. Also it confirms that we see all naughties and we read them correctly.

But I noticed pattern - first two runs fail with some element not being present (sos report taking too long) and third run succeeds, but fails with AVC. I checked how we run `tests-policy` and indeed, we do not check naughties on the last retry. I sent PR which fixes that: https://github.com/cockpit-project/cockpit/pull/13345
(see [log1](https://209.132.184.41:8493/logs/pull-349-20191219-065050-7b1b7320-cockpit-project-cockpit-rhel-7.8-rhel-atomic/log.html#93-2) and [log2](https://209.132.184.41:8493/logs/pull-349-20191218-220000-7b1b7320-cockpit-project-cockpit-rhel-7.8-rhel-atomic/log.html#93-2))

Interestingly, chances of this happening are so slim, yet it happened 4 times in row here. Also, since the test succeeds at least one, this PR is good to land. (And we need to backport that test fix into stable branches as well).

martinpitt:
Thanks @marusak ! Indeed this is good enough to land now.

PR #359, opened 2019-12-20
Add manual scenario cloud-cleanup to be used in weldr/lorax

atodorov:
see https://github.com/weldr/lorax/pull/842


martinpitt:
Blocked on questions in https://github.com/weldr/lorax/pull/842#issuecomment-568159457

marusak:
I think it was agree that this is not going to be done through tests, but through tasks.

PR #375, opened 2019-12-27

cockpituous:


martinpitt:
Apparently this fails to boot after package build.

martinpitt:
We have so many broken image refreshes right now, as a first step let's just do another automatic attempt here.

PR #385, opened 2020-01-02
Image refresh for fedora-31

cockpituous:
Image refresh for fedora-31
 * [x] image-refresh fedora-31


marusak:
There is something wrong with this new VM - every test failes to even boot the VM. I try to create new image refresh to see, if this was just hiccup or if it is a real issue

PR #386, opened 2020-01-02
Image refresh for F31

cockpituous:
Previous refresh created unbootable VM, lets see if it was one off.

 * [x] image-refresh fedora-31


marusak:
Ah, I cannot approve as I opened the issue.

PR #372, opened 2019-12-26
Image refresh for centos-8-stream

cockpituous:
Image refresh for centos-8-stream
 * [x] image-refresh centos-8-stream
 * [x] Add expected message: https://github.com/cockpit-project/cockpit/pull/13353

PR #392, opened 2020-01-07
Image refresh for rhel-8-1

cockpituous:
Image refresh for rhel-8-1
 * [x] image-refresh rhel-8-1


marusak:
distropkg has the same broken tests as the last few refreshes.
`rhel-8-1` has hard time with `testCreate` - the same problem we see in https://github.com/cockpit-project/bots/pull/372 - @KKoukiou could you please take a look at both of these refreshes?

martinpitt:
TBH I think we can ignore our rhel-8-appstream branch. We won't update RHEL 8.1 any more, and for 8.2 it seems easier to have one branch and just do baseos and appstream releases from it (if even necessary). In that sense, I think the creation of the appstream branch was a thinko, or failed experiment.

martinpitt:
I sent PR #414 to drop rhel-8-appstream from testmap, let's retire it.

PR #394, opened 2020-01-07
vm-run: Fix networking for windows-10

martinpitt:
Without a bridge (which requires root privileges on the host),
`vm-run windows-10` ended up with broken networking, as Windows does not
get along with the mcast network.

Introduce a new mode to VirtNetwork that uses standard QEMU user
networking, enabled with `isolate="user"`. Use that mode in vm-run for
the windows image if there is no bridge.


martinpitt:
We don't test vm-run automatically, so I'll only run a few test to show that this doesn't break the default mode.

PR #399, opened 2020-01-09
testmap: only test cockpit-composer on fedora-31

jkozol:
Cockpit-composer will begin using osbuild-composer instead of lorax. This change will cause the tests to fail on distros that aren't fedora-31 so these test environments are now only triggered manually. This PR is required for https://github.com/weldr/cockpit-composer/pull/868

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -136,16 +136,12 @@ REPO_BRANCH_CONTEXT = {
     },
     'weldr/cockpit-composer': {
         'master': [
-            'fedora-30/chrome',
-            'rhel-8-2/firefox',
+            'fedora-31/chrome',
             'fedora-31/edge',
             'fedora-31/firefox',
-            'rhel-7-8/firefox',
-            'rhel-8-1/chrome',
         ],
         # These can be triggered manually with bots/tests-trigger
         '_manual': [
-            'centos-8-stream',
         ],
     },
     'candlepin/subscription-manager': {


marusak:
Thanks :)
So the decision was that we are going to release new versions of cockpit-composer to only Fedora-31 and to nothing else for now? (Later to Fedora-32 and RHEL-8-3(2-1?))
Also, if that is the case, I don't think it makes much sense to keep these contexts in manual - who would trigger them when they are broken? Or you would trigger them in some special occasions?
And just a side note for the next time - on PRs like these, it is great to add `no-test` label, so it does not run all tests :)

jkozol:
> So the decision was that we are going to release new versions of cockpit-composer to only Fedora-31 and to nothing else for now? (Later to Fedora-32 and RHEL-8-3(2-1?))

Yes. For now only fedora 31. 

> I don't think it makes much sense to keep these contexts in manual

I will remove the manual tests. Since we will only release on fedora-31 and once cockpit-composer starts using osbuild these tests will fail I agree there won't be any use in running them manually.

> on PRs like these, it is great to add no-test label

I am currently unable to add labels. In the future I will keep that in mind.

marusak:
huh? And not the tests are again here? Is someone triggering these manually from cmdline?

task/testmap.py
```diff
@@ -136,16 +136,12 @@
...
     'weldr/cockpit-composer': {
         'master': [
-            'fedora-30/chrome',
-            'rhel-8-2/firefox',
+            'fedora-31/chrome',
             'fedora-31/edge',
-            'fedora-31/firefox',
-            'rhel-7-8/firefox',
-            'rhel-8-1/chrome',
+            'fedora-31/firefox',
```
marusak: There are trailing spaces an linter is complaining, can you please fix?
(added no-test label, so we can run only what we need on the next push force)

jkozol: Thanks for pointing this out. Should be fixed now

PR #400, opened 2020-01-09
Migrate weldr/lorax rhel8-branch tests to rhel-8-2

jrusz:
I would like to migrate our rhel8-branch tests to rhel-8-2.
Thanks.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -132,6 +132,14 @@ REPO_BRANCH_CONTEXT = {
         # These can be triggered manually with bots/tests-trigger
         '_manual': [
             'centos-8-stream',
+            'rhel-8-2',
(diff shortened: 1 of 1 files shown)


martinpitt:
Good call, thanks! Did you already manually test these? The usual approach for this kind of move is:

 1) send a PR to add them to the `_manual` branch; that's a fake name that won't auto-trigger, but will allow you to run these tests manually with bots/tests-trigger. This can go in with no trouble.
 2) Trigger the new test in some new lorax PR, and sort out issues
 3) Once everything is green, move them from `_manual` to `master` (or whichever branch they should be triggered for)

jrusz:
Thank you for the advice, I changed the commit to reflect that.

PR #404, opened 2020-01-10

cockpituous:

### DIFF
--- a/images/debian-testing
+++ b/images/debian-testing
@@ -1,1 +1,1 @@
-debian-testing-d2f139e04150834b3c1fd4fa352290a8365a50858be1b6dddfe0ecc0f085117d.qcow2
+debian-testing-074a56be43acbbe27003e47e40ce8188e61ff8bb157995b0e3db7a073ca58a83.qcow2


marusak:
Failed 2 times on the same issue. It sets up udp port 194 and expects that we recognize that it is `irc`. But we don't. Seems like it dropped out from `/etc/services`?

marusak:
Hmm, indeed, in the new `/etc/services` there are many services missing. No idea why.
Current debian-stable: [old_etc_services.txt](https://github.com/cockpit-project/bots/files/4044769/old_etc_services.txt)
Debian-stable from this PR: [new_etc_services.txt](https://github.com/cockpit-project/bots/files/4044770/new_etc_services.txt)

martinpitt:
The removal of irc [was intended](https://tracker.debian.org/news/1089559/accepted-netbase-60-source-into-unstable/), so I suppose we should change the test to use a port that is still being used in this millenium?

marusak:
Sent https://github.com/cockpit-project/cockpit/pull/13381 which fixes this.

martinpitt:
Fix landed in cockpit master, re-running.

PR #408, opened 2020-01-10
Image refresh for fedora-30

cockpituous:
Image refresh for fedora-30
 * [x] image-refresh fedora-30


marusak:
Committing in podman complains about command: https://209.132.184.41:8493/logs/pull-408-20200110-135252-889523b3-cockpit-project-cockpit-podman--fedora-30/TestApplication-testBasicSystem-fedora-30-127.0.0.2-2201-FAIL.png

This does not seem like flake. Needs to be looked at.

marusak:
Reported https://github.com/containers/libpod/issues/4847
I'll change it to simple command in podman.

martinpitt:
> I'll change it to simple command in podman.

For now this looks like an actual regression, so a naughty may be more appropriate? If upstream says that they don't support command lines any more, we can still do that.

marusak:
> For now this looks like an actual regression, so a naughty may be more appropriate? If upstream says that they don't support command lines any more, we can still do that.

Yeah, but that would mean we would not test any committing and also would skip another parts of test.
I thought that HACK in the test would be better (it is less important for us to test if we can pass command with argument in commit than testing the whole committing machinery).

martinpitt:
Thanks for the podman fix, landed. Retrying tests.

PR #416, opened 2020-01-13
images: Introduce rhel-8-2-distropkg

martinpitt:
Declare it as manual test for now, until cockpit's tests get adjusted
for this new context.


martinpitt:
I'm preparing a cockpit branch for this new scenario, but before I'd actually like to land https://github.com/cockpit-project/bots/pull/415 (rhel-8-2 image refresh) -- that will pull in cockpit 210, and thus there will be much fewer special cases.

PR #429, opened 2020-01-17
Image refresh for debian-stable

cockpituous:
Image refresh for debian-stable


martinpitt:
I've seen several debian-stable tests like [this one](https://209.132.184.41:8493/logs/pull-13222-20200117-144747-61ab910a-cockpit-project-cockpit--debian-stable/log.html) which just keep timing out. Something is broken here.

PR #439, opened 2020-01-21
Image refresh for rhel-8-2

cockpituous:
Image refresh for rhel-8-2
 * [x] image-refresh rhel-8-2


marusak:
`sshd` is dumping core. This needs to be reported and then naughtyfied

martinpitt:
Let's build a fresh one, this is getting a bit old.

PR #440, opened 2020-01-21
tests-invoke: Continuous collision detection

Gundersanne:


tests-invoke
```diff
@@ -252,8 +256,21 @@ class PullTask(object):
             cmd = ["timeout", "120m", entry_point]
             if opts.verbose:
                 cmd.append("--verbose")
-            ret = subprocess.call(cmd)
-
+            # Popen instead of call; while not etc... (SubProcess.poll)
```
martinpitt: Please drop this comment. Or better even, say "# cancel test if PR gets updated while test is running"

tests-invoke
```diff
@@ -252,8 +256,21 @@ class PullTask(object):
...
+            if ret:
```
martinpitt: This looks out of place -- this is the "then" branch of `if not ret:`, and it's not reassigned anywhere in between. Probably this whole if is just an overzealous copy&paste?

tests-invoke
```diff
@@ -252,8 +256,21 @@ class PullTask(object):
...
+                sys.stderr.write('Collision detected: {0}\n'.format(ret))
+                return None
+            p = subprocess.Popen(cmd)
+            while p.poll() is None:
+                ret = self.detect_collisions(opts, True)
+                if ret:
```
martinpitt: No need to use the `ret` variable here.

tests-invoke
```diff
@@ -252,8 +256,21 @@ class PullTask(object):
...
+                    sys.stderr.write('Collision detected: {0}\n'.format(ret))
```
martinpitt: Let's please make this a different message. "PR got updated, cancelling this test run."?

tests-invoke
```diff
@@ -252,8 +256,21 @@ class PullTask(object):
...
+                    ret = None
+                    p.terminate()
```
martinpitt: Do a `break` here, or do the sleeping in `else:`. No need to sleep another minute after cancelling.

Also, does the child get reaped automatically, or do we need a `p.wait()` here?

tests-invoke
```diff
@@ -252,8 +256,21 @@ class PullTask(object):
...
+                time.sleep(60)
+            ret = p.returncode
+            # if collision
+            sys.stderr.write("RETURN CODE: {0}", p.returncode)
```
martinpitt: It's useful to log the return code, but the comment doesn't fit.

tests-invoke
Gundersanne: Oh I chose this because it's  similar to the comment on line 225, but I'll add the fact that it's cancelling a running test?

PR #441, opened 2020-01-21
tests-scan: Don't trigger external tests on bots PRs by default

martinpitt:
In practice, very few changes to bots demand testing all external
projects, so that pretty much all PRs are filed with "no-test". It's
more practical for now if a human decides which tests need to run.


marusak:
Could you please use no-test label the next time :)

PR #447, opened 2020-01-23
vm-run: add commandline flags to execute commands

allisonkarlitskaya:
Add --execute and --systemd-start commandline arguments to execute
arbitrary commands or start systemd unit files.  This is pretty useful
for testing repeated incremental changes to an image.
    
These arguments imply --quiet.


Gundersanne:
tests are still failing though :cow:

PR #465, opened 2020-01-28
Add script for creating rel-eng based RHEL images

jikortus:
The existing bots scripts target at creating RHEL images from nightly composes, however, for composer testing we also need rel-eng-based images. The script just makes use of the already existing setup/bootstrap scripts and basically just performs a few URL/repo names replacements.

This is mostly a composer-related script (and it's a bit of a hack :), so if you think it should rather go directly to the lorax repo, I'm fine with it. Otherwise I think it best suits to be located next to the other image handling scripts, which is in the bots repo :).

### DIFF
--- a/image-create-rhel-compose
+++ b/image-create-rhel-compose
@@ -1,0 +1,46 @@
+#!/bin/bash
+set -eu
+
+COMPOSE=${COMPOSE:-""}
+EXTRAS_COMPOSE=${EXTRAS_COMPOSE:-""}
+
+if [ -z "$COMPOSE" ]; then
+    echo "Unspecified compose (use COMPOSE env variable)."
+    exit 1
+fi
+
+# RHEL-x.y(.z)-YYYYMMDD.b -> rhel-x-y
+final_image=${COMPOSE,,}
+final_image=${final_image//./-}
+image=${final_image:0:8}
+
+if [[ "$COMPOSE" =~ RHEL-7 ]]; then
+    if [[ -z "$EXTRAS_COMPOSE" ]]; then
+        EXTRAS_COMPOSE="latest-EXTRAS-7-RHEL-7"
+        echo "EXTRAS_COMPOSE not supplied, using '$EXTRAS_COMPOSE'."
+    fi
+    export SERVER_REPO_URL="http://download.devel.redhat.com/rel-eng/rhel-7/RHEL-7/$COMPOSE/compose/Server/x86_64/os"
+    export EXTRAS_REPO_URL="http://download.devel.redhat.com/rel-eng/rhel-7/EXTRAS-7/$EXTRAS_COMPOSE/compose/Server/x86_64/os"
+    install_url=$SERVER_REPO_URL
+else
+    export BASEOS_REPO_URL="http://download.devel.redhat.com/rel-eng/rhel-8/RHEL-8/$COMPOSE/compose/BaseOS/x86_64/os"
+    export APPSTREAM_REPO_URL="http://download.devel.redhat.com/rel-eng/rhel-8/RHEL-8/$COMPOSE/compose/AppStream/x86_64/os"
+    install_url=$BASEOS_REPO_URL
+fi
+
+bootstrap_script="$(dirname $0)/images/scripts/$image.bootstrap"
+mv $bootstrap_script ${bootstrap_script}.orig
+
+echo "#!/bin/bash
+set -eux
+
+BASE=\$(dirname \$0)
+\$BASE/virt-install-fedora \"\$1\" x86_64 \"$install_url\"
+" > $bootstrap_script
+chmod +x $bootstrap_script
+
+pushd $(dirname $0)
+./image-create $image $@
+mv ${bootstrap_script}.orig $bootstrap_script
+mv images/$image images/$final_image
+popd
--- a/images/scripts/rhel.setup
+++ b/images/scripts/rhel.setup
@@ -35,7 +35,43 @@ set -x
 if [ ! -f "$SKIP_REPO_FLAG" ]; then
     # Configure repositories.
 
-    if [ "$IMAGE" = "rhel-7-7" ]; then
+    if [ -n "$SERVER_REPO_URL" -a -n "$EXTRAS_REPO_URL" ]; then
+    # disable all default repos
+    rm -f --verbose /etc/yum.repos.d/*.repo
+cat <<EOF > /etc/yum.repos.d/rel-eng.repo
+[RHEL-7]
+name=rhel7-server
+baseurl=$SERVER_REPO_URL
+enabled=1
+gpgcheck=0
+
+[EXTRAS-7]
+name=rhel7-extras
+baseurl=$EXTRAS_REPO_URL
+enabled=1
+gpgcheck=0
+EOF
+        $YUM_INSTALL yum-utils
+
+    elif [ -n "$BASEOS_REPO_URL" -a -n "$APPSTREAM_REPO_URL" ]; then
+    # disable all default repos
+    rm -f --verbose /etc/yum.repos.d/*.repo
+cat <<EOF > /etc/yum.repos.d/rel-eng.repo
+[RHEL-8-BaseOS]
+name=rhel8-baseos
+baseurl=$BASEOS_REPO_URL
+enabled=1
+gpgcheck=0
+
+[RHEL-8-AppStream]
+name=rhel8-appstream
+baseurl=$APPSTREAM_REPO_URL
+enabled=1
+gpgcheck=0
+EOF
+        $YUM_INSTALL yum-utils
+
+    elif [ "$IMAGE" = "rhel-7-7" ]; then
         # disable all default repos
         rm -f --verbose /etc/yum.repos.d/*.repo
 cat <<EOF > /etc/yum.repos.d/internal.repo
--- a/image-create
+++ b/image-create
@@ -90,6 +90,10 @@ class MachineBuilder:
             env = {
                 "TEST_OS": self.machine.image,
                 "DO_BUILD": "0" if args.no_build else "1",
+                "SERVER_REPO_URL": os.environ.get("SERVER_REPO_URL", ""),
+                "EXTRAS_REPO_URL": os.environ.get("EXTRAS_REPO_URL", ""),
+                "BASEOS_REPO_URL": os.environ.get("BASEOS_REPO_URL", ""),
+                "APPSTREAM_REPO_URL": os.environ.get("APPSTREAM_REPO_URL", ""),
             }
             self.machine.message("run setup script on guest")
 


jikortus:
Martin,
thanks for the feedback!


> Ugh, this is indeed quite a big hack :-) Let's please not heuristically sed our own code files, it will break sooner or later.

That's true, I guess the most appropriate solution would be to either have scripts that would take into account the possibility of using rel-eng repos or at least provide some placeholders that could be easily replaced in a deterministic way. That would, however, require some non-trivial effort from your side and probably wouldn't be in line with what's needed in Cockpit QE, so naturally I'd like like to avoid that, as, at this point, the need for rel-eng based images is only a matter of composer testing.

> `image-create` is very generic, you shouldn't need to change any of it.
I didn't actually touch it, my intention was just to create custom setup and bootstrap scripts and then feed them to `image-create` to do the rest of the magic.

> Instead, to define a new image like "rhel-7-8-releng" (or -official, or whatnot):
> 
>     * create a new images/scripts/rhel-7-8-releng.bootstrap script by copying rhel-7-8.bootstrap and adjusting the URL, and
That's basically what the script does (well, it recreates the bootstrap script since it looked like a cleaner solution to me, but I can easily change that for a substitution from the original one).

> 
>     * create a new rhel-8-2-releng.setup → rhel.setup symlink and adjust rhel.setup for this new special case (particularly, repo URLs). If you don't need all the cockpit testing stuff in that image, you can also just copy rhel.setup and adjust it to your liking. Probably you still want to share it between your 7.8 and 8.2 variants, i.e . do a rhel-releng.setup and symlink it to -7-8 and -8-2 and such.

Maybe I'm missing some important part here, but I think it's better to copy the original script (let's say, `rhel-7-8.setup`) and modify it so that it contains URLs (and some other necessary changes, e. g. some internal logic that just takes into account existence of `RHEL-x.y` nightlies or older released composes) for a particular compose I intend to test.

If I created just `rhel-7-8-releng` or something along those lines, I'd just be able to to use it for one particular compose (and this changes several times during the testing cycle, so I'd have to update the script manually, which I wanted to avoid by using the script). Other possible outcome of such a solution would be that I'd have to put up with using `-latest` rel-eng compose, which might work most of the times, but wouldn't be as universal as an option to specify a particular compose.

However, maybe each of us talks about some slightly different point of view or use case, so I'd gladly shed some more light on the matter if something I wrote is not clear enough :-).

> 
> 
> I suppose this wouldn't go into image-trigger's `REFRESH`, i. e. this wouldn't live on our image servers for now, and just be a local development tool? (it can be added later on if you need it often)

No, not really, at least not at this moment. Right now it was intended to make our manual testing somewhat easier. We'd appreciate to have images based on particular rel-eng composes available in the future, but I think it's out of scope of this PR :-).

martinpitt:
> scripts that would take into account the possibility of using rel-eng repos

That's what I actually meant. rhel.setup already has an image name → repo mapping, it's easy to add the releng ones there.

> I think it's better to copy the original script (let's say, rhel-7-8.setup) 

Your choice, I guess. It depends on what you want the final image to be. If it's similar to our rhel-7-8 nightly image, i. e. you want the same dependencies, mock etc., then just add the new repo to rhel.setup, it's a robust and simple patch. If OTOH you don't need build capabilities, don't need most of the installed packages, and want to evolve this independently from our "official" 7.8 image, then copying is better.

> We'd appreciate to have images based on particular rel-eng composes

Indeed right now rhel.setup hardcodes "latest". It could be extended to check some env variables to specify a compose, and image-create could grow a -e/--env multi-option to pass extra env variables to the setup script (`run_setup_script()` already passes an env dict, it could be extended there).

jikortus:
OK, thanks a lot for the additional information! If I understand it correctly, an acceptable solution would be to

- patch rhel.setup so that it accepts also rel-eng images (would it also be OK to implement a logic that would identify the compose version right from the supplied image name, as I do it in the helper script in this PR?),
- minimize the helper script to just create a symlink to `rhel.setup` for the particular rel-eng compose,  create a particular `.bootstrap` script and fire `image-create`,
- optionally patch the env variables in `run_setup_script()` (depending on how I'd pass the information about the rel-eng compose to the setup script)?

> > I think it's better to copy the original script (let's say, rhel-7-8.setup)
> 
> Your choice, I guess. It depends on what you want the final image to be. If it's similar to our rhel-7-8 nightly image, i. e. you want the same dependencies, mock etc., then just add the new repo to rhel.setup, it's a robust and simple patch.

I think I could strip `rhel.setup` of some parts that aren't necessary for composer testing or change some of the logic handling the repos altogether, but I'm not convinced this would be a good approach. That way, there would be a need to maintain two distinct scripts, which would either duplicate some work for you (which doesn't make much sense, as this would be a change induced just by a composer need) or compel us to watch for changes in the original `rhel.setup` and port them to the composer version of the setup file. Other reason for not using a separate setup file is that the extra packages not necessary for composer don't interfere with composer testing anyway. 

The best option would be, I guess, to have two setup files sharing the common parts, but that would be quite a complicated task to do, given the current setup file wasn't created with an aim for something like that and therefore doesn't have a structure that would favor such changes.

martinpitt:
> OK, thanks a lot for the additional information! If I understand it correctly, an acceptable solution would be to
> 
>     * patch rhel.setup so that it accepts also rel-eng images

That seems fine. As I said, it depends on what you want these images to be -- a plain minimal compose, or something that is close to the usual nightly cockpit images, except based on a different compose than nightly. In the former case it may make sense to implement your own .setup instead of re-using rhel.setup. But my gut feeling is that modifying rhel.setup is easier.

> (would it also be OK to implement a logic that would identify the compose version right from the supplied image name, as I do it in the helper script in this PR?),

That's a little harder to do right now as image-create looks for `images/scripts/<imagename>.{bootstrap,setup}`, so you'd need to teach it some tricks to ignore a special version suffix. It seems a bit easier to me to pass on a a command line option as environment variable.

>     * minimize the helper script to just create a symlink to `rhel.setup` for the particular rel-eng compose,  create a particular `.bootstrap` script and fire `image-create`,

That would be one way to implement the above, but what would that achieve that you couldn't do with a fixed static name?

>     * optionally patch the env variables in `run_setup_script()` (depending on how I'd pass the information about the rel-eng compose to the setup script)?

I think that's the cleanest and easiest way to parameterize the .setup and .bootstrap scripts, yes. 

> The best option would be, I guess, to have two setup files sharing the common parts, but that would be quite a complicated task to do, given the current setup file wasn't created with an aim for something like that and therefore doesn't have a structure that would favor such changes.

We do something like that for the .install scripts (they are not used for composer, just for cockpit, but nevertheless..) E. g. [debian-testing.install](https://github.com/cockpit-project/bots/tree/master/images/scripts/debian-testing.install) is not a symlink to the generic debian.install, but instead just calls it and then does some extra bits on top of it. Now, this wouldn't really help you with the current rhel.setup, as selecting the repos is pretty central to it -- hence I think it's easiest to parameterize that instead?

martinpitt:
> OK, thanks a lot for the additional information! If I understand it correctly, an acceptable solution would be to
> 
>     * patch rhel.setup so that it accepts also rel-eng images

That seems fine. As I said, it depends on what you want these images to be -- a plain minimal compose, or something that is close to the usual nightly cockpit images, except based on a different compose than nightly. In the former case it may make sense to implement your own .setup instead of re-using rhel.setup. But my gut feeling is that modifying rhel.setup is easier.

> (would it also be OK to implement a logic that would identify the compose version right from the supplied image name, as I do it in the helper script in this PR?),

That's a little harder to do right now as image-create looks for `images/scripts/<imagename>.{bootstrap,setup}`, so you'd need to teach it some tricks to ignore a special version suffix. It seems a bit easier to me to pass on a a command line option as environment variable.

>     * minimize the helper script to just create a symlink to `rhel.setup` for the particular rel-eng compose,  create a particular `.bootstrap` script and fire `image-create`,

That would be one way to implement the above, but what would that achieve that you couldn't do with a fixed static name?

>     * optionally patch the env variables in `run_setup_script()` (depending on how I'd pass the information about the rel-eng compose to the setup script)?

I think that's the cleanest and easiest way to parameterize the .setup and .bootstrap scripts, yes. 

> The best option would be, I guess, to have two setup files sharing the common parts, but that would be quite a complicated task to do, given the current setup file wasn't created with an aim for something like that and therefore doesn't have a structure that would favor such changes.

We do something like that for the .install scripts (they are not used for composer, just for cockpit, but nevertheless..) E. g. [debian-testing.install](https://github.com/cockpit-project/bots/tree/master/images/scripts/debian-testing.install) is not a symlink to the generic debian.install, but instead just calls it and then does some extra bits on top of it. Now, this wouldn't really help you with the current rhel.setup, as selecting the repos is pretty central to it -- hence I think it's easiest to parameterize that instead?

jikortus:
Martin, thanks again for your comment. In the end it still seems to me that using a modified `rhel.setup` to accept rel-eng composes, together with using images with rel-eng-based names should work fine -- I'd still like to have the images with the particular compose in the name, so that I knew at first glance what they were built from.

Please have a look (the commits aren't necessarily in the final shape) and tell me if it could work this way for you. If you think this is still too disruptive compared to the aforementioned approach, I'd implement just building `rhel-x-y`-named images using the env variables, and perform the renaming outside, using the `image-create-custom` script. Thank you.

image-create-custom
```diff
@@ -0,0 +1,82 @@
...
+parse_args() {
+    for arg in $@; do
+            if [[ "$arg" =~ ^rhel(-[[:digit:]]){2}(-([[:digit:]]{8}-[[:digit:]]|latest))$ ]]; then
+                image="$arg"
+                break
+            fi
+    done
+
+    if [ ! "$image" ]; then
+        echo "Missing or unsupported image (rhel-x-y-YYYYMMDD-B or rhel-x-y-latest)."
```
martinpitt: Is it really wise to encode the assumptions about the name format here? This is quite fungible, it seems easier to just accept whatever the user specifies and then fail later on with a 404 not found.

image-create-custom
```diff
@@ -0,0 +1,82 @@
...
+#    echo "Image $image already exists, use '-r' to recreate."
+#    exit 1
+#fi
+
+# determine download server URL prefix (also based on machine location)
+if [[ "$HOSTNAME" =~ brq ]]; then
+    url_prefix="http://download.eng.brq.redhat.com"
+else
+    url_prefix="http://download.eng.bos.redhat.com"
+fi
```
martinpitt: Use download.devel.redhat.com for automatic location based mirror selection.

image-create-custom
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
martinpitt: `set  -eu` please.

image-create-custom
```diff
@@ -0,0 +1,82 @@
...
+compose=${compose//-/.}
+
+# determine image type and set compose ID accordingly
+if [[ "$image" =~ ^rhel-8-.-.{8}-.?$ ]]; then
+    compose="RHEL-$major.$minor.0-$compose"
+elif [[ "$image" =~ ^rhel-7-.-.{8}-.?$ ]]; then
+    compose="RHEL-$major.$minor-$compose"
+else
+    compose="latest-RHEL-$major.$minor"
+fi
```
martinpitt: This also seems brittle -- maybe rather just let the user specify the compose ID, and name the image after that?

image-create-custom
```diff
@@ -0,0 +1,82 @@
...
+    if [ ! "$image" ]; then
+        echo "Missing or unsupported image (rhel-x-y-YYYYMMDD-B or rhel-x-y-latest)."
+        exit 1
+    else
+        echo "Using image '$image'"
+    fi
+}
+
+create_bootstrap() {
+    echo "#!/bin/bash
```
martinpitt: `-eux` please.

image-create-custom
```diff
@@ -0,0 +1,82 @@
...
+
+BASE=\$(dirname \$0)
+\$BASE/virt-install-fedora \"\$1\" x86_64 \"$install_url\"" > $bootstrap_script
```
martinpitt: For that the script should  additionally do `set -o noclobber`.

image-create-custom
```diff
@@ -0,0 +1,82 @@
...
+    url_prefix="http://download.eng.bos.redhat.com"
+fi
+url_prefix+="/rhel-$major/rel-eng/RHEL-$major/$compose/compose"
+
+# determine URL for image installation
+if [[ $major == 7 ]]; then
+    install_url="$url_prefix/Server/x86_64/os/"
+else
+    install_url="$url_prefix/BaseOS/x86_64/os/"
+fi
```
martinpitt: suggestion: If you still retain a case switch between 7 and 8 above, set some variable in it for the the "Server" vs. "BaseOS" path part, and use it here.

image-create-custom
```diff
@@ -0,0 +1,82 @@
...
+
+# create .bootstrap script and .setup and .install links for the particular compose
+# for the required rel-eng compose
+bootstrap_script="$(dirname $0)/images/scripts/$image.bootstrap"
+setup_script="$(dirname $0)/images/scripts/$image.setup"
+install_script="$(dirname $0)/images/scripts/$image.install"
```
martinpitt: Let's leave out the .install script for now. This is only relevant for cockpit, and we don't currently plan on testing on such images.

image-create-custom
```diff
@@ -0,0 +1,82 @@
...
+if [ -L "$setup_script" ]; then
+    rm -f $setup_script
+fi
+if [ -L "$install_script" ]; then
+    rm -f $install_script
+fi
```
martinpitt: Use `ln -sfn` instead of this explicit removal, keeps things simpler.

images/scripts/rhel.setup
```diff
@@ -32,6 +32,31 @@ SKIP_REPO_FLAG="/root/.skip_repos"
...
+    minor=${IMAGE:7:1}
+    compose_build=${IMAGE:9}
+    compose_build=${compose_build/-/.}
+    if [[ "$IMAGE" =~ ^rhel-8-.-.{8}-.?$ ]]; then
+        compose="RHEL-$major.$minor.0-$compose_build"
+    elif [[ "$IMAGE" =~ ^rhel-7-.-.{8}-.?$ ]]; then
+        compose="RHEL-$major.$minor-$compose_build"
+    else
+        compose="latest-RHEL-$major.$minor"
+    fi
```
martinpitt: Please no, this is too much and brittle string manipulation hackery. Please instead pass the repo URLs (which your other script already computed) as environment variables, and change image-create's `run_setup_script` to add them to the `env` variable.

image-create-custom
jikortus: I think the format is pretty much stable, however I simplified the whole process, including removal of the check.

image-create-custom
jikortus: Thanks for pointing this out, I didn't know that.

image-create-custom
jikortus: I changed the script to take a `COMPOSE` (and optionally `EXTRAS_COMPOSE` for RHEL-7) environment variable, and use it to determine the repo URLs and the resulting image name.

image-create-custom
jikortus: OK, good point.

image-create-custom
jikortus: Now I'm using 4 environment variables for passing the URLs (AppStream + BaseOS, or base + Extras) determined from the `COMPOSE`/`EXTRAS_COMPOSE` variables, I hope this is fine. If not, I could simplify it to just pass an URL prefix (or two, including extras repo prefix for RHEL-7).

image-create-custom
jikortus: OK, I saw some failures before without modifications of the .install script, so I included it to the whole process as well.

image-create-custom
jikortus: Thanks for the suggestion!

images/scripts/rhel.setup
jikortus: I simplified the script and did what you suggested in this comment, hopefully this is (close enough to) how you'd expect the script (and modifications to the existing tooling) to look like.

images/scripts/rhel.setup
```diff
@@ -34,8 +34,39 @@ set -x
...
+enabled=1
+gpgcheck=0
+
+[EXTRAS-7]
+name=rhel7-extras
+baseurl=$EXTRAS_REPO_URL
+enabled=1
+gpgcheck=0
+EOF
+        $YUM_INSTALL yum-utils
```
martinpitt: The "regular" rhel-7 cases do the `rm -f --verbose /etc/yum.repos.d/*.repo` thing, to avoid having multiple repos that might collide with the specified compose. I think you want this as well.

image-create-rhel-compose
```diff
@@ -0,0 +1,46 @@
...
+set -eux
+
+BASE=\$(dirname \$0)
+\$BASE/virt-install-fedora \"\$1\" x86_64 \"$install_url\"
+" > $bootstrap_script
+chmod +x $bootstrap_script
+
+pushd $(dirname $0)
+./image-create $image $@
+mv ${bootstrap_script}.orig $bootstrap_script
```
martinpitt: This looks a little dangerous -- if  `$0` is a relative path, then it will be part of `$bootstrap_script`, and after the `pushd` the path will not be valid any more. Make it an absolute path, or drop the `pushd` altogether.

images/scripts/rhel.setup
```diff
@@ -34,8 +34,39 @@ set -x
 
 if [ ! -f "$SKIP_REPO_FLAG" ]; then
     # Configure repositories.
+    if [[ "$SERVER_REPO_URL" && "$EXTRAS_REPO_URL" ]]; then
```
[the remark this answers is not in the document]
jikortus: OK, fixed. Should I also use the POSIX syntax in `image-create-rhel-compose` to keep it consistent with other scripts?

images/scripts/rhel.setup
jikortus: Good point, fixed :). Thanks.

image-create-rhel-compose
jikortus: I'm not sure I can see what can go wrong here? `$0` in `$bootstrap_script` gets expanded when `$bootstrap_script` is executed (not when it gets echoed to it, as it's escaped in `image-create-rhel-compose`), so I think it should work fine.

jikortus:
@martinpitt Thanks for all the suggestions! I fixed two of the problematic spots, but the third one is unclear to me (looks fine as it is to me).

images/scripts/rhel.setup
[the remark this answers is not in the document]
martinpitt: Please use `[ -n "$var" ]`; just `[ "$var" ]` is unsafe in general. Even though the values should always start with a letter, let's be correct just in case. If you want to POSIXify image-create-rhel-compose, that's fine for me, but it's less important.

image-create-rhel-compose
martinpitt: Ah of course, I read it wrong (I missed that it was all part of that long string).

martinpitt:
Thanks @jikortus ! Just one tiny issue left. I leave it up to you whether you want to make image-create-rhel-compose  POSIX shell or keep it at bash -- it's pretty much "your" script/tool anyway.

images/scripts/rhel.setup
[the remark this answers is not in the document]
jikortus: Ah, of course. Fixed :).

jikortus:
Thanks a lot @martinpitt ! Hopefully all is fixed now, I decided to leave the `image-create-rhel-compose` as it is at least for now if you're OK with that.

martinpitt:
@jikortus : Are you still actually using that script? It has never been updated for RHEL 9, and we don't use it in our CI ourselves. I'm asking because it still uses virt-install-fedora. What's the point of this, anyway? Each compose comes along with a cloud image, which are convenient to use and a *lot* faster to get.

jikortus:
@martinpitt: It must be quite a long time since I used it for the last time, and I _guess_ noone else is using it either, since it's outdated, as you mentioned.

PR #467, opened 2020-01-28
testmap: Add manual trigger for new ci test

jrusz:
This is just a temporary trigger with placeholder name that will likely change.

This is for https://github.com/weldr/lorax/pull/960


martinpitt:
@jrusz FYI, no need to do the no-test dance any more.

PR #470, opened 2020-01-29
 github.py: Query issues based on their modifications time

marusak:
`github.py: Query issues based on their status`

This commit should be safe. Honestly I don't see why we would want to always query all issues and then filter them. Is there some good reason for this?

`github.py: Query issues since they were last modified`

Now it also would return issues that were only commented at in the given timeframe. Before it was querying only issues when they were opened/closed. Could this cause some problems? (I need to query issues based on their modification time and not when they were opened/closed)

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -329,24 +329,14 @@ class GitHub(object):
         result = []
         page = 1
         count = 100
-        opened = True
         label = ",".join(labels)
-        while count == 100 and opened:
-            req = "issues?labels={0}&state=all&page={1}&per_page={2}".format(label, page, count)
+        while count == 100:
+            req = "issues?labels={0}&state={1}&page={2}&per_page={3}".format(label, state, page, count)
             issues = self.get(req)
             count = 0
             page += 1
-            opened = False
             for issue in issues:
                 count += 1
-
-                # On each loop of 100 issues we must encounter at least 1 open issue
-                if issue["state"] == "open":
-                    opened = True
-
-                # Make sure the state matches
-                if state != "all" and issue["state"] != state:
-                    continue
 
                 # Check that the issues are past the expected date
                 if since:
--- a/task/github.py
+++ b/task/github.py
@@ -323,31 +323,24 @@ class GitHub(object):
                 count += 1
         return result
 
-    # The since argument is seconds since the issue was either
-    # created (for open issues) or closed (for closed issues)
+    # The since argument is seconds since the issue was last time modified
     def issues(self, labels=["bot"], state="open", since=None):
         result = []
         page = 1
         count = 100
         label = ",".join(labels)
+        if since:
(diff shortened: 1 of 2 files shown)


marusak:
Breaks tests. I'll take a look at it if we agree this is a good change

task/github.py
```diff
@@ -329,25 +329,15 @@ def issues(self, labels=["bot"], state="open", since=None):
         result = []
         page = 1
         count = 100
-        opened = True
         label = ",".join(labels)
-        while count == 100 and opened:
-            req = "issues?labels={0}&state=all&page={1}&per_page={2}".format(label, page, count)
+        while count == 100:
+            req = "issues?labels={0}&state={1}&page={2}&per_page={3}".format(label, state, page, count)
```
martinpitt: This certainly makes sense to leave the state filtering to the github server side. Maybe that was some workaround for traversing through closed states with the paging, but hard to tell -- if that works now, let's by all means use it.

task/github.py
```diff
@@ -323,31 +323,24 @@ def pulls(self, state='open', since=None):
...
-    # The since argument is seconds since the issue was either
-    # created (for open issues) or closed (for closed issues)
+    # The since argument is seconds since the issue was last time modified
     def issues(self, labels=["bot"], state="open", since=None):
         result = []
         page = 1
         count = 100
         label = ",".join(labels)
+        if since:
+            since = "&since={0}".format(time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime(since)))
```
martinpitt: Does since= subsume the created_at/closed_at properties? I. e. these count as "modification" just like posting a comment or changing another issue attribute?

marusak: yes, created_at and closes_at both count as the last modifications. You can see when issue was updated for example at  https://github.com/cockpit-project/cockpit/issues?q=is%3Aissue+sort%3Aupdated-desc

marusak:
Tests fixed

task/test-github
```diff
@@ -73,6 +73,8 @@ def mockServer():
                 issues_ = issues
                 if "state=open" in self.path:
                     issues_ = [i for i in issues_ if i["state"] == "open"]
+                if "since=" in self.path:
+                    issues_ = [i for i in issues_ if "created_at" not in i.keys() and "closed_at" not in i.keys()]
```
martinpitt: That's a little weird, but as we don't track the modification time, good enough for now.

marusak: it is super weird, but that is pretty much the same as it worked until now... But these tests could be greatly improved.

martinpitt:
> Could this cause some problems?

The main thing is that we would now be able to delay an image refresh or something by responding to a previously done one. But that seems rare, so I wouldn't worry about it for now.

PR #488, opened 2020-02-03
images: Don't kill mock after build

martinpitt:


allisonkarlitskaya:
Do we need to do similar on rhel?

martinpitt:
> Do we need to do similar on rhel?

RHEL images also use lib/fedora.install.

PR #493, opened 2020-02-04
test-scan: Unquote contexts

marusak:
When there is '#' in context, run-queue does `pipes.quote(context)` and
wraps the context in pair of `' '`.

TBH. I am not happy with this fix, but nothing better came to my mind quickly. Any better approach is welcomed.

### DIFF
--- a/run-queue
+++ b/run-queue
@@ -26,7 +26,6 @@ import subprocess
 import sys
 import time
 import logging
-import pipes
 
 from task import redhat_network, distributed_queue
 
@@ -59,7 +58,7 @@ def consume_webhook_queue(channel, amqp):
     elif event == 'status':
         repo = request['repository']['full_name']
         sha = request['sha']
-        context = pipes.quote(request['context'])
+        context = request['context']
         cmd = ['./tests-scan', '--sha', sha, '--amqp', amqp, '--context', context, '--repo', repo]
     elif event == 'issues':
         action = request['action']


martinpitt:
The original idea of using `pipes.quote` was that they would get unquoted by the shell in run-queue's main() here: `subprocess.Popen(cmd, shell=type(cmd) is str)`. But I have a feeling that this isn't what happens -- do we perhaps store the tests-scan command as an actual array instead of a string? Then the consuming code would just run the array (it supports array and string mode), and in that case the quoting is indeed unnecessary (and harmful), and we should just drop that instead.

marusak:
This is pretty much what is happening in run-queue:
```
import pipes                                                                    
import subprocess                                                               
                                                                                  
context = pipes.quote("rhel-atomic@bots#491")                                   
# Uncomment the next line for it to work
# context = "rhel-atomic@bots#491"                                                
cmd = ['./tests-scan', '--context', context, '--repo', "cockpit-project/cockpit"]
proc = subprocess.Popen(cmd, stdout = subprocess.PIPE, stderr = subprocess.PIPE )                                        
                                                                     
stdout, stderr = proc.communicate()                                             
print(stdout)                                                                   
print(stderr)
```

> do we perhaps store the tests-scan command as an actual array instead of a string

Yes we do. That is the usual way of doing things.

> and in that case the quoting is indeed unnecessary (and harmful), and we should just drop that instead.

Ah right, `shell=False` should be safe, right?

marusak:
> I went through the code and discussed it with Sanne, and indeed for the webhook events all commands are argv arrays, and at no point in time they get converted to a string. So the correct fix is to drop the pipes.quote() from consume_webhook_queue() instead.

Indeed. This makes so much more sense! Thanks :)
I've updated the code.

PR #497, opened 2020-02-05
Image refresh for fedora-31

cockpituous:
Image refresh for fedora-31
 * [x] image-refresh fedora-31


marusak:
I need to refresh F31 in https://github.com/cockpit-project/bots/pull/501 so if that succeeds, this can be closed

PR #500, opened 2020-02-06
testmap: Define manual triggers for fedora-31

marusak:
We want to switch from Fedora-30 to Fedora-31 which means we need to
test these changes. In order to be able to do that, we need to define
manual contexts.


marusak:
I checked all F30 triggers and created F31 variants that do not exist yet.

PR #501, opened 2020-02-06
Move (most of) tests from F30 to F31

marusak:
* [x] image-refresh fedora-31
 * [x] image-refresh fedora-testing

- [x] Make better use of OS_DEFAULT in Cockpit - https://github.com/cockpit-project/cockpit/pull/13497
- [x] Drop selenium docker - https://github.com/cockpit-project/cockpit/pull/13504
- [x] Drop selenium docker 7.8 - https://github.com/cockpit-project/cockpit/pull/13510
- [x] Drop selenium docker 8.2 - https://github.com/cockpit-project/cockpit/pull/13511
- [x] Update storage message - https://github.com/cockpit-project/cockpit/pull/13512
- [x] Fix windows going to sleep after 30min - https://github.com/cockpit-project/bots/pull/522
- [x] Wait default time for machine to reboot - https://github.com/cockpit-project/cockpit/pull/13517/
- [x] Wait default time for machine to reboot  8.2 - https://github.com/cockpit-project/cockpit/pull/13537 
- [x] Fix testSessionRecordingShell - https://github.com/cockpit-project/cockpit/pull/13539
- [x] testCreateTimer (check_services.TestServices) - https://github.com/cockpit-project/cockpit/pull/13540
- [ ] testConfigureBeforeInstall (check_machines_dbus.TestMachinesDBus)
- [x] cockpit times out while composer is building images - https://github.com/weldr/cockpit-composer/pull/890
 - [x] selenium MachinesBasicTestSuite.testNoVm: https://github.com/cockpit-project/cockpit/pull/13564

Follow ups:
- Move bastion to F31 (we need to replace docker with podman)

PR #502, opened 2020-02-06
Default OS is F31

marusak:


marusak:
This needs to refresh F31 and F-testing. But that would collide with https://github.com/cockpit-project/bots/pull/501 so blocking until then

marusak:
I picked these changes into #501 as it needed refresh anyway

PR #508, opened 2020-02-07
machine: Drop VirtMachine boot diagnostics

### LINKED ISSUE #507: "RETRY due to failure of test harness" takes too long
This [test](https://209.132.184.41:8493/logs/pull-13509-20200207-123434-ba9a8915-debian-stable/log.html#259), when it fails, takes about half an hour (!). There's a retry loop with quadratic timeouts, let's fix that. It gets aggravated further by our "retry tests three times" behaviour.

martinpitt:
This isn't really helpful. In CI it is responsible for a good deal of
long test run times when rebooting fails, and the diagnostics don't tell
us much about *why* a machine is completely frozen or *why* ssh to it is
broken even though it has a login prompt (the usual two cases that we
observe).

For a local developer, it's even less useful -- due to wrapping virsh
console into expect, the shown console does not react to keyboard input
at all, and furthermore blocks a manual `virsh console`, thus *breaking*
debuggability. It is much more helpful to run with --sit and attach to
the VM directly.

Fixes #507

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -516,33 +516,6 @@ class VirtMachine(Machine):
             # Normally only one pass
             break
 
-    def _diagnose_no_address(self):
-        SCRIPT = """
-            spawn virsh -c qemu:///session console $argv
-            set timeout 300
-            expect "Escape character"
-            send "\r"
-            expect " login: "
-            send "root\r"
-            expect "Password: "
-            send "foobar\r"
-            expect " ~]# "
-            send "ip addr\r\n"
-            expect " ~]# "
-            exit 0
-        """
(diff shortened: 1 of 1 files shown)


martinpitt:
Triggering some tests which recently were timing out a lot.

martinpitt:
The [debian-stable](https://209.132.184.41:8493/logs/pull-508-20200207-202121-82e4a57f-debian-stable-cockpit-project-cockpit/log.html), [debian-testing](https://209.132.184.41:8493/logs/pull-508-20200207-202525-82e4a57f-debian-testing-cockpit-project-cockpit/log.html), and [fedora-30](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-508-20200207-202121-82e4a57f-fedora-30-cockpit-project-cockpit/log.html) runs have quite a number of transient boot failures, and now are much more reasonable. They time out after a bit more than 2 minutes rather than > 30 minutes.

PR #513, opened 2020-02-10
testmap: add automatic trigger for rhel8 python test suite

jrusz:
https://github.com/weldr/lorax/pull/960 is merged. I'd like to trigger the test automatically.


jrusz:
Yeah I guess your're right but it's how our make target is called so we're ok with that, thanks :)

PR #505, opened 2020-02-10
Image refresh for continuous-atomic

cockpituous:
Image refresh for continuous-atomic
 * [x] Fix less import in cockpit-ostree: https://github.com/cockpit-project/cockpit-ostree/pull/53


martinpitt:
It seems [cockpit-ostree](https://209.132.184.41:8493/logs/pull-505-20200210-151010-734e0b81-continuous-atomic-cockpit-project-cockpit-ostree/log.html) suffers from the same less update issue than cockpit this morning.

PR #517, opened 2020-02-10
machine: Put SSHConnection.execute() with stdout under timeout

### LINKED ISSUE #516: testCheckpoint takes unreasonably long to fail
[This rhel-8-2-distropkg](https://209.132.184.41:8493/logs/pull-13526-20200210-162626-1a1a80cf-rhel-8-2-distropkg/log.html#149) test run timed out, due to
```
1498 not ok 149 testCheckpoint (check_networking_checkpoints.TestNetworking) # duration: 1498 # RETRY due to failure of test harness or framework
1503 not ok 149 testCheckpoint (check_networking_checkpoints.TestNetworking) # duration: 1503 # RETRY due to failure of test harness or framework
1554 not ok 149 testCheckpoint (check_networking_checkpoints.TestNetworking) # duration: 1554 # RETRY due to failure of test harness or framework
1554 not ok 149 testCheckpoint (check_networking_checkpoints.TestNetworking) # duration: 1554 # RETRY due to failure of test harness or framework
```
I. e. these three retries took 1½ hours. There is clearly a broken test retry loop somewhere.

```
  File "test/verify/check-networking-checkpoints", line 50, in testCheckpoint
    b.click("#confirm-breaking-change-popup button:contains('Keep connection')")
  File "/build/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 213, in click
    self.mouse(selector + ":not([disabled])", "click", 0, 0, 0)
  File "/build/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 209, in mouse
    self.wait_visible(selector)
  File "/build/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 391, in wait_visible
    self.wait_js_func('ph_is_visible', selector)
  File "/build/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 375, in wait_js_func
    self.wait_js_cond("%s(%s)" % (func, ','.join(map(jsquote, args))))
  File "/build/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 372, in wait_js_cond
    self.raise_cdp_exception("timeout\nwait_js_cond", cond, result["exceptionDetails"], trailer)
  File "/build/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 174, in raise_cdp_exception
    raise Error("%s(%s): %s" % (func, arg, msg))
testlib.Error: timeout
wait_js_cond(ph_is_visible("#confirm-breaking-change-popup button:contains('Keep connection'):not([disabled])")):
```

martinpitt:
This caused tests hanging for half an hour on e. g. cleanup commands
like calling "journalctl" in cockpit's `BaseCase.copy_journal()`.

Fixes #516

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -299,11 +299,12 @@ class SSHConnection(object):
             input += script
             command = "<script>"
         command_line = ssh_env + default_ssh_params + additional_ssh_params + env_command + cmd
-        if stdout:
-            subprocess.call(command_line, stdout=stdout)
-            return
 
         with timeoutlib.Timeout(seconds=timeout, error_message="Timed out on '%s'" % command, machine=self):
+            if stdout:
+                subprocess.call(command_line, stdout=stdout)
+                return
+
             output = ""
             proc = subprocess.Popen(command_line, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             stdin_fd = proc.stdin.fileno()


martinpitt:
I'm deliberately triggering loads of tests here to make sure that we don't have legitimate .execute() calls that take longer than 2 minutes. If we do, we need to fix them with an explicit timeout= first.

martinpitt:
cockpit's rhel-8.2 branch [also needs the less fix backported](https://209.132.184.41:8493/logs/pull-517-20200210-203131-f5b42896-rhel-8-2-cockpit-project-cockpit-rhel-8.2/log.html), I sent https://github.com/cockpit-project/cockpit/pull/13529 for that.

martinpitt:
[Here testCheckpoint](https://209.132.184.41:8493/logs/pull-517-20200210-203131-f5b42896-rhel-8-2-cockpit-project-cockpit/log.html#149-2) fails (NM bug?), and it now only takes 140s.

martinpitt:
The [subscription-manager failure](https://209.132.184.41:8493/logs/pull-517-20200210-213939-f5b42896-rhel-8-2-candlepin-subscription-manager/log.html) is unrelated, due to a [less 3.11 regression](https://github.com/less/less.js/issues/3476). Let's give  upstream a day or two to fix it, otherwise I'll send a workaround PR. But I went through integration-tests/check-subscriptions and it doesn't use `.execute(..., stdout=...)`, so this is unaffected by this change.

marusak:
> Here testCheckpoint fails (NM bug?), and it now only takes 140s.

I am actually bit concern about this. A bit afraid that this could bring a lot of flakes when it needs a bit more time. But it seems that when it gets broken, more time won't help, in which case it makes sense.

martinpitt:
@marusak: The timeout can be changed, the default is 2 mins; but normally checkpoint/restore should take about 10s. The failures that we saw so far were due to checkpointing not working at all, not it being slow.

martinpitt:
> The subscription-manager failure

less 3.11.1 got released which fixes that. The test is green now.

However, I'm going to take a look at testCheckpoints on RHEL 8.2. This fails a little too often, and this at least deserves a bug report.

martinpitt:
> A bit afraid that this could bring a lot of flakes when it needs a bit more time.

To clarify: This change does *not* affect the actual test code. It does affect the post-test cleanup that calls `journalctl` to get the logs. That's the bit that uses `Machine.execute()` with the stdout= option. Hence I think this is pretty safe, as the journalctl call is the *only* thing in Cockpit's test suite that uses stdout=.

martinpitt:
@marusak: Right, it would have failed before as well. It's another failure in the exception handling path of the original failure. :-) BTW, https://github.com/cockpit-project/cockpit/pull/13534 is an attempt to actually fix the test on rhel-8-2.

PR #519, opened 2020-02-11
machine: Increase DHCP lease time

martinpitt:
The default is only 1h, which causes e. g. Selenium tests to start
failing if they run longer than that.

Bump it to 4h.


marusak:
Thanks. I picked it into https://github.com/cockpit-project/cockpit/pull/13517 and running tests to see, if that actually helps.

martinpitt:
Wow, the chrome test succeeded on first  try! The 1h timeout could totally explain why the selenium/chrome test has been so unstable lately.

PR #521, opened 2020-02-11
images: Explicitly start firewalld in debian.install

### LINKED ISSUE #520: CI: tests fail prematurely with 'Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory'
```
+ firewall-cmd --add-service=cockpit --permanent
Error: DBUS_ERROR: Failed to connect to socket /var/run/dbus/system_bus_socket: No such file or directory
```
Example: https://209.132.184.41:8493/logs/pull-13488-20200205-111111-b83060e6-ubuntu-1804/log.html

martinpitt:
It can happen that the *.install scripts runs before multi-user.target
is active (it only waits for ssh) and firewalld is actually running.
Explicitly start it (which will just wait if it's already being
started), similar to what fedora.install does.

Fixes #520

### DIFF
--- a/images/scripts/lib/debian.install
+++ b/images/scripts/lib/debian.install
@@ -87,6 +87,7 @@ if [ -n "$do_install" ]; then
     # avoid random dpkg database locks, they break our package related tests
     systemctl disable apt-daily-upgrade.timer
 
+    systemctl start firewalld
     firewall-cmd --add-service=cockpit --permanent
     # not managed by NM, so enable interface manually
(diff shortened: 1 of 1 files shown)


martinpitt:
This should fix failures like [this](https://209.132.184.41:8493/logs/pull-13488-20200205-111111-b83060e6-ubuntu-1804/log.html) or [this](https://209.132.184.41:8493/logs/pull-13534-20200211-142929-49fec250-ubuntu-stable/log.html)

PR #522, opened 2020-02-11
images: Disable automatic suspend on windows-10

martinpitt:
This previously caused the VM to pause and disconnect from the network
after 30 minutes.

See #501


martinpitt:
I gave this a quick local test with cockpit's selenium-edge, but let's first make sure that this fully works. My image is still uploading, I'll trigger the tests once it finished.

PR #534, opened 2020-02-17
tests-scan: Unbreak tests_human

martinpitt:
Commit 9f6d0ece added a new `github_context` argument to tests_invoke(),
but forgot it in tests_human(). This re-fixes `-d`.

### DIFF
--- a/tests-scan
+++ b/tests-scan
@@ -102,7 +102,7 @@ def main():
 
 
 # Prepare a human readable output
-def tests_human(priority, name, number, revision, ref, context, base, original_base, repo, bots_ref):
+def tests_human(priority, name, number, revision, ref, context, base, original_base, repo, bots_ref, github_context):
     if not priority:
         return
     try:


martinpitt:
To reproduce, run `bots/tests-scan -vd`, and see

    TypeError: tests_human() takes 10 positional arguments but 11 were given

tests-scan
```diff
@@ -102,18 +102,19 @@ def main():
...
         priority = int(priority)
     except (ValueError, TypeError):
         pass
-    return "{name:11} {context:25} {revision:10} {priority:2}{repo}{bots_ref}{branches}".format(
+    return "{name:11} {context:25}{github_context} {revision:10} {priority:2}{repo}{bots_ref}{branches}".format(
         priority=priority,
         revision=revision[0:7],
         context=context,
         name=name,
+        github_context=github_context == "master" and "" or (" " + github_context),
```
marusak: github_context will never be "master". It is literally the name of the test in github with all parts.

martinpitt: TBH I didn't look too much into what it means, but I started with just printing the value, and `bots/tests-scan -vd` just showed "master" everywhere. I just sent PR #13561 which has a bots PR test, but it doesn't show that.

So I adjusted this PR to simply accept and ignore the argument for now.

martinpitt: Argh, I had the argument in the wrong place, I really saw bots_ref, sorry.

PR #556, opened 2020-02-23
Image refresh for rhel-7-8

cockpituous:
Image refresh for rhel-7-8
 * [x] image-refresh rhel-7-8


martinpitt:
"No package beakerlib available." -- did this fall out of rhel 7.8 now?

martinpitt:
Ah no, beakerlib gets installed from a third-party source -- see e. g. https://logs.cockpit-project.org/logs/pull-967-20200220-131017-21483a0c-rhel-7-8-vmware/log . So that third-party source simply disappeared, and the lorax test needs to get updated -- @atodorov , FYI.

So we can ignore that one for now, but we still need to get the composer tests to green.

martinpitt:
Nice, a retry worked,  so it seems there was just a temporary outage of the external beaker repos.

martinpitt:
@henrywang , can you please have a look at the composer rhel-8 errors?

henrywang:
@martinpitt, all green now!

PR #561, opened 2020-02-24
tests-policy: Drop default value of $COCKPIT_LEARN_SERVICE_HOST

martinpitt:
This service is not currently active, and we waste a failing http
request for each test failure right now.

Skip guessFlake() if no learn service is available.

If/when this comes back, we can revert this, or set the variable in the
cloud config.

### DIFF
--- a/tests-policy
+++ b/tests-policy
@@ -137,6 +137,12 @@ def parseName(output):
 # Flakiness Checks
 
 def guessFlake(output, context, verbose=False):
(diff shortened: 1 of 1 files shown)


martinpitt:
This will avoid the nasty "predict" error messages in the [test logs](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-13589-20200221-073845-c2bdc268-debian-stable/log.html)

PR #564, opened 2020-02-24
testmap: Move starter-kit testing from Fedora 30 to 31

martinpitt:
I ran `TEST_OS=fedora-31 make check` locally to make sure that this works. If you'd rather go via `_manual` first, I'm happy to do that, too.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -68,11 +68,10 @@ REPO_BRANCH_CONTEXT = {
     'cockpit-project/starter-kit': {
         'master': [
             'centos-7',
-            'fedora-30',
+            'fedora-31',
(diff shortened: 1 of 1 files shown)


task/testmap.py
```diff
@@ -68,7 +68,7 @@
     'cockpit-project/starter-kit': {
         'master': [
             'centos-7',
-            'fedora-30',
+            'fedora-31',
             'centos-8-stream',
         ],
         '_manual': [
```
marusak: It is in manual, so it should be tested here and also dropped from manual.

martinpitt: Argh, sorry. Fixed/done.

PR #580, opened 2020-02-26
Image refresh for rhel-7-7

cockpituous:
Image refresh for rhel-7-7
 * [x] image-refresh rhel-7-7


martinpitt:
[sosreport](https://logs.cockpit-project.org/logs/pull-580-20200226-152948-8feef16b-rhel-atomic-cockpit-project-cockpit-rhel-7.8/log.html) failures, but these shouldn't really get triggered by a rhel-7-7 image refresh, that's just for building the packages. In both cases this was just a timeout waiting for the report to be generated. This was fixed by PR #13609, which we backported to rhel-8.2 branch, but apparently not to -7.8. I'll do that.

martinpitt:
Ah no, PR #13609 was about selenium tests, not verify ones. Just retrying then, but I'll still backport the selenium fix.

PR #589, opened 2020-02-28
Image refresh for rhel-atomic

cockpituous:
Image refresh for rhel-atomic


martinpitt:
This [subscription-manager test failure](https://logs.cockpit-project.org/logs/pull-589-20200228-202458-3d43d23a-rhel-atomic-cockpit-project-cockpit-rhel-7.8/log.html#186-2) looks serious, but retrying once for comparison.

PR #612, opened 2020-03-06
images: Add ubuntu-2004

martinpitt:
 * [x] adjust cockpit for ubuntu-2004 tests: https://github.com/cockpit-project/cockpit/pull/13666


cockpituous:
Log: https://logs.cockpit-project.org/logs/image-refresh-612-20200306-070000/

cockpituous:
image-refresh ubuntu-2004 done: https://github.com/cockpituous/bots/commits/image-refresh-ubuntu-2004-20200306-072929

martinpitt:
https://github.com/cockpit-project/cockpit/pull/13666 adjusts cockpit's tests for this new image, so this is being tested there.

task/testmap.py
```diff
@@ -33,6 +33,7 @@
             'debian-stable',
             'debian-testing',
             'ubuntu-1804',
+            'ubuntu-2004',
```
marusak: Don't forget to drop it from manual :)

task/testmap.py
```diff
@@ -179,7 +180,7 @@
         "fedora-31@cockpit-project/cockpit",
         "fedora-31/selenium-firefox@cockpit-project/cockpit",
         "fedora-31/selenium-chrome@cockpit-project/cockpit",
-        "ubuntu-1804@cockpit-project/cockpit",
+        "ubuntu-2004@cockpit-project/cockpit",
```
marusak: Why?

martinpitt: Initially, because I'd like to drop the ubuntu-1804 image soon. But indeed this should be a separate PR. I'll revert that. Moreover, there's no freeipa on ubuntu-2004 ATM, so we couldn't test that part. I think we better test that on debian-stable then.

task/testmap.py
martinpitt: Ack, fixed.

marusak:
Naughties 65,66 and 70 were not applied not even once. Makes me wonder if we need to copy them over.

martinpitt:
Interesting, let's try without these then. We can add them back if necessary. I was fairly sure that we need https://github.com/cockpit-project/bots/issues/66 as it affects *all* OSes (it's just how NM works these days).

PR #627, opened 2020-03-10
task: Add manual starter-kit test case for firefox scenario

martinpitt:


martinpitt:
Tested `TEST_OS=fedora-32 make check` locally, and that works fine.

PR #632, opened 2020-03-12
Image refresh for fedora-31

cockpituous:
Image refresh for fedora-31
 * [x] image-refresh fedora-31


martinpitt:
Looks like we need to backport some selenium kdump fix to the rhel-7.8 branch? https://logs.cockpit-project.org/logs/pull-632-20200312-154609-d96927a7-fedora-31-selenium-chrome-cockpit-project-cockpit-rhel-7.8/log.html I see the same failure with the current image (https://github.com/cockpit-project/cockpit/pull/13707), so let's not block the image refresh on that.

PR #649, opened 2020-03-18
image-download: download rhel and windows from VPN

allisonkarlitskaya:
If an image contains 'rhel' or 'windows' in the name, skip checking the
PUBLIC_STORES and go right to REDHAT_STORES.  These images are never
available on the public servers.

Similarly (but perhaps more controversially): don't download non-private
images from the VPN.  The public servers will be faster.

### DIFF
--- a/image-download
+++ b/image-download
@@ -125,6 +125,8 @@ def find(name, stores, latest, quiet):
 
 
 def download(dest, force, state, quiet, stores):
+    name = os.path.basename(dest)
+
     if not stores:
         config = os.path.expanduser(CONFIG)
         if os.path.exists(config):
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
I need a bit of help with the "controversial" half of the patch.  Will this prevent propagation of images from the internal to the external servers?

allisonkarlitskaya:
My issue is that the amount of time taken to timeout on the 10.x.x.x host is too long.  The networks I use don't actively blackhole this, so I always end up waiting for the timeout...

marusak:
@martinpitt @allisonkarlitskaya any agreement? :)

martinpitt:
The "good" half is now part of #909. The other is what we don't want. Can you please test with #909 and check what's still broken for you? Thanks!

PR #667, opened 2020-03-22
images: Pin ubuntu-stable to 19.10 (eoan) until 20.10 gets opened

martinpitt:
After 20.04 LTS gets released, we don't want ubuntu-stable to move to
that, as we already have a separate ubuntu-2004 image for that. Let's
rather keep testing 19.10, until 20.10 opens.

### DIFF
--- a/images/scripts/ubuntu-stable.bootstrap
+++ b/images/scripts/ubuntu-stable.bootstrap
@@ -15,4 +15,7 @@ fi
 # release name is the last part of the URL
 rel=${rel##*/}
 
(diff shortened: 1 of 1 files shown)


martinpitt:
As you can see from the [most recent image refresh](https://logs.cockpit-project.org/logs/image-refresh-646-20200317-004343/log):

    + rel=eoan

this PR does not currently change anything, so I don't rebuild the image. I just don't want to forget about it in a month when 20.04 LTS releases.

PR #671, opened 2020-03-23
Remove ubuntu-1804 from image build scripts, naughties and test scripts

KKoukiou:
Since release 216 we will stop backporting cockpit in ubuntu-1804 so
we can now drop all the related testing bits.


task/testmap.py
```diff
@@ -189,7 +188,6 @@
         "fedora-31@cockpit-project/cockpit",
         "fedora-31/selenium-firefox@cockpit-project/cockpit",
         "fedora-31/selenium-chrome@cockpit-project/cockpit",
-        "ubuntu-1804@cockpit-project/cockpit",
```
martinpitt: Let's move that to ubuntu-2004 instead.

martinpitt:
@KKoukiou , can you please rebase this to current master and do the services image fix above? Thanks!

PR #636, opened 2020-03-29
Image refresh for rhel-7-8

cockpituous:
Image refresh for rhel-7-8
 * [x] image-refresh rhel-7-8


marusak:
[openstack on lorax](https://logs.cockpit-project.org/logs/pull-636-20200329-154726-a1694181-rhel-7-8-openstack-weldr-lorax-rhel7-extras/log.html) failed. Once more?

marusak:
ok, so openstack failed at least 3 times, @atodorov PTAL. Is this related to the refresh?

atodorov:
@marusak - NO, the failure doesn't seem related at first glance:

```
:: [ 05:27:01 ] :: [   PASS   ] :: Command 'pip install ansible openstacksdk' (Expected 0, got 0)

:: [ 09:37:51 ] :: [   INFO   ] :: localhost | FAILED! => {
    "changed": false, 
    "msg": "openstacksdk is required for this module"
}
```

So pip tells us that openstacksdk has been installed and later ansible complains that it is missing. Weird.

marusak:
> So pip tells us that openstacksdk has been installed and later ansible complains that it is missing. 

that could be related to the refresh surely. Could you please try if you can reproduce this locally?

atodorov:
> Could you please try if you can reproduce this locally?

You need to tell me which version of 7.8 that is. I will give it a try with the latest I can get my hands on.

marusak:
I mean, if you could use this exact image from this PR, that would be awesome.
But otherwise it is the newest compose we found at http://download.devel.redhat.com/rhel-7/nightly/RHEL-7/latest-RHEL-7.8/compose/Server/x86_64/os/

atodorov:
@marusak - the same error is reproducible locally with the latest OS version. However I have the feeling the OS version doesn't actually matter b/c we're using a virtualenv to install ansible and its dependencies. 

I've opened this with Ansible here https://github.com/ansible/ansible/issues/68616. Looking in their issues it looks like there are some similar problems, but not exactly the same.

marusak:
So it really is related to the refresh as the new version is missing the `queue` dependency, right?
Not sure how the tests are structured, but can you install the missing dependency on your side into the VM so we can unblock it?

martinpitt:
Unrelated to image refresh and fixed by https://github.com/weldr/lorax/pull/982 . Thanks @atodorov!

PR #699, opened 2020-03-30
Image refresh for fedora-32

cockpituous:
Image refresh for fedora-32
 * [x] image-refresh fedora-32


marusak:
The same issue as in #700, fixed in https://github.com/cockpit-project/cockpit/pull/13828

marusak:
Argh, so my fix works on fedora-testing (f31) but forgot to trigger also f-32 in that PR and it is not working.
```
+ semanage boolean -D && semanage fcontext -D && semanage module -D
Re-declaration of type ipa_custodia_t
Failed to create node
Bad type declaration at /var/lib/selinux/targeted/tmp/modules/100/ipa_custodia/cil:1
OSError: [Errno 0] Error
```
raises two questions:
1. Why does it try to re-declare it? Does fedora-testing do the same - if so, this will be prone to flakes (but it did not fail even once and I would guess it would be flake we would hit almost every time)
2. Is the declaration in `/var/lib/selinux/targeted/tmp/modules/100/ipa_custodia/cil:1` broken on Fedora 32?

marusak:
Reported in https://bugzilla.redhat.com/show_bug.cgi?id=1819626
Lets give it a day or two, maybe there is some actuall workaround and this is not a bug. If no reply or it is actual bug, then I'll create naughty out of it.

marusak:
Possible workaround: `semodule -e ipa_custodia`

martinpitt:
@marusak: seems ok as a naughty for now? The bz sounds like it is a packaging and/or policy bug.

marusak:
yeah, I'll send one.

PR #711, opened 2020-04-02
images: Remove fedora-coreos "core" user from wheel

martinpitt:
Otherwise we get spurious unexpected messages when polkit tries to
authenticate an action against all available administrators.

### DIFF
--- a/images/scripts/fedora-coreos.install
+++ b/images/scripts/fedora-coreos.install
@@ -4,6 +4,9 @@ set -eux
 # Enable ssh password authentication
 sed -i '/PasswordAuthentication no/d' /etc/ssh/sshd_config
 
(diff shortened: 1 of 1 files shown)


martinpitt:
This should help with [this](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-13819-20200330-154405-fae701ae-fedora-coreos/log.html) or [this](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-13837-20200402-185635-157ca9c7-fedora-coreos/log.html) failure. I'm not 100% sure yet whether we need to remove the user completely, but that should do.

martinpitt:
Yay, no failures any more due to this message, only the known networking timeout

PR #733, opened 2020-04-08
machine: Avoid warnings from cockpit startup

martinpitt:
Use systemctl stop with --quiet to avoid

    Warning: Stopping cockpit.service, but it can still be activated by:
      cockpit.socket

Don't use systemctl cat, as that complains with

    # Warning: cockpit.service changed on disk, the version systemd has loaded is outdated.
    # This output shows the current version of the unit's original fragment and drop-in files.
    # If fragments or drop-ins were added or removed, they are not properly reflected in this output.
    # Run 'systemctl daemon-reload' to reload units.
    Failed to cat /etc/systemd/system/cockpit.service.d/notls.conf: No such file or directory

This is a lot of unnecessary noise for each test.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -246,7 +246,7 @@ class Machine(ssh_connection.SSHConnection):
             rm -f /etc/systemd/system/cockpit.service.d/notls.conf &&
             systemctl reset-failed 'cockpit*' &&
             systemctl daemon-reload &&
-            systemctl stop cockpit.service &&
+            systemctl stop --quiet cockpit.service &&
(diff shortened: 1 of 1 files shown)


martinpitt:
The [log](https://logs.cockpit-project.org/logs/pull-733-20200408-162411-06a83386-rhel-8-2-cockpit-project-cockpit/log.html) confirms that the large warning at the top is gone now. There's still the "Stopping cockpit.service, but it can still be activated by" at the end, that's fixed by https://github.com/cockpit-project/cockpit/pull/13875

PR #738, opened 2020-04-09
Change RHEL-7 version to 7.9 for weldr/lorax tests

jikortus:
I think it's time to switch weldr/lorax tests to RHEL-7.9 images. Can we switch over, or are there still some obstacles preventing this @atodorov ? Thanks!

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -139,6 +139,14 @@ REPO_BRANCH_CONTEXT = {
         # These can be triggered manually with bots/tests-trigger
         '_manual': [
             'centos-8-stream',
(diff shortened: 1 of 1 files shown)


martinpitt:
If you are not sure, add them to `_manual` first, run them on some PR, and if it works switch it over like in this current PR.

atodorov:
Other than what @martinpitt says I don't see why not switch.

jikortus:
OK, thanks. Let's do the change in `_manual` context first, although I ran the test suite on RHEL-7.9 last week and didn't notice any 7.9-related issues there.

PR #755, opened 2020-04-16
Image refresh for debian-testing - fixed in #757

cockpituous:


martinpitt:
timesyncd was recently split out, needs adjustment in the build scripts:
```
dpkg: dependency problems prevent removal of chrony:
 systemd depends on systemd-timesyncd | time-daemon; however:
```

PR #756, opened 2020-04-16
task: Add CDP test on chromium and firefox

henrywang:
Let's manual first. The PR [#931](https://github.com/osbuild/cockpit-composer/pull/931) should work with this PR.
I do not forget the correct work flow. :)

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -165,6 +165,7 @@ REPO_BRANCH_CONTEXT = {
         ],
         # These can be triggered manually with bots/tests-trigger
         '_manual': [
+            'fedora-31',
         ],
     },
     'candlepin/subscription-manager': {


task/testmap.py
```diff
@@ -165,6 +165,8 @@
         ],
         # These can be triggered manually with bots/tests-trigger
         '_manual': [
+            'fedora-31',
+            'fedora-31/firefox',
```
marusak, line 169: `fedora-31/firefox` already exists in `master`. Is the plan to remove selenium straight away or keep it side by side with cdp tests? If the latter, then I think there should be some better distinction in those names.

marusak: Looking at https://github.com/osbuild/cockpit-composer/pull/931 I see that you decided to remove selenium right away, nice!
Have you covered all the functionality that selenium tests had already in verify ones?

Anyway, now only valid contexts are really these two you just posted. (Or you can have `fedora-31/chromium` but I think `fedora-31` is better) :)
Please just drop the firefox one as it is duplicate from master.

henrywang: `fedora-31/firefox` removed. Thanks for your review! :+1:

PR #759, opened 2020-04-16
machine: Specify backing format for QEMU disk overlay

martinpitt:
Without this, current (Fedora 32) libvirt fails to auto-detect the
backing format of the transient overlay, and vm-run fails with

    libvirt.libvirtError: Requested operation is not valid:
    format of backing image 'test/images/fedora-31' of image 'tmp/run/qj8pw2wn.qcow2' was not specified in the image metadata

See https://bugzilla.redhat.com/show_bug.cgi?id=1798148 for details.


martinpitt:
@garrett: This is what breaks vm-run and tests in Fedora 32.

martinpitt:
Now we use the same approach for the temporary overlay as for [cockpit's image-prepare persistent overlay](https://github.com/cockpit-project/cockpit/blob/master/test/image-prepare#L119).

PR #767, opened 2020-04-19
Image refresh for fedora-30

cockpituous:
Image refresh for fedora-30
 * [x] image-refresh fedora-30


marusak:
Fedora 30 will be EOL in a month or so. Maybe we can remove it, as chances we would actually deliver anything into it are rather slim.

martinpitt:
@marusak: Indeed +1 to removing it now. We didn't actually need it so far, and we can always build an image locally if we really do need it for something.

PR #769, opened 2020-04-19
Image refresh for rhel-8-2

cockpituous:
Image refresh for rhel-8-2
 * [x] image-refresh rhel-8-2


marusak:
4 times lorax failures. Alibaba is out of resources, the 3 remaining not sure, @atodorov can you please take a look and let us know, if it is related to this image refresh?

atodorov:
- Alibaba is a known issue, we don't know what's going on there. 
- vmware fix is in https://github.com/weldr/lorax/pull/986
- For live-iso and tar I see in the logs that everything from the test script reports PASS but then it fails with some qemu error, I think when we try to boot the resulting image so we can ssh into it. I can't tell why that is.

martinpitt:
live-iso/tar fail with

    qemu-img: /build/tmp/run/u9ym4myz.qcow2: Image is not in qcow2 format
    subprocess.CalledProcessError: Command '('qemu-img', 'create', '-q', '-f', 'qcow2', '-o', 'backing_file=/var/tmp/lorax-test.2u1_8z9_/live.iso,backing_fmt=qcow2', '/build/tmp/run/u9ym4myz.qcow2')' returned non-zero exit status 1.

This smells like fallout from PR #759 --- it seems this uses the bots Machine architecture for image overlays which are raw or another format. I suppose it shouldn't use a temporary image overlay for an *.iso. Either way, this isn't related to the image refresh itself.

martinpitt:
I sent PR #778 to fix the broken lorax tests, sorry about that!

PR #778, opened 2020-04-20
machine: Detect qcow2 overlay format

martinpitt:
Commit 0603b382dd7e always specified `backing_fmt=qcow2` for a Machine
overlay, but this broke lorax tests -- they use overlays with raw and
iso files. Thus check the header of the image file if it is actually
qcow2, and only specify the option  then.

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -360,8 +360,13 @@ class VirtMachine(Machine):
         image_to_use = self.image_file
         if not self.maintain:
             (unused, self._transient_image) = tempfile.mkstemp(suffix='.qcow2', prefix="", dir=self.run_dir)
+            options = ""
+            # check if it is qcow2 (libvirt complains otherwise)
+            with open(self.image_file, "rb") as f:
+                if f.read(3) == b"QFI":
+                    options += ",backing_fmt=qcow2"
             execute("qemu-img", "create", "-q", "-f", "qcow2",
-                    "-o", "backing_file=%s,backing_fmt=qcow2" % self.image_file, self._transient_image)
+                    "-o", "backing_file=%s%s" % (self.image_file, options), self._transient_image)
             image_to_use = self._transient_image
 
         keys = {


martinpitt:
I tested this locally on F32 and confirm that vm-run still works. Triggering a few representative tests, in particular the two broken (and one previously working) lorax tests.

machine/machine_core/machine_virtual.py
```diff
@@ -360,8 +360,13 @@ def execute(*args):
         image_to_use = self.image_file
         if not self.maintain:
             (unused, self._transient_image) = tempfile.mkstemp(suffix='.qcow2', prefix="", dir=self.run_dir)
+            options = ""
+            # check if it is qcow2 (libvirt complains otherwise)
+            with open(self.image_file, "rb") as f:
+                if b"qcow2" in f.read(128):
```
marusak, line 366: out of interest, is it given that first 128 chars always contains type of format?

martinpitt: I tried it on a couple of qcow images that I had laying around, and it looks quite regular:
```
❱❱❱ od -A x -t x1z -v test/images/fedora-32 | head
000000 51 46 49 fb 00 00 00 03 00 00 00 00 00 00 01 48  >QFI............H<
000010 00 00 00 70 00 00 00 10 00 00 00 03 00 00 00 00  >...p............<
000020 00 00 00 00 00 00 00 18 00 00 00 00 00 03 00 00  >................<
000030 00 00 00 00 00 01 00 00 00 00 00 01 00 00 00 00  >................<
000040 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000050 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
000060 00 00 00 04 00 00 00 68 e2 79 2a ca 00 00 00 05  >.......h.y*.....<
000070 71 63 6f 77 32 00 00 00 68 03 f8 57 00 00 00 c0  >qcow2...h..W....<
000080 00 00 64 69 72 74 79 20 62 69 74 00 00 00 00 00  >..dirty bit.....<
000090 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  >................<
```

martinpitt:
Somehow I botched that manual test request with a magic space or so -- triggered the two real tests.

machine/machine_core/machine_virtual.py
martinpitt: Actually, our cirros standard image doesn't have it -- let's rather check for `QFI` then.

martinpitt: See https://people.gnome.org/~markmc/qcow-image-format.html, the `QFI` magic is fixed.

PR #779, opened 2020-04-21
testmap: switch RHEL-7.9 weldr/lorax tests to automatic mode

jikortus:
Issues with weldr/lorax tests switchover to RHEL-7.9 have been found in a dummy PR for [weldr/lorax#990](https://github.com/weldr/lorax/pull/990) and most of them has been fixed. The only remaining one with live ISO is waiting for a PR ([weldr/lorax#994](https://github.com/weldr/lorax/pull/994)) to be merged.

When it's merged, we can switch the manual 7.9 tests to automatic mode.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -127,21 +127,6 @@ REPO_BRANCH_CONTEXT = {
             'rhel-8-2/ci',
         ],
         'rhel7-extras': [
-            'rhel-7-8',
-            'rhel-7-8/live-iso',
-            'rhel-7-8/qcow2',
-            'rhel-7-8/aws',
-            'rhel-7-8/azure',
-            'rhel-7-8/openstack',
-            'rhel-7-8/vmware',
-            'rhel-7-8/tar',
-        ],
-        # These can be triggered manually with bots/tests-trigger
-        '_manual': [
-            # alibaba keeps failing with "resource is out of stock in the specified zone"
-            'fedora-31/alibaba',
-            'rhel-8-2/alibaba',
-            'centos-8-stream',
             'rhel-7-9',
             'rhel-7-9/live-iso',
             'rhel-7-9/qcow2',
@@ -150,6 +135,13 @@ REPO_BRANCH_CONTEXT = {
             'rhel-7-9/openstack',
             'rhel-7-9/vmware',
             'rhel-7-9/tar',
+        ],
+        # These can be triggered manually with bots/tests-trigger
+        '_manual': [
+            # alibaba keeps failing with "resource is out of stock in the specified zone"
+            'fedora-31/alibaba',
+            'rhel-8-2/alibaba',
+            'centos-8-stream',
             'rhel-8-3',
             'rhel-8-3/live-iso',
             'rhel-8-3/qcow2',


martinpitt:
https://github.com/weldr/lorax/pull/994 just landed, triggering all the tests here as 994 only triggered one.

martinpitt:
@jikortus: This fails everywhere :/

jikortus:
@martinpitt It looks like I created a mismatch between the `_manual` and `rhel7-extras` contexts by a mistake, so I hope that was the real issue. However, from what I could find in the logs, the failure was in that the tests (including the lorax-composer VM preparation script, `test/vm.install`) from lorax-composer _master_ branch were fired against a rhel-7-9 VM, and the script failed when trying to resize the root volume (this part is not present in the rhel7-extras branch).

Can you please fire the remaining tests again for this PR? Thanks!

martinpitt:
@jikortus : Sure! Done.

martinpitt:
@jikortus : FTR, I did

    ./tests-trigger 779 rhel-7-9{,/live-iso,/qcow2,/aws,/azure,/openstack,/vmware,/tar}@weldr/lorax

You should be able to do the same.

martinpitt:
@jikortus : [azure](https://logs.cockpit-project.org/logs/pull-779-20200430-153426-242b38d1-rhel-7-9-azure-weldr-lorax/log.html), and all other tests failed with some sfdisk error.

jikortus:
Thanks a lot for your help, @martinpitt! Now there's another issue, which should soon be solved by weldr/lorax#1004.

Also thanks for the tip with retriggering the tests manually, however unfortunately it didn't work for me:
`779 is not a pull request.`

I didn't have much luck either with a specified `--repo`:
```
...
task.github.GitHubError: Error accessing /repos/cockpit-project/bots/statuses/242b38d1a1e02ba250326e2f715c370fd076b63c
  Status: 404
  Reason: Not Found
  Response: {"message":"Not Found","documentation_url":"https://developer.github.com/v3/repos/statuses/#create-a-status"}
```
Maybe I'm not authorized to do this or miss some piece of configuration on my side?

martinpitt:
@jikortus : Right, that sounds like an authorization problem. I do see you in the team, do you have a ~/.config/github-token?

Blocking this PR on the repo fix in weldr.

jikortus:
@martinpitt The PR should be good to go now, as the fixes in weldr/lorax#1004 have been merged.

I do have the github token, triggering weldr/lorax tests works fine for me. I don't think it's worth spending time investigating this, as I don't anticipate any need for further changes in bots repo related to composer testing from my side :).

martinpitt:
Thanks! Retrying tests, let's see if it still works. We got new image refreshes since then, so it *may* be that this will need another rebase.

martinpitt:
@jikortus : Seems it still happens: :cry: 
```
sfdisk: trailing junk after number
sfdisk: bad input
```

jikortus:
@martinpitt Thanks! I think the results you were looking at were the old ones, executed previously on master branch. The new ones executed on rhel7-extras branch seem to be fine with one exception (rhel-7-9@weldr/lorax/rhel7-extras). My gut feeling is that it might be caused by lorax-composer accessing content from RHN, which can be quite slow at times.

So if my understanding related to test execution on rhel7-extras vs. master branch is correct, I'd merge the changes so that we can run new CI tests on RHEL-7.9. @atodorov, what do you think about that?

PR #780, opened 2020-04-21

cockpituous:

### DIFF
--- a/images/fedora-coreos
+++ b/images/fedora-coreos
@@ -1,1 +1,1 @@
-fedora-coreos-13f1c43f6962b6bc5f81688f049ca21384c0a90a31126557b38b476dae3239ed.qcow2
+fedora-coreos-18d4d9b76216cf6ff4e3c5b2fd43b090ff878ce1144c14c1ccefbfb14dae8b36.qcow2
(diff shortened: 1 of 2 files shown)


marusak:
Load of failures like:
```
Error: error creating container storage: the container name "ws" is already in use by "7211a43ae80fae52111a166fdec31673a4f3e1ac0a27ae4489913c32c8444b4f". You have to remove that container to be able to reuse that name.: that name is already in use
Error: `/proc/self/exe run -d --name ws --privileged --pid=host -v /:/host localhost/cockpit/ws:latest /container/atomic-run --local-ssh --no-tls` failed: exit status 125
```

martinpitt:
Hah, this is a fun one -- in the current master image, the `podman container runlabel ` command doesn't respect the `--name NAME`, and it always generates a new container. With this image refresh, the name is now always `ws` as intended, and thus the collision.

marusak:
oh, is that regression in podman?

martinpitt:
@marusak: More like a progression; however, podman itself and crun are the same version, I wonder if it was just the container refresh itself from the last cockpit release? Anyway, fix pushed.

PR #796, opened 2020-04-27
images: Drop obsolete rhel-7-7 and rhel-8-1 images

martinpitt:
7.8 and 8.2 are released. Switch rhel-atomic build image to RHEL 7.8 to
reflect reality.

Also add the missing rhel-8-3 to the refresh table.


martinpitt:
@marusak: Indeed! I'll let the current test runs finish to make sure rhel-atomic built on 7.8 works alright (looking good so far), and then force-push just with the `host` test, if that's alright with you.

marusak:
> ... and then force-push just with the host test, if that's alright with you.

Sure! Can you please link results here before you force push, thanks.

martinpitt:
Some [TestSOS](https://logs.cockpit-project.org/logs/pull-796-20200427-044403-ef8dcc7e-rhel-atomic-cockpit-project-cockpit-rhel-7.9/log.html) timeouts, I don't see how they are related to building the rpm. I just re-run them.

martinpitt:
Still SOS timeouts. This needs to be figured out independently :sob:

marusak:
> Still SOS timeouts. This needs to be figured out independently

That is unrelated. The sos on rhel-7 is collecting all modules, which is slow. We changed this on rhel-8 and master to collect only 2 or 3 modules, which is super quick. I think that retrying is your best friend (or going into these branches and bumping up the timeout for these tests).
Either way I think you can update the PR, push force and hope for a better luck :)

martinpitt:
> The sos on rhel-7 is collecting all modules, which is slow

No, we already [have that backported](https://github.com/cockpit-project/cockpit/blob/rhel-7.8/test/verify/check-sosreport#L38). It just seems to get stuck. It succeeded on 7.8 now, but seems I need to retry once more on 7.9.

marusak:
> No, we already have that backported.

I think that this was introduced in some newer version of sosreport, so it is likely that it is ignored in rhel-7 versions.

PR #803, opened 2020-04-28
testmap: Move f32 from manual to master and add f32/firefox

henrywang:
f32 test passed in PR [#919](https://github.com/osbuild/cockpit-composer/pull/919)


martinpitt:
@henrywang : Want to try and enable rhel-8-3 as well?

Issue #805, opened 2020-04-28
Podman new favourit activity is to dump core


marusak:
Podman lately started to dump cores. They might point to the same issue, but they seem rather different. Opening only one issue as they happen in the same test (well two, once for root and once for user).
Upstream report: https://github.com/containers/libpod/issues/6021
Upstream report: https://github.com/containers/libpod/issues/6018

marusak:
Bump

marusak:
verify/fedora-32
Ooops, it happened again<hr>
```
# testRunImageUser (__main__.TestApplication)
Warning: Stopping io.podman.service, but it can still be activated by:
  io.podman.socket
mkdir: cannot create directory '/home/admin/.ssh': File exists
ControlSocket /build/tmp/ssh-127.0.0.2-2201-admin-1397 already exists, disabling multiplexing
Warning: Stopping io.podman.service, but it can still be activated by:
  io.podman.socket

DevTools listening on ws://127.0.0.1:9483/devtools/browser/5b02c5f6-3093-4f65-a88a-1f61f68e4419
[0526/074255.249995:ERROR:egl_util.cc(60)] Failed to load GLES library: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: cannot open shared object file: No such file or directory
[0526/074255.251558:ERROR:viz_main_impl.cc(159)] Exiting GPU process due to errors during initialization
CDP: {"source":"security","level":"error","text":"Refused to apply inline style because it violates the following Content Security Policy directive: \"default-src 'self' http://127.0.0.2:9091\". Either the 'unsafe-inline' keyword, a hash ('sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.\n","timestamp":1590478976312.63,"url":"http://127.0.0.2:9091/cockpit/$1b00be25adc122ba9d5adb63c5530aea584548f07e7909684fe3f7b4ddc8c3e4/shell/index.js","lineNumber":14,"stackTrace":"(minified)"}
> warning: grep: /sys/class/dmi/id/product_serial: Permission denied
grep: /sys/class/dmi/id/power/autosuspend_delay_ms: Input/output error
grep: /sys/class/dmi/id/chassis_serial: Permission denied
[...]

PR #806, opened 2020-04-28
naughty: Add pattern for podman coredumps

marusak:
See known issue #805

In one case we get the actual traceback from `exe` where word
`rootlessport` seems to be the most important one.

In the other case there is not really much we can match to. The varlink
dies, thus disconnects us. Also we save coredump and it always fails
waiting for 'No Containers' string.

### DIFF
--- a/naughty/fedora-31/805-podman-dumps-core
+++ b/naughty/fedora-31/805-podman-dumps-core
@@ -1,0 +1,3 @@
+Process * (exe) of user 1000 dumped core.
+*
+*rootlessport*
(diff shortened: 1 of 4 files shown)


marusak:
Example logs:
https://logs.cockpit-project.org/logs/pull-370-20200428-120538-fb42efc7-fedora-31-bots-804/log.html
https://logs.cockpit-project.org/logs/pull-370-20200428-120501-fb42efc7-fedora-32/log.html

PR #804, opened 2020-04-28
Image refresh for fedora-31

cockpituous:
Image refresh for fedora-31
 * [x] image-refresh fedora-31

PR #807, opened 2020-04-28
Image refresh for fedora-31

cockpituous:
Image refresh for fedora-31
 * [x] image-refresh fedora-31


marusak:
We refreshed f31 today in #804 but 'bot' label was dropped by accident.

PR #812, opened 2020-05-03
Image refresh for ubuntu-2004

cockpituous:
Image refresh for ubuntu-2004


martinpitt:
[machines testCreate](https://logs.cockpit-project.org/logs/pull-812-20200504-065034-16b55d12-ubuntu-2004-cockpit-project-cockpit/log.html#84-2) failed several times in a row now. The [screenshot](https://logs.cockpit-project.org/logs/pull-812-20200504-065034-16b55d12-ubuntu-2004-cockpit-project-cockpit/TestMachinesDBus-testCreate-ubuntu-2004-127.0.0.2-2201-FAIL.png) says "You need to select the most closely matching OS", and it tries to put "/var/lib/libvirt/novell.iso" into the input line, which doesn't seem right? @KKoukiou , any idea?

martinpitt:
The debian-testing refresh in PR #818 fails on exactly the same problem.

martinpitt:
OK, this failure isn't new, it happens on current images as well. So retry galore. But we really need to fix this..

PR #827, opened 2020-05-04
images: Disable cron and kdump on services image

martinpitt:
They both start at boot, and kdump.service rebuilds the initrd (which
takes 45s). This may help with some timeouts that we see in
check-realms.

 * [x] image-refresh services


cockpituous:
Log: https://logs.cockpit-project.org/logs/image-refresh-827-20200504-072727/

martinpitt:
This is a stab into the dark. It may not actually help much with the check-realms flakes investigated in https://github.com/cockpit-project/cockpit/pull/14009 , but it's a nice optimization either way.

cockpituous:
image-refresh services done: https://github.com/cockpituous/bots/commits/image-refresh-services-20200504-081010

martinpitt:
This doesn't help much with the test instability, but it still reduces what the test has to do, so I think we should land this anyway. But I'll keep digging.

PR #834, opened 2020-05-05
machine: Expand ~ in images-data-dir git config

martinpitt:
Use git config --type=path to ensure that any ~ included in the pathname
is properly expanded.

### DIFF
--- a/machine/machine_core/directories.py
+++ b/machine/machine_core/directories.py
@@ -24,14 +24,14 @@ _images_data_dir = None
 _temp_dir = None
 
 
-def get_git_config(variable):
+def get_git_config(*args):
     if not os.path.exists(GIT_DIR):
         return None
 
     try:
         myenv = os.environ.copy()
         myenv["GIT_DIR"] = GIT_DIR
-        return subprocess.check_output(["git", "config", variable], universal_newlines=True, env=myenv).strip()
+        return subprocess.check_output(("git", "config") + args, universal_newlines=True, env=myenv).strip()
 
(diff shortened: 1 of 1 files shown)


martinpitt:
This really needs to be tested manually, as we don't use that git config in CI. Today I replaced the `/home/martin` in my ~/.gitconfig with `~` as I want to put that config into a container. Then on master, it failed like this:
```
❱❱❱ ./image-prune -c --force 
  File "./image-prune", line 223, in <module>
    sys.exit(main())
  File "./image-prune", line 212, in main
    prune_images(args.force, args.dryrun, quiet=args.quiet,
  File "./image-prune", line 160, in prune_images
    for filename in os.listdir(get_images_data_dir()):
FileNotFoundError: [Errno 2] No such file or directory: '~/.cache/cockpit'
```

allisonkarlitskaya:
Here's a counter-proposal that has three improvements:

  - doesn't use the word "literal" with the opposite meaning in the commit message
  - doesn't try to expand ~ found in environment variables (which would be weird)
  - uses the standard git-config feature for this

martinpitt:
Hah, patch-bait reliably works -- give a developer a bad patch, and they invariably feel the urge to fix it :grin: Thanks! I wasn't aware of `--type=path`, that's nice! I actually deliberately applied it to the env var as well, but  (1) indeed it's unusual, and (2) I don't really care. I validated that just with your patch it works. So I tossed out my patch and just kept yours.

Of course review-wise that's now a little odd, as it's my PR with your patch, but :+1 from me. You have to formally approve, though.

PR #836, opened 2020-05-06
naughty: Add IPv4 variant of iptables-restore crash on debian-testing

martinpitt:


martinpitt:
Example: https://logs.cockpit-project.org/logs/pull-14035-20200506-042636-7f01329f-debian-testing/log.html#151-2

PR #848, opened 2020-05-11
naughty: Close 622: check-machines-dbus TestMachinesDBus.testCreate fails on ubuntu-2004 in CI

### LINKED ISSUE #622: check-machines-dbus TestMachinesDBus.testCreate fails on ubuntu-2004 in CI
This test case fails all the time in CI, like [here](https://logs.cockpit-project.org/logs/pull-13666-20200309-071745-a0f56f82-ubuntu-2004-bots-612/log.html). I can't reproduce this locally, so this will require some more in-depth debugging.

For now, let's treat this as a naughty so that we can start testing all the other bits of cockpit on Ubuntu 20.04, and unblock the bots PR.

cockpituous:
Known issue which has not occurred in 23 days

check-machines-dbus TestMachinesDBus.testCreate fails on ubuntu-2004 in CI

Fixes #622

### DIFF
--- a/naughty/ubuntu-2004/622-Machines-testCreate
+++ b/naughty/ubuntu-2004/622-Machines-testCreate
@@ -1,7 +1,0 @@
-Traceback (most recent call last):
-  File "test/verify/machineslib.py", line *, in testCreate
-    start_vm=True, delete=False))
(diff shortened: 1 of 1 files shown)


cockpituous:
Please comment on the upstream distro bug manually before accepting this pull request.

If you wish to keep this known issue open, then update it

PR #853, opened 2020-05-11
Image refresh for rhel-8-3

cockpituous:
Image refresh for rhel-8-3
 * [x] image-refresh rhel-8-3


marusak:
All machine tests fail with:
```
audit: type=1400 audit(1589263657.467:126): avc:  denied  { read write } for  pid=404889 comm="qemu-kvm" path="/dev/mapper/control" dev="devtmpfs" ino=11494 scontext=system_u:system_r:svirt_tcg_t:s0:c157,c654 tcontext=system_u:object_r:lvm_control_t:s0 tclass=chr_file permissive=0
```

I think we already saw this somewhere, but not really sure.

martinpitt:
Let's get the latest one to try and reproduce a PAM issue that we see downstream.

PR #855, opened 2020-05-12

cockpituous:


marusak:
beforeInstall and detachDisk refuse to get green on  centos-8-stream. These two are common flake, but they are really stubborn in this case.

PR #862, opened 2020-05-13
Fix URLs in image-create-rhel-compose

jikortus:


martinpitt:
@jikortus : You can forget about `[no-test]` thing for bots, it's no longer necessary. We will now decide on a PR-by-PR basis which tests to run.

PR #864, opened 2020-05-14
Image refresh for rhel-8-3

cockpituous:
Image refresh for rhel-8-3
 * [x] image-refresh rhel-8-3

### DIFF
--- a/images/rhel-8-3
+++ b/images/rhel-8-3
@@ -1,1 +1,1 @@
-rhel-8-3-abd9178c25d8445e172b50bf6c76957a03fc8566d94b7bbbcaf1521c954978f9.qcow2
+rhel-8-3-b504eb6bdf2fbbcfaf42f63f6bf9fca461b59aef60eed0efaa15dfe2a51ecb4c.qcow2


martinpitt:
I filed the new SELinux violation at https://bugzilla.redhat.com/show_bug.cgi?id=1835909 and sent https://github.com/cockpit-project/cockpit/pull/14090 to deal with it.

martinpitt:
The /dev/mapper SELinux failures are sorted out now, but [current run](https://logs.cockpit-project.org/logs/pull-864-20200515-065923-c43ff2a7-rhel-8-3-cockpit-project-cockpit/log.html) still seems to break realms and [testConfigureBeforeInstall](https://logs.cockpit-project.org/logs/pull-864-20200515-065923-c43ff2a7-rhel-8-3-cockpit-project-cockpit/log.html#84-2). @KKoukiou , does the latter ring a bell? That seems different from the failure you fixed yesterday, right? I've already seen it in PR #14090, so its' not just a flake, but something that actually got broken (or at least different) in RHEL 8.3.

I'm retrying once to compare check-realms results.

martinpitt:
Yes, these all seem to be real :(

martinpitt:
Reproduced locally - that's just the same PAM symbol linking issue again. Rebuilding image.

Issue #867, opened 2020-05-14
Libvirt race resulting in 'domain XYZ already being removed' when stopping Vm installation


KKoukiou:
testConfigureBeforeInstall often fails with: ERROR operation failed: domain 'VmNotInstalled' is already being removed

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1780591

cockpituous:
verify/centos-8-stream
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testConfigureBeforeInstall (__main__.TestMachines)
error: Failed to start network default
error: Requested operation is not valid: network is already active

DevTools listening on ws://127.0.0.1:9907/devtools/browser/b70a0311-1f13-422e-88ed-f3b129a5b84f
[0514/131637.700829:ERROR:egl_util.cc(60)] Failed to load GLES library: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: cannot open shared object file: No such file or directory
[0514/131637.702372:ERROR:viz_main_impl.cc(161)] Exiting GPU process due to errors during initialization
CDP: {"source":"security","level":"error","text":"Refused to apply inline style because it violates the following Content Security Policy directive: \"default-src 'self' http://127.0.0.2:9091\". Either the 'unsafe-inline' keyword, a hash ('sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.\n","timestamp":1589462199296.368,"url":"http://127.0.0.2:9091/cockpit/$b0386c0d3440f9073e503535dc6b966c5de02f553a81e217df0b5d86da00e318/shell/index.js","lineNumber":14,"stackTrace":"(minified)"}
> warning: grep: /sys/class/dmi/id/power/autosuspend_delay_ms: Input/output error

> warning: initResource action failed: No such method 'ListInterfaces'
> warning: getAllInterfaces action failed: No such method 'ListInterfaces'
> warning: initResource action failed: No such method 'ListInterfaces'
> warning: getAllInterfaces action failed: No such method 'ListInterfaces'
[...]

PR #868, opened 2020-05-14
Add naughty for libvirt being racy when stopping an installation

KKoukiou:
See known issue #867

### DIFF
--- a/naughty/rhel-8/867-libvirt-race-install
+++ b/naughty/rhel-8/867-libvirt-race-install
@@ -1,0 +1,5 @@
+* ERROR    operation failed: domain * is already being removed
+*
+Traceback (most recent call last):
+  File "test/verify/check-machines", line *, in testConfigureBeforeInstall
+    b.wait_in_text("#vm-VmNotInstalled-state", "shut off")


KKoukiou:
> This is very generic and can ignore all sorts of test results. Can you please at least include the test name in the pattern? Also, is there a workaround? Something like a `until virsh` loop to wait for the domain to really become undefined before going on?

No actually the problem was not what we discussed yesterday so anything , sleep or polling between the virsh undefine and virt-install won't help, because the problem is happening within virt-install itself.

What virt-install does is it starts an installation, then waits for the installation to reboot and then it changes the XML so that it does not boot into the installation.
In our case it's not rebooting but we shut it fown from the UI and the effect is the same.
virt-install will replace the installation XML with the real XML.
What libvirt people said is that there's possibly a race here wrt shutdown of the QEMU process and the subsequent attempt to define the domain with the real XML.
(the bugzilla is not updated yet with all these details, I am still in discussion with the libvirt people)


So, there is not way to WA it, it's really a race, inherently timing issue.

KKoukiou:
Please don't merge yet _ I think I found a WA for our tests.

KKoukiou:
Closing in favour of https://github.com/cockpit-project/cockpit/pull/14089

PR #877, opened 2020-05-18
machine_core: Fix Machine.set_address()

martinpitt:
Require the `nmcli con up` to succeed. It is absolutely not okay to
fail, as it will just cause weird test failures later on, due to
connection failures.

Robustify this by first stopping the interface's current connection, so
that the new one can succeed. Otherwise it happened in slow VMs that the
"eth1" connection was still active, and thus the provision's manual
address never appeared.

----

[Example](https://logs.cockpit-project.org/logs/pull-14101-20200518-055533-acbfe4f2-fedora-coreos/log.html#205-2) -- one of our most common flakes right now.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -295,10 +295,11 @@ class Machine(ssh_connection.SSHConnection):
 
     def set_address(self, address, mac='52:54:01'):
         """Set IP address for the network interface with given mac prefix"""
-        cmd = "nmcli con add type ethernet autoconnect yes con-name static-{mac} " \
-              "ifname \"$(grep -l '{mac}' /sys/class/net/*/address | cut -d / -f 5)\"" \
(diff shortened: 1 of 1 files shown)


martinpitt:
This reproduces well locally with running two of these in parallel:

    while TEST_OS=fedora-coreos test/verify/check-shutdown-restart -stv TestShutdownRestart.testBasic; do sleep 1; done

martinpitt:
Only cockpit's tests use `provision` or `set_address`, thus only triggering those.

martinpitt:
Whoops, the con down *can* [fail](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-877-20200518-084022-66860641-fedora-31-selenium-firefox-cockpit-project-cockpit/log.html). Just the `up` should never fail.

PR #884, opened 2020-05-19
naughty: Add more patterns for subscription-manager failure

martinpitt:
With the pkg/lib/packagekit.js backports from [1], the tests pass the
System page notification checks again, but still fail on the
Subscriptions page (as that still uses the old and broken API).

Thus add two more patterns to represent that. The older patterns can be
removed after merging [1].

[1] https://github.com/cockpit-project/cockpit/pull/14113


martinpitt:
Tested in https://github.com/cockpit-project/cockpit/pull/14113 . This is rather urgent, to untangle an ever-growing dependency chain...

PR #890, opened 2020-05-22

cockpituous:
 * [x] Adjust tests for test VMs > 1 GiB RAM: https://github.com/cockpit-project/cockpit/pull/14158


martinpitt:
There's [something wrong](https://logs.cockpit-project.org/logs/pull-890-20200522-132954-cf1e79b8-debian-testing-cockpit-project-cockpit/log.html#90-2) with qemu/libvirt.

martinpitt:
I filed the libvirt failure with QEMU 5.0 as https://bugs.debian.org/961575, and a QEMU regression (which is possibly responsible for that) as https://bugs.debian.org/961574.

I don't yet know a workaround -- I don't think we want to skip all Machines tests. Let's give the bug reports a day or two for responses, and then we can re-consider what to do.

@KKoukiou , FYI.

martinpitt:
I found that this only happens if the host has 1 GiB of RAM. If it has even just a tad more, it works. I updated the downstream bug and added a commit here which bumps the default RAM. I initially thought to special-case debian-testing, but QEMU 5 will arrive in Fedora and RHEL at some point as well.

martinpitt:
Ah, some of the tests make assumptions about the host's available memory. Let's fix that, this will also fail on downstream tests with existing testbeds.

martinpitt:
The test fixes landed, retrying.

PR #857, opened 2020-05-12
Add pr-tests: Summarize test status of a PR

martinpitt:
I threw this together as I'm currently working mostly without a real browser. But this seems useful for us in general, so let's see what others think:
```
❱❱❱ bots/pr-tests 14071
9 tests in state success: ubuntu-stable fedora-32 ubuntu-2004 debian-stable debian-testing fedora-31/selenium-edge fed$
ra-31/selenium-firefox semaphoreci fedora-31/container-bastion

Failed tests
============

fedora-31/firefox:
  https://logs.cockpit-project.org/logs/pull-14071-20200512-064525-7ef4f698-fedora-31-firefox/log.html
  not ok 85 test/verify/check-machines TestMachines.testCreate [ND] # RETRY 1 (be robust against unstable tests)
  not ok 85 test/verify/check-machines TestMachines.testCreate [ND] # RETRY 2 (be robust against unstable tests)
  not ok 85 test/verify/check-machines TestMachines.testCreate [ND]# ---------------------------------------------------------------------
  not ok 264 test/verify/check-superuser TestSuperuserOldShell.test # RETRY 1 (be robust against unstable tests)

fedora-31:
  https://logs.cockpit-project.org/logs/pull-14071-20200512-064525-7ef4f698-fedora-31/log.html
  not ok 59 test/verify/check-journal TestJournal.testBasic # RETRY 1 (be robust against unstable tests)
  not ok 59 test/verify/check-journal TestJournal.testBasic # RETRY 2 (be robust against unstable tests)
  not ok 122 test/verify/check-multi-machine TestMultiMachineAdd.testBasic # RETRY 1 (be robust against unstable tests)
  not ok 122 test/verify/check-multi-machine TestMultiMachineAdd.testBasic # RETRY 2 (be robust against unstable tests)
  not ok 122 test/verify/check-multi-machine TestMultiMachineAdd.testBasic# ---------------------------------------------------------------------

rhel-8-2-distropkg:
  https://logs.cockpit-project.org/logs/pull-14071-20200512-064525-7ef4f698-rhel-8-2-distropkg/log.html
  not ok 145 test/verify/check-networking-firewall TestFirewall.testNetworkingPage [ND] # RETRY 1 (be robust against unstable tests)
  not ok 84 test/verify/check-machines TestMachines.testConfigureBeforeInstall [ND] # RETRY 1 (be robust against unstable tests)
[...]

PR #894, opened 2020-05-25
Introduce script for storing flaked tests

marusak:
@martinpitt This is the first part. It needs to actually store into DB but for now I just print it on cmdline.

### DIFF
--- a/store-failures
+++ b/store-failures
@@ -94,7 +94,6 @@ def init_db(cursor):
                     (testname TEXT NOT NULL,
                      retry_reason TEXT,
                      run INTEGER NOT NULL,
-                     UNIQUE (testname, retry_reason, run)
                      FOREIGN KEY (run) REFERENCES TestRuns(id))
                    """)
 
@@ -108,7 +107,7 @@ def insert_run(cursor, project, context, url):
 
 
 def insert_failure(cursor, run_id, testname, retry_reason):
-    cursor.execute("INSERT OR REPLACE INTO Failures VALUES (?, ?, ?)", (testname, retry_reason, run_id))
+    cursor.execute("INSERT INTO Failures VALUES (?, ?, ?)", (testname, retry_reason, run_id))
 
 
 def find_failed_tests(fp):
--- a/store-failures
+++ b/store-failures
@@ -24,9 +24,13 @@ import sqlite3
 import urllib.request
 import argparse
 import logging
+import re
 
 from task import github
 from machine.machine_core.constants import IMAGES_DIR
+
+# regexp for retry reason in a "not ok" line
+re_retry = re.compile(r"# RETRY .* \(([^)]+)\)")
 
 
 def main():
@@ -113,15 +117,15 @@ def find_failed_tests(fp):
         if line.startswith("not ok "):
             # Result line has form like:
             # not ok 24 test/verify/check-foobar TestExample.testBasic [ND] # RETRY 2 (be robust against unstable tests)
+            # older logs look like this instead:
+            # not ok 78 testNodeNavigation (check_openshift.TestOpenshift) # duration: 172s
             fields = line.split(" ")
             # sometimes line breaks are missing:
             # not ok 117 test/verify/check-metrics TestMetrics.testPcp# ------
             fields[4] = fields[4].rstrip("#")
             testname = "{0} {1}".format(fields[3], fields[4])
-            retry_reason = None
-            if "(" in line and ")" in line:
-                retry_reason = line[line.rfind("(") + 1:line.rfind(")")]
-            yield (testname, retry_reason)
+            m = re_retry.search(line)
+            yield (testname, m and m.group(1) or None)
 
 
 if __name__ == '__main__':
--- a/store-failures
+++ b/store-failures
@@ -70,8 +70,14 @@ def main():
                 # Can contain branch, as it diffs between master and branches (e.g `cockpit-project/cockpit/rhel-7.9`)
                 repo = parts[1]
 
+        row = cursor.execute("SELECT id FROM TestRuns WHERE url = ?", (url, )).fetchone()
+        if row is not None:
+            logging.warning("Test run %s already exists as ID %i, skipping", url, row[0])
+            continue
+
         raw_log = url[:-5] if url.endswith(".html") else url
         logging.debug("scanning log %s", raw_log)
+
         with urllib.request.urlopen(raw_log, context=ctx if "logs.cockpit-project.org" in raw_log else None) as fp:
             for (testname, retry_reason) in find_failed_tests(fp):
                 if run_id is None:
@@ -99,10 +105,6 @@ def init_db(cursor):
 
 
 def insert_run(cursor, project, context, url):
-    row = cursor.execute("SELECT id FROM TestRuns WHERE url = ?", (url, )).fetchone()
-    if row is not None:
-        logging.debug("Test run %s already exists as ID %i, skipping", url, row[0])
-        return row[0]
     return cursor.execute("INSERT INTO TestRuns VALUES (null, ?, ?, ?)", (project, context, url)).lastrowid
 
 
--- a/store-failures
+++ b/store-failures
@@ -1,0 +1,110 @@
+#!/usr/bin/env python3
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import ssl
+import sys
+import sqlite3
+import urllib.request
+import argparse
+
+from task import github
+from machine.machine_core.constants import IMAGES_DIR
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Store information about failed tests')
+    parser.add_argument("--db", default="cockpit-failures.db", help="Database name")
+    parser.add_argument("--repo", help="Repository from which to process failures")
+    parser.add_argument("revision", help="SHA for which failures should be stored")
+    opts = parser.parse_args()
+
+    ca = os.path.join(IMAGES_DIR, "files", "ca.pem")
+    ctx = ssl.create_default_context()
+    ctx.load_verify_locations(cafile=ca)
+
+    db_conn = sqlite3.connect(opts.db)
+    cursor = db_conn.cursor()
+    init_db(cursor)
+
+    api = github.GitHub(repo=opts.repo)
+    statuses = api.statuses(opts.revision)
+
+    for status in statuses:
+        run_id = None
+        repo = api.repo
+        context = status
+        url = statuses[status]["target_url"]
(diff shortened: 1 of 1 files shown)


marusak:
Added storing into sqlite DB

store_flakes
```diff
@@ -0,0 +1,81 @@
...
+    cursor = db_conn.cursor()
+
+    init_db(cursor)
+
+    ctx = ssl.create_default_context()
+    if opts.no_check_certificate:
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+
+    if opts.log.startswith("http"):
```
martinpitt, line 44: urllib can deal with file:// URLs, which may make this more uniform.

store_flakes
```diff
@@ -0,0 +1,81 @@
...
+        fp = urllib.request.urlopen(opts.log, context=ctx)
+    else:
+        fp = open(opts.log, "r")
+
+    for line in find_failed_tests(fp):
+        (test, name, retry) = process_line(line)
+        insert(cursor, test, name, 1 if retry else 0)
+
+    fp.close()
```
martinpitt, line 53: If you use `with open...`, you won't need this.

store_flakes
```diff
@@ -0,0 +1,81 @@
...
+
+
+def insert(cursor, test, name, retry):
+    cursor.execute("INSERT INTO CockpitFlakes VALUES ('{0}', '{1}', '{2}')".format(test, name, retry))
+
+
+def find_failed_tests(fp):
+    for line in fp:
+        if isinstance(line, bytes):
+            line = line.decode('utf-8')
```
martinpitt, line 70: If you always use urllib, you don't need this special case, you can always work with bytes. Or, open the file in binary mode.

store_flakes
```diff
@@ -0,0 +1,81 @@
...
+        insert(cursor, test, name, 1 if retry else 0)
+
+    fp.close()
+
+    db_conn.commit()
+    db_conn.close()
+
+
+def init_db(cursor):
+    cursor.execute("CREATE TABLE if not exists CockpitFlakes (test text, name text, retry integer)")
```
martinpitt, line 60: It's not clear what "test" and "name" mean. This is also not just for "Cockpit".

I think for meaningful queries we want to store some more information. E. g. we want something like "get me all the test logs where TestFoo.testBar() failed". So we should have one table `TestRuns` like
```sql
id INTEGER PRIMARY_KEY AUTOINCREMENT,
project CHAR[50],
context CHAR[50],
url VARCHAR(1000)
```

It could use our sink identifiers like `pull-895-20200526-085151-0f1159c6-fedora-31-aws-weldr-lorax` as ID, but just an auto-generated integer will be is faster for processing, and ultimately easier to use.

Then we want to record which tests failed in these runs, in a `Failures` table:
```sql
testname VARCHAR(100) NOT NULL,
retry_reason VARCHAR(1000),
run INTEGER NOT NULL,
FOREIGN KEY (run) REFERENCES TestRuns(id)
```
This also turns the retry reason from a boolean into a string which can be `NULL` (no retry), and we don't lose the information why it was retried (infrastructure or test failure)

WDYT?

store_flakes
```diff
@@ -0,0 +1,81 @@
...
+
+def insert(cursor, test, name, retry):
+    cursor.execute("INSERT INTO CockpitFlakes VALUES ('{0}', '{1}', '{2}')".format(test, name, retry))
+
+
+def find_failed_tests(fp):
+    for line in fp:
+        if isinstance(line, bytes):
+            line = line.decode('utf-8')
+        if line.startswith("not ok "):
```
martinpitt, line 71: Note: This currently doesn't work for external projects which moved to cockpit's testlib 219 (as that broke TAP output), but we shouldn't be concerned about that here - that just needs to be fixed.

store_flakes
```diff
@@ -0,0 +1,81 @@
...
+            yield line
+
+
+def process_line(line):
+    fields = line.split(" ")
+    return (fields[3], fields[4], "RETRY" in line)
```
martinpitt, line 77: This is underdocumented and sounds overly general -- let's please move the splitting into `find_failed_tests()` and directly return the relevant information as a tuple.

marusak:
@martinpitt changed according to your review.
Now when I want to store results from https://github.com/cockpit-project/cockpit/pull/14162 I can do:
`./store-failures --repo=cockpit-project/cockpit 716b589d33a41d4751937b520f2447f5a8c1823a`
and then inspect DB and I see: (just dummy queries, no joins...)
```
 $ sqlite3 cockpit-failures.db
SQLite version 3.31.1 2020-01-27 19:55:54
Enter ".help" for usage hints.
sqlite> select * from TestRuns;
1|cockpit-project/cockpit|centos-8-stream|https://logs.cockpit-project.org/logs/pull-14162-20200528-082123-716b589d-centos-8-stream/log.html
2|cockpit-project/cockpit|ubuntu-stable|https://logs.cockpit-project.org/logs/pull-14162-20200528-082123-716b589d-ubuntu-stable/log.html
3|cockpit-project/cockpit|rhel-8-3|https://logs.cockpit-project.org/logs/pull-14162-20200528-082123-716b589d-rhel-8-3/log.html
4|cockpit-project/cockpit|fedora-31|https://logs.cockpit-project.org/logs/pull-14162-20200528-082123-716b589d-fedora-31/log.html
5|cockpit-project/cockpit|rhel-8-2-distropkg|https://logs.cockpit-project.org/logs/pull-14162-20200528-082123-716b589d-rhel-8-2-distropkg/log.html
6|cockpit-project/cockpit|debian-testing|https://logs.cockpit-project.org/logs/pull-14162-20200528-082123-716b589d-debian-testing/log.html
sqlite> select * from Failures;
test/verify/check-networking-mac TestNetworking.testMac|be robust against unstable tests|1
test/verify/check-networking-team TestTeam.testBasic|be robust against unstable tests|2
test/verify/check-networking-team TestTeam.testActive|be robust against unstable tests|2
test/verify/check-networking-team TestTeam.testActive|be robust against unstable tests|2
test/verify/check-networking-team TestTeam.testActive|None|2
test/verify/check-machines TestMachines.testDiskEdit|be robust against unstable tests|2
test/verify/check-storage-vdo TestStorage.testVdo|be robust against unstable tests|3
test/verify/check-storage-vdo TestStorage.testVdo|be robust against unstable tests|3
test/verify/check-machines TestMachines.testCreateThenInstall|be robust against unstable tests|4
test/verify/check-packagekit TestUpdates.testSecurityOnly|be robust against unstable tests|4
test/verify/check-machines TestMachines.testCreateThenInstall|be robust against unstable tests|5
test/verify/check-storage-resize TestStorage.testGrowShrinkEncryptedHelp|be robust against unstable tests|6
test/verify/check-networking-bond TestBonding.testActive|be robust against unstable tests|6
test/verify/check-accounts TestAccounts.testUnprivileged|be robust against unstable tests|6
test/verify/check-accounts TestAccounts.testBasic|be robust against unstable tests|6
sqlite> 
```

Checked with actual results and it all is correct.

store-failures
```diff
@@ -0,0 +1,112 @@
...
+        url = statuses[status]["target_url"]
+        if "@" in status:
+            parts = status.split("@")
+            context = parts[0]
+            if not parts[1].startswith("bots"):
+                # Can contain branch, as it diffs between master and branches (e.g `cockpit-project/cockpit/rhel-7.9`)
+                repo = parts[1]
+
+        raw_log = url[:-5] if url.endswith(".html") else url
+        with urllib.request.urlopen(raw_log, context=ctx) as fp:
```
martinpitt, line 62: This won't work for CentOS CI logs, similarly to PR #857. logs-https-cockpit.. uses the official SSL cert from CentOS CI.

store-failures
```diff
@@ -0,0 +1,112 @@
...
+    cursor.execute("""CREATE TABLE if not exists Failures
+                    (testname TEXT NOT NULL,
+                     retry_reason TEXT,
+                     run INTEGER NOT NULL,
+                     FOREIGN KEY (run) REFERENCES TestRuns(id))
+                   """)
+
+
+def insert_run(cursor, project, context, url):
+    return cursor.execute("INSERT INTO TestRuns VALUES (null, '{0}', '{1}', '{2}')"
```
martinpitt, line 88: Ah, is `AUTOINCREMENT` implied for a primary key when giving `null`?

Please don't use str.format() for DB commands, but let the cursor.execute() function do its own parameter substitution -- this avoids SQL injection errors when names or URLs happen to contain `'`. I. e. use `?` values in the command string, and supply their values as  separate arguments to execute(). (See https://docs.python.org/2/library/sqlite3.html)

store-failures
marusak: Done

martinpitt:
I tried it on #950 first, and that looks odd:
```
sqlite> select * from Failures;
testNodeNavigation (check_openshift.TestOpenshift)|check_openshift.TestOpenshift|1
testNodeNavigation (check_openshift.TestOpenshift)|check_openshift.TestOpenshift|1
testNodeNavigation (check_openshift.TestOpenshift)|check_openshift.TestOpenshift|1
```
I. e. it doesn't get along well with the different output format of the rhel-7.9 branch. The test name is put into `retry_reason`. The rhel-7.9 branch doesn't have a retry reason. We could restrict this to master only of course, but maybe there's an easy way to fix that -- either by adjusting the parser, or by backporting a test/common/testlib.py change that updates the format to be the same as in master.

Also, the entry is triplicated; even in master, in the example that you show above. Perhaps this should declare `UNIQUE(test_name, retry_reason, run)` and we either ignore "exists" failures on `INSERT`, or use [REPLACE](https://www.sqlitetutorial.net/sqlite-replace-statement/) instead (probably cleaner and simpler).

The same problem happens on `TestRuns`, if you scan the same hash twice (which can happen if you e. g. add and remove a commit). But we can't use `REPLACE` there as that would increment the primary key and break references -- we really need to check there first.

There is also no error if retrieving the SHA fails -- I think we should raise some alarm bell there, as it would indicate a bug in our infrastructure.

On

    ./store-failures --repo cockpit-project/cockpit 34695ff51ad8b10206ae581ecd765eb0492c63e8

from PR #14200 I noticed these:
```
test/verify/check-metrics TestMetrics.testInternal|be robust against unstable tests|7
test/verify/check-metrics TestMetrics.testInternal#||7
test/verify/check-metrics TestMetrics.testPcp|be robust against unstable tests|7
test/verify/check-metrics TestMetrics.testPcp#||7
```
That's due to some bug/race in the logging where it sometimes is missing the line break:
```
# 1 TEST FAILED [32s on 3-ci-srv-01]
not ok 117 test/verify/check-metrics TestMetrics.testPcp# ----------------------------------------------------------------------
```
log.html does that by stripping off `test_header_start`.

I added FIXUPs for all but the first problems, and some extra tweaks. Let me know what you think!

martinpitt:
I now also pushed a FIXUP to get along with the old log format.

store-failures
```diff
@@ -80,6 +80,7 @@ def init_db(cursor):
                     (testname TEXT NOT NULL,
                      retry_reason TEXT,
                      run INTEGER NOT NULL,
+                     UNIQUE (testname, retry_reason, run)
```
marusak, line 97: Thanks @martinpitt for the fixes and improvements!
I am not convinced it we want `Avoid duplicates in Failures ` commit - we are interested to see, how many times test fails for statistical reasons. So having the same record twice makes sense, as it points to the fact, it was retried twice and not only once. We could put there retry number or some other identifier, but I am not sure if it is really needed.

martinpitt: That's a good point, yes -- so please kick it out again on a rebase. It just somehow needs to make sure that scanning a log twice doesn't end up duplicating these entries. So maybe the transaction should start with a 

    DELETE FROM Failures WHERE run = run_id

?

martinpitt:
This looks good, thanks! I feel we should write some tests for this, by feeding a reduced version of some "typical" old and new logs into it (like the ones I linked to from above), but that doesn't need to block this PR. Please squash it, and I'll approve.

marusak:
> Please squash it, and I'll approve.

I thought we could squash on merging.

PR #893, opened 2020-05-25

cockpituous:


martinpitt:
[breaks sssd authentication](https://logs.cockpit-project.org/logs/pull-893-20200525-105936-be5f941b-fedora-testing-cockpit-project-cockpit/log.html#194-2)

martinpitt:
This is a regression in sssd, I filed https://bugzilla.redhat.com/show_bug.cgi?id=1839805 about it, and https://github.com/cockpit-project/bots/issues/897 to track it. Unfortunately the bodhi update already landed, so next image refresh (https://github.com/cockpit-project/bots/pull/895 already?) will also be affected.

abbra:
I don't think it is a regression in SSSD. Rather, this looks like your AD environment only allows RC4-HMAC keys for users  (default state in AD) while Fedora 32+ disabled use of RC4-HMAC via default crypto policy.

martinpitt:
@abbra: Thanks for the pointer! On the [server side](https://github.com/cockpit-project/bots/blob/master/images/scripts/services.setup#L66) we use the [nowsci/samba-domain](https://hub.docker.com/r/nowsci/samba-domain) container. But this is an F31 image, not 32.

PR #898, opened 2020-05-25

cockpituous:


martinpitt:
Retrying, cockpit-ostree fix landed.

PR #908, opened 2020-05-26
Image refresh for rhel-7-9

cockpituous:
Image refresh for rhel-7-9
 * [x] image-refresh rhel-7-9


martinpitt:
Sent https://github.com/weldr/lorax/pull/1023 to compare logs if the lorax failure is a regression.

PR #909, opened 2020-05-27
image-download fixes/improvements

martinpitt:
This makes image-download a lot more informative about what's going on.

### DIFF
--- a/image-download
+++ b/image-download
@@ -81,9 +81,13 @@ def find(name, stores, latest, quiet):
                                                    "--fail", "--cacert", ca, source]
                     start = time.time()
                     output = subprocess.check_output(cmd, universal_newlines=True)
-                    found.append((cmd, output, store, time.time() - start))
+                    duration = time.time() - start
+                    found.append((cmd, output, store, duration))
                     if not quiet:
-                        sys.stderr.write("{0} {1}\n".format(SYMBOLS['present'], store))
+                        sys.stderr.write("{0} {1} [{2}ms]\n".format(
+                            SYMBOLS['present'],
+                            store,
+                            int(duration * 1000 + 0.5)))
                     return True
                 except subprocess.CalledProcessError:
                     return False
--- a/image-download
+++ b/image-download
@@ -68,20 +68,17 @@ def find(name, stores, latest, quiet):
             ai = socket.getaddrinfo(url.hostname, url.port or defport, socket.AF_INET, 0, socket.IPPROTO_TCP)
         except socket.gaierror:
             ai = []
-            message = store
 
         for (family, socktype, proto, canonname, sockaddr) in ai:
-            message = "{scheme}://{0}:{1}{path}".format(*sockaddr, scheme=url.scheme, path=url.path)
-
             def curl(*args):
                 try:
                     cmd = ["curl"] + list(args) + ["--head", "--connect-timeout", "10", "--silent",
                                                    "--fail", "--cacert", ca, source]
                     start = time.time()
                     output = subprocess.check_output(cmd, universal_newlines=True)
-                    found.append((cmd, output, message, time.time() - start))
+                    found.append((cmd, output, store, time.time() - start))
                     if not quiet:
-                        sys.stderr.write(" > {0}\n".format(message))
+                        sys.stderr.write(" > {0}\n".format(store))
                     return True
                 except subprocess.CalledProcessError:
                     return False
@@ -99,7 +96,7 @@ def find(name, stores, latest, quiet):
                 continue
 
             if not quiet:
-                sys.stderr.write(" x {0}\n".format(message))
+                sys.stderr.write(" x {0}\n".format(store))
 
     # If we couldn't find the file, but it exists, we're good
     if not found:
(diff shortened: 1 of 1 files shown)


martinpitt:
The last image refresh that we landed was two days ago, so by now all our bots should have all images cached. Still triggered a test for good measure, but this needs to be tested manually mostly. Once #908 lands, we can trigger a few tests on the refreshed rhel-7.9 image and check that it downloads properly.

martinpitt:
Looks like this now for me. Public image:
```
❱❱❱ ./image-download fedora-32
 ✔ https://images-cockpit.apps.ci.centos.org/ [2746ms] 
 ⨯ https://cockpit-11.e2e.bos.redhat.com:8493
 ✔ https://10.29.163.169:8493/ [870ms] 🐌
❯❯ https://images-cockpit.apps.ci.centos.org/fedora-32-09c1ab4c1d8422b96ca2222f2957b4d4b6dec4fd55c2d0f4bca32522c3f51dec.qcow2
```

Internal images skip the public image store testing, and prefer cockpit-11 over AWS:
```
❱❱❱ ./image-download rhel-8-2
 ✔ https://cockpit-11.e2e.bos.redhat.com:8493 [665ms] 
 ✔ https://10.29.163.169:8493/ [570ms] 🐌
❯❯ https://cockpit-11.e2e.bos.redhat.com:8493/rhel-8-2-da12a57b3312474745b484942c6fee8734199277d69a379a589027ad7b7109e0.qcow2
```

martinpitt:
I took out the penalty thing for now, I'll check if I can meaningfully measure download speed instead which offsets the ~ 2.5s latency cost of the CentOS CI server. This leaves the actual main problem open again, but let's do that in a follow-up. The other fixes should be mostly straightforward.

image-download
```diff
@@ -54,6 +54,11 @@ CONFIG = "~/.config/image-stores"
 DEVNULL = open("/dev/null", "r+")
 EPOCH = "Thu, 1 Jan 1970 00:00:00 GMT"
 
+if os.isatty(1):
+    SYMBOLS = {'present': ' ✔', 'absent': ' ⨯', 'selected': '❯❯'}
```
allisonkarlitskaya, line 58: have to admit I'm a bit sad to see the snail go :rofl:

image-download
```diff
@@ -68,20 +73,21 @@ def find(name, stores, latest, quiet):
             ai = socket.getaddrinfo(url.hostname, url.port or defport, socket.AF_INET, 0, socket.IPPROTO_TCP)
         except socket.gaierror:
             ai = []
-            message = store
 
         for (family, socktype, proto, canonname, sockaddr) in ai:
-            message = "{scheme}://{0}:{1}{path}".format(*sockaddr, scheme=url.scheme, path=url.path)
```
allisonkarlitskaya, line 74: So the reason we were showing the IP address is because we clearly anticipated multiple IPs per name here.  If that's no longer the case, then we ought to simplify this code much more substantially.

image-download
```diff
@@ -68,20 +73,21 @@ def find(name, stores, latest, quiet):
...
                     output = subprocess.check_output(cmd, universal_newlines=True)
-                    found.append((cmd, output, message, time.time() - start))
+                    duration = time.time() - start
+                    found.append((cmd, output, store, duration))
                     if not quiet:
-                        sys.stderr.write(" > {0}\n".format(message))
+                        sys.stderr.write("{0} {1} [{2}ms]\n".format(
+                            SYMBOLS['present'],
+                            store,
+                            int(duration * 1000 + 0.5)))
```
allisonkarlitskaya, line 90: This addition is pretty nice, actually.  It gives a bit of a hint as to the criteria in use.

image-download
```diff
@@ -120,19 +126,23 @@ def find(name, stores, latest, quiet):
...
     if not stores:
         config = os.path.expanduser(CONFIG)
         if os.path.exists(config):
             with open(config, 'r') as fp:
                 stores = fp.read().strip().split("\n")
         else:
             stores = []
-        stores += PUBLIC_STORES
+        # skip testing public stores for private images
+        if not name.startswith("rhel") and not name.startswith("windows"):
```
allisonkarlitskaya, line 144: thanks :)

image-download
martinpitt: As I said I'm perfectly happy to leave the penalty thing as an imperfect solution for now, until we figure out something better. I started to work on parallel downloads and such, but it will look rather messy.

image-download
martinpitt: I'm not sure about the history here. I think that was from a time when we had OpenShift running ourselves, and used http://cockpit-images.verify.svc.cluster.local . This was dropped in e6d417cc65644a8195c0c8130840a75d48d20374 . But I honestly don't know why it was necessary in the first place -- services and routes should take care of the load balancing, and I'm not sure why one needs to pin down an IP for that. I'll try to drop it.

martinpitt: Ah, of course -- we use that with `curl --resolve` (so that we can use the "cockpit-test" SNI), which requires an IP. So this has to do a DNS resolution.

PR #910, opened 2020-05-28
image-download: Select store by transfer speed instead of latency

martinpitt:
Measuring the duration of the HEAD request is not a good metric, as it
measures latency, not throughput. The CentOS CI store takes some extra 2.5s
to connect (OpenShift route/service overhead?), but otherwise is
lighting fast. OTOH our AWS store has very little latency, but only a
laughable bandwidth.

Thus image-download usually selected the AWS mirror, causing downloads
to take several hours.

Fix this by starting the download from all stores in parallel, and
decide after 7s from which one we can pull the fastest. Keep that one,
and kill all other downloads.


garrett:
This is working well for me! Thanks! Looking forward to it being merged.

PR #912, opened 2020-05-28
image-create: Put VM image in /var/tmp/

martinpitt:
At least in our CI, testvm.get_temp_dir() points to a tmpfs file, so
that tests can put VM overlays into RAM and are not affected by slow
I/O. But that doesn't work for image-create, as that's the entire image,
not just the overlay. This often caused the VM to crash due to ENOFS,
particularly during the "fill disk with zeros" phase.

Use /var/tmp/ instead, which is meant for large files.

----

This should fix the large set of failed image builds. I'll trigger one image here as an example, then we can retry the others after landing this.

### DIFF
--- a/image-create
+++ b/image-create
@@ -55,14 +55,11 @@ if args.image in ["continuous-atomic", "rhel-atomic", "fedora-coreos", "services
 
 class MachineBuilder:
     def __init__(self, machine):
-        tempdir = testvm.get_temp_dir()
         self.machine = machine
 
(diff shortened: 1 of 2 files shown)


cockpituous:
image-refresh in progress on 4-ci-srv-04.
Log: https://logs.cockpit-project.org/logs/image-refresh-912-20200528-091919/

cockpituous:
image-refresh continuous-atomic done: https://github.com/cockpituous/bots/commits/image-refresh-continuous-atomic-20200528-102525

PR #872, opened 2020-05-28
Image refresh for rhel-atomic

cockpituous:
Image refresh for rhel-atomic


martinpitt:
[sosreport failure](https://logs.cockpit-project.org/logs/pull-872-20200528-122143-12131968-rhel-atomic-cockpit-project-cockpit-rhel-7.9/log.html), but I think that's a common flake.

PR #926, opened 2020-06-02
machine: Support specifying an explicit overlay dir for VirtMachine

martinpitt:
The default tmpfs overlay dir is not appropriate for all projects.
E.g. lorax' tests needs huge overlays which overflow /tmp.

See https://github.com/weldr/lorax/pull/1027

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -286,7 +286,8 @@ class VirtMachine(Machine):
     memory_mb = None
     cpus = None
 
-    def __init__(self, image, networking=None, maintain=False, memory_mb=None, cpus=None, graphics=False, **args):
+    def __init__(self, image, networking=None, maintain=False, memory_mb=None, cpus=None,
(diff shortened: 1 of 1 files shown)


martinpitt:
I triggered a lorax tests that uses this in https://github.com/weldr/lorax/pull/1027

martinpitt:
Ugh, we don't regularly run starter-kit on f32 yet -- seems [we broke it](https://logs.cockpit-project.org/logs/pull-926-20200602-060123-a945dedc-fedora-32-cockpit-project-starter-kit/log.html). But that's unrelated to this PR, I'll look into that separately.

marusak:
The broken strater-kit is due to new navigation.
I changed this in cockpit tests, see for example [here](https://github.com/cockpit-project/cockpit/pull/13663/files#diff-d506bc1e0ed906165b3f2490e3e0795aR116)

martinpitt:
Unrelated starter-kit failure fixed in https://github.com/cockpit-project/starter-kit/pull/325

PR #927, opened 2020-06-02
testmap: Move starter-kit to fedora-32

martinpitt:


martinpitt:
I fixed fedora-32 in https://github.com/cockpit-project/starter-kit/pull/325 , but I'll also trigger a firefox test - let's see how that works.

martinpitt:
I just switched starter-kit's branch protection rules to require fedora-32 and not require fedora-31 any more.

PR #936, opened 2020-06-04

cockpituous:


martinpitt:
Seems libvirt lost its default network, needs investigation.

martinpitt:
Rather old now, let's just try this again.

PR #943, opened 2020-06-04
Eliminate $TEST_DATA

martinpitt:
This variable influenced too many files which don't belong together.
Replace its usage with more sensible defaults and better storage
behaviour:

 * Downloaded images: These are precious as a local developer, so put
   them into the XDG cache dir by default: ~/.cache/cockpit-images/.
   With that, there is no configuration needed, and one can clean git
   trees, have multiple checkouts etc. without problems. The path can
   still be changed with either a symlink or the
   cockpit.bots/images-data-dir git config.

   In CI the images are usually on a persistent external volume.
   ~/.cache/cockpit-images now *only* contains images, so it's easier to
   set up the volume without putting a lot of irrelevant crap into it
   (such as overlays, built rpms, lock files, etc.)

 * Ephemeral VM overlays: These should never be put into a container
   volume, and they should be on lightning fast storage (ideally, on
   tmpfs). By default, put them into tmp/run/ to keep the current
   behaviour (without `$TEST_DATA`), but allow bots to change the
   location with `$TEST_OVERLAY_DIR`.

 * GitHub cache: This is safe to share with other bots, but they are not
   particularly precious nor big. So we don't need to special-case them
   for CI. Keep them in ~/.cache/github/, there's no need to override
   that.

 * SSH master control sockets: These must *never ever* be shared with
   other bots. Put them into /tmp/.cockpit-test-resources/, where
   the image/port lock files already live.

Eliminate the usage of get_temp_dir(). It's too unspecific and can't fit
all the diverging use cases above. Still keep a backwards compatible
shim until cockpit drops its usage in all of its branches (see
https://github.com/cockpit-project/cockpit/pull/14190).

----

This can't yet land for CI, but it does work fine locally, so I'm interested in
some feedback.

 - [x] Adjust cockpituous and cockpit/tasks container for ~/.cache/cockpit-images and `$TEST_OVERLAY_DIR`: https://github.com/cockpit-project/cockpituous/pull/329
 - [x] [Announce the change to cockpit-devel](https://lists.fedorahosted.org/archives/list/cockpit-devel@lists.fedorahosted.org/thread/FETLZZHRNMJKZYCKLDUIINXCK4X5ZUSU/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,19 +49,12 @@ then please [allow][1] `qemu-bridge-helper` to access the bridge settings.
 
 ## Image location
 
-Downloaded images are stored into ./images/ by default. This is often
-impractical as you cannot easily clean your tree without losing them, nor can
-they be easily put into a separate container volume.
-
-You can set the `cockpit.bots.images-data-dir` variable with
-`git config` to a directory where to store the pristine virtual machine images.
-For example:
+Downloaded images are stored into ~/.cache/cockpit-images/ by default. If you
+want to change that, you can set the `cockpit.bots.images-data-dir` variable
+with `git config` to a directory where to store the pristine virtual machine
+images.  For example:
 
     $ git config cockpit.bots.images-data-dir /srv/cockpit/images
-
-Alternatively, you can set the `$TEST_DATA` environment variable, then images
-will be put into `$TEST_DATA/images/`. However, this is deprecated and will go
-away soon.
 
 ## Tests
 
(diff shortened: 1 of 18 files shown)


martinpitt:
@marusak: firefox/composer looks pretty bad -- presumably we need to update it with the same user.prefs things?

martinpitt:
Hmm, and cockpit/firefox isn't that happy either

martinpitt:
FTR, the triggered tests above are *not* related to this PR, but for the new firefox introduced in https://github.com/cockpit-project/cockpituous/pull/329 -- I just happened to use this PR to run test tests.

PR #945, opened 2020-06-04
Image refresh for centos-7

cockpituous:
Image refresh for centos-7
 * [x] image-refresh centos-7


martinpitt:
Ugh, starter-kit failure:
```
RPM needs to be updated
ERROR You need to update rpm to handle:
rpmlib(PayloadIsZstd) <= 5.4.18-1 is needed by cockpit-starter-kit-1-1.fc32.noarch
```

PR #946, opened 2020-06-04
naughty: Drop exact line number from 897 pattern

martinpitt:
The test often changes, and then this won't match any more.


martinpitt:
@KKoukiou : Yes, the "fedora-32@bots#946" test, and it [succeeded](https://logs.cockpit-project.org/logs/pull-14187-20200604-155854-0593dbc0-fedora-32-bots-946/log.html).

Issue #948, opened 2020-06-05
"ip_finish_output: general protection fault" kernel crash on fedora-coreos


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1844318

cockpituous:
verify/fedora-coreos
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testBasic (__main__.TestSystemInfo)

DevTools listening on ws://127.0.0.1:10088/devtools/browser/1599fbb4-6a8a-410c-bf2d-d0b8b5564c35
[0605/114623.187564:ERROR:egl_util.cc(60)] Failed to load GLES library: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: cannot open shared object file: No such file or directory
[0605/114623.189806:ERROR:viz_main_impl.cc(159)] Exiting GPU process due to errors during initialization
CDP: {"source":"security","level":"error","text":"Refused to apply style from 'http://127.0.0.2:9291/cockpit/$b47917fc39265635e84707ed591912a5bb81f55caa7d4249c87eb5f0b3a1fb49/shell/nav.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.","timestamp":1591357584638.384,"url":"http://127.0.0.2:9291/system"}
CDP: {"source":"security","level":"error","text":"Refused to apply inline style because it violates the following Content Security Policy directive: \"default-src 'self' http://127.0.0.2:9291\". Either the 'unsafe-inline' keyword, a hash ('sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.\n","timestamp":1591357584813.9219,"url":"http://127.0.0.2:9291/cockpit/$b47917fc39265635e84707ed591912a5bb81f55caa7d4249c87eb5f0b3a1fb49/shell/index.js","lineNumber":14,"stackTrace":"(minified)"}
> warning: grep: /sys/class/dmi/id/power/autosuspend_delay_ms: Input/output error

> warning: # localhost:22 SSH-2.0-OpenSSH_8.3
# localhost:22 SSH-2.0-OpenSSH_8.3
[...]

PR #949, opened 2020-06-05
naughty: Add pattern for kernel oops on fedora-coreos

martinpitt:
Some sshd operation in TestSystemInfo.testBasic causes a kernel oops on
Fedora CoreOS 32. It's not predictable which exact check fails, so just
match on any failure in that particular test for now.

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1844318
Known issue #948

### DIFF
--- a/naughty/fedora-coreos/948-ip_finish_output-oops
+++ b/naughty/fedora-coreos/948-ip_finish_output-oops
@@ -1,0 +1,2 @@
+Traceback (most recent call last):
+  File "test/verify/check-system-info", line *, in testBasic


martinpitt:
Fixes [this failure](https://logs.cockpit-project.org/logs/pull-14187-20200604-192153-0593dbc0-fedora-coreos/log.html#67-2), which was introduced in the recent fedora-coreos refresh. That moved the image from Fedora 31 to 32 and a new kernel/podman/etc.

martinpitt:
Bah, and of course *this* test run succeeded. I tested the pattern locally though (I have no trouble reproducing the oops), and it matched fine.

PR #951, opened 2020-06-05
Fix BOTS_DIR imports harder

martinpitt:
Commit 8a96eba0c50f did not catch all of it.

---

This didn't actually break most scripts, but it did break make-checkout really hard: https://logs.cockpit-project.org/logs/pull-950-20200605-080349-8c850103-fedora-32-cockpit-project-cockpit/log.html

### DIFF
--- a/make-checkout
+++ b/make-checkout
@@ -23,7 +23,7 @@ import shutil
 import subprocess
 import sys
 from task import github
-from machine.machine_core.directories import BASE_DIR
+from machine.machine_core.constants import BASE_DIR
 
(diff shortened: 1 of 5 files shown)


marusak:
hmm, not:
```
  File "./make-checkout", line 26, in <module>
    from machine.machine_core.directories import BASE_DIR
ImportError: cannot import name 'BASE_DIR' from 'machine.machine_core.directories' (/build/cockpit-project/bots/machine/machine_core/directories.py)
```

PR #957, opened 2020-06-08
testmap: Drop RHEL 7 contexts from cockpit-composer RHEL 8 branch

martinpitt:
These don't work any more due to incompatible rpm format, and
structurally don't make sense anyway.


martinpitt:
They currently fail [like this](https://logs.cockpit-project.org/logs/pull-955-20200607-171358-8757b34c-rhel-7-8-firefox-osbuild-cockpit-composer-rhel-8/log.html) and block image refreshes, see PR #956

PR #958, opened 2020-06-08
testmap: Build rhel-atomic rpms on rhel-7-9

martinpitt:
This makes sure that it works with the next stable release. This is also
the last remaining consumer of the rhel-7-8 image.


martinpitt:
Needs https://github.com/cockpit-project/cockpit-ostree/pull/97

PR #960, opened 2020-06-08
testmap: Move subscription-manager to rhel-8-3

martinpitt:
I triggered an initial test in https://github.com/candlepin/subscription-manager/pull/2274 . Once that works, I'll trigger it here, too.


martinpitt:
Yep, [works](https://logs.cockpit-project.org/logs/pull-2274-20200608-101748-9c0f3c05-rhel-8-3/log.html)

PR #961, opened 2020-06-08
testmap: Move lorax rhel8-branch to rhel-8-3

martinpitt:


martinpitt:
anaconda on RHEL 8.3 is unhappy.. that could be the critical bug that was discussed today.

bcl:
I've filed this:

https://bugzilla.redhat.com/show_bug.cgi?id=1846156

martinpitt:
Thanks @bcl! There's an anaconda fix for now, and this points out a real bug in the distro, so it's not a broken image.

PR #965, opened 2020-06-09
Image refresh for fedora-32

cockpituous:
Image refresh for fedora-32
 * [x] image-refresh fedora-32


martinpitt:
[Also FIPS error](https://logs.cockpit-project.org/logs/pull-965-20200609-094904-37fda8de-fedora-32-cockpit-project-cockpit/log.html#129-2), like in #968.

martinpitt:
Same issue as in #968:

    remotectl: /etc/cockpit/ws-certs.d/0-self-signed.cert: Could not parse PEM certificate: The request is invalid.

Something in FIPS got stricter to not accept our certififcate any more.

martinpitt:
Tracked in https://github.com/cockpit-project/cockpit/issues/14074 , moving to blocked instead.

martinpitt:
I investigated/reported the GnuTLS regression and added a naughty pattern.

martinpitt:
You already approved the same pattern for fedora-testing, but for formality's sake.. :)

PR #966, opened 2020-06-09
Storing flakes, webhook part

marusak:
Depends on:

- [x] https://github.com/cockpit-project/cockpituous/pull/331
- [x] https://github.com/cockpit-project/bots/pull/894

### DIFF
--- a/run-queue
+++ b/run-queue
@@ -32,6 +32,8 @@ from task import redhat_network, distributed_queue
 
 logging.basicConfig(level=logging.INFO)
 
+statistics_queue = os.environ.get("RUN_STATISTICS_QUEUE")
+
 # Returns a command to execute and the delivery tag needed to ack the message
 
 
@@ -75,7 +77,7 @@ def consume_webhook_queue(channel, amqp):
 # Returns a command to execute and the delivery tag needed to ack the message
 
 
-def consume_task_queue(channel, amqp, declare_public_result, declare_rhel_result):
+def consume_task_queue(channel, amqp, declare_public_result, declare_rhel_result, declare_stats_result):
     queue = 'public'
     if redhat_network():
         # Try the rhel queue if the public queue is empty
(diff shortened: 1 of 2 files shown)


run-queue
```diff
@@ -85,6 +87,9 @@ def consume_task_queue(channel, amqp, declare_public_result, declare_rhel_result
         elif declare_rhel_result.method.message_count > 0:
             queue = ['public', 'rhel'][random.randrange(2)]
 
+    if statistics_queue and declare_stats_result.method.message_count > 0:
+        queue = [queue, 'statistics'][random.randrange(2)]
```
martinpitt, line 105: This will prefer running things from the statistics queue (50% of the time, and public and rhel get 25% each). That's actually fine -- that queue is fast to process, and we *want* to prefer it. I just wonder if we shouldn't even bother with the 50% chance, and always process statistics first (same like the webhook queue, which is also drained first).

run-queue
```diff
@@ -58,6 +60,18 @@ def consume_webhook_queue(channel, amqp):
         if request['action'] in ['opened', 'synchronize']:
```
martinpitt, line 60: Would you mind changing this to `action`, so that all `if`s use the same variable? (It's just aliased above). Also, can we please change this to `elif`? It doesn't make a practical difference, but it more clearly points out that these are non-overlapping alternatives.

run-queue
marusak: Yeah, I realized that this would prefer stats queue. But I agree that it is fine and also kind-of wanted.
I would not go fully with draining stats queue first as it is **not that** important.

run-queue
marusak: done

task/distributed_queue.py
```diff
@@ -46,6 +46,9 @@
     'public': {
         "x-max-priority": MAX_PRIORITY
     },
+    'statistical': {
```
martinpitt, line 49: This should be `statistics`.

marusak: good catch, fixed

PR #979, opened 2020-06-14
task: Stop testing starter-kit on centos-7

martinpitt:
RPMs built on Fedora 32 do not work any more in CentOS 7 due to changed
RPM binary format (zstd compression).

starter-kit is already tested on CentOS 8, which is enough.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -59,7 +59,6 @@ REPO_BRANCH_CONTEXT = {
     },
     'cockpit-project/starter-kit': {
         'master': [
-            'centos-7',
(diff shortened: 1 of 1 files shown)


martinpitt:
[This failure](https://logs.cockpit-project.org/logs/pull-329-20200613-104740-9794d28a-centos-7/log.html) was fallout from a recent update of our tasks container, F32 introduced new zstd RPM compression. Similar to the [recent cockpit-ostree update](https://github.com/cockpit-project/cockpit-ostree/pull/97).

Issue #971, opened 2020-06-10
gnutls 3.6.14 broken in FIPS mode: FIPS140-2 self testing part 2 failed


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1845806

cockpituous:
verify/fedora-32
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testFIPS (__main__.TestMultiMachine)
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.

DevTools listening on ws://127.0.0.1:10093/devtools/browser/f7a9058a-058a-476d-bd31-a6b6b08f425e
[0610/070624.292410:ERROR:egl_util.cc(60)] Failed to load GLES library: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: cannot open shared object file: No such file or directory
[0610/070624.294030:ERROR:viz_main_impl.cc(159)] Exiting GPU process due to errors during initialization
Traceback (most recent call last):
  File "/usr/lib64/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "test/verify/check-multi-machine", line 475, in testFIPS
    self.login_and_go("/dashboard")
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 925, in login_and_go
    self.browser.login_and_go(path, user=user, host=host, authorized=authorized, superuser=superuser, urlroot=urlroot, tls=tls)
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 554, in login_and_go
    self.try_login(user, password, authorized=authorized, superuser=superuser)
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 536, in try_login
    self.wait_visible("#login")
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 392, in wait_visible
    self.wait_present(selector)
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 383, in wait_present
    self.wait_js_func('ph_is_present', selector)
[...]

PR #987, opened 2020-06-16
Image refresh for fedora-31

cockpituous:
Image refresh for fedora-31
 * [x] image-refresh fedora-31
 * [x] https://github.com/osbuild/cockpit-composer/pull/1007


martinpitt:
[osbuild-composer fails](https://logs.cockpit-project.org/logs/pull-987-20200617-043019-3496877a-fedora-31-osbuild-cockpit-composer/log.html);
```
+ pvcreate /dev/vda3
  Physical volume "/dev/vda3" successfully created.
+ vgextend fedora_ibm-p8-kvm-03-guest-02 /dev/vda3
  Volume group "fedora_ibm-p8-kvm-03-guest-02" not found.
  Cannot process volume group fedora_ibm-p8-kvm-03-guest-02
Command '/var/tmp/vm.install' returned non-zero exit status 5.
```
It did that in #989 as well. @henrywang , @larskarlitski , does that ring a bell? Some change in the default partitioning in Anaconda?

larskarlitski:
> It did that in #989 as well. @henrywang , @larskarlitski , does that ring a bell? Some change in the default partitioning in Anaconda?

Isn't there a better way to make this work across distros? It seems brittle to me that we need to mention the device and LVM names there.

martinpitt:
The device name is hard to figure out, as it's not quite predictable what that `fdisk` command does. But at least the VG name could be detected through `vgs --noheadings -o vg_name`.

martinpitt:
I sent https://github.com/osbuild/cockpit-composer/pull/1007 to fix this.

martinpitt:
[two failures](https://logs.cockpit-project.org/logs/pull-987-20200617-110931-3496877a-fedora-31-firefox-cockpit-project-cockpit/log.html), but on firefox only, which suggests it's not generally broken. Retrying to compare.

marusak:
pretty sure it is just https://github.com/cockpit-project/bots/issues/971 but has a slightly different backtrace on firefox.

martinpitt:
@marusak: Ah right, it's in expect_load() instead of wait_visible(), which is sensitive to the browser. However, your new pattern is more general than the existing one, so it would make more sense to just take out the `wait_visibile()` line from the existing pattern. But no need to push again just for that.

marusak:
Current, chrome, pattern:
```
  File "test/verify/check-multi-machine", line *, in testFIPS                   
    self.login_and_go("/dashboard")                                             
*                                                                               
    self.wait_visible("#login")                                                 
*                                                                               
testlib.Error: timeout 
```
What we get from firefox:
```
  File "test/verify/check-multi-machine", line 475, in testFIPS
    self.login_and_go("/dashboard")
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 925, in login_and_go
    self.browser.login_and_go(path, user=user, host=host, authorized=authorized, superuser=superuser, urlroot=urlroot, tls=tls)
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 552, in login_and_go
    self.open(href, tls=tls)
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 126, in open
    self.expect_load()
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/testlib.py", line 140, in expect_load
    self.cdp.command('expectLoad(%i)' % (self.cdp.timeout * 1000))
  File "/home/user/cockpit-project/bots/make-checkout-workdir/test/common/cdp.py", line 136, in command
    raise RuntimeError(res["error"])
RuntimeError: timed out waiting for page load
```

> it would make more sense to just take out the `wait_visibile()` line from the existing pattern.

I would need to take also out `testlib.Error: timeout`, which would make the pattern even more generic. But I can add `self.expect_load()` into the firefox pattern, if you think it would be better.

martinpitt:
@marusak: Ah,  right. Let's not overthink this, it's good enough and hopefully the gnutls bug will be fixed soon anyway.

PR #988, opened 2020-06-16
Image refresh for ubuntu-2004

cockpituous:
Image refresh for ubuntu-2004


marusak:
[TestServices.testBasic](https://logs.cockpit-project.org/logs/pull-988-20200616-222132-9581542f-ubuntu-2004-cockpit-project-cockpit/log.html) seems unhappy, retrying to compare.

marusak:
Ah, just looked at the test, and it should be fine. The test has one problem - it checks if next trigger of timer is 'Today', but the unit has `OnBootSec=100min`. So if we bootup the VM after 22:20, it will show 'Tommorow'. Similarly `test/timer` triggers every 2 minutes, so in a very edge cases it might fail in a similar fashion, but that is very unlikely (and retry should cover it).

PR #999, opened 2020-06-22
tests: Start using Fedora 32 for weldr/lorax@master

atodorov:


marusak:
Those F32s are not tested. Better workflow is to put all the new tests into manual, we can merge that, then you can open PR in weldr, trigger those F32 tests and if they succeed you can move them into `master` with reference to such PR.

atodorov:
Updated but what happened with the automating kick-off of linked test runs when the map changes ? I was expecting to see all of these started against lorax.

martinpitt:
@atodorov : A while ago we switched to triggering relevant tests for bots PRs manually. We have too many projects now to *always* run *all* tests for each bots PR, it takes too long and is just a waste.

PR #989, opened 2020-06-16
Image refresh for fedora-32

cockpituous:
Image refresh for fedora-32
 * [x] image-refresh fedora-32
 * [x] https://github.com/cockpit-project/cockpit/pull/14242
 * [x] https://github.com/osbuild/cockpit-composer/pull/1007
 * [x] https://github.com/cockpit-project/cockpit/pull/14245

PR #1006, opened 2020-06-24
Image refresh for rhel-8-2

cockpituous:
Image refresh for rhel-8-2
 * [x] image-refresh rhel-8-2


marusak:
Hmm, c-composer on rhel-8 branch has some problem but only on firefox. @henrywang can you please take a look and let us know if it is a problem we should block this refresh on? (this branch seems like the old lorax API, not even sure if you still want to test this).

henrywang:
Thanks @marusak. It's because of [issue #989 ](https://github.com/osbuild/cockpit-composer/issues/989). And it's not fixed yet. I think we can land this PR.

PR #1007, opened 2020-06-24
Image refresh for rhel-8-3

cockpituous:
Image refresh for rhel-8-3
 * [x] image-refresh rhel-8-3


marusak:
bah, `tuned-adm` is missing some library and thus fails with `ModuleNotFoundError: No module named 'utilist'`...

PR #1018, opened 2020-06-29
machine: Add append, owner, and permission parameters to SSHConnection.write()

martinpitt:
These are a common operations in tests, let's make them declarative.

---

I tested this in a local rewrite of https://github.com/cockpit-project/cockpit/pull/14264 .

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -437,17 +437,24 @@ class SSHConnection(object):
         except:
             self.message("Error while downloading directory '{0}'".format(source))
 
-    def write(self, dest, content):
+    def write(self, dest, content, append=False, owner=None, perm=None):
         """Write a file into the test machine
 
         Arguments:
             content: Raw data to write to file
             dest: The file name in the machine to write to
+            append: If True, append to existing file instead of replacing it
+            owner: If set, call chown on the file with the given owner string
+            perm: Optional file permission as chmod shell string (e.g. "0600")
         """
         assert dest
         assert self.ssh_address
 
-        cmd = "cat > '%s'" % dest
+        cmd = "cat %s '%s'" % (append and '>>' or '>', dest)
+        if owner:
+            cmd += " && chown '%s' '%s'" % (owner, dest)
+        if perm:
+            cmd += " && chmod '%s' '%s'" % (perm, dest)
         self.execute(command=cmd, input=content)
 
     def spawn(self, shell_cmd, log_id):


machine/machine_core/ssh_connection.py
```diff
@@ -437,17 +437,21 @@ def download_dir(self, source, dest, relative_dir="."):
         except:
             self.message("Error while downloading directory '{0}'".format(source))
 
-    def write(self, dest, content):
+    def write(self, dest, content, append=False, mode=None):
```
allisonkarlitskaya, line 440: `chmod` might be a better name than `mode`.  it's less accurate, but it's much clearer.  `mode` is too overloaded.  `perm` is another possibility.  `mode` is also fine.  your choice :)

`owner=` (or similarly, `chown=`) string e.g. `admin:admin` or so, might be a nice addition.

machine/machine_core/ssh_connection.py
```diff
@@ -437,17 +437,21 @@ def download_dir(self, source, dest, relative_dir="."):
...
         """Write a file into the test machine
 
         Arguments:
             content: Raw data to write to file
             dest: The file name in the machine to write to
+            append: If True, append to existing file instead of replacing it
+            mode: Optional file access mode as shell string (e.g. "0600")
```
allisonkarlitskaya, line 447: `append` and `chmod` together is strange to imagine...

machine/machine_core/ssh_connection.py
```diff
@@ -437,17 +437,21 @@ def download_dir(self, source, dest, relative_dir="."):
...
         """
         assert dest
         assert self.ssh_address
 
-        cmd = "cat > '%s'" % dest
+        cmd = "cat %s '%s'" % (append and '>>' or '>', dest)
+        if mode:
+            cmd += ' && chmod %s %s' % (mode, dest)
```
allisonkarlitskaya, line 454: Please use an equivalent amount of quoting for both appearances of `dest`

machine/machine_core/ssh_connection.py
martinpitt: Right, perm= makes sense, and I added owner= as well. We don't use it anywhere right now, but at least it's there when we want to.

machine/machine_core/ssh_connection.py
martinpitt: True that, same for append+owner, but *technically* it works.

machine/machine_core/ssh_connection.py
martinpitt: Well spotted, done.

PR #1027, opened 2020-06-30
Generate page with graphs for the most common test failures

marusak:
This is just a rough draft to show the direction. @martinpitt WDYT of this concept?

The script for generating would be probably part of `store-failures` or at least would call it.

It is just tables for now. For some timeline graphs we would need time information which we don't store in DB. We could have some pie charts or so. Up for discussion.
Surely I want to give it a bit of css love.

How to run:
`./generate-stats --db test-failures.db > data.js`
`firefox index.html`

### DIFF
--- a/failure-stats.html
+++ b/failure-stats.html
@@ -1,0 +1,75 @@
+<!DOCTYPE html>
+<meta charset="utf8" />
+<html>
+  <script type="text/javascript" src="failure-stats.js"></script>
+  <style>
+    body {
+        display: grid;
+        grid-template-columns: auto auto;
+    }
+    body > div {
+        margin: 10px;
+    }
+    table {
+        width: 100%;
+    }
+    tr:nth-child(even) {
+        background-color: #f2f2f2;
+    }
+    th, td {
+        padding: 10px;
+        text-align: left;
+    }
+    a:first-child {
+        padding-right: 10px;
+    }
+  </style>
+  <body>
+  </body>
+  <script type="text/javascript">
+    const body = document.querySelector("body");
+    tables.forEach(t => {
+        const wrapper = document.createElement("DIV");
+
+        // Create title of the table
+        const title = document.createElement("H2");
+        title.textContent = t.name;
+        wrapper.appendChild(title);
+
+        // Create table itself
+        const table = document.createElement("TABLE");
+
+        // Create heading
+        let row = document.createElement("TR");
+        t.columns.forEach(c => {
+            const th = document.createElement("TH");
+            th.textContent = c;
+            row.appendChild(th);
+        });
+        table.appendChild(row);
+
+        // Fill the table with data
+        t.data.forEach(r => {
+            row = document.createElement("TR");
+            r.forEach(d => {
+                // Array with logs
+                const td = document.createElement("TD");
+                if (typeof d === "object")
+                    d.forEach(a => {
+                        const ea = document.createElement("A");
+                        ea.textContent = "log";
+                        ea.href = a;
+                        td.appendChild(ea);
+                    });
+                else
+                    td.textContent = d;
+                row.appendChild(td);
+            });
+            table.appendChild(row);
+        });
+
+        wrapper.appendChild(table);
+        body.appendChild(wrapper);
+    });
+  </script>
+</html>
--- a/generate-stats
+++ b/generate-stats
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import sys
+import time
+import sqlite3
+import argparse
+import logging
+import json
+
+# Seconds in two weeks
+TWO_WEEKS = 1209600
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Generate statistics about failed tests')
+    parser.add_argument("--db", default="cockpit-failures.db", help="Database name")
+    parser.add_argument("--verbose", action="store_true", help="Enable verbose logging")
+    opts = parser.parse_args()
+
+    if opts.verbose:
+        logging.basicConfig(level=logging.DEBUG)
+
+    db_conn = sqlite3.connect(opts.db)
+    cursor = db_conn.cursor()
+
+    all_tables = []
+    since = time.time() - TWO_WEEKS
+
+    # Get all failures sorted by number of occurrences
+    top10 = []
+    rows = cursor.execute("""\
+            SELECT testname, COUNT(run)
+            FROM Failures
+            JOIN TestRuns ON Failures.run = TestRuns.id
+            WHERE TestRuns.time > ?
+            GROUP BY testname
+            ORDER BY COUNT(run) DESC
+            LIMIT 10;""", (since, )).fetchall()
+    for row in rows:
+        urls = cursor.execute("""\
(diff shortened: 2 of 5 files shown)


marusak:
gaphz

generate-stats
```diff
@@ -0,0 +1,83 @@
...
+import argparse
+import logging
+import re
+import json
+
+from task import github
+from machine.machine_core.constants import IMAGES_DIR
+
+def main():
+    parser = argparse.ArgumentParser(description='Store information about failed tests')
```
martinpitt, line 34: How about "generate statistics about failed tests"?

generate-stats
```diff
@@ -0,0 +1,83 @@
...
+import ssl
+import sys
+import sqlite3
+import urllib.request
+import argparse
+import logging
+import re
+import json
+
+from task import github
```
martinpitt, line 30: Not needed.

generate-stats
```diff
@@ -0,0 +1,83 @@
...
+        logging.basicConfig(level=logging.DEBUG)
+
+    db_conn = sqlite3.connect(opts.db)
+    cursor = db_conn.cursor()
+
+    all_tables = []
+
+    # Get all failures sorted by number of occurrences
+    top10 = []
+    rows = cursor.execute("select testname, COUNT(run) from Failures GROUP BY testname ORDER BY COUNT(run) DESC LIMIT 10;").fetchall()
```
martinpitt, line 49: I think we should limit this by default to the last 7 days.  Otherwise the "sins of the past" will haunt us forever, and we won't see the effect of test fixes.

generate-stats
```diff
@@ -0,0 +1,83 @@
...
+    for row in rows:
+        runs = cursor.execute("SELECT DISTINCT run from Failures WHERE testname = ? ORDER BY run DESC LIMIT 2;", (row[0], )).fetchall()
+        urls = cursor.execute("SELECT url from TestRuns WHERE id = ? or id = ?;", (runs[0][0], runs[1][0])).fetchall()
```
martinpitt, line 52: It could happen that there is only one run, and then `runs[1]` isn't defined, right? I think it'd be more robust to let SQL do the join, and select the urls with it.

generate-stats
```diff
@@ -0,0 +1,83 @@
...
+        top10.append([row[0], row[1], [urls[0][0], urls[1][0]]])
+
+    all_tables.append({"name": "Top failures", "columns": ["Test name", "Count", "Logs"], "data": top10})
+
+    # Get failures by OS
+    oses = cursor.execute("SELECT DISTINCT context FROM TestRuns;").fetchall();
```
martinpitt, line 58: let's call them context; for consistency, and it's also things like "-distropkg" or "/firefox"

index.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<meta charset="utf8" />
+<html>
```
martinpitt, line 3: Let's  please rename this failure-stats.html or something more specific than "index". Also, how about keeping render.js and style.css inline? This is simple enough that a single file is easier to handle IMHO. If you want to keep them separate, please rename them.

index.html
```diff
@@ -0,0 +1,9 @@
...
+  <script type="text/javascript" src="data.js"></script>
```
martinpitt, line 4: likewise, failure-stats.json or so? Does it have to be a JS wrapper, or could this use XMLHttpRequest to load a .json document directly?

index.html
```diff
@@ -0,0 +1,9 @@
...
+  <script type="text/javascript" src="render.js"></script>
+  <link rel="stylesheet" type="text/css" href="style.css">
+  <body id="the-body">
```
martinpitt, line 7: body is unique anyway, so `querySelector("body")` should be fine?

render.js
```diff
@@ -0,0 +1,47 @@
+document.addEventListener("DOMContentLoaded", () => {
+    const body = document.getElementById("the-body");
+
+    tables.forEach(t => {
+        const wrapper = document.createElement("DIV");
+
+        // Create title of the table
+        const title = document.createElement("H2");
```
martinpitt, line 8: Oh, the good old times without React :grin:

generate-stats
marusak: Well, we don't have any timestamp in the DB. We can either add it or do different limitation like "last 1000 failed tests".

marusak: Sent https://github.com/cockpit-project/bots/pull/1034, lets block this for a bit and then use with timestamp.

generate-stats
```diff
@@ -0,0 +1,107 @@
...
+
+import sys
+import time
+import sqlite3
+import argparse
+import logging
+import json
+
+# Seconds in two weeks
+TWO_WEEKS = 1209600
```
marusak, line 28: Went with two weeks, I feel like 7 days is too short time period to see any trends.

issues-review
```diff
@@ -56,8 +56,8 @@ def close_issues(api, opts):
         count = len(issues)
         for issue in issues:
             labels = api.get("issues/%i/labels" % issue["number"])
-            enhancement = [l for l in labels if l["name"] == "enhancement"]
-            bug = [l for l in labels if l["name"] == "bug"]
+            enhancement = [la for la in labels if la["name"] == "enhancement"]
+            bug = [la for la in labels if la["name"] == "bug"]
```
marusak, line 60: now it is "la la land" :smile:

run-queue
```diff
@@ -64,12 +64,21 @@ def consume_webhook_queue(channel, amqp):
...
             db = os.path.join(get_images_data_dir(), "test-failures.db")
+            data = os.path.join(get_images_data_dir(), "failure-stats.js")
             cmd = ['./store-failures', "--db", db, "--repo", repo, sha]
             body = {
                 "command": " ".join(cmd),
             }
             channel.basic_publish('', 'statistics', json.dumps(body),
                                   properties=pika.BasicProperties(priority=distributed_queue.MAX_PRIORITY))
+
+            cmd = ['./generate-stats', "--db", db, ">", data]
```
marusak, line 75: This is "racy", but I think it is fine, we don't necessary need to have new with the very last failures.
Also not sure about the `>` part...

index.html
marusak: It (probably) does not need to be, but I kind of like it more like that, than using HttpRequest...

issues-review
martinpitt: Is that because pycodestyle doesn't like the variable name `l`, as it's too ambiguous with `1`?

marusak: https://www.flake8rules.com/rules/E741.html

run-queue
martinpitt: It looks very confusing as an argv list, one would rather expect to use `stdout=open(data)` somewhere. But this is an actual shell string. Please skip building a list and plug in the string straight into `body = { "command": "./generate..." }` so that it's immediately clear that this is *not* an argv. The same is true for the existing ./store-failures call.

As for the race, that's indeed nasty as this may be done by a different bot. Why not simply do it in one command?

    ./store-failures ... && ./generate-stats

failure-stats.html
```diff
@@ -0,0 +1,75 @@
...
+            th.textContent = c;
+            row.appendChild(th);
+        });
+        table.appendChild(row);
+
+        // Fill the table with data
+        t.data.forEach(r => {
+            row = document.createElement("TR");
+            r.forEach(d => {
+                // Array indicates logs
```
martinpitt, line 55: "indicates" → "contains" or simply "with"?

generate-stats
```diff
@@ -0,0 +1,107 @@
...
+    all_tables = []
+
+    # Get all failures sorted by number of occurrences
+    top10 = []
+    rows = cursor.execute("""\
+            SELECT testname, COUNT(run)
+            FROM Failures
+            GROUP BY testname
+            ORDER BY COUNT(run)
+            DESC LIMIT 10;""").fetchall()
```
martinpitt, line 52: This is confusing -- the `DESC` belongs to `ORDER BY`, please move it up.

generate-stats
```diff
@@ -0,0 +1,107 @@
...
+    cursor = db_conn.cursor()
+
+    all_tables = []
+
+    # Get all failures sorted by number of occurrences
+    top10 = []
+    rows = cursor.execute("""\
+            SELECT testname, COUNT(run)
+            FROM Failures
+            GROUP BY testname
```
martinpitt, line 50: Let's restrict this to "last two weeks" as well, for consistency? Or was this deliberate for some reason? If so, please point that out in the headings.

marusak: Oh right, that was just forgotten. Fixed.

run-queue
marusak: Right, Having just one string with `&&` seems like the best option. Updated to that.

PR #1033, opened 2020-06-30
Image refresh for rhel-8-2

cockpituous:
Image refresh for rhel-8-2
 * [x] image-refresh rhel-8-2


martinpitt:
@larskarlitski , @henrywang : It seems cockpit-composer's rhel-8 branch is broken, there are dozens of eslint errors like

  155:5  error  Prop type `object` is forbidden  react/forbid-prop-types

Aside from that, this looks a bit stale: I suggest to re-sync rhel-8 branch to the tag from master that you put into 8.3, and run tests on rhel-8-3.

larskarlitski:
> @larskarlitski , @henrywang : It seems cockpit-composer's rhel-8 branch is broken, there are dozens of eslint errors like
> 
> 155:5 error Prop type `object` is forbidden react/forbid-prop-types

cockpit-composer didn't fix eslint versions, and new error messages started appearing in recent versions. @jkozol fixed this in https://github.com/osbuild/cockpit-composer/pull/1024. The fix is not yet in a release, though.

> Aside from that, this looks a bit stale: I suggest to re-sync rhel-8 branch to the tag from master that you put into 8.3, and run tests on rhel-8-3.

Hm, the rhel-8 branch only tests on 8.2, and master is tested on 8.3. I don't think we want to sync 8.2 with master? Should I maybe just backport the above fix onto the 8.2 branch?

martinpitt:
@larskarlitski : I just commented on https://github.com/osbuild/cockpit-composer/pull/1024 . This IMHO shouldn't be backported.

> I don't think we want to sync 8.2 with master

I meant that from a dev point of view, RHEL 8.2 is "done", and you most likely will not need the rhel-8 branch for that any more. It may make more sense to use the rhel-8 branch for 8.3 now, and thus reset it to the last release that actually is in 8.3.

henrywang:
@martinpitt, @larskarlitski. I updated eslint and eslint-* to latest and fixed lots of errors in [#1028](https://github.com/osbuild/cockpit-composer/pull/1028)

martinpitt:
Thanks @henrywang ! So that still leaves the question what to do with the current rhel-8 branch. In it's current form it's broken, but IMHO also not useful any more.

henrywang:
@martinpitt, Yes, I totally agree with you. I do not think cockpit-composer have patch on rhel 8.[0-2] any more.

jkozol:
> I meant that from a dev point of view, RHEL 8.2 is "done", and you most likely will not need the rhel-8 branch for that any more. It may make more sense to use the rhel-8 branch for 8.3 now, and thus reset it to the last release that actually is in 8.3.

@martinpitt I agree that it makes sense to bump the rhel-8 branch to what we have released in 8.3. @jgiardino @larskarlitski Do you agree?

martinpitt:
PR #1057 moved this branch away from rhel-8-2, and PR #1058 drops the image. The latter will close this PR.

PR #1034, opened 2020-07-01
Store also datetime for failed tests

marusak:
This will fail with the current DB. We can either drop it, or migrate it. @martinpitt WDYT?

### DIFF
--- a/store-failures
+++ b/store-failures
@@ -93,7 +93,7 @@ def main():
         with urllib.request.urlopen(raw_log, context=ctx if "logs.cockpit-project.org" in raw_log else None) as fp:
             for (testname, retry_reason) in find_failed_tests(fp):
                 if run_id is None:
-                    run_id = insert_run(cursor, repo, context, url, time)
+                    run_id = insert_run(cursor, repo, opts.revision, context, url, time)
                 insert_failure(cursor, run_id, testname, retry_reason)
 
     db_conn.commit()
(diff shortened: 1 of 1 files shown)


store-failures
```diff
@@ -78,10 +78,20 @@ def main():
...
         logging.debug("scanning log %s", raw_log)
 
+        # Parse time from log file name. Wrap in try/except block if the url would be unexpected
+        time = "now"
+        try:
+            parts = url.split("/")[-2].split("-")
+            day = parts[2]
+            hours = parts[3]
+            time = "{0}-{1}-{2} {3}:{4}:{5}".format(day[0:4], day[4:6], day[6:], hours[0:2], hours[2:4], hours[4:])
+        except:
```
martinpitt, line 88: No blanket exceptions, please; I think what we can reasonably expect is `IndexError`.

Also, I'd rather just store it as a unix timestamp, as a number. Makes age selection/calculation much easier. You can use strftime() for that.

store-failures
```diff
@@ -94,6 +104,7 @@ def init_db(cursor):
                      project TEXT,
                      context TEXT,
                      url TEXT,
+                     time DATETIME,
```
martinpitt, line 107: ... and use TIMESTAMP (not much difference in SQLite, but better be explicit)

store-failures
marusak: changed

martinpitt:
Idea from IRC discussion while we are changing the DB format: change `TestRuns.id INTEGER` → `TestRuns.ref TEXT`, and store the full SHA. Then the ID is something meaningful, and we can easily re-import/recreate the DB from that list.

PR #1061, opened 2020-07-12
Image refresh for rhel-7-9

cockpituous:
Image refresh for rhel-7-9
 * [x] image-refresh rhel-7-9


marusak:
[lorax vmware](https://logs.cockpit-project.org/logs/pull-1061-20200712-055003-4461da87-rhel-7-9-vmware-weldr-lorax-rhel7-extras/log.html), @atodorov

atodorov:
> [lorax vmware](https://logs.cockpit-project.org/logs/pull-1061-20200712-055003-4461da87-rhel-7-9-vmware-weldr-lorax-rhel7-extras/log.html), @atodorov

@marusak booting a VM failed for some reason but IDK why. Can you restart so we can see if this happens consistently?

marusak:
> Can you restart so we can see if this happens consistently?

I've tried it 3 times before I commented. But sure, I'll spin it once more now.

martinpitt:
FTR, I already retried it twice yesterday already.

atodorov:
> FTR, I already retried it twice yesterday already.

@chris1984, @jikortus do you have any idea why tests in vcenter are failing with the image refresh ?

jikortus:
@atodorov The error (traceback) from the log looks familiar to me. I hit this problem at least twice in the past, we tried to debug this a bit with @chris1984, without any specific success. The problem is that the VM is not able to get/connect to a free switch port, however at the same time I don't know what exactly that means and/or what the real, underlying problem is. From what I can remember, Chris applied some patches on the hypervisor and restarted it, and then the test finished with pass (several times). The workaround seems to be to just restart the hypervisor, personally I don't think it would be worth spending any significant efforts on debugging this deeper at this point.

martinpitt:
Thanks @jikortus . This seems to be related to the remote vmware end then, not to the local RHEL 7.9 VM?

jikortus:
> Thanks @jikortus . This seems to be related to the remote vmware end then, not to the local RHEL 7.9 VM?

Yes, most likely.

chris1984:
@jikortus @atodorov ill restart the hosts again and see what is going on. thanks for letting me know vmware--

PR #1088, opened 2020-07-23
Image refresh for fedora-32

cockpituous:
Image refresh for fedora-32
 * [x] image-refresh fedora-32

PR #1091, opened 2020-07-23
Image refresh for fedora-31

cockpituous:
Image refresh for fedora-31
 * [x] image-refresh fedora-31
 * [x] add naughty pattern for kdump regression: PR #1088
 * [x] drop selenium-kdump test from master: https://github.com/cockpit-project/cockpit/pull/14401
 * [x] backport that to the rhel-8.3 stable branch: https://github.com/cockpit-project/cockpit/pull/14408
 * [ ] somehow adjust the rhel-7.9 branch


martinpitt:
Looks like F31 now [breaks kdump](https://logs.cockpit-project.org/logs/pull-1091-20200724-042624-0c7f4fd7-fedora-31-selenium-firefox-cockpit-project-cockpit/log.html) similarly to F32. Meh..

martinpitt:
#1088 adds the naughty pattern, but we also need to fix the selenium tests. For cockpit master and rhel-8.3 branches I think we can just remove the kdump tests, but for the rhel-7.9 branch I don't have a good idea what to do there. Possibly moving the "start kdump" into the selenium-kdump.py test so that at least not every other test fails?

martinpitt:
I propose to just ignore the rhel-7.9 selenium kdump failure for now. We don't currently need that branch, I don't want to drop the test entirely, and there is no workaround for the  crash.

martinpitt:
Rebased to pick up #1088. Still blocked on https://github.com/cockpit-project/cockpit/pull/14408

PR #1107, opened 2020-07-29
fix the direct handling to all ssh methods

jscotka:
Allow to overwrite execution method  and use direct via env var `SSH_DIRECT` and also fix the handling of it in all methods in class instead of just execute.

Reason why:
 1. using Master connection sometimes fails for my tests (maybe timeout of connection)
 2. I have troubles to use master connection inside podman. Unable to find the root cause, maybe some selinux or any restrains of container.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -30,6 +30,8 @@ from . import timeout as timeoutlib
 
 
 class SSHConnection(object):
+    ssh_default_opts = ["-o", "StrictHostKeyChecking=no", "-o", "UserKnownHostsFile=/dev/null", "-o", "BatchMode=yes"]
+
     def __init__(self, user, address, ssh_port, identity_file, verbose=False):
         self.verbose = verbose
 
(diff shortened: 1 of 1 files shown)


machine/machine_core/ssh_connection.py
```diff
@@ -231,6 +229,16 @@ def _ensure_ssh_master(self):
         if not self._check_ssh_master():
             self._start_ssh_master()
 
+    def __ssh_direct_opt_var(self, direct=False):
+        return os.getenv("SSH_DIRECT", direct)
```
martinpitt, line 233: For consistency, can you please call this `$TEST_SSH_DIRECT`?

jscotka:
> Thanks! It sounds a bit hackish -- we use ssh masters all the time in docker and podman containers, but if it helps you I don't mind that. Just one naming nitpick.

Exactly, that's why I don't understand why it causes troubles to me, when cockpit uses it in this way.
my usecase where master connection failed is: I have RHEL8 test system -> via ssh connected on some fedora machine in openshift -> there I create podman machine fedora:latest with cockpit test dependencies -> this podman container will test that RHEL8 test system. 

and I found the issue that normal ssh connection is no problem, after several iterations of various issues :-) 
, but from some reason ssh master connection failed to create, maybe unable to create socket file there.

I found more issues, e.g. that chromium-headless was not enough for testing, and I had to install also chromium, there were missing some .so lib ("swing" or how it was called what is part of chromium package not chrome-headless) maybe this is issue in fedora:latest for podman container

martinpitt:
@jscotka : For chromium you need to pass some options to make it run headless: https://github.com/martinpitt/cockpit/blob/master/test/common/cdp.py#L174

martinpitt:
Looks good. Triggering project tests to make sure this works everywhere. Thanks!

jscotka:
> @jscotka : For chromium you need to pass some options to make it run headless: https://github.com/martinpitt/cockpit/blob/master/test/common/cdp.py#L174

Ah thanks a lot that you've explained me these options and I was not so familiar with all of them.

PR #1132, opened 2020-08-06
machine: Tolerate "domain is not running" errors on destroying

martinpitt:
In some cases, `VirtMachine.shutdown()` fails with

    libvirt.libvirtError: Requested operation is not valid: domain is not running

Ignore "not running" errors similarly to how it already ignores "not
found" errors, as these are transient domains.


martinpitt:
This may help with @garrett 's [woes](https://gist.githubusercontent.com/garrett/ac835591b777bb9c274700c77f4755eb/raw/e745fc30008d70e2d3338cef0f5d9ddb1e6d3b70/image-prepare.txt)?

garrett:
Seems to work here!

VMs are building properly again and I can even run them. (Not sure if coincidental or due to this PR.) This PR definitely doesn't hurt though! :wink:

PR #1133, opened 2020-08-06
testmap: Automatically run fedora-32/rawhide test for cockpit-podman

martinpitt:
This is prone to failure, but we at least want to know about it.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -87,6 +87,7 @@ REPO_BRANCH_CONTEXT = {
         'master': [
             'fedora-31',
             'fedora-32',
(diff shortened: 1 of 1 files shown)


martinpitt:
See https://github.com/cockpit-project/cockpit-podman/pull/495

martinpitt:
WTH -- several `npm install` failures on [@patternfly-react-table](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-1133-20200806-100644-abaa5891-fedora-32-rawhide-cockpit-project-cockpit-podman/log.html).

PR #1137, opened 2020-08-07
Image refresh for ubuntu-2004

cockpituous:
Image refresh for ubuntu-2004


martinpitt:
Something happened to libvirtd or libvirt-dbus.

marusak:
Current image:
```
ls -l /var/run/libvirt/libvirt-sock
srw-rw-rw- 1 root root 0 Aug 12 08:41 /var/run/libvirt/libvirt-sock
```

This refresh:
```
root@ubuntu:~# ls -l /var/run/libvirt/libvirt-sock
srw-rw---- 1 root libvirt 0 Aug 12 08:21 /var/run/libvirt/libvirt-sock
```

Now it fails with `Failed to connect socket to '/var/run/libvirt/libvirt-sock': Permission denied`.

Doing:
```
diff --git test/verify/check-machines test/verify/check-machines
index 36cafff6a..035b73769 100755
--- test/verify/check-machines
+++ test/verify/check-machines
@@ -68,6 +68,8 @@ class TestMachines(MachineCase, StorageHelpers, NetworkHelpers):
         self.restore_dir("/var/lib/libvirt")
         self.restore_dir("/etc/libvirt")
 
+        self.machine.execute("chmod o+rw /var/run/libvirt/libvirt-sock")
```

Fixes the problem, however this would be just hack for our tests. I will report this into ubuntu as this renders this version of libvirt unusable for cockpit users.

marusak:
Reported https://bugs.launchpad.net/ubuntu/+source/libvirt/+bug/1891311

I suggest blocking this and hopefully the report will move somewhere.

marusak:
Ah the isssue is wont-fix :/
But shouldn't we use `$XDG_RUNTIME_DIR/libvirt/libvirt-sock` in c-machines for user session anyway?

marusak:
This is maybe something for @KKoukiou to figure out? :)

marusak:
[ubuntu-stable](https://logs-https-cockpit.apps.ci.centos.org/logs/pull-1161-20200819-060625-53532b6a-ubuntu-stable-cockpit-project-cockpit/log.html) is affected as well. Next refresh will get stuck on the same thing.

KKoukiou:
The permissions for the socket are fetched from the socket systemd file.
This is in this ubuntu refresh - notice SocketMode=0660
On fedora systems, I guess also in older ubuntu that was 0666. This is what broke the tests. I will figure out what the right one.

```
root@ubuntu:~# systemctl cat libvirtd.socket
# /lib/systemd/system/libvirtd.socket
[Unit]
Description=Libvirt local socket
Before=libvirtd.service


[Socket]
# The directory must match the /etc/libvirt/libvirtd.conf unix_sock_dir setting
# when using systemd version < 227
ListenStream=/run/libvirt/libvirt-sock
Service=libvirtd.service
SocketMode=0660
SocketUser=root
SocketGroup=libvirt

[Install]
WantedBy=sockets.target
```

this

KKoukiou:
So final update on this,
ubuntu has expected 0660 permission, that's due to a security fix in ubuntu. fedora uses polkit for auth so uses mode 0666.  ubuntu disables polkit and uses group membership so sets it 0660. Before this image update, heir bug was that they had mistakenly left it 0666 previously which was a security flaw.

The issue here is that libvirt-dbus system process runs under `libvirtdbus` user which is not allowed to access the `/var/run/libvirt/libvirtd.socket`

`usermod -a -G libvirt libvirtdbus` will solve the issue, so that libvirtdbus user, which is the user under which libvirt-dbus system process runs as, can access the socket.
Of course it's horrible for the users to be expected to set groups to undocumented users, but right now I have no better suggestion.

martinpitt:
https://github.com/cockpit-project/cockpit/pull/14507 landed, retrying.

martinpitt:
[too old images](https://logs.cockpit-project.org/logs/pull-1137-20200826-053628-d818d29b-ubuntu-2004-cockpit-project-cockpit/log.html), rebased.

martinpitt (approved):

PR #1141, opened 2020-08-10
sink: Exit 0 in case of a collision

Gundersanne:
- [x] https://github.com/cockpit-project/cockpituous/pull/337 -> sink from there is deployed and currently running

### DIFF
--- a/task/sink.py
+++ b/task/sink.py
@@ -85,10 +85,14 @@ class Sink(object):
         shutil.rmtree(self.attachments)
 
         # All done sending output
-        self.ssh.stdin.close()
+        try:
+            self.ssh.stdin.close()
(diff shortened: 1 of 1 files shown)


task/sink.py
```diff
@@ -85,10 +85,14 @@ def flush(self, status=None):
         shutil.rmtree(self.attachments)
 
         # All done sending output
-        self.ssh.stdin.close()
+        try:
+            self.ssh.stdin.close()
+        except BrokenPipeError:
```
Gundersanne, line 90: Should I print something here to stderr?

marusak: I think it is fine I think to not to bother.

PR #1157, opened 2020-08-17
Image refresh for rhel-7-9

cockpituous:
Image refresh for rhel-7-9
 * [x] image-refresh rhel-7-9


marusak:
@atodorov [azure](https://logs.cockpit-project.org/logs/pull-1157-20200818-061332-8e67e587-rhel-7-9-azure-weldr-lorax-rhel7-extras/log.html) failed 3 times in a row. Is this due to exceeded quota or something that could be related to this refresh?

atodorov:
@marusak I can see in the logs:

```
:: [ 02:24:59 ] :: [   LOG    ] :: STDERR: /tmp/venv.90v/lib/python2.7/site-packages/cryptography/__init__.py:39: CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in a future release.
```

then a super non-human readable traceback which causes the azure commands to fail. So even though this is a deprecation warning it looks like a custom one and is probably not handled properly. That looks like a bug in either the cryptography Python package or the Azure tools using it.

marusak:
> That looks like a bug in either the cryptography Python package or the Azure tools using it.

Would you be interested in figuring it exactly and reporting it? This seems like it causes regression for your product and should be dealt with.

atodorov:
> > That looks like a bug in either the cryptography Python package or the Azure tools using it.
> 
> Would you be interested in figuring it exactly and reporting it? This seems like it causes regression for your product and should be dealt with.

I can do this once I deal with more urgent tasks. We don't have a specific demand for 7.9 testing ATM and capacity is down to the bare minimum.

marusak:
> I can do this once I deal with more urgent tasks. We don't have a specific demand for 7.9 testing ATM and capacity is down to the bare minimum.

Ok, then I'll merge this. I think it is more about seeing that there is problem that users will face than about having working VM for testing. But if the lorax team does not mind that lorax won't work on rhel 7.9, then I don't see any reason to block this refresh.

PR #1168, opened 2020-08-21
Check for github teams in appropriate org

marusak:
When PR is opened in a different organization than `cockpit-project` and
we check if user is in `Contributors` group, we should check this group
in the given org and not always in `cockpit-project`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,11 +133,11 @@ You can also invoke bots/tests/trigger from any project checkout, in which case
 you don't need the explicit `--repo` -- it will default to the GitHub origin of
 the current directory's project.
 
-### Testing a pull request by a non-whitelisted user
+### Testing a pull request by a non-allowed user
 
-If you want to run all tests on pull request #1234 that has been
-opened by someone who is not in our white-list, run tests-trigger
-with `--allow`:
+If you want to run all tests on pull request #1234 that has been opened by someone
+who does not have push access to the repository nor isn't in the 'Contributors' team,
+run tests-trigger with `--allow`:
 
     $ ./tests-trigger --allow [...]
 
--- a/issue-scan
+++ b/issue-scan
@@ -128,13 +128,12 @@ def tasks_for_issues(issues_data):
     else:
         api = github.GitHub()
         issues = api.issues(state="open")
-    whitelist = api.whitelist()
 
     for issue in issues:
         if issue["title"].strip().startswith("WIP"):
             continue
         login = issue.get("user", {}).get("login", {})
-        if login not in whitelist:
+        if not api.is_user_allowed(login):
             continue
 
         #
--- a/task/github.py
+++ b/task/github.py
@@ -380,6 +380,15 @@ class GitHub(object):
         else:
             raise KeyError("Team {0} not found".format(name))
 
+    def is_user_allowed(self, user):
+        # Firstly check if user has push access
+        data = self.get("/repos/{0}/collaborators/{1}/permission".format(self.repo, user)) or {}
+        if data.get("permission") in ["admin", "write"]:
+            return True
+
+        # User does not have push access, lets check if in `Contributors` group
+        return user in self.whitelist()
+
 
 class Checklist(object):
     def __init__(self, body=None):
--- a/tests-scan
+++ b/tests-scan
@@ -292,7 +292,6 @@ def cockpit_tasks(api, update, contexts, repo, pull_data, pull_number, sha, amqp
     else:
         pulls = api.pulls()
 
-    whitelist = api.whitelist()
     for pull in pulls:
         title = pull["title"]
         number = pull["number"]
@@ -300,6 +299,8 @@ def cockpit_tasks(api, update, contexts, repo, pull_data, pull_number, sha, amqp
         statuses = api.statuses(revision)
         login = pull["head"]["user"]["login"]
         base = pull["base"]["ref"]  # The branch this pull request targets
+
+        allowed = api.is_user_allowed(login)
 
         logging.info("Processing #{0} titled '{1}' on revision {2}".format(number, title, revision))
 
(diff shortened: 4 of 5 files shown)


marusak:
We will need to make sure that there is `Contributors` group in all orgs that we consider and that `cockpit` user has admin rights to all of them.

Now when I do 
```$ ./tests-trigger --repo osbuild/cockpit-composer 1088
task.github.GitHubError: Error accessing /orgs/osbuild/teams
  Status: 403
  Reason: Forbidden
  Response: {"message":"Must have admin rights to Repository.","documentation_url":"https://docs.github.com/rest/reference/teams#list-teams"}
```

When `cockpit` user does the same:
```
$ ./tests-trigger --repo osbuild/cockpit-composer 1088                                                                                                                                                                                                                                                                                                                                       
KeyError: 'Team Contributors not found'   
```

Orgs to check:

- [ ] cockpit-project
- [ ] weldr
- [ ] osbuild
- [ ] candlepin
- [ ] skobyda
- [ ] martinpitt

The last two are not orgs but users. That means this might get a bit trickier. Maybe we need some advanced logic for this?
@Gundersanne @martinpitt WDYT?

Motivation behind this: @teg opened PR against `osbuild/cockpit-composer`. The CI was not triggered becase he is not in `Commiters` group in `cockpit-project`. Then he tried to `tests-trigger --allow` it and it worked as he has push access to this repository. My first thought was to trigger tests for users that have push access - I even implemented it only then I realized that it would mean we would have to grant push access to everyone we want to trigger CI automatically for - which is not ideal.
This was the second idea - to parse Contributors from the given org, which is surely the right thing to do anyway but since some "orgs" are not orgs but users and I am not sure if we want to now force all orgs to set up this team with all the members, maybe we could go with combination of both, e.g:
```py
def is_user_allowed(self, user):
    # Firstly check if user has push access
    data = self.get("/repos/{0}/collaborators/{1}/permission".format(self.repo, user)) or {})
    if data.get("permission") in ["admin", "write"]:
        return True
    
    # User does not have push access, lets check if in `Contributors` group
    return user in api.whitelist() # with this patch + not failing when group not found
```

marusak:
Ok, so I actually implemented that. Some testing:
1. Triggering any random PR in cockpit-project org. Should work fine as I have both push access and am In the group. - OK
```
$ ./tests-trigger -f --repo=cockpit-project/cockpit 14468 debian-stable 
debian-stable: triggering on pull request 14468
```
2. Trying to trigger for someone who is not allowed. First without `--allow` then with. OK
```
$ ./tests-trigger -f --repo=cockpit-project/cockpit-podman 515 fedora-32/rawhide
Pull request author 'ezeeyahoo' isn't allowed. Override with --allow.
$ ./tests-trigger -f --repo=cockpit-project/cockpit-podman --allow 515 fedora-32/rawhide
fedora-32/rawhide: triggering on pull request 515
```
3. Trying to trigger PR in `martinpitt/performance-graphs` repo. I have push access there. It is user not org. OK
```
$ ./tests-trigger -f --repo=martinpitt/performance-graphs 7 fedora-32
fedora-32: triggering on pull request 7
```
4. Trying to trigger PR in `candlepin/subscription-manager`. I don't have push access. Previously my CI would get triggered on my PRs (as I am in cockpit-project contributors team) but I could not trigger tests (even on my own PRs). So I am loosing a bit with this, but I think it is correct as organizations should decide who can and cannot do stuff and not `cockpit-project`. - OK, maybe a nicer error would be great :) 
```
$ ./tests-trigger -f --repo=candlepin/subscription-manager 2176 rhel-8-1
  File "./tests-trigger", line 88, in <module>
    sys.exit(main())
  File "./tests-trigger", line 84, in main
    return trigger_pull(api, opts)
  File "./tests-trigger", line 21, in trigger_pull
    if not opts.allow and not api.is_user_allowed(login):
  File "/home/mmarusak/cockpit/bots/task/github.py", line 387, in is_user_allowed
    data = self.get("/repos/{0}/collaborators/{1}/permission".format(self.repo, user)) or {}
  File "/home/mmarusak/cockpit/bots/task/github.py", line 259, in get
    raise GitHubError(self.qualify(resource), response)
task.github.GitHubError: Error accessing /repos/candlepin/subscription-manager/collaborators/candlepin/permission
  Status: 403
  Reason: Forbidden
  Response: {"message":"Must have push access to view collaborator permission.","documentation_url":"https://docs.github.com/rest/reference/repos#get-repository-permissions-for-a-user"}
```
5.  Trying `osbuild-composer`. I have push access, it is org but does not have `Contributors` group. OK
```
$ ./tests-trigger -f --repo=osbuild/cockpit-composer 1082 fedora-31
fedora-31: triggering on pull request 1082
```
6. User that is in `Contributors` group for some other repo yet does not have push access... We don't have anyone like that. We might have someone for `cockpit-project` like that. A bit trickier to test as they would have to open PR (as they are not allowed to read user permissions thus not allowed to trigger tests). I guess lets just hope it still works the same :)
7. cockpit user can trigger all following orgs and users: TBD
- [ ] cockpit-project
- [ ] weldr
- [ ] osbuild
- [ ] candlepin
- [ ] skobyda
- [ ] martinpitt
8. `candlepin` user magic that I dropped TBD

teg:
Thanks for looking into this @marusak!

martinpitt:
TBH I don't particularly like this approach. This hardcoding was deliberate, as it is *our* infrastructure, and we want to retain some control/audit who is actually using it. In practice, https://github.com/orgs/cockpit-project/teams/contributors has always been "people who can run tests on Cockpit's CI", and that's what the team description says. Maybe we can rename /contributors to /ci-users even to make this less confusing? People in that team can *not* commit to Cockpit, so it's double-confusing.

Is it possible to add teams to other teams, like adding the "weldr" team to ci-users? https://github.com/orgs/cockpit-project/teams/contributors/members has a tab "0 child team members" which sounds promising, but I think that's limited to teams from the same org. I suppose we may instead need a list of allowed teams in bots?

marusak:
Right, it is really confusing, as you said, that 'Contributors' cannot actually contribute (as in commit directly) but bots trigger tests on their PRs (no matter what organization). It does not scale greatly, but oh well, it makes sense to have control over this.
However, this PR has two parts - the first being not using Contributors only from c-p/c, which seems is not the best idea. The other is that we actually trigger tests if the user has at least write access - as such user can do `tests-trigger` anyway, we don't prevent anything of any use here. And I would say, that in majority of cases we want to trigger tests for users who have this write access and there are only really just exceptions for any other cases. What do you think of that part?

martinpitt:
@marusak: Yes, I do agree to the third commit (push access → trigger CI), this is nice! For the other, WDYT about renaming the contributors team and adjusting the name in the code?

marusak:
Dropped the other commits.

martinpitt (approved):
Nice, thank you!

PR #1222, opened 2020-09-15
Image refresh for centos-7

cockpituous:
Image refresh for centos-7
 * [x] image-refresh centos-7


martinpitt:
We should really run a cockpit master test for centos-7 refreshes -- this already bit us once. I sent PR # about it, and ran `./tests-trigger 1222 image:centos-7` to validate it and trigger the test manually.

martinpitt (approved):

PR #1225, opened 2020-09-15
testmap: Trigger cockpit master tests on centos-7 refreshes

martinpitt:
check-multi-os and check-superuser run centos-7 as auxiliary image to
validate protocol backwards compatibility.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -17,6 +17,7 @@
 
 import os.path
 
+from machine.machine_core.constants import TEST_OS_DEFAULT
 
 REPO_BRANCH_CONTEXT = {
     'cockpit-project/bots': {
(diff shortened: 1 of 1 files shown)


martinpitt:
See PR #1222, I ran tests-trigger on this branch and it correctly triggered the cockpit master run.

marusak (approved):
Yeah, that is true, thanks for spotting it out! But shouldn't we pick the default OS, which is now F32.

martinpitt:
@marusak: I was trying to avoid having to bump it too often -- but indeed, importing machine/machine_core/constants.py and generating the context dynamically from `TEST_OS_DEFAULT` is an excellent idea. Let me try that.

martinpitt:
@marusak: Changed accordingly, and tested on #1222 again.

marusak (approved):
Thanks!

PR #1237, opened 2020-09-17
images: Drop obsolete centos macro definition with --rhel

martinpitt:
It's been a long time since we built RHEL packages in EPEL; builds
happen in proper RHEL mocks. These should work without the extra macro
definition hack.


martinpitt:
rhel-7.9 cockpit branch fails subscriptions, needs work.

martinpitt:
Actually, on rhel-7-9 image the mock build env really *is* EPEL. So, let's just forget about this for now, it's unimportant cleanup.

PR #1252, opened 2020-09-22
Image refresh for rhel-7-9

cockpituous:
Image refresh for rhel-7-9
 * [x] image-refresh rhel-7-9
 * [x] https://github.com/cockpit-project/cockpit/pull/14656


mvollmer:
`testNodeNavigation (check_openshift.TestOpenshift)` is failing pretty reliably. I'll investigate.

mvollmer:
@atodorov, could you have a look at the `rhel-7-9/azure@weldr/lorax/rhel7-extras` failure?

It fails during "Upload image to Azure" with 
```
Failed to import the required Python library (msrestazure) on m1.cockpit.lan's Python /tmp/venv.EL9/bin/python.
```

and I guess the rest of the failures are just fallout from that.

Logs here https://logs.cockpit-project.org/logs/pull-1252-20200923-075315-34669f8e-rhel-7-9-azure-weldr-lorax-rhel7-extras/log.html

mvollmer:
> `testNodeNavigation (check_openshift.TestOpenshift)` is failing pretty reliably. I'll investigate.

The test was looking for the "Login failed" message in the wrong place, otherwise the test makes sense.
I have no idea why it looks in the wrong place.

atodorov:
> @atodorov, could you have a look at the `rhel-7-9/azure@weldr/lorax/rhel7-extras` failure?
> 
> It fails during "Upload image to Azure" with
> 
> ```
> Failed to import the required Python library (msrestazure) on m1.cockpit.lan's Python /tmp/venv.EL9/bin/python.
> ```

Above that I see: `ImportError: No module named msrest.serialization`

and before that this test does `pip install ansible[azure]`. So probably something in the Azure dependencies for Ansible changed or was left behind in some of the upstream packages.

> 
> and I guess the rest of the failures are just fallout from that.
> 

Correct.

mvollmer:
> So probably something in the Azure dependencies for Ansible changed or was left behind in some of the upstream packages.

@atodorov, how do you propose we fix this? Would this need changes to the weldr/lorax repository? I am at a loss as to what to do next.

atodorov:
> 
> @atodorov, how do you propose we fix this? Would this need changes to the weldr/lorax repository? I am at a loss as to what to do next.

Open an issue against the lorax repository so this doesn't get forgotten and merge as-is. The composer team has very limited QE capacity at the moment and given we're not actively testing the rhel7 branch right now this will not be a big priority for us.

mvollmer:
> The composer team has very limited QE capacity at the moment and given we're not actively testing the rhel7 branch right now this will not be a big priority for us.

Ok, I see! In that case, I propose that we stop testing your rhel7 branch when we make new rhel-7 images.

mvollmer (approved):

atodorov:
> Ok, I see! In that case, I propose that we stop testing your rhel7 branch when we make new rhel-7 images.

that works too.

PR #1254, opened 2020-09-23
testmap: Don't test the "azure" scenario with weldr/lorax

mvollmer:
It has started failing and is not maintained right now by weldr/lorax.

See https://github.com/weldr/lorax/issues/1084

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -117,13 +117,13 @@ REPO_BRANCH_CONTEXT = {
             'rhel-7-9/live-iso',
             'rhel-7-9/qcow2',
             'rhel-7-9/aws',
-            'rhel-7-9/azure',
             'rhel-7-9/openstack',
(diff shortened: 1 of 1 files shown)


task/testmap.py
```diff
@@ -117,7 +117,6 @@
             'rhel-7-9/live-iso',
             'rhel-7-9/qcow2',
             'rhel-7-9/aws',
-            'rhel-7-9/azure',
```
martinpitt, line 120: I'd suggest to move it to `_manual` instead, so that it's possible to run `tests-trigger` on it to investigate/fix it.

martinpitt (changes requested):

task/testmap.py
mvollmer: Good point!

martinpitt (approved):

PR #1256, opened 2020-09-24
Image refresh for rhel-8-3

cockpituous:
Image refresh for rhel-8-3
 * [x] image-refresh rhel-8-3


mvollmer:
rhel-8-3@martinpitt/performance-graphs failure got fixed by https://github.com/martinpitt/performance-graphs/pull/56

martinpitt:
@henrywang : The cockpit-composer tests look rather grim -- is that a regression in osbuild or something else in the underlying image, or does it happen with the current image as well?

mvollmer:
> @henrywang : The cockpit-composer tests look rather grim -- is that a regression in osbuild or something else in the underlying image, or does it happen with the current image as well?

@larskarlitski, @jkozol, fyi.

henrywang:
@martinpitt, looks something stuck at loading packages. Do not have this kind of issue in https://github.com/osbuild/cockpit-composer/pull/1108

mvollmer:
I can see this in the Browser console:

```
Error in fetchBlueprintContentsSaga {…}
  ​message: "Bad Request"
​  options: Object { body: "", path: "/api/v0/projects/info/openssh-server" }
  ​problem: null
​```

Gundersanne:
This is an issue with valid baseos/appstream repos for osbuild-composer on cockpit-composer, and needs to be fixed there (both of them, as cockpit-composer overwrites osbuild-composer's repos). The nightlies used there during the tests have switched to production certificates, and as such can no longer be accessed, the invalid subscription results in the bad request error.

So either naughty for now, and then make this fix a priority? Or we hold off on merging this, but it might be good to reflect this change in cockpit's image asap :)

henrywang:
> So either naughty for now, and then make this fix a priority in cockpit-composer and osbuild-composer? Or we hold off on merging this, but it might be good to reflect this change in cockpit's image asap :)

Thanks @Gundersanne. Does `RHEL-8.3.0-Snapshot-3.1` repo fix that issue? Thanks.

martinpitt:
@Gundersanne , @henrywang : That does sound like it would affect tests on the currrent image as well, though? https://github.com/osbuild/cockpit-composer/pull/1108 is already a week old. I sent https://github.com/osbuild/cockpit-composer/pull/1112 to do a run against the current image, so that we can compare.

henrywang:
@martinpitt, @Gundersanne, Looks renaming `rhel-8-beta.json` to `rhel-8.json` works. I think we can land this first then I send a PR to update https://github.com/osbuild/cockpit-composer/blob/master/test/files/rhel-8-beta.json.

martinpitt:
@henrywang , nice thanks! we are still waiting for some other tests here, and on the results for https://github.com/osbuild/cockpit-composer/pull/1112 . Unfortunately most of our test machines are down now, I'm setting up AWS fallback.

martinpitt:
ah, we can't re-run the lorax tests here -- the [image cannot be downloaded](https://logs.cockpit-project.org/logs/pull-1256-20200925-051132-a51e429f-rhel-8-3-qcow2-weldr-lorax-rhel8-branch/log.html), as it was only on cockpit-11. If e2e machines don't come back soon, I'll just close this and then the AWS runner can try its luck again.

martinpitt:
https://github.com/osbuild/cockpit-composer/pull/1112 has shown that this *does* depend on the base image somehow. So, let's close this, let the bots build a new one, then @henrywang has something to test against.

mvollmer:
> (maybe not the most helpful error, something worth improving also).

Additionally, the error only appears in the Browser console, the dialog stays open and spins endlessly. I'll open an issue about that.

Gundersanne:
@henrywang 

> Looks renaming `rhel-8-beta.json` to `rhel-8.json` works. I think we can land this first then I send a PR to update https://github.com/osbuild/cockpit-composer/blob/master/test/files/rhel-8-beta.json.

Doesn't that just avoid the copy here https://github.com/osbuild/cockpit-composer/blob/master/test/vm.install#L15 which overwrites the osbuild-composer's repositories?

PR #1258, opened 2020-09-24
testmap: Introduce `fedora-testing/dnf-copr` context

marusak:
This context will stay in `manual` and it will run tests with dnf from
nightly repos. Since it is based on `fedora-testing` repo it will be run
on every refresh of this image.


martinpitt (approved):
\o/

PR #1263, opened 2020-09-27
testmap: Automatically trigger fedora-testing/dnf-copr scenario

martinpitt:
See https://github.com/cockpit-project/cockpit/pull/14664

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -182,7 +182,8 @@ OSTREE_BUILD_IMAGE = {
 # only put auxiliary images here; triggers for primary OS images are computed from testmap
 IMAGE_REFRESH_TRIGGERS = {
     "fedora-testing": [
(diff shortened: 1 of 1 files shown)


task/testmap.py
```diff
@@ -182,7 +182,8 @@
 # only put auxiliary images here; triggers for primary OS images are computed from testmap
 IMAGE_REFRESH_TRIGGERS = {
     "fedora-testing": [
-        "fedora-testing@cockpit-project/cockpit"
+        "fedora-testing@cockpit-project/cockpit",
```
marusak, line 185: Is this necessary? I though that it would be computed. Or we don't consider `__manual` when computing contexts to be triggered?

martinpitt: Correct, we ignore the `_manual` ones, because they are, well, *manual* :-)

marusak (approved):

PR #1266, opened 2020-09-28

cockpituous:
 * [ ] image-refresh ubuntu-stable

### DIFF
--- a/images/ubuntu-stable
+++ b/images/ubuntu-stable
@@ -1,1 +1,1 @@
-ubuntu-stable-ae08ede9447c4567d66cf5980002306ee4e5368847c100aef79a4b66c3ef02dd.qcow2
+ubuntu-stable-b385c30cf3e2a7bc81ef1abf8f8adb778fe8fda0fd5f8e31a972e85a2cb4e9eb.qcow2


martinpitt:
There's something [seriously wrong](https://logs.cockpit-project.org/logs/pull-1266-20200929-044544-af3be7a4-ubuntu-stable-cockpit-project-cockpit/log.html#250-2) here. I already retried once, that wasn't just an one-off "WTH?"

mvollmer:
The many parted issues are due to a change in the 5.8 kernel, where the BLKPG_DEL_PARTITION ioctl syscall now returns ENOMEM instead of ENXIO for non-existing partitions. This trips up parted, partprobe, etc.

See https://gitlab.gnome.org/GNOME/gparted/-/issues/111, https://github.com/torvalds/linux/commit/88ce2a530cc9865a894454b2e40eba5957a60e1a

martinpitt:
There's a new kernel in the pipeline: https://bugs.launchpad.net/kernel-sru-workflow/+bug/1896668 . Might be worth trying that (from -proposed) to see if that fixes it already; if so, we can just stall this until that lands.

martinpitt:
From #ubuntu-devel:
```
sil2100 | pitti: yes please! Since I know the current linux in release was not supposed to be released
sil2100 | pitti: it migrated by accident, only the -proposed one is the right one (hopefully migrating soon)
sil2100 | We were waiting with beta images until that lands 
```

mvollmer:
The kernel tracking bug went to "Fixed Released", I don't know what that means but let's try a new image.

martinpitt:
@mvollmer: I checked this morning and it wasn't published yet. Let's try tomorrow morning to be sure that it propagated to the mirrors.

martinpitt:
OK, it's time to try a new build.

PR #1273, opened 2020-09-30
lorax: Only run composer-cli tests with osbuild-composer

bcl:
lorax-composer has been removed in Fedora 34, only run the tests with
osbuild-composer.


marusak (approved):
You said it was removed in F34 yet you stop testing on F32.
But I am fine with this, just saying :)

PR #1293, opened 2020-10-08
testmap: Test only F32 and F33 for certificates

marusak:
See https://github.com/skobyda/cockpit-certificates/pull/25
Should land together


Gundersanne (approved):
Oki, I'll let you or simon do the landing then :)

PR #1296, opened 2020-10-08
issue-scan: Add the --repo option

Gundersanne:
This is needed so the webhook can iterate over projects.

---

Tested with `./issue-scan -v --repo "cockpit-project/cockpit"`. Output:
```
issue-14702 npm-update     9
issue-14696 po-refresh     9
```

To make sure the `--issues-data` option still works I also ran `./issue-scan -v --issues-data $(<issues-data)`. Output:
```
issue-1295 image-refresh centos-8-stream    9
```

### DIFF
--- a/issue-scan
+++ b/issue-scan
@@ -67,6 +67,8 @@ def main():
                         help='The host:port of the AMQP server to publish to (format host:port)')
     parser.add_argument('--issues-data', default=None,
                         help='issues or pull request event GitHub JSON data to evaluate')
+    parser.add_argument('--repo', default=None,
+                        help='Repository to scan and checkout.')
     opts = parser.parse_args()
 
     if opts.amqp and no_amqp:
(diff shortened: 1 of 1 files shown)


issue-scan
```diff
@@ -125,6 +127,9 @@ def tasks_for_issues(issues_data):
         if 'bot' in labels:
             issues.append(issue)
         api = github.GitHub(repo=repo)
+    elif opts_repo:
+        api = github.GitHub(repo=opts_repo)
+        issues = api.issues(state="open")
     else:
         api = github.GitHub()
         issues = api.issues(state="open")
```
martinpitt, line 135: Please DRY. if `opts_repo` is None, that's totally legit, and the second line is identical. So you can just drop the `elif` check.

martinpitt (changes requested):
That approach looks fine, thanks! Just a nitpick for simplification.

issue-scan
Gundersanne: ugh of course ><

marusak (approved):
Thanks!

martinpitt (approved):
Dank je wel!

PR #1299, opened 2020-10-09
issue-scan: Log the task's command when publishing to amqp

Gundersanne:
Also some whitespace cleanup, always good :)


martinpitt (approved):
That seems harmless enough, thanks!

PR #1301, opened 2020-10-13
Add script to upload GitHub action secrets

martinpitt:
These encrypted secrets can be used by GitHub actions. By default we
want them at the organization level, so that all projects of the org can
share them. For testing, non-generic secrets, or third-party projects
without an organization they can also be uploaded to a repository.

See https://developer.github.com/v3/actions/secrets for details.

### DIFF
--- a/github-upload-action-secrets
+++ b/github-upload-action-secrets
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+#
+# Upload encrypted secrets to GitHub
+# With this the secrets can be used by GitHub actions:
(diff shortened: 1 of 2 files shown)


martinpitt:
I committed a standalone version of this to our internal secrets repo, but it makes much more sense (for re-use) to have it here. Once that lands, I'll drop it from the secrets repo and adjust the documentation.

Gundersanne (approved):
Tried this out on my own repo and it worked quite well, thank you!

image

PR #1315, opened 2020-10-16
task: Deal with tokens that cannot access the /user API

martinpitt:
This is the case with the restricted default tokens in GitHub actions.
With these we want to push the fork to the original repo anyway, as
that's much easier to set up especially for third-party projects (you
need zero credentials setup).

### DIFF
--- a/task/__init__.py
+++ b/task/__init__.py
@@ -406,8 +406,16 @@ def branch(context, message, pathspec=".", issue=None, branch=None, push=True, *
     except subprocess.CalledProcessError:
         raise RuntimeError("Couldn't configure git config with our API token")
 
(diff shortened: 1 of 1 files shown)


martinpitt:
This doesn't change the behaviour for our current cockpituous token/webhook setup, but unblocks running bots in github actions, see https://github.com/cockpit-project/cockpit-ostree/pull/154 . I successfully tested that [in this run](https://github.com/martinpitt/cockpit-ostree/runs/1263315300?check_suite_focus=true) and it produced https://github.com/martinpitt/cockpit-ostree/pull/9

marusak (approved):
Nice, thanks!

PR #1320, opened 2020-10-20
github-upload-action-secrets: Trim whitespace from secrets

martinpitt:
So that you can use the secrets as part of a string like
`token:${{ secrets.MY_TOKEN }}@github.com` without having to worry about
line breaks.


martinpitt:
I re-uploaded our cockpituous token with this fixed version, and the [po-refresh run succeeded](https://github.com/cockpit-project/cockpit-podman/runs/1279898608) now.

marusak (approved):

PR #1334, opened 2020-10-24
Run maintenance tasks in a GitHub workflow

martinpitt:
We want to deprecate the .tasks mechanism, this is the next-to-last
user.

The default GitHub token suffices for this, as
image-refresh and naughty-prune don't care who created the issue with
the task.

Let these tasks run daily, which roughly coincides with how often the
bots run them through .tasks.


martinpitt:
I gave this some very light testing on my own fork, to rule out missing dependencies and broken credentials setup. But my fork doesn't have any issues and such, so it didn't really do anything. As previously, I propose to land this, run it in production, and I either fix it quickly or we revert it.

martinpitt:
@marusak: After that, the only remaining .tasks consumer is cockpit's urls-check. Do you want to look at that, or want me to?

marusak (approved):
Thanks!

marusak:
> After that, the only remaining .tasks consumer is cockpit's urls-check. Do you want to look at that, or want me to?

I have it on my TODO list :) I'll try to do it tomorrow :crossed_fingers:

martinpitt:
Landing. I'll check the result tomorrow, and stand by for fixing/reverting if anything blows up.

.github/workflows/daily.yml
```diff
@@ -0,0 +1,21 @@
+name: daily
+on:
+  schedule:
+    - cron: '1 * * * *'
+  # can be run manually on https://github.com/cockpit-project/bots/actions
+  workflow_dispatch:
+jobs:
+  npm-update:
```
martinpitt, line 8: Argh, this should have been a more generic name than "npm-update"; I'll send a follow-up fix.

martinpitt:
I [triggered a manual run](https://github.com/cockpit-project/bots/runs/1305701767?check_suite_focus=true). It didn't really do anything, but at least did not error.

PR #1335, opened 2020-10-25
workflows: Fix daily job name

martinpitt:
This job does not only run npm-update, so rename it to something more
generic.


martinpitt:
Follow-up from PR #1334.

marusak (approved):

PR #1336, opened 2020-10-26
workflows: Fix cron of daily workflow

martinpitt:
Actually run this every day only, instead of hourly.


martinpitt:
See https://github.com/cockpit-project/bots/actions

marusak (approved):
thanks!

PR #1344, opened 2020-10-28
task: Add the ocp4 public image store

Gundersanne:


task/__init__.py
```diff
@@ -54,6 +54,7 @@
 # Servers which have public images
 PUBLIC_STORES = [
     "https://images-cockpit.apps.ci.centos.org/",
+    "https://images-frontdoor.apps.ocp.ci.centos.org/",
```
martinpitt, line 57: Please move this up, as it should be the primary store now. Otherwise image-upload will prefer uploading to ocp3.

martinpitt (changes requested):
Please adjust the commit message to drop the "remove", as we don't yet.

martinpitt (approved):
Dank je wel!

PR #1345, opened 2020-10-29
Drop fedora-31

marusak:
We can merge this once @jkozol acks that cockpit-composer no longer tests against F31


jkozol:
@marusak cockpit-composer no longer needs to test against f31. And once https://github.com/osbuild/cockpit-composer/pull/1145 merges we will no longer have fedora 31 in our cockpituous release. Thanks for removing f31 from the cockpit-composer testmap.

Gundersanne (approved):

PR #1352, opened 2020-11-02
Run our unit tests in a GitHub workflow

martinpitt:


martinpitt:
If you like this, I'll switch the required tests around. Note that OCP4 is currently down, so the webhook is broken, and the old "hosts" test doesn't currently get run.

.github/workflows/test.yml
```diff
@@ -0,0 +1,14 @@
+name: test
```
marusak, line 1: `test / tests (pull_request) ` it shows up like this in the list of tests. Could we give it better name? (Not that I know of any :) )

marusak (dismissed):
Nice!
I like the idea that we run as much as we can outside of our own infra and this is a great candidate!
One nitpick, but not really a blocker :)

.github/workflows/test.yml
martinpitt: Indeed - as per IRC, I renamed it to "bots", to coincide with `./test-bots` (the script that runs).

marusak (approved):
Thanks!

PR #1353, opened 2020-11-02
testmap: Add manual ubuntu-stable test for cockpit-podman

martinpitt:
cockpit-podman is in Ubuntu now, so let's make sure it stays working.

We'll start with running this through Travis [1], but with that we can
compare results on our infrastructure.

[1] https://github.com/cockpit-project/cockpit-podman/pull/586


marusak (approved):
Thanks!

PR #1363, opened 2020-11-03
Image refresh for rhel-7-9

cockpituous:
Image refresh for rhel-7-9
 * [x] image-refresh rhel-7-9


martinpitt:
@KKoukiou : FYI, ignore the lorax tests. Alex said they are not currently looking at them, so I suggest to send a bots testmap PR to move these to `_manual` for the time being.

KKoukiou (approved):
Apart from the lorax tests everything else succeeded

PR #1371, opened 2020-11-04
make-checkout: Drop --reference-if-able

Gundersanne:
This only creates "warning" noise in logs, since repos are no longer
cached by the tasks container.

### DIFF
--- a/make-checkout
+++ b/make-checkout
@@ -63,9 +63,7 @@ def main():
     if os.path.exists(TARGET_DIR):
         shutil.rmtree(TARGET_DIR)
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Cheers! Please run a small test to make sure this works.

Gundersanne:
@martinpitt tested locally with `./make-checkout master` could also do `run-queue` from this bots ref on a task pod if you prefer.

martinpitt:
Oh right, make-checkout isn't self-testing. Looks safe enough, anyway, and thanks for the local test!

Issue #1372, opened 2020-11-04
In debian-stable images libvirt crashes when freeing up resources in TestMachines tests


KKoukiou:
Upstream bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=973758

cockpituous:
verify/debian-stable
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testAddDiskNFS (__main__.TestMachines)
error: failed to connect to the hypervisor
error: Failed to connect socket to '/var/run/libvirt/libvirt-sock': No such file or directory
error: failed to connect to the hypervisor
error: Failed to connect socket to '/var/run/libvirt/libvirt-sock': No such file or directory
error: Failed to start network default
error: Requested operation is not valid: network is already active

DevTools listening on ws://127.0.0.1:9989/devtools/browser/b3c5d4a9-70bb-4e72-ad43-e38c59b2cf1e
[1104/171903.903392:ERROR:egl_util.cc(70)] Failed to load GLES library: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: cannot open shared object file: No such file or directory
[1104/171903.905209:ERROR:viz_main_impl.cc(150)] Exiting GPU process due to errors during initialization
CDP: {"source":"security","level":"error","text":"Refused to apply style from 'http://127.0.0.2:9191/cockpit/$645c9615cf0d85b1976488c95cae2fa9714562855765727b89ab95553e77554d/shell/nav.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.","timestamp":1604510344807.062,"url":"http://127.0.0.2:9191/machines"}
> warning: grep: /sys/class/dmi/id/power/autosuspend_delay_ms: Input/output error

> warning: LIST_DOMAIN_SNAPSHOTS action failed for domain %s: %s /org/libvirt/QEMU/domain/_87533007_6004_46e8_963e_5040f4e58188 {"problem":null,"name":"org.freedesktop.DBus.Error.UnknownMethod","message":"No such method “ListDomainSnapshots”"}
nfs-pool nfs-volume-0
[...]

PR #1373, opened 2020-11-04
Add naughty for libvirt crash in debian stable

KKoukiou:
See known issue #1372

### DIFF
--- a/naughty/debian-stable/1372-libvirt-crashes-on-test-teardown
+++ b/naughty/debian-stable/1372-libvirt-crashes-on-test-teardown
@@ -1,0 +1,8 @@
+*TestMachines*
+*
+Process * (libvirtd) of user * dumped core.
+*
+#1  * virQEMUDriverGetConfig (libvirt_driver_qemu.so)
+*
+#3  * virStateStop (libvirt.so.0)
+*


KKoukiou:
> See known issue #1372

Related failure https://logs.cockpit-project.org/logs/pull-1350-20201103-090725-fe2ff21e-debian-stable-cockpit-project-cockpit/
This is from where I fetched the core files also for the bug report.

KKoukiou:
@marusak I forgot how to locally test test output agains naughties to check that the regexp applies. Do you remember the command to run for that?

martinpitt:
@KKoukiou: copy the raw test output into a file, then go into bots, and run `cat /tmp/output | ./tests-policy -o centos-8-stream` (or whichever image). It should turn a FAIL into a SKIP if a pattern matches.

naughty/debian-stable/1372-libvirt-crashes-on-test-teardown
```diff
@@ -0,0 +1,9 @@
+*testAddDiskNFS*
+*
+Process * (libvirtd) of user * dumped core.
+*
+#0  * __GI___pthread_mutex_lock (libpthread.so.0)
+#1  * virQEMUDriverGetConfig (libvirt_driver_qemu.so)
+#2  * n/a (libvirt_driver_qemu.so)
```
martinpitt, line 7: I suggest to skip lines #0 and #2, as there might be too much fuzz. Just matching on #1 and #3 should be enough.

KKoukiou: It was actually skipped as expected - see https://logs.cockpit-project.org/logs/pull-1373-20201104-165549-f6b48ae6-debian-stable-cockpit-project-cockpit/log.html

Shall I still change it or keep it as specific as it get?

KKoukiou: HA testInlineConsoleWithUrlRoot fails with the exact same way - I think i will need to remove the test pattern match.

martinpitt (approved):
Bots were down last night, Sanne fixed them. So re-triggering tests. Looks great now, thanks!

PR #1375, opened 2020-11-06
Add naughty for libvirt crash in RHEL

### LINKED ISSUE #1374: In rhel-8-3 images libvirt crashes when freeing up resources in TestMachines tests
Very similar to https://github.com/cockpit-project/bots/issues/1372 but I the backtrace is slightly different and the libvirt versions differ as well, thus I would like to keep it separate.

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1895359

KKoukiou:
See known issue #1374
Fixes #1374

Note: this is same as #1372 but the regexp differs because rhel images
do not have the debuginfo packages installed.

Should catch crashes similar to https://logs.cockpit-project.org/logs/pull-14869-20201106-110814-e9c545ec-rhel-8-3/log.html#288

### DIFF
--- a/naughty/rhel-8/1374-libvirt-crashes-on-test-teardown
+++ b/naughty/rhel-8/1374-libvirt-crashes-on-test-teardown
@@ -1,0 +1,7 @@
+*TestMachines*
+*
+Process * (libvirtd) of user * dumped core.
+*
+#1  * virCondWait (libvirt.so.0)
+#2  * virThreadPoolWorker (libvirt.so.0)
+#3  * virThreadHelper (libvirt.so.0)


KKoukiou:
Oh this is the exact same crash like https://github.com/cockpit-project/bots/issues/798

marusak:
> Oh this is the exact same crash like #798

How do you mean? If it is the same one, then why creating a new one?

KKoukiou:
> > Oh this is the exact same crash like #798
> 
> How do you mean? If it is the same one, then why creating a new one?

Oh sorry - just saw you answered. So fedora images dont have debuginfo installed for libvirt. RHEL do, so the stacktrace looks completely different. A common regexp  match would be hard.
Also the upstream bug reports, are separate. 
My  


> > Oh this is the exact same crash like #798
> 
> How do you mean? If it is the same one, then why creating a new one?

Ah sorry I only now saw the review.
fedora images don't have debuginfo and generate completely different stacktrace.
So having two seemingly at least different stacktraces, two different bug report, since for downstream and fedora it's different anyway, and for sure different libvirt versions makes me think that it worths to keep also the naughties separate.
Lastly, it's my intuition that that's the same bug, the libvirt devels did not even confirm on the bug report yet.

marusak (approved):
Thanks!

PR #1387, opened 2020-11-12

cockpituous:


martinpitt:
Uh, what? ubuntu-stable has SELinux enabled now??

allisonkarlitskaya:
1) I've rebuilt the image locally and I can't get the fail

2) I don't think this is related to selinux in any way.  The failure code is 32, which is documented in `mount(8)` to mean 

```
       32     mount failure
```

martinpitt:
@allisonkarlitskaya : I mean that now `if selinuxenabled` obviously is true, and that's unexpected. So it is somehow related. Would be interesting to investigate that locally?

marusak:
https://logs.cockpit-project.org/logs/pull-14876-20201112-093728-21a2742b-rhel-8-3/log.html#305-2 happens the same issue also on other OSes like 8.3

allisonkarlitskaya:
> @allisonkarlitskaya : I mean that now `if selinuxenabled` obviously is true, and that's unexpected. So it is somehow related. Would be interesting to investigate that locally?

I think maybe you read the code incorrectly

```
if selinuxenabled 2>/dev/null; then chcon --reference /proc/cmdline /run/cmdline; fi && mount --bind /run/cmdline /proc/cmdline
```

ie:
```
if selinuxenabled 2>/dev/null; then chcon --reference /proc/cmdline /run/cmdline; fi
```

followed by

```
mount --bind /run/cmdline /proc/cmdline
```

(note: `if`/`then`/`fi` branch returns `0` in case the condition is false, or the return value of the `then` branch, otherwise, and that seems to be what was intended here).

So, the error 32 is definitely coming either from `chcon` (which is extremely unlikely) or from `mount`, irrespective of selinux (which is extremely likely).

allisonkarlitskaya:
Put another way:

```
[lis@x1 ~]$ if false; then true; fi && mount /x /y
mount: /y: mount point does not exist.
[lis@x1 ~]$ echo $?
32
```

martinpitt:
@allisonkarlitskaya : Of course you are right, I missed the `;fi`. Brown paperbag totally! :blush:

martinpitt:
So, this does seem to be some valid race condition, but not specific to this refresh. Retrying test.

allisonkarlitskaya (approved):

PR #1395, opened 2020-11-16
image-trigger: update fedora-testing every 3 days

allisonkarlitskaya:
Suggested by Martin.  This will give us a better chance to notice
breaking changes before they make it into the official distro.

### DIFF
--- a/image-trigger
+++ b/image-trigger
@@ -26,7 +26,7 @@ REFRESH = {
     "debian-stable": {},
     "fedora-32": {},
     "fedora-33": {},
-    "fedora-testing": {},
+    "fedora-testing": {"refresh-days": 3},
     "fedora-coreos": {},
     "ubuntu-2004": {},
     "ubuntu-stable": {},


marusak:
Not that huge fan, TBH. Image refreshes are already noisy and take always some time to review. And there is fedora gating that should catch failures, not Cockpit (even though I feel that Cockpits CI caught more failures than fedora gating due to teams not setting up gating or just ignoring it...)

If we could go full "CD" and if all tests pass it would get merged automatically? That would be cool! :)

martinpitt:
I'd be totally on board with writing a gh action that auto-approves/lands image refreshes (and maybe even translation updates in cockpit, but *not* npm) if all tests passed and there is no nack.

martinpitt:
I think this is better, as the amount of work is usually *smaller* when we manage to catch a regression and then -1 it in bodhi, *before* it creeps into updates and then breaks fedora-3x. As long as image refreshes succeed, they are virtually no work.

allisonkarlitskaya:
> I'd be totally on board with writing a gh action that auto-approves/lands image refreshes (and maybe even translation updates in cockpit, but _not_ npm) if all tests passed and there is no nack.

As we discussed in 1:1, let's be careful: this is how Skynet gets started.

allisonkarlitskaya:
I feel like maybe we're approaching this wrong.  What we actually want is to run our tests against fedora-testing more often.  What we don't want is the thrash in the bots/ repository of updating the image so often (plus the small amount of human work required to do that).

Add to this the fact that fedora-testing is not even an automatically-tested image as per testmap.

Why don't we arrange to build a new fedora-testing image and run the tests against it, but not treat it as an image refresh?  We could easily do that once per day, even, if we wanted.  What would be involved in that?

marusak:
> I'd be totally on board with writing a gh action that auto-approves/lands image refreshes (and maybe even translation updates in cockpit, but not npm) if all tests passed and there is no nack.

Right, image refreshes and po refreshes (where we should just really trigger default context). `npm` is not safe in general indeed. There are some packages where it would be safe, but I think we can handle that.

> I think this is better, as the amount of work is usually smaller when we manage to catch a regression and then -1 it in bodhi, before it creeps into updates and then breaks fedora-3x. As long as image refreshes succeed, they are virtually no work.

Agree with this. But not that often we see broken refresh (/ME knocks on wood or whatever one is supposed to do)

> I feel like maybe we're approaching this wrong. What we actually want is to run our tests against fedora-testing more often. What we don't want is the thrash in the bots/ repository of updating the image so often (plus the small amount of human work required to do that).

absolutely.

> Why don't we arrange to build a new fedora-testing image and run the tests against it, but not treat it as an image refresh? We could easily do that once per day, even, if we wanted. What would be involved in that?

We would not be able to reproduce failures. Or might not be able to reproduce failures with the model of repositories and refreshes that fedora is using.

martinpitt:
> Why don't we arrange to build a new fedora-testing image and run the tests against it, but not treat it as an image refresh?

That would be totally fine for me, and also reduce the feedback loop that we can give to bodhi. We can try this for a while, and if we don't like it, go back to committing them.

> We would not be able to reproduce failures

First, the image logs  do show package versions, so usually one can reproduce a failure by image-prepare and then `vm-run --maintain` with manually installing stuff from updates-testing. Second, that nightly run could still image-upload the build *if it fails*, so that it remains in the cache for a few days until it gets pruned.

martinpitt:
FTR, I'm totally fine with this PR as-is, so assume my :+1:  . Moving review to @marusak though, as I don't want to decide that over his head (as acting tech lead).

marusak (approved):
ok, lets land this. All other ideas are out of scope of this PR. We will see how it behaves and if it will become annoying - then we can think of other mentioned approaches.

PR #1397, opened 2020-11-18
Image refresh for centos-7

cockpituous:
Image refresh for centos-7
 * [x] image-refresh centos-7

### DIFF
--- a/images/centos-7
+++ b/images/centos-7
@@ -1,1 +1,1 @@
-centos-7-194a6ed90c17f8e1c719c1cca0f6096aa7649cf2e8fee0c2fd12276bb0cff2db.qcow2
+centos-7-beacc5fba1a4377e345264aec2da284f02dfd0175006a8e6acc791d9bc750a04.qcow2


martinpitt:
@allisonkarlitskaya : Did you manually trigger the "centos-7@cockpit-project/cockpit" context here? That can't work. If so, then we can just ignore this. If our machinery triggered this automatically, we need to teach it not to. I just ran `./tests-trigger 1397 image:centos-7` and that did the right thing.

allisonkarlitskaya:
> @allisonkarlitskaya : Did you manually trigger the "centos-7@cockpit-project/cockpit" context here? That can't work. If so, then we can just ignore this. If our machinery triggered this automatically, we need to teach it not to. I just ran `./tests-trigger 1397 image:centos-7` and that did the right thing.

This was me, yes.  :)  Didn't realise we need a branch for this one too (but it makes sense, of course).

allisonkarlitskaya:
Probable fix in https://github.com/cockpit-project/cockpit/pull/14947

allisonkarlitskaya (approved):

PR #1410, opened 2020-11-24
image-diff: new command for comparing images

allisonkarlitskaya:
This is a hacky/experimental new command for displaying the differences
in installed packages and versions on two different images.
    
That's mostly useful when something broke because of an image refresh
and you'd like to find out what changed.
    
You'd usually want to use this with the full filename of a before- and 
after-refresh version of the same image, for example:
    
  $ ./image-diff fedora-33-[oldhash].qcow2 fedora-33-[newhash].qcow2

### DIFF
--- a/image-diff
+++ b/image-diff
@@ -1,0 +1,74 @@
+#!/usr/bin/python3
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2020 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+
+from machine import testvm
+
+
+def get_packages(machine):
+    # List all packages, irrespective of package manager (rpm or dpkg).
+    # If both are missing, then the command will fail.
+    #
+    # We'd ideally like to get source packages everywhere, but it's a
+    # bit more difficult on RPM. (TODO)
+    pkgcmd = """dpkg-query -W -f='${source:Package}\t${version}\n' 2>/dev/null ||
+                rpm -qa --qf '%{NAME}\t%{EVR}\n' 2>/dev/null"""
+
+    output = machine.execute(pkgcmd)
+
+    return dict(line.split('\t') for line in output.splitlines())
+
+
+parser = argparse.ArgumentParser(description='Compare package versions on VM images')
+parser.add_argument('old', help='the "old" image to compare')
+parser.add_argument('new', help='the "new" image to compare')
+args = parser.parse_args()
+
+# boot the machines in parallel
+old_vm = testvm.VirtMachine(image=args.old)
+new_vm = testvm.VirtMachine(image=args.new)
+
+old_vm.start()
+new_vm.start()
+
+old_vm.wait_boot()
+new_vm.wait_boot()
+
+old_pkgs = get_packages(old_vm)
+new_pkgs = get_packages(new_vm)
+
+old_vm.kill()
+new_vm.kill()
+
+print('Removed:')
+for name in sorted(set(old_pkgs) - set(new_pkgs)):
+    print(f'  {name} ({old_pkgs[name]})')
+print()
+
+print('Added:')
+for name in sorted(set(new_pkgs) - set(old_pkgs)):
+    print(f'  {name} ({new_pkgs[name]})')
+print()
+
+print('Changed:')
+for name in sorted(set.intersection(set(new_pkgs), set(old_pkgs))):
+    if new_pkgs[name] != old_pkgs[name]:
+        print(f'  {name} ({old_pkgs[name]} -> {new_pkgs[name]})')
+print()


allisonkarlitskaya:
A couple of possible TODOs if we want to clean this up a bit:

 * allow running images by filename
 * proper commandline argument parsing
 * run the two images concurrently

allisonkarlitskaya:
Example output:

```
sid:~/src/cockpit$ bots/image-diff fedora-32-fd6c09bcd1d569307f8d5d88e62e3a19606e3d823ffd1db26754f75c8153a8bf.qcow2 fedora-32-a094f4fe6f63482be085de99ed1cf743d7501b0354714bd92ceb1458e06f4705.qcow2
Removed:

Added:
  jq (1.6-4.fc32)
  oniguruma (6.9.5-4.rev1.fc32)

Changed:
  binutils (2.34-5.fc32 -> 2.34-6.fc32)
  binutils-gold (2.34-5.fc32 -> 2.34-6.fc32)
  chromium-common (85.0.4183.121-1.fc32 -> 86.0.4240.111-1.fc32)
  chromium-headless (85.0.4183.121-1.fc32 -> 86.0.4240.111-1.fc32)
  clevis (14-4.fc32 -> 15-2.fc32)
  clevis-luks (14-4.fc32 -> 15-2.fc32)
  corosynclib (3.0.4-1.fc32 -> 3.0.4-2.fc32)
  kernel (5.8.16-200.fc32 -> 5.6.6-300.fc32)
  kernel-core (5.6.6-300.fc32 -> 5.8.18-200.fc32)
  kernel-modules (5.6.6-300.fc32 -> 5.8.18-200.fc32)
  kernel-tools (5.8.11-200.fc32 -> 5.8.18-200.fc32)
  kernel-tools-libs (5.8.11-200.fc32 -> 5.8.18-200.fc32)
  libnfsidmap (1:2.5.1-4.rc4.fc32 -> 1:2.5.2-0.fc32)
  libsmbclient (2:4.12.8-1.fc32 -> 2:4.12.10-0.fc32)
  libteam (1.30-2.fc32 -> 1.31-2.fc32)
  libwbclient (2:4.12.8-1.fc32 -> 2:4.12.10-0.fc32)
  linux-firmware (20200918-112.fc32 -> 20201022-113.fc32)
  linux-firmware-whence (20200918-112.fc32 -> 20201022-113.fc32)
  nfs-utils (1:2.5.1-4.rc4.fc32 -> 1:2.5.2-0.fc32)
  pam (1.3.1-26.fc32 -> 1.3.1-27.fc32)
  pcre (8.44-1.fc32 -> 8.44-2.fc32)
  pcre2 (10.35-7.fc32 -> 10.35-8.fc32)
  pcre2-syntax (10.35-7.fc32 -> 10.35-8.fc32)
  pcre2-utf32 (10.35-7.fc32 -> 10.35-8.fc32)
  python3-ldap (3.1.0-9.fc32 -> 3.3.1-1.fc32)
  python3-perf (5.8.11-200.fc32 -> 5.8.18-200.fc32)
  python3-pytz (2020.1-1.fc32 -> 2020.4-1.fc32)
  python3-urwid (2.1.0-1.fc32 -> 2.1.2-1.fc32)
  samba-client-libs (2:4.12.8-1.fc32 -> 2:4.12.10-0.fc32)
  samba-common (2:4.12.8-1.fc32 -> 2:4.12.10-0.fc32)
  samba-common-libs (2:4.12.8-1.fc32 -> 2:4.12.10-0.fc32)
  teamd (1.30-2.fc32 -> 1.31-2.fc32)
  vim-minimal (2:8.2.1885-1.fc32 -> 2:8.2.1941-1.fc32)
  xen-libs (4.13.1-7.fc32 -> 4.13.2-1.fc32)
  xen-licenses (4.13.1-7.fc32 -> 4.13.2-1.fc32)

```

vm-run
```diff
@@ -160,6 +160,9 @@ try:
     # execute any -e and -s commands
     if args.execute:
         for cmd in args.execute:
+            if cmd == 'poweroff -f':
+                machine.kill()
```
martinpitt, line 164: I wonder about this special case -- what goes wrong with running this directly through `execute()`? We often do this through our tests, even though we usually do some hacks like `m.spawn('sleep 1; poweroff')`. This is a bit too magic for my taste, I'd rather add a `--power-off` option  to it.

But more importantly, why don't you use bots/image-customize for this? This is meant for this kind of sort-running automated/batched processes. We use it all the time in starter-kit and friends.

    ./image-customize -v -r 'rpm -qa' fedora-33

That *almost* gives you the right result -- you need `-v` to show stdout, and then you also get some `set -x` output. The latter goes to stderr, so that's harmless; the only thing you need to fix is to print the machine's UUID to stderr instead of stdout.

image-diff
```diff
@@ -0,0 +1,66 @@
...
+    if any(image.startswith(prefix) for prefix in ['rhel', 'fedora', 'centos']):
+        # TODO: want source packages, but SOURCERPM is not great and the others don't work
+        pkgcmd = "rpm -qa --qf 'PACKAGE\t%{NAME}\t%{EVR}\n'"
+    elif any(image.startswith(prefix) for prefix in ['debian', 'ubuntu']):
+        pkgcmd = "dpkg-query -W -f='PACKAGE\t${source:Package}\t${version}\n'"
+    else:
+        sys.exit("Sorry, don't know how to list packages on " + image)
+
+    vm_run = os.path.join(os.path.dirname(__file__), 'vm-run')
+    output = subprocess.check_output([vm_run, '-e', pkgcmd, '-e', 'poweroff -f', image])
```
martinpitt, line 34: This is already a Python script, so it'd be cleaner and easier to use the machines API directly. Check out machine/testvm.py and machine/machine_core/cli.py or https://cockpit-project.org/blog/cockpit-custom-test-framework.html about the details. No need to go via a cumbersome CLI and parsing stdout for these.

martinpitt (changes requested):
Great idea! I'd just like this to use our Python machines API for simplicity and robustness.

vm-run
allisonkarlitskaya: I got rid of this and followed your suggestion to use the python API.  Thanks for this idea, it's obviously *way* better.

machine/machine_core/constants.py
```diff
@@ -21,6 +21,10 @@
 ATOMIC_IMAGES = ["rhel-atomic"]
 OSTREE_IMAGES = ATOMIC_IMAGES + ["fedora-coreos"]
 
+# Package manager for a given image
+DPKG_IMAGE_PREFIXES = ["debian-", "ubuntu-"]
+RPM_IMAGE_PREFIXES = ["rhel-", "fedora-", "centos-"]
```
martinpitt, line 26: IMHO this should not be public API. These lists are incomplete -- it's perfectly reasonable to diff the services or openshift image, for example. I think we want to remain able to change the detection implementation, and not third-party projects to use this,

machine/machine_core/machine.py
```diff
@@ -74,6 +75,9 @@ def __init__(self, address="127.0.0.1", image="unknown", verbose=False, label=No
         self.image = image
         self.atomic_image = self.image in ATOMIC_IMAGES
         self.ostree_image = self.image in OSTREE_IMAGES
+        self.dpkg_image = any(self.image.startswith(pfx) for pfx in DPKG_IMAGE_PREFIXES)
+        self.rpm_image = any(self.image.startswith(pfx) for pfx in RPM_IMAGE_PREFIXES)
```
martinpitt, line 79: Would you consider a runtime detection instead? `image-customize` does that in `install_packages()`. Perhaps *that* bit would warrant a new public Machines API, `get_package_manager()`? Then we can extend this in the future for Alpine, zypper, etc, and the command would also work for our other images like "services".

machine/machine_core/machine.py
```diff
@@ -214,6 +218,29 @@ def get_admin_group(self):
         else:
             return "wheel"
 
+    def get_packages(self, source=False):
```
martinpitt, line 221: IMHO we should start with putting that logic into imagediff, and only factor it out once we actually have an use case for it.

image-diff
```diff
@@ -0,0 +1,56 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import sys
+
+from machine import testvm
+
+_, old_img, new_img = sys.argv
```
martinpitt, line 24: Not the most friendly failure mode, but as that's a developer tool, it may be acceptable  :grin: Maybe add an `if len(sys.argv) != 2:` help message, or use `argparse` if you want to get fancy :)

martinpitt (changes requested):
Nice! I think we should call this in the generated command in `image-trigger`, so that we get a package diff in the automatic image refresh logs. This will also make sure that this script actually keeps working.

I would *not* add it to `image-refresh` directly, as that is often used by human developers, and  there may not be an old image to compare against.

machine/machine_core/machine.py
allisonkarlitskaya: Ya.  I agreed.  I'll move it back.

image-diff
allisonkarlitskaya: I wanna get fancy.  This is still a draft for a reason :)

mvollmer:
I sometimes want to know which version of a package is in a given image.  I guess this tool can trivially do that, too, right?  Kind of like a diff between an empty image and a given one.

marusak:
> I sometimes want to know which version of a package is in a given image. 

You can just do something like: (sure, special command would be nice, just wanted to point out)
```
$ ./bots/image-customize -v -r "rpm -q kernel" fedora-32
a1139753-5ce4-4052-a0ad-87959bb103bf
+ rpm -q kernel
kernel-5.6.6-300.fc32.x86_64
kernel-5.8.18-200.fc32.x86_64
killing ssh master process 68871
```

allisonkarlitskaya:
I have some vague ideas about eventually expanding this to some kind of a `image-query` tool which would include different subcommands like `image-query diff`, `image-query pkgversion`, etc.  That can wait, though.

image-diff
```diff
@@ -0,0 +1,80 @@
...
+
+import argparse
+
+from machine import testvm
+
+
+def get_packages(machine):
+    if any(machine.image.startswith(p) for p in ['debian', 'ubuntu']):
+        pkgcmd = "dpkg-query -W -f='PACKAGE\t${source:Package}\t${version}\n'"
+    elif any(machine.image.startswith(p) for p in ['rhel', 'fedora', 'centos']):
```
martinpitt, line 28: That still won't catch images like "openshift" or "services". I really suggest to detect this dynamically; as a first suggestion:
```sh
if dpkg -s bash >/dev/null  2>&1; then
    ... it's dpkg;
elif rpm -q bash >/dev/null 2>&1; then
   ... it's rpm;
else
   ... unknown package system, error out;
fi
```

image-diff
```diff
@@ -0,0 +1,80 @@
...
+        # SOURCERPM is not great and the others don't work
+        pkgcmd = "rpm -qa --qf 'PACKAGE\t%{NAME}\t%{EVR}\n'"
+    else:
+        raise ValueError("Can only list packages on rpm/dpkg, not " + machine.image)
+
+    output = machine.execute(pkgcmd)
+
+    packages = {}
+    for line in output.splitlines():
+        if line.startswith('PACKAGE'):
```
martinpitt, line 38: Seems redundant to have this PACKAGE prefix? These commands are meant to print exactly what you ask them to print, not anything extra. (But that's not a veto)

image-diff
```diff
@@ -0,0 +1,80 @@
...
+            _, name, version = line.split('\t')
+            packages[name] = version
+
+    return packages
+
+
+parser = argparse.ArgumentParser(description='Compare CI images')
```
martinpitt, line 45: "Compare package versions on VM images"?

image-diff
```diff
@@ -0,0 +1,80 @@
...
+new_vm.start()
+
+old_vm.wait_boot()
+new_vm.wait_boot()
+
+old_pkgs = get_packages(old_vm)
+new_pkgs = get_packages(new_vm)
+
+old_vm.kill()
+new_vm.kill()
```
martinpitt, line 64: DRY -- the VM handling could move to get_packages() or a new function, and then you just call it with the two images.

martinpitt (changes requested):
Nice! A few more nitpicks, but this is well into the "fun bantering" part now :grin: 

To exercise this code path regularly, and to make life easier for our pilot, I'd like to integrate this into image-trigger/image-refresh (the latter will need a `--diff` option or so), so that we get that package diff in the image build logs. That can of course  be a follow-up.

image-diff
allisonkarlitskaya: I added this when I was dealing with `vm-run` and getting some extra output.  I could probably remove it now.

image-diff
allisonkarlitskaya: sure

image-diff
allisonkarlitskaya: I can't, though: I want to start the VMs in parallel and wait for them to boot in parallel, otherwise this takes twice as long.

I could probably move the `.wait_boot()` to inside `get_packages()` though.

image-diff
allisonkarlitskaya: naja.  i was actually thinking to run both commands on every host.  the list of rpm packages installed on a debian system is gonna be pretty small :)

martinpitt: That's a great idea as well. One *can* install packages from a foreign package manager, but (1) it's not going to be the case for our CI images, and (2) *if* it is, then it's actually correct to include them. You just need to account for one or the other executable not being present at all, but at the same time we shouldn't ignore errors if the dpkg/rpm command fails.

image-diff
martinpitt: Ah, good point. So both the original and your new idea of moving just wait_boot() seem fine.

image-diff
allisonkarlitskaya: I went with a fairly reasonable approach for now:  `dpkg 2>/dev/null || rpm 2>/dev/null`.  Will give us the output, either way, and will fail if neither is installed.  If both happen to be installed, then we'll have a problem ya...

image-diff
allisonkarlitskaya: This resulted in a *very* nice cleanup.  Thanks!

image-diff
allisonkarlitskaya: I moved `wait_boot()` since it's a proper precondition of the `.execute()` in `get_packages()`, but `.kill()` feels too weird, so i'll leave it duplicated.

I briefly entertained putting both VMs into a list or a tuple and using some weird iterator notation to call methods on all items in the list but this just made things really unreadable.

allisonkarlitskaya: I take it back.  I actually prefer `.wait_boot()` also split out.  I've changed the comment above to emphasize that a bit more.

martinpitt (approved):
'nuff bikeshedding, thanks!

PR #1415, opened 2020-11-26
po-refresh: Don't run all tests

marusak:
Run just default context. Also mark as release blockers.

Example (dummy) run: https://github.com/marusak/cockpit/pull/35 (notice labels and triggered test, all done by just this task)

### DIFF
--- a/po-refresh
+++ b/po-refresh
@@ -26,6 +26,7 @@ import sys
 sys.dont_write_bytecode = True
 
 import task
+from task import github, testmap
 from machine.machine_core.constants import BASE_DIR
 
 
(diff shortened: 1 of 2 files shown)


po-refresh
```diff
@@ -168,19 +169,25 @@ def run(context, verbose=False, **kwargs):
...
+        branch = "{0}:{1}".format(user, local_branch)
+
+    # Create a pull request from these changes
+    pull = task.pull(branch, labels=['bot', 'no-test'], run_tests=False, **kwargs)
+
+    # Trigger this pull request
+    api = github.GitHub()
+    head = pull["head"]["sha"]
+    api.post("statuses/{0}".format(head), {"state": "pending", "context": TEST_OS_DEFAULT,
+                                           "description": github.NOT_TESTED_DIRECT})
```
martinpitt, line 190: Did you check that this is true for all our projects? E.g. cockpit-ostree comes to my mind, which can't run on fedora-32. Possibly we need to keep only the `no-test` bit and trigger the tests from the per-project workflows?

marusak: good point!
I think we can do some helper function in `testmap.py` and the logic is going to be simple - for `cockpit-project/cockpit` trigger the default context, for anything else trigger the `master` branch context. WDYT?

marusak: Updated

po-refresh
```diff
@@ -168,19 +169,28 @@ def run(context, verbose=False, **kwargs):
...
+        branch = "{0}:{1}".format(user, local_branch)
+
+    # Create a pull request from these changes
+    pull = task.pull(branch, labels=['bot', 'no-test'], run_tests=False, **kwargs)
+
+    # Trigger this pull request
+    api = github.GitHub()
+    head = pull["head"]["sha"]
+
+    triggers = testmap.tests_for_po_refresh(api.repo)
```
martinpitt, line 190: To avoid having to maintain another map, how about this: *if* TEST_OS_DEFAULT is in the existing map, then only run this, otherwise run all contexts (i.e. don't set a no-test label)

marusak: That assumes that every project that uses default context also wants just that triggered on po refreshes. I don't want to do this assumption.  I think we want that refresh to be treated as any other PR, unless we tell differently - thus the map.

martinpitt: OK, fair enough. We can change that in the future if we want to.

martinpitt (approved):
Thanks!

PR #1430, opened 2020-12-02
testmap: Test cockpit-podman master on rhel-8, not the branch

mvollmer:
We don't really use the rhel-8 branch for anything.  Releases of
cockpit-podman to RHEL are made by a third party by copying packages
from Fedora into some RHEL modules, apparently.

But it's always good to not break upstream cockpit-podman on RHEL, so
test master there.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -81,13 +81,12 @@ REPO_BRANCH_CONTEXT = {
     },
     'cockpit-project/cockpit-podman': {
         'master': [
+            'rhel-8-3',
+            'rhel-8-4',
             'fedora-32',
             'fedora-32/rawhide',
             'fedora-33',
             'debian-testing',
-        ],
-        'rhel8': [
-            'rhel-8-3',
         ],
         '_manual': [
             'centos-8-stream',


marusak:
For this to land, we first need to test that rhel-8-3 and rhel-8-4 work on c-podman master.
There are two ways how to do this:
1. First add these context into `__manual__`, we can land it, test it and if it succeeds, then we can test it in master
2. Keep this PR, jump into webhook container into bots checkout, pick this commit, `test-scan` this PR and then cleanup after yourself

The first one is the lazy way and also the "correct" way, the other one is the faster way with a bit of manual work. The second one can become annoying if there are needed some fixups in c-podman first.

mvollmer:
> For this to land, we first need to test that rhel-8-3 and rhel-8-4 work on c-podman master.

Yes, good point, let's do it the lazy way.

mvollmer:
It looks like a longer journey to get master green on rhel-8-3 and rhel-8-4, see https://github.com/cockpit-project/cockpit-podman/pull/617. Let's replace this with #1433

PR #1427, opened 2020-12-02
Image refresh for rhel-8-4

cockpituous:
Image refresh for rhel-8-4
 * [x] image-refresh rhel-8-4

PR #1434, opened 2020-12-03
Image refresh for rhel-8-4 (+naughty)

allisonkarlitskaya:
I can't seem to push to #1427, so here's a new one.


marusak (approved):
Thanks!
Can you please trigger all relevant tests and once they are green (and bots will comment on #1435) then this can land :)

PR #1444, opened 2020-12-04
task: Query default branch via api

Gundersanne:
This is a requirement to make pull requests created by scripts work with
different default branches.

----

`osbuild/image-builder-frontend` recently switched to `main` as a default branch and the `npm-update` workflow stopped working because it would try to create a PR against a branch which doesn't exist. With this patch it'll query the default branch from the api.

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -169,14 +169,16 @@ class GitHub(object):
         if not self._url:
             if not self._base:
                 netloc = os.environ.get("GITHUB_API", "https://api.github.com")
-                self._base = "{0}/repos/{1}/".format(netloc, self.repo)
+                self._base = "{0}/repos/{1}".format(netloc, self.repo)
 
             self._url = urllib.parse.urlparse(self._base)
 
         return self._url
 
     def qualify(self, resource):
-        return urllib.parse.urljoin(self.url.path, resource)
+        if resource is None:
+            return self.url.path
+        return urllib.parse.urljoin("{0}/".format(self.url.path), resource)
 
     def request(self, method, resource, data="", headers=None):
         if headers is None:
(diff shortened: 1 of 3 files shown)


Gundersanne:
forced an npm-update run: https://github.com/osbuild/image-builder-frontend/pull/61

task/github.py
```diff
@@ -169,14 +169,16 @@ def url(self):
...
+                self._base = "{0}/repos/{1}".format(netloc, self.repo)
 
             self._url = urllib.parse.urlparse(self._base)
 
         return self._url
 
     def qualify(self, resource):
-        return urllib.parse.urljoin(self.url.path, resource)
+        if resource is None:
+            return self.url.path
```
martinpitt, line 180: That looks unrelated -- what's that for? Should it be a separate commit?

martinpitt (commented):
Thanks, this is nice! Just one question.

task/github.py
Gundersanne: ah yea, this should be a separate commit. Doing that :)

Gundersanne: Wait on second thought it shouldn't be, but it is kinda confusing. I can add a bit more explanation when squashing.

So `/repos/:owner/:repo` is also an endpoint, this has the `default_branch` value in it. And `qualify()` always appends an `/`, which breaks that particular endpoint.

This way `api.get()` will query the `/repos/:owner/:repo` endpoint.

martinpitt: Thanks. I can't say that I fully understand this, but fine to land then with some extra explanation.

martinpitt (approved):

task/github.py
Gundersanne: So an api call to `/repos/cockpit-project/cockpit` returns an object; an api call to `/repos/cockpit-project/cockpit/` returns None

martinpitt: Yes, I get that part, that's line 172. I meant the default fallback of "resource" here.

PR #1446, opened 2020-12-05
Image refresh for fedora-33

cockpituous:
Image refresh for fedora-33
 * [x] image-refresh fedora-33

PR #1445, opened 2020-12-05
Image refresh for fedora-32

cockpituous:
Image refresh for fedora-32
 * [x] image-refresh fedora-32

PR #1450, opened 2020-12-07

cockpituous:


marusak:
Same as #1445 and #1446

martinpitt:
#1445 is done, but failing on kdump now, so keeping needswork

PR #1425, opened 2020-12-01
images: Bootstrap rhel-8-3 from latest-RHEL-8.3.1

mvollmer:
The latest-RHEL-8.3 locations seems to be broken right now for use
with virt-install, as it is missing the bootloader images.  The
latest-RHEL-8.3.1 location seems to work.

 * [x] image-refresh rhel-8-3

PR #1452, opened 2020-12-07
Revert "images: Bootstrap rhel-8-3 from latest-RHEL-8.3.1"

mvollmer:
It is correct to use latest-RHEL-8.3 in order to follow along with the
latest 8.3 nightly bits.

We keep using rhel-8.3.1-build with mock, since there doesn't seem to
be no "build stuff that corresponds to latest RHEL 8.3".

This reverts most of commit 0b7b1ffb2a042c7755f8bdbf690e4bf099aad309.

 * [x] image-refresh rhel-8-3


mvollmer:
In #1425 we started using latest-RHEL-8.3.1 for our rhel-8-3 images since latest-RHEL-8.3 was broken. This is being fixed and the fix will come to the mirrors eventually, and then we might want to switch back to latest-RHEL-8.3 so that we will automatically catch RHEL 8.3.2 when it happens.

However: we still need to manually change the mock config to rhel-8.3.2 at that point. Also, I have low confidence that the latest-RHEL-8.3 symlink will not break again when RHEL 8.3.2 happens. So I think we might just as well keep using latest-RHEL-8.3.1.

We might have a little daily task that scans the RHEL mirrors for new stuff.  (And for inconsistencies such as broken latest-RHEL-X.Y links that we can then report.)

mvollmer:
Or, of course!, we let the image creation scripts do the mirror scanning when they create the image.

cockpituous:
image-refresh in progress on 4-ci-srv-02.
Log: https://logs.cockpit-project.org/logs/image-refresh-1452-20201207-094747/

cockpituous:
image-refresh rhel-8-3 done: https://github.com/cockpituous/bots/commits/image-refresh-rhel-8-3-20201207-100606

mvollmer:
Oh look, that worked already.

martinpitt:
I close this for now, as it cause automatic refreshes to completely stop. As you noticed, we need to bump the mock buildroot URL anyway, and I also want to move to qcow2 images (which will change the URL anyway).

PR #1459, opened 2020-12-09
testmap: enable rhel-8-4 testing for cockpit-composer

jkozol:
cockpit-composer will now be releasing into rhel 8.4 so the 8.4 tests are now enabled. I will update the `rhel-8` branch tests to 8.4 once I release into the nightlies and update the branch. https://github.com/osbuild/cockpit-composer/pull/1189 shows a successful rhel-8-4 test run.

Also, what is the purpose of the `_manual` tests section? I was able to manually trigger the rhel-8-4 test even though it is not in the `_manual` tests section.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -128,8 +128,7 @@ REPO_BRANCH_CONTEXT = {
             'fedora-32',
             'fedora-32/firefox',
             'fedora-33',
-            'rhel-8-3',
-            'rhel-8-3/firefox',
+            'rhel-8-4',
         ],
         'rhel-8': [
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
You can manually trigger "rhel-8-3" because that is already in the "rhel-8" branch. It just needs to be on *any* branch, and we just use `_manual` as a convention as that is (practically) guaranteed to not be a real branch name on git projects. It has no particularly special behaviour otherwise. So you wouldn't strictly need to add it to the `_manual` entry again, but it doesn't hurt either.

PR #1463, opened 2020-12-11
tests: Move TEST_OS_DEFAULT to fedora-33

mvollmer:


martinpitt (dismissed):
Thanks!

martinpitt:
Note that we shoudl test this: It influences the debug/production build of cockpit at least, so we should validate that a fedora 33 debug build works as expected. Also, all the multi-machine and networking tests use that as auxiliary image.

martinpitt (approved):
Ah, needed another rebase? Thanks!

PR #1478, opened 2020-12-17
testmap: Introduce f-33/rawhide context

marusak:
This will use the real `rawhide`. F-32/rawhide is really just f-33
preview.

Whats more, in F34 a new podman version 3 is being created. I think we want to test this ASAP.


martinpitt (approved):
Thanks!

PR #1511, opened 2021-01-04
images: Build rhel-7-9 from official qcow image

martinpitt:
Starting from the cloud image is faster, more robust, and more realistic
(using an official product) than our own virt-install bootstrap.

 * [x] image-refresh rhel-7-9


cockpituous:
image-refresh in progress on 4-ci-srv-06.
Log: https://logs.cockpit-project.org/logs/image-refresh-1511-20210104-081616/

cockpituous:
image-refresh rhel-7-9 done: https://github.com/cockpituous/bots/commits/image-refresh-rhel-7-9-20210104-083030

martinpitt:
[previous refresh](https://logs.cockpit-project.org/logs/image-refresh-1507-20210103-200808/log) took 1414s, image size is 2.3 GB.  This refresh took just 818s (almost half the time!). Image size shrank to 2.0 GB, due to dropping node mostly, I guess.

A local image build is much more fun that way, particularly because our cloudimage.bootstrap locally caches the downloaded image. So iterating is much faster.

Of course I'll also apply that to the rhel-8 images, but we need to sort out the rhel-8-4 refresh failure first.

martinpitt:
This [regresses the docker tests](https://logs.cockpit-project.org/logs/pull-1511-20210104-101950-febe7ad8-rhel-7-9-cockpit-project-cockpit-rhel-7.9/log.html), so they'll need some adjustment to the (presumably) changed partitioning layout etc. That's the kind of "test official product" thing that I want to achieve with this.

martinpitt:
Meh -- fixing `TestDockerStorage.testBasic` is relatively easy:
```diff
--- test/verify/check-docker-storage
+++ test/verify/check-docker-storage
@@ -132,7 +132,18 @@ class TestDockerStorage(MachineCase):
         m = self.machines["machine1"]
         b = self.browser
 
-        m.execute("systemctl start docker")
+        # create an LVM docker pool, and copy an image into it
+        m.add_disk("300M", serial="DISK0")
+        m.execute('''
+        systemctl start docker
+        docker save -o /tmp/busybox.img docker.io/busybox:latest
+        systemctl stop docker
+        printf "VG=containers\nDEVS=$(ls /dev/disk/by-id/*QEMU*_DISK0)\nMIN_DATA_SIZE=100M\n" > /etc/sysconfig/docker-storage-setup
+        rm /etc/sysconfig/docker-storage
+        container-storage-setup
+        systemctl start docker
+        docker load < /tmp/busybox.img
+        ''')
 
         self.login()
```

But `TestDockerStorage.testDevicemapper` is a huge PITA: It makes assumptions about `atomic storage reset`, and the latter just does something bogus on an image which is already overlay2. It's way too late to fix `atomic storage` at this point, so I think I call rhel-7-9 a lost cause for now, and rather look at rhel-8*.

PR #1513, opened 2021-01-04
po-refresh: Don't fail when nothing to update

marusak:
There are 3 cases we need to count with:
1. Some language was added/dropped
2. No language was added/dropped but some language was updated
3. Some language was added/dropped and some language was updated

All of these are handled correctly. However there is special case that
nothing was added/dropped nor there is nothing to update. In this case
it would fail, as `branch` would be `None` and when we would try to
create pull request we would get error message:
```
{'message': "Invalid request.\n\nFor 'properties/head', nil is not a string.", 'documentation_url': 'https://docs.github.com/rest/reference/pulls#create-a-pull-request'}
```
and we would try to get `number` property from this.

This was broken in 234053cfc4. (See the `if branch/elif local_branch)

Fixes broken po refresh in c-podman: https://github.com/cockpit-project/cockpit-podman/actions/runs/461677922

### DIFF
--- a/task/__init__.py
+++ b/task/__init__.py
@@ -460,13 +460,7 @@ def pull(branch, body=None, issue=None, base=None, labels=['bot'], run_tests=Tru
         if body:
             data["body"] = body
 
-    pull = api.post("pulls", data, accept=[422])
-
-    # If we were refused to grant maintainer_can_modify, then try without
-    if "errors" in pull:
-        if pull["errors"][0]["field"] == "fork_collab":
-            data["maintainer_can_modify"] = False
-        pull = api.post("pulls", data)
+    pull = api.post("pulls", data)
 
     # Update the pull request
     label(pull, labels)
--- a/po-refresh
+++ b/po-refresh
@@ -180,6 +180,10 @@ def run(context, verbose=False, **kwargs):
         task.push_branch(user, local_branch)
         branch = "{0}:{1}".format(user, local_branch)
 
+    if not branch:
+        # Nothing to update
+        return
+
     # Create a pull request from these changes
     pull = task.pull(branch, labels=['bot', 'no-test', 'release-blocker'], run_tests=False, **kwargs)
 


marusak:
We accept error code 422 when pull fails to be created. Looking now in history why is that (the history is in cockpit, so a bit more archeology)
If we would drop that, we would get this error message:
```
$ ./bots/po-refresh 
.................................................................................. done.
............................................................................... done.
............................................................................... done.
................................................................................... done.
echo "Updated"
Updated
rm ./po/*.po
cp tmp/weblate-repo/*.po ./po/
HEAD is now at a1cca62d po: Update from Fedora Weblate
M	Makefile
po-refresh: Error accessing /repos/cockpit-project/cockpit/pulls
  Status: 422
  Reason: Unprocessable Entity
  Response: {"message":"Invalid request.\n\nFor 'properties/head', nil is not a string.","documentation_url":"https://docs.github.com/rest/reference/pulls#create-a-pull-request"}
```
which is so much more readable.

marusak:
ok, the culprit https://github.com/cockpit-project/cockpit/commit/4c4160e875292d89bf3d97cad1e3a96c2b4e4067

I wonder if this is still actually needed. I got this error while playing around:
```
    if pull["errors"][0]["field"] == "fork_collab":                             
KeyError: 'field'         
```
so makes me think if we ever hit this.

marusak:
Hmm, thinking about it, it might be the reason why we sometimes cannot push into bots' PRs. So this might be used rather often.

marusak:
Ok, I think I found up a solution how not to break that behavior and also get the nicely readable error message. Pushed.

marusak:
With just the second commit it would end up with:
```
$ ./bots/po-refresh 
.............................................................................. done.
.................................................................................. done.
............................................................................... done.
..................................................................................... done.
echo "Updated"
Updated
rm ./po/*.po
cp tmp/weblate-repo/*.po ./po/
HEAD is now at a1cca62d po: Update from Fedora Weblate
M	Makefile
po-refresh: Error accessing /repos/cockpit-project/cockpit/pulls
  Status: 422
  Reason: Unprocessable Entity
  Response: {"message":"Invalid request.\n\nFor 'properties/head', nil is not a string.","documentation_url":"https://docs.github.com/rest/reference/pulls#create-a-pull-request"}
```
instead of with:
```
...
  File "/home/runner/work/cockpit-podman/cockpit-podman/bots/task/__init__.py", line 285, in run
    ret = function(context, **kwargs)
  File "bots/po-refresh", line 184, in run
    pull = task.pull(branch, labels=['bot', 'no-test', 'release-blocker'], run_tests=False, **kwargs)
  File "/home/runner/work/cockpit-podman/cockpit-podman/bots/task/__init__.py", line 472, in pull
    label(pull, labels)
  File "/home/runner/work/cockpit-podman/cockpit-podman/bots/task/__init__.py", line 509, in label
    resource = "issues/{0}/labels".format(issue["number"])
KeyError: 'number'
po-refresh: Task threw an exception
```
```

File: task/__init__.py [hunk 1]
```diff
@@ -460,13 +460,17 @@ def pull(branch, body=None, issue=None, base=None, labels=['bot'], run_tests=Tru
...
-    # If we were refused to grant maintainer_can_modify, then try without
-    if "errors" in pull:
-        if pull["errors"][0]["field"] == "fork_collab":
-            data["maintainer_can_modify"] = False
+    pull = {}
+    try:
         pull = api.post("pulls", data)
+    except RuntimeError as e:
+        # If we were refused to grant maintainer_can_modify, then try without
+        if "errors" in pull:
```
martinpitt, line 468: This looks rather pointless -- if the `except` happens, then `pull` is *guaranteed* to not get assigned from the api.post result, as there is no result. I.e. it will *always* be `{}` here from line 463. Does the error `e` have anything useful that we could check? If not, I'm ok with just blindly re-trying without fork_collab.

In the next weeks I also would really like to get rid of our cockpituous forks, and let such workflows push directly into origin. This avoids a lot of moving parts, permission problems, is much more portable to new projects, and avoids this pile-up of thousands of branches in the cockpituous forks.

martinpitt (changes requested):
Thanks for figuring out the root cause of this! I'm glad to see that it's "mostly harmless" and would not break actual updates. So let's dwell on this a little bit.

task/__init__.py
marusak: > This looks rather pointless -- if the except happens, then pull is guaranteed to not get assigned from the api.post result, as there is no result. I.e. it will always be {} here from line 463.

oh, you are right!

>  Does the error e have anything useful that we could check?

I cannot reproduce this (or just not sure how). I believe that the message in reason should be usable, but I don't know how it really looks like.

> I'm ok with just blindly re-trying without fork_collab.

What about doing now just blind retry with the same data and if it fails with `fork_collab` twice in a row, we will see the error message and we can put it back in (or check if we can bump up rights to bots in some way if it is happening always).

martinpitt: I don't believe it is a race condition, but let's do that indeed -- if it fails, let's investigate it properly.

marusak: Fair enough, if we don't think it is a race, then lets not do that ugly try/except but lets just go with one try only. If that fails we can take a look why it is happening and what we can do.

martinpitt (approved):
Agreed, let's try this. Thanks!

PR #1524, opened 2021-01-07
testmap updates for rhel-8-4

martinpitt:
cockpit rhel-8-4-distropkg landed yesterday: https://github.com/cockpit-project/cockpit/pull/15103

But still triggering it here again for good measure.


marusak (approved):
Thanks!

PR #1526, opened 2021-01-07
testmap: Automatically run cockpit-podman/ubuntu-stable

martinpitt:
We can't run anything on Travis any more with the free plan, so run
ubuntu-stable tests on our infra instead.


marusak (approved):
Thanks!

PR #1528, opened 2021-01-07
task: Retry when cannot grant maintainer_can_modify

marusak:
As seen in https://github.com/cockpit-project/bots/issues/1522


martinpitt (approved):
Thanks! Approving with a sigh :grin:

PR #1529, opened 2021-01-07
images: Apply crypttab workaround to rhel-8-4-distropkg

martinpitt:
Commit ca7ef1fba0a which introduced rhel-8-4-distropkg landed in
parallel with commit e61562e3ad6d9d which moved the rhel-8-4 image to a
cloud image. The latter introduced a workaround which was missing in
thhe -distropkg version.

### DIFF
--- a/images/scripts/rhel-8-4-distropkg.install
+++ b/images/scripts/rhel-8-4-distropkg.install
@@ -3,3 +3,6 @@
 set -e
 
 /var/lib/testvm/fedora.install --rhel "$@"
+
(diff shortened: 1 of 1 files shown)


martinpitt:
This causes all cockpit PRs to [fail like this](https://logs.cockpit-project.org/logs/pull-15108-20210107-161744-393942a3-rhel-8-4-distropkg/log.html). Thus this is a bit urgent.

marusak (approved):
Thanks!

PR #1531, opened 2021-01-08
Image refresh for debian-testing - fixed in #1544

cockpituous:


marusak:
podman fails with `Failed to get unit file state for docker.service: No such file or directory`
But it seems like it comes from it being installed. Do we have some thing for docker in debian?

marusak:
Ok, so that docker thing is something we should get rid of, but it actually fails on:
```
Error: default OCI runtime "crun" not found: invalid argument
```

marusak:
And also for  cockpit a new message:
```
audit: type=1400 audit(1610088271.168:30): apparmor="DENIED" operation="capable" profile="libvirtd" pid=1915 comm="rpc-worker" capability=39  capname="bpf"
```

PR #1534, opened 2021-01-08
Drop unused trigger scripts

marusak:
We don't use this triggering scripts anymore and we directly call
associate actions from GH actions.

Also we use `naughty-trigger`. We could just call `naughty-prune` directly. If you feel it would run way too often, we can dial it back with creating a separate GH action and run that one once a week. Especially since naughty-trigger just opens issue and then closes it back up. It does not even reuse it to create PR as `image-trigger` does. @martinpitt WDYT?


martinpitt (approved):
Thanks for the cleanup!

Indeed it makes sense to run naughty-prune directly. It will need a little bit of care as it currently requires the cockpituous token to create PRs, not just the github one. Of course I'd like it to just create branches in origin, with the github token -- but we again we run the bots unit tests in github as well, and the token does not have the permission to do that (although it's not really necessary in that case). I.e. essentially we have the same situation as npm-update.

So in short, I'm all for it, but it needs some care, and maybe we should figure this out with npm-update or po-refresh first?

marusak:
> we should figure this out with npm-update or po-refresh first?

Agree.

PR #1535, opened 2021-01-08
testmap: Use F33 for podman rawhide testing

marusak:
Succesful run in https://github.com/cockpit-project/cockpit-podman/pull/648


martinpitt (approved):
Yay!

PR #1544, opened 2021-01-12
Allow naughty apparmor denial message for new capname bpf + debian-testing refresh

### LINKED ISSUE #1531: Image refresh for debian-testing
 * [ ] image-refresh debian-testing

skobyda:
- [x] blocked on https://github.com/cockpit-project/cockpit-podman/pull/651

Since apparmor 2.13, the new capabilities 'bpf' and 'perfmon' has been
added. This has been causing apparmor denial when a VM is defined and
started

Closes #1531


marusak (dismissed):
Can you please first very if works? Ideally by running CI with it. We cannot test two bots PR against each other, so if you could puhs this commit into https://github.com/cockpit-project/bots/pull/1531 that would be great. If you don't have permissions to do that (it happens sometimes) then, pick the commit from #1531 here and then run all related tests. Thanks!

martinpitt:
Nice job on the Debian downstream report! With simple reproducer and all :+1:

martinpitt:
Yes, I think you can't push into #1531, at least I can't. I recommend the following:
 - grab the image refresh commit from #1531 and commit it here
 - Add "Closes #1531" to the commit message
 - append to #1531 title something like "- fixed in #1544"

Thanks!

marusak:
...

- append to #1531 title something like "- fixed in #1544"
- trigger all related tests (see tests-trigger --help for the `image:` syntax)

martinpitt:
@skobyda: It seems [podman tests](https://logs.cockpit-project.org/logs/pull-1544-20210112-202906-d0dfafa4-debian-testing-cockpit-project-cockpit-podman/log.html) are blocked on https://github.com/cockpit-project/cockpit-podman/pull/651 (which is not working yet). Please make a `[ ] blocked on ...` list in the description on such PRs and label them with "blocked", so that it's easy to see at a glance what else needs to be done first (often enough there is more than one dependency).

It also seems that the naughty [does not work yet](https://logs.cockpit-project.org/logs/pull-1544-20210112-202906-d0dfafa4-debian-testing-cockpit-project-cockpit/log.html#124). The standard way how to test this quickly is like this:

 - Copy the raw log of the failed test into some file, e.g. /tmp/t
 - Adjust your bots directory for the new naughty
 - In cockpit, run `bots/tests-policy -o debian-testing < /tmp/t`. That should turn the `not ok` into `ok ... SKIP #12345` on success, otherwise you need to refine the pattern.

martinpitt:
@skobyda Nice, that seems to work now. I keep forgetting about when we use regexps (in `allow_journal_message()`) and when glob patterns (in naughty).

naughty/debian-testing/1543-apparmor-denial-capnames-perfmon-and-bpf
```diff
@@ -0,0 +1 @@
+*apparmor="DENIED" operation="capable" profile="libvirtd" * comm="rpc-worker" capability=39  capname="bpf"
```
marusak, line 1: The commit message talks about `bpf and perfmon` but here I only see `bpf`.

marusak:
> Nice, that seems to work now.

If you say that based on a green test, then you are for a surprise when you check the run :) This is just `test/bots` test once again.

@skobyda this context is misleading in bots PR. You need to specify which repo you want to test.
What you want to trigger is:
```
 ./bots/tests-trigger --repo cockpit-project/bots 1544 image:debian-testing
```

skobyda:
Thanks, retrying

naughty/debian-testing/1543-apparmor-denial-capnames-perfmon-and-bpf
skobyda: I will reword the commit. (I think adding the second pattern would be unnecessary)

martinpitt:
Retrying podman, the fix just landed. This should work now. Nice work, @skobyda ! :100:

martinpitt (dismissed):

martinpitt:
Argh, [still a podman failure](https://logs.cockpit-project.org/logs/pull-1544-20210114-061426-870a4fee-debian-testing-cockpit-project-cockpit-podman/log.html#7-2), this looks rather different, though.

marusak:
debian-testing now can show usage for rootless container. We just need to add it into list of supported systems.

@skobyda It is always good to trigger tests with new bots, when you are fixing refresh. That way you would get run with this image in the cockpit-podman PR and could see that there is more work needed. `tests-trigger --bots-pr` FWIW

marusak:
Retrying tests

marusak (approved):
Thanks!

PR #1548, opened 2021-01-13
images: Fix priority between mcast1 and static-{mac} interfaces on fedora-coreos

martinpitt:
`Machine.set_address()` generates a `static-{mac}` NetworkManager
connection for our multicast interface with a real IP. This collided/was
racing with our static "mcast1" connection (which we normally use to
avoid timing out DHCP, as a single VM is not connected to anything):
NM just threw a dice which connection to activate, and if it was
"mcast1", a multi-VM test was not connected to its peers.

This caused TestShutdownRestart.testBasic() to fail on "Cockpit could
not contact the given host" about half of the time, our top flake on
<https://images-frontdoor.apps.ocp.ci.centos.org/failure-stats.html>.

Fix this by giving `mcast1` a very low priority, so that any
runtime-created connection with default priority 0 will win.


cockpituous:
image-refresh in progress on 3-cockpit-10.
Log: https://logs.cockpit-project.org/logs/image-refresh-1548-20210113-153636/

martinpitt:
I can reproduce that failure fairly easily locally. With this adjustment I got 5/5 successes.

cockpituous:
image-refresh fedora-coreos done: https://github.com/cockpituous/bots/commits/image-refresh-fedora-coreos-20210113-154242

martinpitt:
Voilà, zero retries.

marusak (approved):
Nice, thanks!

PR #1551, opened 2021-01-15
Image refresh for rhel-8-4 (being replaced by #1572)

cockpituous:
Image refresh for rhel-8-4
 * [x] image-refresh rhel-8-4


marusak:
`Command 'timedatectl set-ntp off' returned non-zero exit status 1.` needs work

marusak:
```
timedatectl set-ntp off
Failed to set ntp: NTP not supported
```

marusak:
aha, `timedatex` is missing from this new image. If I install it, then it all works as expected.
In the current image, the only package that has some dep. is chrony with recommending. This has not changed with the new package.
I though that a new dnf version is present that does not resolve it the same way, but dnf has not changed.
So I think that it was dropped from the base image by RCM. I would have thought, that if they have dropped it other tools count on that, which does not seem to be the case. I'll open BZ and ask around.

marusak:
Actually, before I do that, let's try a new refresh, maybe some tool was not updated when we did this refresh.

martinpitt:
@marusak: That means we could also adjust our tests to check for and get along with "NTP service: n/a", and skip the NTP parts then? After all, disabling NTP is not necessary when it's not running in the first place :-)

marusak:
> That means we could also adjust our tests to check for and get along with "NTP service: n/a", and skip the NTP parts then? After all, disabling NTP is not necessary when it's not running in the first place :-)

Right, we can do that, but the image seems broken anyway. And if it stops using timedatex we can enable test in test/verify/check-system-info 

But if it does not work yet, I'll add then HACK in our code, to just work with NTP being n/a

PR #1554, opened 2021-01-17
Image refresh for ubuntu-2004

cockpituous:
Image refresh for ubuntu-2004


marusak:
testJournal flaked twice, I just merged PR fixing it, so I'll retry just to see, it helps :)
Then testAddDiskDirPool failed, which is currently the worst offender.
In other words, this should be fine, but I'll respin it anyway.

marusak (approved):
Very good. check-journal is now fine, and the most common flakes are present. We are working on them.

PR #1557, opened 2021-01-19
Show top reasons tests failed unexpectedly

marusak:
Adds a new table that looks like this:
unexpectedmessages

These data are currently (and for the next  two weeks won't be usable) due the bug that is being in the second commit.
The second commit I tested as:
```
diff --git store-failures store-failures
index eb88bfe6b..58623ae00 100755
--- store-failures
+++ store-failures
@@ -92,9 +92,7 @@ def main():
 
         with urllib.request.urlopen(raw_log, context=ctx if "logs.cockpit-project.org" in raw_log else None) as fp:
             for (testname, retry_reason) in find_failed_tests(fp):
-                if run_id is None:
-                    run_id = insert_run(cursor, repo, opts.revision, context, url, time)
-                insert_failure(cursor, run_id, testname, retry_reason)
+                print("{0} {1} {2}".format(context, testname, retry_reason))
```
and then running `./store-failures 4113c0923fc98acbbf623b455649735a6fdf5926` and `./store-failures 572864529c9c19fdf9a5627304d63343ff728f4d` both with the new fix and without. One confirms that it does not save known issues and the other one that it still saves unexpected messages on real failures.

The tables, as I said, is unreliable, but there are definitely some messages we should take a look at.

### DIFF
--- a/generate-stats
+++ b/generate-stats
@@ -17,12 +17,14 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
+import re
 import sys
 import time
 import sqlite3
 import argparse
 import logging
 import json
+from collections import Counter
 
 # Seconds in two weeks
 TWO_WEEKS = 1209600
@@ -99,6 +101,34 @@ def main():
 
     # Sort tables
     all_tables = sorted(all_tables, key=lambda x: x["data"][0][1], reverse=True)
+
+    # Select unexpected messages
+    messages = cursor.execute("""\
+            SELECT retry_reason, TestRuns.url
+            FROM Failures
+            JOIN TestRuns ON Failures.run = TestRuns.id
+            WHERE TestRuns.time > ? AND
+                retry_reason IS NOT null AND
+                retry_reason IS NOT 'be robust against unstable tests' AND
+                retry_reason IS NOT 'due to failure of test harness or framework';""", (since,)).fetchall()
+
+    messages = [m for m in messages if m[0] is not None]
+    # Replace any numbers > 100 with 0 (usually time stamps or PIDs)
+    messages = [(re.sub(r'\d\d[\d\.]+', '0', m[0]), m[1]) for m in messages]
+
+    # Replace any partial paths like dracut.xf8hG with dracut.xxx
+    messages = [(re.sub(r'(/[^/ \t]{3,}\.)[^/ \t]+', r'\1xxx', m[0]), m[1]) for m in messages]
+
+    data = []
+    for c in Counter([m[0] for m in messages]).most_common():
+        if (c[1] > 1):  # Don't show message that happened just once
+            logs = [m[1] for m in messages if m[0] == c[0]][:2]
+            data.append([c[0], c[1], logs])
+
+    if data:
+        all_tables.insert(0, {"name": "Unexpected messages",
+                              "columns": ["Message", "Count", "Logs"],
+                              "data": data})
 
     print("const tables = {0};".format(json.dumps(all_tables)))
 
--- a/store-failures
+++ b/store-failures
@@ -138,6 +138,11 @@ def find_failed_tests(fp):
             save_message = False
             continue
 
+        # If the issue is known, don't save unexpected messages
+        if line.startswith("ok ") and "SKIP Known issue" in line:
+            save_message = False
+            last_msg = ""
+
         if "FAIL: Test completed, but found unexpected" in line and "raise" not in line:
             save_message = True
 


store-failures
```diff
@@ -138,6 +138,11 @@ def find_failed_tests(fp):
             save_message = False
             continue
 
+        # If the issue is known, don't save unexpected messages
+        if line.startswith("ok ") and "SKIP Known issue" in line:
```
martinpitt, line 142: Technically the "SKIP Known issue" is redundant IMHO. If we consider the test as passed, then we shouldn't second-guess it here.

generate-stats
```diff
@@ -100,6 +102,34 @@ def main():
...
 
+    # Select unexpected messages
+    messages = cursor.execute("""\
+            SELECT retry_reason, TestRuns.url
+            FROM Failures
+            JOIN TestRuns ON Failures.run = TestRuns.id
+            WHERE TestRuns.time > ? AND
+                retry_reason IS NOT null AND
+                retry_reason IS NOT 'be robust against unstable tests' AND
+                retry_reason IS NOT 'due to failure of test harness or framework';""", (since,)).fetchall()
```
martinpitt, line 113: I'd rather have a positive check here, but this may not be entirely trivial. Since this code is totally fungible, let's start with that.

generate-stats
```diff
@@ -100,6 +102,34 @@ def main():
...
+
+    messages = [m for m in messages if m[0] is not None]
+    # Replace any numbers > 100 with 0 (usually time stamps or PIDs)
+    messages = [(re.sub(r'\d\d[\d\.]+', '0', m[0]), m[1]) for m in messages]
```
martinpitt, line 117: Totally unimportant, but `r\d{3,}` says "at least 3 repetitions" more directly.

generate-stats
```diff
@@ -100,6 +102,34 @@ def main():
...
+
+    # Replace any partial paths like dracut.xf8hG with dracut.xxx
+    messages = [(re.sub(r'\/([^\/]*)\.[^\/]*\/', r'/\1.xxx/', m[0]), m[1]) for m in messages]
```
martinpitt, line 120: No need to escape the slashes. However, this is way too vague -- it matches on non-contiguous paths/multiple words as well, such as `/ (has failed) - Exiting.`. Does it hurt much to drop this for now?

If it creates too much noise, please refine the RE to only match `/[^/ \t]{7,}\.`, i.e. non-trivial absolute paths without spaces.

generate-stats
```diff
@@ -100,6 +102,34 @@ def main():
...
+
+    data = []
+    for c in Counter([m[0] for m in messages]).most_common():
+        if (c[1] > 1):  # Don't show message that happened just once
```
martinpitt, line 124: Good call! I think after we establish some percentages for the flaking tests, we should also not show trivial ones (below the "error budget"), so that the page will turn into a proper TODO list.

martinpitt (changes requested):
Thanks for working on this!

store-failures
marusak: Yeah, I was just not sure, if `ok ` is strong enough patter and it could not appear as part of something else as well.

generate-stats
marusak: omitting it all together creates way too much noise and messages are not group almost at all.

Where does the `7` come from? With this pattern nothing again is matched, as the worst ofender is `dracut.asdfgh`, where neither part is 7 characters long.

I can change it to `/[^/ \t]{3,}\.)[^/ \t]+` which not sure if is better readable really. I want to show it is part of path, thus starts with /, then contains something, then dot and then there is some another part, that we want to replace.

generate-stats
marusak: is decimal point included in `\d`? I don't think it is.

store-failures
martinpitt: Oh, good point - this isn't the last line of the test only, it's "anywhere in the log". So indeed, let's leave it as-is.

martinpitt: Oh, good point - this isn't the last line of the test only, it's "anywhere in the log". So indeed, let's leave it as-is.

generate-stats
martinpitt: Right, my bad.

generate-stats
martinpitt: Right, but `[^/]*` is a terribly wide pattern which can basically be *anything*. Hence let's at least require that it's a single word.

The 7 was fairly arbitrary, to make sure we are not accidentally screwing up very short words, or things which are not paths at all.

marusak: Updated

martinpitt: This still looks a bit sloppy, as it replaces "static/readable" names like /foo.bar wit .xxx as well. But again, this is fungible at any time and not an API, so it's fine. Thanks!

martinpitt (approved):
Dakujem!

PR #1558, opened 2021-01-19
Use more memory by default

marusak:
I am trying to take a look at testAddDiskDirPool which is out top no.1 flake, but cannot reproduce locally, so I'll try some ideas in CI.

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -33,7 +33,7 @@ from .constants import TEST_DIR, BOTS_DIR
 
 sys.path.insert(1, BOTS_DIR)
 
-MEMORY_MB = 1152
+MEMORY_MB = 2048
 
 
 # based on http://stackoverflow.com/a/17753573


marusak:
once [ubunut-2004](https://logs.cockpit-project.org/logs/pull-1558-20210119-191239-e0bef7bb-ubuntu-2004-cockpit-project-cockpit/log.html#100) green, trying again with a few more OSes.

marusak:
So this actually seems to help, which confirms that it is dying due to OOM killer.
Interestingly, mostly just one test is affected, which somehow suggest that this test requires more resources then other tests. Also it happens just on ubuntu* and rhel-8-4*. We may need to limit resource this test is using or maybe split it into two tests if it is too complex. @KKoukiou

marusak:
I'll keep this open for a bit if we need some more experiments, but most likely I'll close it soon-ish.

martinpitt:
testAddDiskDirPool now seems fixed with https://github.com/cockpit-project/cockpit/commit/e170ba73832f33f427fd5bbb35c80cecf9588b53

PR #1567, opened 2021-01-21
testmap: Add manual test for cockpituous

Gundersanne:


task/testmap.py
```diff
@@ -154,6 +154,11 @@
             'centos-8-stream',
         ]
     },
+    'cockpit-project/cockpituous': {
+        '_manual': [
+            'host',
```
martinpitt, line 159: Hmm, don't we need some `@cockpit-project/bots` project selector here?

martinpitt: Ah no, tests-scan pieces that together by itself, doesn't it.

marusak: Depends what you are trying to do.
This means: "You can trigger 'host' context for cockpit-project/cockpituous repository".
That means you can trigger 'host' in cockpit-project/cockpituous or for example in bots PR you can do `host@cockpit-project/cockpituous`. This does not mean you can trigger `host` for bots from cockpitious. You can already do that just by triggering `host@cockpit-project/bots` because `host` is defined for `cockpit-project/bots`.

Gundersanne: oh i just named it host for no reason, I wasn't intending to run that test there atm but we definitely could do it that way, it would probably make sense to do so.

martinpitt: Right, ignore me -- I originally thought you were going to trigger the bots tests, but triggering cockpituous' tests makes sense. We already run its unit tests in a workflow after all, and the tasks containers should be able to run them as well.

marusak:
@Gundersanne Why is it blocked and needswork?

Gundersanne:
@marusak I think i want to name it something else which makes it more obvious what it is, but i'm not sure what yet :p suggestions welcomed

martinpitt:
"Canary"? Although that is a bit cruel, and not vegan :green_heart: "mock-test"? Although I still think it should do something useful and really run `make check`, and thus `unit-tests` would actually be correct.

marusak (approved):

PR #1571, opened 2021-01-22
Enable picking number of days for stats to be shown

marusak:
If you want to see how many failures there has been in the last 3 days you can just simply do:
```
wget --no-check-certificate https://images-frontdoor.apps.ocp.ci.centos.org/test-failures.db
./generate-stats --db test-failures.db --days 3 > failure-stats.js
firefox failure-stats.html
```

### DIFF
--- a/generate-stats
+++ b/generate-stats
@@ -69,6 +69,9 @@ def main():
             links.append(urls[1][0])
         top10.append([row[0], row[1], links])
 
+    if not top10:
+        top10.append(["No failures in the last {0} days".format(opts.days), 0, []])
+
     all_tables.append({"name": "Top failures", "columns": ["Test name", "Count", "Logs"], "data": top10})
 
     # Get failures by OS
--- a/generate-stats
+++ b/generate-stats
@@ -26,13 +26,15 @@ import logging
 import json
 from collections import Counter
 
-# Seconds in two weeks
-TWO_WEEKS = 1209600
+# Seconds in a day
+DAY = 86400
 
 
 def main():
     parser = argparse.ArgumentParser(description='Generate statistics about failed tests')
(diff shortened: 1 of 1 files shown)


generate-stats
```diff
@@ -26,13 +26,14 @@ import logging
...
-# Seconds in two weeks
-TWO_WEEKS = 1209600
+# Seconds in a day
+DAY = 86400
 
 
 def main():
     parser = argparse.ArgumentParser(description='Generate statistics about failed tests')
     parser.add_argument("--db", default="cockpit-failures.db", help="Database name")
+    parser.add_argument("--days", default=14, type=int, help="Number of days to take statistics from. Default is 14.")
```
martinpitt, line 36: Nitpick: `Default is %(default)s.`

generate-stats
```diff
@@ -68,6 +68,9 @@ def main():
             links.append(urls[1][0])
         top10.append([row[0], row[1], links])
 
+    if not top10:
+        top10.append(["No failures in the last {0} days".format(opts.days), 0, []])
```
martinpitt, line 72: This soo needs to add a 🍻 ! :grin:

martinpitt (approved):
Totally unimportant "improve your python", so good enough as-is.

generate-stats
marusak: :smile: If we ever get to that state, lets do it! But now (and for some time) it will mean mostly (You picked too few days and we haven't ran any tests in that timeframe or your DB is too old).

generate-stats
marusak: oh nice! Didn't know about this, updated

martinpitt (approved):
Dakujem!

PR #1578, opened 2021-01-24
testmap: Move lorax rhel8-branch testing to RHEL 8.4

martinpitt:


martinpitt:
@atodorov , @bcl: Seems lorax is not very happy on RHEL 8.4? I strongly recommend to test new RHEL versions as early as possible. I retried already, these failures seem real.

atodorov:
@martinpitt yes, this was an issue with how osbuild-composer starts the services. We didn't discover this until last week.

task/testmap.py
```diff
@@ -88,15 +88,15 @@
...
-            'rhel-8-3/ci',
+            'rhel-8-4',
+            'rhel-8-4/live-iso',
+            'rhel-8-4/qcow2',
+            'rhel-8-4/aws',
+            'rhel-8-4/azure',
+            'rhel-8-4/openstack',
+            'rhel-8-4/vmware',
+            'rhel-8-4/tar',
+            'rhel-8-4/ci',
```
atodorov, line 99: FTR some of these scenarios are already covered on the osbuild-composer side. 

@bcl do we want to keep having the same scenarios executed on the lorax side as well? My guess is they could still fail and we don't have the capacity to keep fixing them.

martinpitt: Sure, retiring obsolete scenarios is of course even better. :grin: But the ones which are still relevant should be fixed soon.

bcl: I think the rhel-8-4 one is probably ok, but I don't think we need any of the image specific tests anymore.

martinpitt: Sure! I honestly don't have any useful opinion here. @atodorov , @bcl, feel free to send a PR of your own or just force-push mine.

atodorov: I can't force push to your repo so opened https://github.com/cockpit-project/bots/pull/1587. Also can't close this PR so please close it.

martinpitt: Thanks!

PR #1586, opened 2021-01-26
testmap: Update cockpit-composer rhel-8 branch tests

henrywang:
rhel-8 branch has been reset to master tag 28, which is the 2nd rhel 8.4 release, test should be update to rhel-8-4 and rhel-8-4/firefox


martinpitt (approved):
Awesome, thank you!

PR #1588, opened 2021-01-26
testmap: Update cockpit-composer test

henrywang:
- Add fedora-33/firefox and remove fedora-32/firefox
- Add rhel-8-4/firefox
- Remove rhel-8-4/firefox from manual

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -124,9 +124,9 @@ REPO_BRANCH_CONTEXT = {
     'osbuild/cockpit-composer': {
         'master': [
             'fedora-32',
-            'fedora-32/firefox',
             'fedora-33',
             'rhel-8-4',
+            'rhel-8-4/firefox',
         ],
         'rhel-8': [
             'rhel-8-4',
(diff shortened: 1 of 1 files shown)


task/testmap.py
```diff
@@ -124,17 +124,17 @@
     'osbuild/cockpit-composer': {
         'master': [
             'fedora-32',
-            'fedora-32/firefox',
             'fedora-33',
+            'fedora-33/firefox',
```
marusak, line 128: This is a new context, so you cannot trigger it in this PR.
It first needs to go into `_manual`, then you can test it and then we can move it to `master`.

henrywang: Make sense to me. Thanks @marusak!

marusak (approved):
Thanks!
Let's wait for `rhel-8-4/firefox` result. (or if you have some run by hand, please post it here)
When this is merged, please send another PR where you move `fedora-33/firefox` to `master` ideally with test result, but if you believe it will just work, then feel free to just run it in the PR itself.

henrywang:
@marusak, Thanks. I'll send another PR to move `fedora-33/firefox` to `master`. Let's land this PR first.

PR #1593, opened 2021-01-28
Fix npm-update for patternfly

martinpitt:
This avoids the repeated mess that we are getting with trying to update individual PF dependencies, like in https://github.com/cockpit-project/cockpit-podman/pull/661 or https://github.com/cockpit-project/cockpit/pull/15184 .

On current cockpit-podman I get:
```
$ bots/npm-update --dry -v
+ @patternfly/react-table@4.20.15
Updated package '@patternfly/react-table' is in group @patternfly
Ignoring 'docker-names' as it does not belong to current update group @patternfly

+ @patternfly/react-core@4.90.2
Ignoring 'xterm' as it does not belong to current update group @patternfly
Ignoring 'react' as it does not belong to current update group @patternfly
Ignoring 'react-dom' as it does not belong to current update group @patternfly
Ignoring 'prop-types' as it does not belong to current update group @patternfly
Ignoring 'moment' as it does not belong to current update group @patternfly
Ignoring 'react-scrollable-anchor' as it does not belong to current update group @patternfly

+ @patternfly/patternfly@4.80.3
Ignoring 'throttle-debounce' as it does not belong to current update group @patternfly

$ git diff
```
```diff
--- package.json
+++ package.json
@@ -48,9 +48,9 @@
     "webpack-cli": "^3.1.0"
   },
   "dependencies": {
-    "@patternfly/patternfly": "4.70.2",
-    "@patternfly/react-core": "4.84.4",
-    "@patternfly/react-table": "4.19.45",
+    "@patternfly/patternfly": "4.80.3",
+    "@patternfly/react-core": "4.90.2",
+    "@patternfly/react-table": "4.20.15",
     "docker-names": "1.1.1",
     "moment": "2.29.1",
     "prop-types": "15.7.2",
```

After committing that and a moment downgrade, I get

```
+ moment@2.29.1
```
and it updates only that.

I ran both on my fork without `--dry`, and you can see the results here:

- https://github.com/martinpitt/cockpit-podman/pull/4
- https://github.com/martinpitt/cockpit-podman/pull/6

I also ran it again when everything is current, and it is a successful no-op as expected.

### DIFF
--- a/npm-update
+++ b/npm-update
@@ -25,7 +25,11 @@
 # Dependencies where minor updates break, only bump microversion
 FRAGILE = [
     "jquery",
-    "patternfly",
+]
+
+# Dependencies which have to be updated in lockstep, not individually
+GROUP = [
+    "@patternfly"
 ]
 
 import collections
@@ -107,8 +111,17 @@ def run(specified_package, verbose=False, **kwargs):
             return "^" + version
         return "~" + version
 
+    group = None
+
     for package in packages:
         if package in pending_updates:
+            continue
+
+        # once we updated the first member of a group, only consider other group members
+        if group and group not in package:
+            if task.verbose:
+                sys.stderr.write("Ignoring '{0}' as it does not belong to current update group {1}\n"
+                                 .format(package, group))
             continue
 
         orig_version = orig_package_json["dependencies"][package]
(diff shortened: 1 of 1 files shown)


npm-update
```diff
@@ -25,7 +25,11 @@
 # Dependencies where minor updates break, only bump microversion
 FRAGILE = [
     "jquery",
-    "patternfly",
```
marusak, line 29: This is PF3, `@patternfly` is PF4. This means, that PF3 is now not fragile an PF4 is updated as a group.

marusak (commented):
Looks nice! Easy enough, thought it would be harder to implement :)
I have one concern about PF3 vs PF4, maybe I have missed something?

npm-update
martinpitt: Right, most projects don't use pf3 any more (all but cockpit itself). If we leave this, it will cause PF4 to only get updated to microversions, which is not what we want IMHO. Indeed this will now attempt to update PF3 to new minorversions also, but I think that's ok? We need to visually validate it anyway, and I doubt that there will be many (if any) new minors in the first place.

martinpitt:
No-change FP to tickle the bots unit-tests workflow.

marusak (approved):
> Indeed this will now attempt to update PF3 to new minorversions also, but I think that's ok? We need to visually validate it anyway, and I doubt that there will be many (if any) new minors in the first place.

Fair enough! Thanks :)

PR #1596, opened 2021-01-28
publish-wrapper: Cleanup unneeded detection collision

marusak:
This was copied over back from when tests were polling and no amqp queue
was used. This therefore never happens (or not unless someone else
publishes status). I grabbed all logs we currently have and this
detection never appeared.


martinpitt (approved):
Thanks for checking the logs!

PR #1598, opened 2021-01-28
machine: fix shell quoting on m.execute()

allisonkarlitskaya:
SSHConnection.execute() has existing support for passing the command as
a list instead of a string.  This seems like it might get a
subprocess-like interface for protecting you from the shell, but in
practice the list items are just passed as separate commandline
arguments to ssh, which will concatenate them and run them as a shell
command anyway.

Fix that by using shlex.join() to properly escape each string, and pass
the result as a single argument to ssh.

This behaviour was previously undocumented, so add it to the docstring.

 * [ ] cockpit-project/cockpit#15239

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -22,6 +22,7 @@ import socket
 import subprocess
 import select
 import errno
+import shlex
 import sys
 import tempfile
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Nice fix, thanks! Good to land on green.

marusak:
> Good to land on green.

fedora-33@cockpit-project/cockpit Tests failed with code 1

martinpitt:
Ugh, yes -- it crept into master now, see https://github.com/cockpit-project/cockpit/pull/15241

martinpitt:
No wait, that *was* the place that 426bd42b31a1fb71af fixed. I'm confused. *This* PR ran with the old version still, but in https://github.com/cockpit-project/cockpit/pull/15241 the *new* code now fails.

PR #1602, opened 2021-01-29
Revert "machine: fix shell quoting on m.execute()"

marusak:
This reverts commit 13eff0d5743b75c3a7ea53ffd87ac081a947285e.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -22,7 +22,6 @@ import socket
 import subprocess
 import select
 import errno
-import shlex
 import sys
 import tempfile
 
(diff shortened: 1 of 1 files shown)


marusak:
@allisonkarlitskaya btw as now master fails with: https://logs.cockpit-project.org/logs/pull-15234-20210129-062252-e208e601-fedora-32/log.html

martinpitt (dismissed):
I suppose we need to fix the check-realms test a little harder, and then reapply this.

martinpitt:
Sent check-realms fix in https://github.com/cockpit-project/cockpit/pull/15243 , it was pretty obvious after all (and also the correct thing to do).

allisonkarlitskaya:
pitti fixed that with cockpit-project/cockpit#15243, just landed

PR #1603, opened 2021-01-29
image-upload: Fix fallback upload to http store

martinpitt:
We never hit this in practice, but it does happen when trying to run
this in a local podman pod deployment (where https SSI works somwhere
between "poorly" and "not at all").

---

Totally unimportant, but now that I saw it my OCD just can't let it slide..


martinpitt:
Nevermind, our nginx config doesn't actually support http only (by design), so this is rather moot.

PR #1604, opened 2021-01-29
image-download: Fix crash with --state

martinpitt:
Commit 2efccf29fab8b23 removed the fourth "latency" item tuple entry
from the "found" list, but forgot to adjust `header_date()` accordingly.
This is only being used with `--state`, which we don't currently use
much.

### DIFF
--- a/image-download
+++ b/image-download
@@ -104,7 +104,7 @@ def find(name, stores, latest, quiet):
 
     # Find the most recent version of this file
     def header_date(args):
-        cmd, output, message, latency = args
+        cmd, output, message = args
         try:
             reply_line, headers_alone = output.split('\n', 1)
             last_modified = email.message_from_file(io.StringIO(headers_alone)).get("Last-Modified", "")


martinpitt:
Not running tests, as they don't use that code path (otherwise we'd have known long ago). This can be reproduced and checked with
```
❱❱❱ ./image-download --state tests-train-1.jsonl.gz
 ⨯ https://images-frontdoor.apps.ocp.ci.centos.org/
 ⨯ https://images-cockpit.apps.ci.centos.org/
 ✔ https://cockpit-11.e2e.bos.redhat.com:8493
 ⨯ https://10.29.163.169:8493/
image-download: not enough values to unpack (expected 4, got 3)
```

martinpitt:
I noticed that when writing integration tests for image-{upload,download} in cockpituous' run-local.sh. So it soon *will* have test coverage.

allisonkarlitskaya (approved):
...or we could remove `--state`?  This would be a really significant simplification.

PR #1613, opened 2021-02-01
Image refresh for rhel-8-3

cockpituous:
Image refresh for rhel-8-3
 * [x] image-refresh rhel-8-3


martinpitt (approved):
Heh, easy one, only sub-man left.

PR #1614, opened 2021-02-01
Remove learn container related scripts

martinpitt:
We have not used them in more than two years, they have been broken for
a while, and the learn container isn't coming back any time soon.


martinpitt:
I cleaned up tests-policy as well. I tested it locally, but running a cockpit test now to validate it.

marusak (approved):
Nice, -1090 lines :)
Thanks!

PR #1616, opened 2021-02-01
Stop using cockpituous forks

martinpitt:
The cockpituous forks of a project need to get set up manually, pile up
hundreds of stale branches over time, make it difficult to test changes
on  forks, and are an unnecessary part now.

starter-kit, cockpit-{podman,ostree,composer} don't use them at all any
more: they use po-refresh and npm-update with the default GitHub token
and push branches straight to origin. From there they get auto-deleted
when landing pull requests, which avoids piling up cruft.

Let's do the same for the remaining two forks for cockpit (npm-update
and po-refresh) and bots (image-refresh and naughty-prune) for
consistency and simplification . In `task.branch()`, drop the
opportunistic detection of a fork, and always push to origin. Eliminate
the now unnecessary `user` and `push_repo` arguments to `push_branch()`.

-----

 * [x] naughty-prune -- nothing to actually do.. but it uses the same code as image-refresh and npm-update.
 * [x] image-refresh ubuntu-2004 : correctly produced [branch on origin](https://github.com/cockpit-project/bots/commits/image-refresh-ubuntu-2004-20210201-184848)
 - [x] I ran `bots/npm-update` on my fork after committing a moment downgrade, and it produced https://github.com/martinpitt/cockpit/pull/8
 - [x] Run bots/po-refresh (accidentally on origin, not my fork), and it produced https://github.com/cockpit-project/cockpit/pull/15257 -- as expected with a branch on origin, not cockpituous.

### DIFF
--- a/po-refresh
+++ b/po-refresh
@@ -172,12 +172,14 @@ def run(context, verbose=False, **kwargs):
     branch = task.branch(context, "po: Update from Fedora Weblate", pathspec="po/",
                          branch=local_branch, **kwargs)
 
+    api = github.GitHub()
+
     if local_branch and not branch:  # We only introduced/dropped language but otherwise no updates
         if kwargs["issue"]:
-            clean = "https://github.com/{0}".format(task.find_our_fork(user))
+            clean = "https://github.com/{0}".format(api.repo)
             task.comment_done(kwargs["issue"], "po-refresh", clean, local_branch, context)
 
-        task.push_branch(user, local_branch)
+        task.push_branch(local_branch)
         branch = "{0}:{1}".format(user, local_branch)
 
     if not branch:
@@ -188,7 +190,6 @@ def run(context, verbose=False, **kwargs):
     pull = task.pull(branch, labels=['bot', 'no-test', 'release-blocker'], run_tests=False, **kwargs)
 
     # Trigger this pull request
-    api = github.GitHub()
     head = pull["head"]["sha"]
 
     triggers = testmap.tests_for_po_refresh(api.repo)
--- a/task/__init__.py
+++ b/task/__init__.py
@@ -370,24 +370,8 @@ def execute(*args):
     return output
 
 
-def find_our_fork(user):
-    repos = api.get("/users/{0}/repos".format(user))
-    for r in repos:
-        if r["full_name"] == api.repo:
-            # We actually own the origin repo, so use that.
-            return api.repo
-        if r["fork"]:
-            full = api.get("/repos/{0}/{1}".format(user, r["name"]))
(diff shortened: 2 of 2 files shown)


cockpituous:
naughty-prune in progress on 4-cockpit-8.
Log: https://logs.cockpit-project.org/logs/naughty-prune-1616-20210201-183030/

cockpituous:
image-refresh in progress on 3-cockpit-8.
Log: https://logs.cockpit-project.org/logs/image-refresh-1616-20210201-183131/

cockpituous:
image-refresh ubuntu-2004 done: https://github.com/cockpit-project/bots/commits/image-refresh-ubuntu-2004-20210201-184848

marusak:
Nice one! I have one question.
We have `user` in `branch()`. This function is used in a few places, but string that it returns (in form `user:branch`) is used only in two ways:
1. Directly passed into `pull()`
2. A few things happen to it in `po-refresh` but it just splits the user or concats it back in. Eventually it just passes it into `pull()` as well

`pull()` also gets rid of the user part. The **only** place where this `user:` part is kept and used is in `pull()` method in `data` as `head`.
From [docs](https://docs.github.com/en/rest/reference/pulls#create-a-pull-request):
```
The name of the branch where your changes are implemented. For cross-repository pull requests in the same network, namespace head with a user like this: username:branch
```
I am not sure, if we actually need the `username` part since we don't do cross-repository pull-request anymore. But if we do, then we might just move `user = api.repo.split('/')[0]` from `branch()` to `pull()` and we can get rid off all of those `split`s.

Do I make sense or have I missed here something? If I do, then separate commit (for easier review) would be nice :)

martinpitt:
@marusak: You do understand this correctly, we *can* do some more serious simplification here. I didn't yet, because I first wanted to double-check if this really works with cockpit and bots once this is on master. And such a simplification needs to happen after this commit, not before. So I thought I'd rather do that as a follow-up. But I can do it right here if you prefer.

marusak (approved):
> You do understand this correctly, we can do some more serious simplification here. I didn't yet, because I first wanted to double-check if this really works with cockpit and bots once this is on master. And such a simplification needs to happen after this commit, not before. So I thought I'd rather do that as a follow-up. But I can do it right here if you prefer.

Sure, follow up works, I just thought that if we won't clean it up now when we have it fresh in our minds, we never will :) But this PR is of course fine as is :) Thanks!

martinpitt:
At least [cockpit po-refresh](https://github.com/cockpit-project/cockpit/pull/15260) worked fine :+1:

martinpitt:
@marusak: user cleanup sent to PR #1618

PR #1617, opened 2021-02-01
Metrics new generation

marusak:
This does bunch of things:
- Storing also tests runs, that were retried
- Stored data about runs now also includes state, which retry it was and how long the run took (and how long it was in a queue)
- Also successful tests are stored in DB
- No pre-generating with python is done, but DB is read directly from JS
- On page showing % and not number of fails
- New design including graphs, date-range and stuff :)

Live demo (I created DB from some random PRs, don't take those data seriously): https://mmarusak.fedorapeople.org/tests.html

I would love to hear feedback! We can show more metrics on the page as followup, this is kind of where I think we can start. If you think of some metric we should save in DB as well, now is the right time to speak. I need to validate all the numbers I show on the page if I haven't done some mistake and cleanup commits. The code is not the cleanest things, I admit it openly, but I would say good for start :) And I want to get at least the storing merged so we can start collecting new data.


run-queue
```diff
@@ -64,7 +64,7 @@ def consume_webhook_queue(channel, amqp):
         elif action == 'closed' and pull_request.get('merged', False):
             sha = pull_request['head']['sha']
             db = os.path.join(get_images_data_dir(), "test-results.db")
-            cmd = "./store-failures --db {0} --repo {1} {2}".format(db, repo, sha)
+            cmd = "./store-tests --db {0} --repo {1} {2}".format(db, repo, sha)
```
martinpitt: This could be squashed into the previous commit -- that is already rather inconsistent as it mixes "-failures" with "tests.db" and such.

tests.html
```diff
@@ -138,13 +139,14 @@ <h2 id="g1-title"></h2>
                         WHERE TestRuns.time > ${since.getTime() / 1000} AND TestRuns.time < ${to.getTime() / 1000}`);
                 labels.unshift(to.getDate() + "." + (to.getMonth() + 1));
                 points.unshift(failed[0].values[0]);
+                limit_points.push(1);
```
martinpitt, line 244: That seems too coarse. Limit should be per measurement, i.e. queue time vs. run time vs. failure rate. Current limits [are here](https://github.com/cockpit-project/cockpit/wiki/DevelopmentPrinciples#service-level-objectives)

tests.html
```diff
@@ -43,6 +43,12 @@ <h2 id="g1-title"></h2>
                     <canvas id="g1" style="width: 100%; height: 256px"></canvas>
                 </div>
             </div>
+            <div class="row">
```
martinpitt, line 46: Commit "Show averge time test waited in queueu to be picked up" .. a little too many typos :)

store-failures
```diff
@@ -55,46 +55,79 @@ def main():
...
-            for (testname, retry_reason) in find_failed_tests(fp):
-                if run_id is None:
-                    run_id = insert_run(cursor, repo, opts.revision, context, url, time)
-                insert_failure(cursor, run_id, testname, retry_reason)
+        if context not in contexts:
+            contexts[context] = []
+
+        if status["state"] == "pending":
+            if "Testing in progress" in status["description"]:
+                contexts[context][-1]["url"] = status["target_url"]
```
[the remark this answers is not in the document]
marusak: Right, that would mean there would not be never 'Not yet tested' status, only 'Testing in progress'. If that would happen, we would have a few more problems (like missing 'posted'). So I think it is fine if we crash on it, so we know something is very wrong with this logic we use (as then we would possibly be overwritting old statuses anyway).

store-failures
```diff
@@ -55,46 +55,79 @@ def main():
...
+            contexts[context][-1]["state"] = status["state"]
+            if status["state"] == "error":
+                contexts[context][-1]["description"] = status["description"]
+        else:
+            logging.warning("Status has unexpected state '{0}', skipping.".format(status["state"]))
+
+    for context, statuses in contexts.items():
+        for retry, status in enumerate(statuses):
+            if not status.get("finished"):
+                # TODO: Store partial, store as non-finished?
```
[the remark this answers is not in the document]
marusak: Indeed. This would point to those 'Never picked up, we need to requeue it manually' cases, which surely contribute to us exceeding error budget by needing to manually intervene. It goes towards "All PRs finish up in 90 minutes" (or how we defined it).
But I want to do this as follow up, as it has a bit different logic then rest of finished tests.

task/github.py
```diff
@@ -313,6 +313,18 @@ def statuses(self, revision):
                 count = len(data["statuses"])
         return result
 
+    def allStatuses(self, revision):
```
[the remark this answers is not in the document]
marusak: > So the commits/SHA/status API only shows current ones, while statuses/SHA shows the entire history?

correct

> /statuses is documented as a POST-only API; but it's what statuses_url points to, so I guess it's legit.

ha, good point! I indeed got it from `statuses_url` at first and then just checked it existed in docs, completely missing on the fact it is post. But seems to work :)

store-failures
```diff
@@ -148,9 +163,10 @@ def init_db(cursor):
                      description TEXT,
                      UNIQUE (url))
                    """)
-    cursor.execute("""CREATE TABLE if not exists Failures
+    cursor.execute("""CREATE TABLE if not exists Tests
                     (testname TEXT NOT NULL,
                      retry_reason TEXT,
+                     failed INTEGER,
```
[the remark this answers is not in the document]
marusak: https://sqlite.org/datatype3.html
There is no boolean type.

store-failures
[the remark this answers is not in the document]
martinpitt: Correct. I was mostly worried about GitHub pruning old statuses. But let's start with that, I'd rather want to know when it happens than being overly defensive here.

tests.html
marusak: This is branch for failure rate, so the limit is for this metric. I know it is a bit stupid to do it like this, but I would say good enough for now :) (I am bumping it to 10% though according to what we picked)

store-failures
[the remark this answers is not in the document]
martinpitt: SQLite doesn't have it (or rather, maps it to int), see "3.1.1. Affinity Name Examples"; but many other databases do.

tests.html
martinpitt: Ah, ok -- the gray line in "Average time waiting in queue" is also only a few seconds (possibly 1?). I think setting that to something like 15 minutes seems reasonable.

marusak: yeah, I think I put there 1s, which was just random number.
15 minutes seems to long to me, especially since we want `Test results are available within 60 minutes in 95% of runs`. Maybe then 5 minutes? Or I'll just ditch this metric and rather have how long it took, not how long it waited (but waited indicates length of queue and availability of CI)

martinpitt: I think *recording* both the queue and test run length is important -- these are both useful SLIs for figuring out where the problems are. I'm less sure about the graph -- should it just show the queue+run sum, or two graphs with the individual metrics. Is it possible to draw more than one curve into the graph, and use different scales? (seconds for queue time, minutes for run time)?

But I suppose we also don't need to overthink this. We want to know about the outliers which spend > 5 mins in the queue or > 1 hour in the test, and show the precise log and time stamp for investigation. So maybe a table is good enough for that as well.

But all of that is fungible, let's not block the data collection and initial PR on that. I'd say, for this PR just set it to 5 mins, yes.

store-failures
```diff
@@ -52,46 +54,86 @@ def main():
...
-        run_id = None
+    contexts = {}
+
+    # Read the "status story"
+    # First we get 'Not yet tested' or 'Not yet tested (direct trigger)'
+    # Then when it is picked up we get "Testing in progress" and finally it is either red or green
+    # Putting it all together can give us data like number of retries, how long it waited to get
+    # testes, how long testing took...
+    # Create map from 'context' to array of 'testrun's where each is a map containing: (order in the
+    # array indicates retries, first is original run, first is the first retry...)
```
martinpitt: The first first and the second first one? :grin

run-queue
```diff
@@ -60,12 +60,12 @@ def consume_webhook_queue(channel, amqp):
...
             cmd = ['./tests-scan', '--pull-data', json.dumps(request), '--amqp', amqp, '--repo', repo]
-        # When PR was merged, generate task for storing failures
+        # When PR was merged, generate task for storing tests
         elif action == 'closed' and pull_request.get('merged', False):
             sha = pull_request['head']['sha']
-            db = os.path.join(get_images_data_dir(), "test-failures.db")
+            db = os.path.join(get_images_data_dir(), "test-results.db")
             data = os.path.join(get_images_data_dir(), "failure-stats.js")
-            cmd = "./store-failures --db {0} --repo {1} {2} && ./generate-stats --db {0} > {3}".format(
+            cmd = "./store-tests --db {0} --repo {1} {2} && ./generate-stats --db {0} > {3}".format(
```
martinpitt: Where will the html actually be copied? I expect that we'll still change this quite a bit, so having to copy it manually is a bit cumbersome.

martinpitt (approved):
Thanks! Almost there, just one question.

run-queue
marusak: For now it is manually copied. Once that becomes burden, we can surely do it somewhere in bots.

martinpitt (approved):

PR #1618, opened 2021-02-02
task: Drop user from branch()/pull()

martinpitt:
Commit 1b5a07419eb23e7 dropped using forks for our bots. After that, we
can simplify task.branch() to not parse and return the user name any
more, as it will always be the origin.

Most bots just directly pass the return value to task.pull(), adjust
that to not expect an "username:" prefix any more. po-refresh does some
more processing of the result, adjust that accordingly.

### DIFF
--- a/po-refresh
+++ b/po-refresh
@@ -69,7 +69,6 @@ def run(context, verbose=False, **kwargs):
     subprocess.check_call(cmd, cwd=work)
 
     local_branch = None
-    user = None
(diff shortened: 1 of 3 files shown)


martinpitt:
I tested this on my fork, and po-refresh [produced a correct PR](https://github.com/martinpitt/cockpit/pull/9).

marusak (approved):
Great, thanks! :)

PR #1619, opened 2021-02-02
tests-scan: Use direct test trigger without --amqp

martinpitt:
The primary way of calling tests-scan is through run-queue: This will
supply an --amqp server and queue the tests by itself. This uses status
`github.NOT_TESTED`, which run-queue will ignore (to avoid
triggering/queueing them again).

However, we also occasionally call tests-scan by hand, for examaple when
removing the no-test label from a PR. That does not really work, as the
tests will then never get into the queue. So far the workaround was to
run an additional `tests-trigger --force` on the PR. But we don't really
need this extra roundtrip -- when calling tests-scan without `--amqp`,
we *always* need something that flips the NO_TESTED to
NOT_TESTED_DIRECT, so tests-scan may just as well do that on its own.

This will also allow us to move test triggering from a webhook to a
workflow, as we plan to do for cockpit soon.

### DIFF
--- a/tests-scan
+++ b/tests-scan
@@ -200,7 +200,7 @@ def queue_test(priority, name, number, revision, ref, context, base,
         logging.info("Published '{0}' on '{1}' with command: '{2}'".format(name, revision, command))
 
 
-def prioritize(status, title, labels, priority, context, number):
+def prioritize(status, title, labels, priority, context, number, direct):
     state = status.get("state", None)
     update = {"state": "pending"}
 
@@ -247,7 +247,7 @@ def prioritize(status, title, labels, priority, context, number):
             logging.info("Not updating status for '{0}' on #{1} because of low priority".format(context, number))
             update = None
         else:
-            update["description"] = github.NOT_TESTED
+            update["description"] = github.NOT_TESTED_DIRECT if direct else github.NOT_TESTED
 
     return [priority, update]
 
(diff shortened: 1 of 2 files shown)


martinpitt:
I tested that with `bots/tests-scan -p 9` on my fork, and it [triggered the right statuses](https://github.com/martinpitt/cockpit/pull/9). Of course I don't have webhooks on my fork, so it didn't actually *run* any tests.

@Gundersanne , I'd appreciate if you could review this -- I think you came up with the _DIRECT thing, and I don't fully trust my reasoning here. Dank je wel!

tests-scan
```diff
@@ -200,7 +200,7 @@ def queue_test(priority, name, number, revision, ref, context, base,
         logging.info("Published '{0}' on '{1}' with command: '{2}'".format(name, revision, command))
 
 
-def prioritize(status, title, labels, priority, context, number):
+def prioritize(status, title, labels, priority, context, number, direct):
```
Gundersanne, line 203: So I guess the by now ancient polling method for `bots/tests-scan` would also be considered direct.

Gundersanne (approved):
Hm I'm a bit confused as to how this will help in future? As tests-scan was able to post statuses before.

martinpitt:
>  As tests-scan was able to post statuses before.

Yes, but they were "dead", as the webhook ignored them. You have to "enable" them by running something like `bots/tests-scan -f -p 1234`.

Gundersanne:
@martinpitt I see, I think i'm good with merging this.

PR #1620, opened 2021-02-02
weather: Don't fail if there is no unexpected message

marusak:


martinpitt (approved):
Nicer weather! :sun_behind_small_cloud:

PR #1621, opened 2021-02-02
weather: Show graphs from today

marusak:
Logic was to start from today midnight and go back, but we can actually
start from right now and go back.


martinpitt (approved):
Danke! Tested locally, works fine.

PR #1622, opened 2021-02-02
testmap: Add manual unit test for bots

martinpitt:
Similar to commit 1c9fcaec954, we soon want to introduce self-validation
for bots PRs with cockpituous' run-local.sh. Declare a new manual
"unit-tests-local" test so that tests-scan has something to chew on.

 - [ ] Requires https://github.com/cockpit-project/cockpituous/pull/376

### DIFF
--- a/naughty/debian-testing/1635-api-commit-HOME
+++ b/naughty/debian-testing/1635-api-commit-HOME
@@ -1,0 +1,1 @@
+> error: Failed to commit container *: CommitFailure: error resolving name "*": could not get either XDG_CONFIG_HOME or HOME: Internal Server Error


martinpitt:
I'll put the self-validating workflow into this PR -- after all, that's the whole point, testing such changes should *not* require a new rollout.

martinpitt:
Heck, that is tricky -- seems our real bots keep jumping on the "unit-tests" thing before my local deployment can:
```
DEBUG:root:event: status, path: /fake/issues
DEBUG:root:repository: cockpit-project/bots
INFO:root:Handling status event
INFO:root:repo: cockpit-project/bots; sha: 5a2b2a197e2b9158535e622e3731b55be1149c67; state: pending; description: Not yet tested (direct trigger)
```
PR #1619 was supposed to do that, so it shouldn't happen here yet. I just saw that run-local.sh calls `tests-trigger`, which is unnatural -- the "real" bots don't do that, they call tests-scan only. That ought to suffice, I'll test that separately.

PR #1623, opened 2021-02-03
testmap: Add testmap injection

martinpitt:
This is useful for cockpituous' self-tests so that they can use the same
tests-scan workflow as production [1], without having to play extra tricks
with tests-scan. That approach would not work at all for writing similar
integration self-tests for bots, as doing tests-trigger races with the
*real* bots (the bots repo has an issues webhook trigger, unlike
cockpituous).

This replaces the `_manual` trigger for cockpituous.

[1] https://github.com/cockpit-project/cockpituous/pull/376

 -  [x] get it to work with https://github.com/cockpit-project/cockpituous/pull/376

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -139,11 +139,6 @@ REPO_BRANCH_CONTEXT = {
             'centos-8-stream',
         ]
     },
-    'cockpit-project/cockpituous': {
-        '_manual': [
-            'unit-tests',
-        ]
-    }
 }
 
 # The OSTree variants can't build their own packages, so we build in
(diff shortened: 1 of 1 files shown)


martinpitt:
Unblocked. This works fine with https://github.com/cockpit-project/cockpituous/pull/376 , I just wasted 1.5 hours on an unrelated issue, which turned out to be a very delicate timing issue :exploding_head: (https://github.com/cockpit-project/cockpituous/pull/378)

marusak (approved):
It edits context for each project I was somehow expecting it to inject it to pre-defined project. But since this is really just for testing, look good to me :)
Thanks!

martinpitt:
@marusak: I actually started with statically defining a bots/unit-tests, in PR #1622. But that *completely* falls on its nose. This only works for cockpituous because that doesn't have any webhooks, but it is fundamentally incompatible with a project that does. Hence all my wrangling this morning with https://github.com/cockpit-project/cockpituous/pull/376 and others :grin: 

I don't land this yet, as landing that will break cockpituous PRs until https://github.com/cockpit-project/cockpituous/pull/376 lands.

PR #1624, opened 2021-02-03
weather: Show PR health

marusak:
See demo https://mmarusak.fedorapeople.org/tests.html

It shows sadly how bad we are at it now :/ We are nowhere near our not-so-high goal. (Of course take those data with pinch of salt, but still...)


martinpitt (approved):
Thanks! Also, it's hopefully not *that* bad -- we have very few data points right now, and a single highly unstable test can make the difference between "all green" and "always red".

PR #1626, opened 2021-02-03
weather: Parse number of days as integer

marusak:
JS is amazing with dealing with this kind of situations:
```
> "5" +1
"51"

> "5" * 10
50
```

This manifested itself on the page - I was using `days` with just
multiplication and ti worked fine, but there is now also addition and it
does not behave as expected. Reproducible by changing number of days and
then limit lines in graphs disappear.

### DIFF
--- a/tests.html
+++ b/tests.html
@@ -55,7 +55,7 @@
   <script>
     const DAY = 86400;
     const params = new URLSearchParams(window.location.search);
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Ugh, what a head → desk :boom:  Thanks!

PR #1633, opened 2021-02-04
image-customize: Add --no-network option

martinpitt:
Similar to vm-run's option. This allows projects to enforce that test
image setup remains reproducible and not dependent on online package
installation.

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -44,6 +44,7 @@ parser.add_argument('-u', '--upload', action='append', dest="uploadlist", defaul
                     help='Upload file/dir to destination file/dir separated by ":" example: -u file.txt:/var/lib')
 parser.add_argument('--base-image', help='Base image name, if "image" does not match a standard Cockpit VM image name')
 parser.add_argument('--resize', help="Resize the image. Size in bytes with using K, M, or G suffix.")
+parser.add_argument('-n', '--no-network', action='store_true', help='Do not connect the machine to the Internet')
 parser.add_argument('image', help='The image to use (destination name when using --base-image)')
 args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


martinpitt:
We don't use this anywhere just yet -- I was going to use it in podman and ostree, but unfortunately both still pull some container images, so do need networking still. But we could split this and run the rpm package installation without internet, to at least make sure that stays reproducible.

Test with
```
❱❱❱ bots/image-customize -r 'curl https://google.com' --no-network -v fedora-33
ab8e8b0e-0449-4f27-8e44-e7dca0886383
+ curl https://google.com
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (6) Could not resolve host: google.com
Command 'curl https://google.com' returned non-zero exit status 6.
```

The normal mode still works:
```
❱❱❱ bots/image-customize -r 'curl https://google.com' -v fedora-33
f654d7df-603c-469d-9470-1a7087988092
+ curl https://google.com
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   220  100   220    0     0   1301      0 <HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="https://www.google.com/">here</A>.
...
```

marusak (approved):
Thanks!
We would indeed need to split podman prepare script for this.

Issue #1625, opened 2021-02-03
weather: PR health might hve misleading data when retried in differen days


marusak:
If there is PR where tests run on Monday and on Tuesday we retry it, it would be represented in graph as:
- For Monday it would be counted as red PR as on Monday statuses were green + some reds without retry
- For Tuesday it would be counted as grenn PR with correct number of retries.

This can of course be taken to extremes, where you would retry in more days of a week and from one PR would get even more fake PRs.

Two options come to mind:
- Store also into DB when it was merged (using `time.now()` when `./store-tests` is being run). Not ideal, but easiest
- Do sub-query to find SHAs that have the newest time in the given range and then only consider those SHAs for the given day. A bit more complicated, but we don't need to store anything else, no need to update db...

PR #1634, opened 2021-02-04
weather: Show real counts + dont render points for empty values

marusak:
Live demo https://mmarusak.fedorapeople.org/tests.html

A few changes:
- On top of the page, there is overview with how many data we are working (I want to have above that general status)
- Labels in graphs show from how many data they are being counted
- In graphs line is not drawn when there were no data in that given day

Options?

### DIFF
--- a/tests.html
+++ b/tests.html
@@ -11,6 +11,10 @@
                 <h4 class="my-0 mr-md-auto font-weight-normal">CI Weather Report</h4>
                 <small>Brought to you by <b>MM</b></small>
             </div>
+            <div>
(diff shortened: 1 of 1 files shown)


marusak:
Before someone points out that "from 13 PRs" but if you count it in the PR health graph it is 14 - https://github.com/cockpit-project/bots/issues/1625

martinpitt (approved):
This is great! Also, my gut feeling keeps underestimating how many tests we run. :grinning:

marusak:
> Also, my gut feeling keeps underestimating how many tests we run

Exactly, I always also think in few hundreds tests and in two (normal) days we ran around 40k :)

PR #1638, opened 2021-02-04
workflows: Run local cockpituous deployment against proposed branch

martinpitt:
This will cover changes to the part that always run from master in
production. Like run-queue, make-checkout, or publish-wrapper.

The `push` event test does not actually do much except for setting up
the scaffolding; but we may expand cockpituous' run-local.sh to do
something with bots other than just run-queue checking an empty queue;
so keep this running for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,3 +12,56 @@ jobs:
 
       - name: Run test
         run: test/run
+
+  cockpituous:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          # need this to get origin/master for git diff
+          fetch-depth: 0
+
+      - name: Rebase to current master
+        run: |
+          git config user.name github-actions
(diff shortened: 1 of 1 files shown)


martinpitt:
Oh, I see -- the run [fails](https://github.com/cockpit-project/bots/pull/1638/checks?check_run_id=1829998811) because the default github token is not powerful enough:
```
+ ./tests-scan -p 1638 --amqp localhost:5671 --repo cockpit-project/bots
ERROR:root:tests-scan: Error accessing /repos/cockpit-project/bots/pulls/1638
  Status: 401
  Reason: Unauthorized
  Response: {"message":"Bad credentials","documentation_url":"https://docs.github.com/rest"}
```

So we should make the workflow use the cockpituous token (just like cockpituous PRs), and just do the normal "no PR" test when the PR comes from a fork.

martinpitt:
This [worked as expected](https://github.com/cockpit-project/bots/pull/1638/checks?check_run_id=1830948253). I'll re-submit this as a PR from origin now, to be able to use the token.

PR #1639, opened 2021-02-04
workflows: Run local cockpituous deployment against proposed branch

martinpitt:
This will cover changes to the part that always run from master in
production. Like run-queue, make-checkout, or publish-wrapper.

The `push` event test does not actually do much except for setting up
the scaffolding; but we may expand cockpituous' run-local.sh to do
something with bots other than just run-queue checking an empty queue;
so keep this running for now.

 - [x] Fix run-tests.sh for non-cockpituous tests: https://github.com/cockpit-project/cockpituous/pull/382

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,3 +12,56 @@ jobs:
 
       - name: Run test
         run: test/run
+
+  cockpituous:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          # need this to get origin/master for git diff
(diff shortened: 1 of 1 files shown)


martinpitt:
Hah, [bazinga](https://github.com/cockpit-project/bots/pull/1639/checks?check_run_id=1853734332)! I sent https://github.com/cockpit-project/cockpituous/pull/382 to fix this, once this lands I can clean this up.

martinpitt:
@Gundersanne , @marusak : Friendly review ping? *bats eyelashes* Thanks!

marusak (approved):
Nice one, thanks!

PR #1640, opened 2021-02-04
weather: Shorter names, repo selection and fixes

marusak:
As always, preview: https://mmarusak.fedorapeople.org/tests.html

I have fixed mine DB in ^. DB we have in production is a bit broken. I can provide fixed DB and we can put that into images store.

### DIFF
--- a/tests.html
+++ b/tests.html
@@ -57,6 +57,11 @@
         </div>
   </body>
   <script>
+    // Pretty format test names
+    function normalize(name) {
+        return name.slice(name.lastIndexOf("/") + 1);
+    }
+
     const DAY = 86400;
     const params = new URLSearchParams(window.location.search);
     const days = Number(params.get('days') || 14);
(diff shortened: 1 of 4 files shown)


tests.html
```diff
@@ -245,7 +250,7 @@ <h2 id="g2-title"></h2>
                         ORDER BY run DESC
                         LIMIT 2`);
                     const links = urls[0].values.map(u => u[0]);
-                    top10.push([{name: row[0], key: "test"}, row[1].toFixed(2), links]);
+                    top10.push([{name: normalize(row[0]), key_name: row[0], key: "test"}, row[1].toFixed(2), links]);
```
martinpitt: Any chance we can add the project here? check-application is rather vague, our current top failures are from cockpit-certificates. The new selector surely helps with that, though!

martinpitt (approved):
Whee, every round this becomes so much better, thank you! Let's get this in ASAP to not further collect the wrongly unified contexts.

tests.html
marusak: Good idea! Pushed additional commit. (And deployed it on my page as well for reviewing).

martinpitt (approved):
Cool, thank you!

PR #1642, opened 2021-02-05
machine: Avoid shlex.join()

martinpitt:
Commit 13eff0d5743b introduced that to properly handle argvs. However,
`shlex.join()` only exists since Python 3.8, and thus not yet in
RHEL/CentOS 8.

Replace with the official definition of the join function.
`shlex.quote()` has existed since Python 3.3, and thus is safe to use.

 - [x] unbreak cockpit master tests: https://github.com/cockpit-project/cockpit/pull/15285

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -293,7 +293,8 @@ class SSHConnection(object):
                 if not quiet:
                     self.message("+", command)
             else:
-                cmd.append(shlex.join(command))
+                # use shlex.join() once Python 3.8 is available everywhere
+                cmd.append(' '.join(shlex.quote(arg) for arg in command))
                 if not quiet:
                     self.message("+", *command)
         else:


marusak (approved):
Thanks!

martinpitt:
@KKoukiou , @marusak : Ugh, [this is a most curious failure](https://logs.cockpit-project.org/logs/pull-1642-20210205-062142-208b46ce-fedora-33-cockpit-project-cockpit/log.html#19-2):
```
AssertionError: 'div id="firewall"' not found in '<!DOCTYPE html>[...] <div class="ct-page-fill" id="firewall">...'
```

It's clear *how* to fix this, but *why* did that change now? We don't use `ct-page-fill` anywhere in our code, and yet the "ct-" suggests that this is ours. The HTML doesn't have it for sure:

    pkg/networkmanager/firewall.html:      <div id="firewall"></div>

Does that ring any bell by chance?

marusak:
https://github.com/cockpit-project/cockpit/pull/15277/files#diff-362bc2c6d348c48e7f398b90d6064e4fabeab7a6c23bf60cda7c114ec844be66

ah damn, it ran with no-test label :/ We always need to run at least one test

marusak:
Sent https://github.com/cockpit-project/cockpit/pull/15285

PR #1645, opened 2021-02-07
Image refresh for rhel-8-4

cockpituous:
Image refresh for rhel-8-4
 * [x] image-refresh rhel-8-4


marusak:
composer fails with unexpected message:
```
audit: type=1400 audit(1612680326.559:4): avc:  denied  { unlink } for  pid=1 comm="systemd" name="machine-id" dev="vda3" ino=551464 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:tmp_t:s0 tclass=lnk_file permissive=0
```

podman is sad as well. Let's retry it after https://github.com/cockpit-project/cockpit-podman/pull/669 and then address it.

marusak:
```
Error: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: rootfs_linux.go:59: mounting "sysfs" to rootfs at "/sys" caused: operation not permitted: OCI permission denied
```
podman failing. Needs to be looked at

marusak:
https://github.com/cockpit-project/bots/commit/cfd677e542e152723759bef4adc2be74d976c29b
hmm, so for the podman problem we had naughty before, but we already closed it. So either 1. We haven't had enough 8-4 runs in 4 weeks (possibly broken refresh + no podman PRs) or it is regression in podman.

KKoukiou:
Oh this won't work - the error happens in an m.execute, and seems to be ignored by the naughty pattern.

marusak:
> the error happens in an m.execute, and seems to be ignored by the naughty pattern.

If we see it in logs, then so does tests-policy. Maybe it changed slightly?

KKoukiou:
> > the error happens in an m.execute, and seems to be ignored by the naughty pattern.
> 
> If we see it in logs, then so does tests-policy. Maybe it changed slightly?

It did, and I adjusted the naughty therefore from the past one, but now it's really exact match, I can CTRL+f the string in the logs.

KKoukiou:
The machines failure is real, PXE boot in the tests stopped working.
Screen Shot 2021-02-10 at 18 25 51

martinpitt (approved):
Ugh, so many regressions :crying_cat_face: Thanks for tracking and reporting these! Approving pending green tests.

marusak:
> Ugh, so many regressions

Its a bit ironic, that the closer 8.4 release is getting, the more regressions are there

PR #1649, opened 2021-02-07
images: Drop ubuntu-stable hardcoded version

martinpitt:
Ubuntu 20.10 ("groovy") has been stable for several months, so drop the
hardcoding. Once 21.04 releases, ubuntu-stable will then once again move
to it automatically, as intended.


martinpitt:
This is a no-op right now, so not rebuilding image.

marusak (approved):

PR #1652, opened 2021-02-08
Weather fixes and improvements (once again)

marusak:
Preview on https://mmarusak.fedorapeople.org/tests.html
I am using the same DB that https://images-frontdoor.apps.ocp.ci.centos.org/tests.html uses, without that fix in `store-tests`.

Not super sure about the new graph, ideas are welcomed :)

### DIFF
--- a/tests.html
+++ b/tests.html
@@ -53,6 +53,12 @@
                     <canvas id="g2" style="width: 100%; height: 256px"></canvas>
                 </div>
             </div>
+            <div class="row">
+                <div class="col-md-12">
+                    <h2 id="g3-title"></h2>
+                    <canvas id="g3" style="width: 100%; height: 256px"></canvas>
+                </div>
+            </div>
         </div>
   </body>
   <script>
@@ -288,6 +294,7 @@
                 createTable("t1", "Top failures", ["Test name", "Fail rate (%)", "Logs"], top10);
 
                 // Show average waiting time in queue per day
+                const time_points = [];
                 const wait_points = [];
                 const wait_labels = [];
                 const pr_labels = [];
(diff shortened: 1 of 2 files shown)


martinpitt (dismissed):
Nice improvements, thanks! The gaps didn't actually bother me much, but indeed that looks a bit nicer now. Interesting that we sometimes exceed our 60 minute budget already -- it would be good to somehow get a link to the logs which did, so that we can investigate. (It may just because that was a PR which triggered a lot of retries?)

marusak:
> it would be good to somehow get a link to the logs which did, so that we can investigate. (It may just because that was a PR which triggered a lot of retries?)

Right, do you have some idea how to show this on the page? Interestingly, 8 out of 12 centos-8-stream runs exceeded 60 minutes. Then a few rhel-8-4 and a few F-33s.

martinpitt:
@marusak : I don't think we should cram it into the graph. For my sake, a simple `<p><a href="...">log</a> <a>...</a></p>` would be fine (to minimize space usage).  Bonus if the "log" link name becomes a date+time+context

marusak:
I've added it, you can see it on my page.

tests.html
```diff
@@ -73,6 +74,10 @@ <h2 id="g3-title"></h2>
         return name;
     }
 
+    function formateDate(date) {
```
martinpitt, line 77: It's "formatDate"

martinpitt (approved):
It would make Garrett's eyes pop out :grin: , but great to have this info now, thanks! Something is off somewhere, though -- the [log](https://logs.cockpit-project.org/logs/pull-15216-20210202-165243-5bd2c4b1-centos-8-stream/log) says "3039s on 1-cockpit-10, 151 serial tests: 0: 2448s, 1: 2720s" which is well below one hour. So this was either a long queue time, or something takes a long time to update statuses?

marusak:
> It would make Garrett's eyes pop out

:smile: 

> Something is off somewhere, though -- the log says "3039s on 1-cockpit-10, 151 serial tests: 0: 2448s, 1: 2720s" which is well below one hour. So this was either a long queue time, or something takes a long time to update statuses?

Indeed. This number is from trigger to final status being posted.
It waited 75s in a queue and took 3566 seconds to post the final status.
Do those `3039` seconds include build? But also I very often see that some test has finished, but it takes minutes until it is also posted to github. I guess it takes longer time to copy all logs, screenshots...

tests.html
marusak: Ah thanks, fixed

martinpitt:
> Do those 3039 seconds include build? 

No, they don't, that's exactly it. More hooray to "let's not rebuild the webpack umpteen times" :grin:

martinpitt (approved):

PR #1658, opened 2021-02-09
images: Fix debian.install to get along with tar.xz dist tarballs

martinpitt:
https://github.com/cockpit-project/cockpit/pull/15302 switches
cockpit's `image-prepare` to produce tar.xz tarballs instead of tar.gz.
Adjust debian.install to get along with both formats until that lands.


martinpitt:
I did a debian-stable run against this in https://github.com/cockpit-project/cockpit/pull/15302 to demonstrate that it also works with .tar.xz now.

marusak (approved):
Thanks!

PR #1660, opened 2021-02-10
Fix npm-update for grouped modules

KKoukiou:
@patternfly is a group and all modules inside that group should be
updated within the same PR

There was an issue with the following scenario:
    "@patternfly/patternfly": "4.80.3" <- this does not have an update
    "@patternfly/react-console": "4.2.18" <- this can update to 4.2.19

In the first iteration "@patternfly/patternfly" module is checked.
The `package.json` gets updated to contain @patternfly/patternfly: "^4.80.3",
the script does not find update and moves to the second iteration.

In the second iteration @patternfly/react-console package is checked.
The package.json gets updated to @patternfly/react-console: "^4.2.18",
the script *does* find the update, so the new package.json has
@patternfly/react-console: "4.2.19".
Now the `group` is updated to "@patternfly" and the `orig_package_json` get's
set to the updated `package.json`.

The updated `package.json` however contains the @patternfly/patternfly: "^4.80.3"
from the first iteration, which we don't want to include in the update.

This commit fixes this scenario, by making sure that package.json does
not ever contain "^~" symbols.

### DIFF
--- a/npm-update
+++ b/npm-update
@@ -159,6 +159,9 @@ def run(specified_package, verbose=False, **kwargs):
             else:
                 # only update one dep, so that updates can be tested one by one
                 break
+        else:
+            # if the version was not updated revert package.json changes
+            package_json(orig_package_json, package, orig_version)
 
     if updated_packages and not kwargs["dry"]:
         # Create a pull request from these changes


marusak:
Thanks! Have you tried to run this in you fork and you have result for it?

marusak:
` test / cockpituous (pull_request) ` failed twice now. This does not seem correct, But I think this PR is unrelated. @martinpitt do you have an idea?

KKoukiou:
@marusak I tested it locally and it worked as expected for this example.

martinpitt:
@marusak: Sorry, I guess I broke PRs from forks, I'll have a look. Unrelated to this change for sure.

martinpitt:
Can reproduce localaly with

    COCKPIT_BOTS_REPO=https://github.com/KKoukiou/bots COCKPIT_BOTS_BRANCH=npm-update-fix tasks/run-local.sh

in current cockpituous master.

martinpitt (approved):
I can reproduce the bug locally with `bots/npm-update --dry -v` in cockpit:
```diff
[...]
-    "@patternfly/react-table": "4.20.15",
+    "@patternfly/react-table": "^4.20.15",
```

The change makes sense, and above command now works properly. Cheers!

I'll still wait until the [cockpituous test bug fix](https://github.com/cockpit-project/cockpituous/pull/385) lands, before landing this PR (no red PRs, right?) Should be < 1 hour now.

PR #1661, opened 2021-02-10
DONTMERGE debug cockpituous test failure

martinpitt:
Test [fails repeatedly](https://github.com/cockpit-project/bots/pull/1660/checks?check_run_id=1871579295) with

    curl: (35) error:0407008A:rsa routines:RSA_padding_check_PKCS1_type_1:invalid padding

in the image-upload test, even though the explicit curl succeeded. I
have serious trouble reproducing this locally, debugging on the gh
machine.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,6 +44,8 @@ jobs:
           repository: cockpit-project/cockpituous
           path: cockpituous
 
+      - uses: mxschmitt/action-tmate@v3
+
       - name: Test local CI deployment
         if: steps.changes.outputs.changed
         run: |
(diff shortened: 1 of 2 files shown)


martinpitt:
OK, got it -- it's because this is not enough:

        until [ -f bots/tests-trigger ]; do echo "waiting for tasks to initialize"; sleep 5; done

KKoukiou/bots master branch is ancient, and cockpit-tasks always clones master. I need to fix `update_bots` to reliably clone the right branch.

martinpitt:
Fix sent in https://github.com/cockpit-project/cockpituous/pull/385

PR #1665, opened 2021-02-11
make-checkout: Fix cleanup failure on non-writable subdirectories

martinpitt:
If `$TARGET_DIR` contains any read-only directories, then `rm -rf` will
silently ignore them, but eventually fail with

    rm: cannot remove 'make-checkout-workdir': Directory not empty

Avoid that by ensuring all files are writable first.

----

Spotted [here](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-15315-20210211-101034-fce8fdf9-debian-stable/log.html)

### DIFF
--- a/make-checkout
+++ b/make-checkout
@@ -30,7 +30,11 @@ REF="$1"
 REV=${2:-"FETCH_HEAD"}
 
 TARGET_DIR="make-checkout-workdir"
-rm -rf "$TARGET_DIR"
+# avoid failures with non-writable subdirectories
+if [ -e "$TARGET_DIR" ]; then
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
:+1:

martinpitt:
Hah, our new tests [caught a bug](https://github.com/cockpit-project/bots/actions/runs/557818722):

    chmod: cannot access 'make-checkout-workdir': No such file or directory

I'm glad we didn't land that the old way :grin:

marusak (approved):
Thanks!

PR #1666, opened 2021-02-11
Weather weather weather weather

marusak:
https://mmarusak.fedorapeople.org/tests.html (with the current DB)

### DIFF
--- a/store-tests
+++ b/store-tests
@@ -96,6 +96,7 @@ def main():
                 cc[-1]["started"] = datetime.strptime(status["created_at"], DATE_FORMAT).timestamp()
             elif "Not yet tested" in desc:
                 new_run = {"posted": datetime.strptime(status["created_at"], DATE_FORMAT).timestamp()}
+                new_run["state"] = "pending"
                 cc.append(new_run)
             else:
                 logging.warning("Pending status has unexpected description '{0}', skipping.".format(desc))
@@ -117,14 +118,11 @@ def main():
                 repo = parts[1]
 
         for retry, status in enumerate(statuses):
-            if not status.get("finished"):
-                # TODO: Store partial, store as non-finished?
-                logging.debug("Test hasn't finished, skipping")
-                continue
-
-            url = status["url"]
-            waited = status["started"] - status["posted"]
-            took = status["finished"] - status["started"]
+            posted = status["posted"]  # This must be always present, lets fails if it isn't
+            url = status.get("url")
+            started = status.get("started", posted)
+            waited = started - posted
+            took = status.get("finished", started) - started
             logging.debug("Processing {0}{1} as {2} that waited {3}s and took {4}s in {5}. {6}".format(
                 context,
                 "(retry no." + str(retry) + ")" if retry else "",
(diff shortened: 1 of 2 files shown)


store-tests
```diff
@@ -82,7 +82,7 @@ def main():
     # 'repo', 'url', 'state', 'description'
     # 'started' - when this run was picked by bot
     # 'posted' - when this run was triggered
-    # 'finished' - when this run finished - 'state' is then either 'success', 'failure' or 'error'
+    # 'finished' - when this run finished - 'state' is then either 'pending', 'success', 'failure' or 'error'
```
martinpitt, line 85: That comment is a bit confusing -- The "finished" field does not exist for state "pending"; move the comment to "posted"?

store-tests
```diff
@@ -117,36 +118,39 @@ def main():
...
+            waited = started - posted
+            took = status.get("finished", started) - started
             logging.debug("Processing {0}{1} as {2} that waited {3}s and took {4}s in {5}. {6}".format(
                 context,
                 "(retry no." + str(retry) + ")" if retry else "",
                 status["state"] + (" (" + status["description"] + ")" if status.get("description") else ""),
                 waited, took, repo, url
             ))
 
+            # URL is our primary key`, so if we don't have one, just fake something unique
```
martinpitt, line 133: stray `

store-tests
```diff
@@ -117,36 +118,39 @@ def main():
...
+            if not url:
+                url = "-".join(["xxx", opts.revision, context, str(retry)])
```
martinpitt, line 135: xxx → queued or pending?

store-tests
```diff
@@ -117,36 +118,39 @@ def main():
...
                 continue
 
             run_id = insert_run(cursor, repo, opts.revision, context, url,
-                                status["posted"], retry, waited, took, status["state"],
+                                posted, retry, waited, took, status["state"],
                                 status.get("description"))
 
-            raw_log = url[:-5] if url.endswith(".html") else url
-            logging.debug("scanning log %s", raw_log)
+            if (status.get("url")):
```
martinpitt, line 146: My first instinct here was "this should be `if url:`, but of course we don't want to scan fake URLs. A small comment may be helpful for our future selves. Or better yet, move this up as `else:` to the place that sets the fake url., and then conditionalize this on `raw_url`.

BTW, this is python, no need to wrap the condition into ().

tests.html
```diff
@@ -276,7 +276,7 @@ <h2 id="g4-title"></h2>
                             SELECT testname, COUNT(run) as failed
                             FROM Tests
                             JOIN TestRuns ON Tests.run = TestRuns.id
-                            WHERE TestRuns.time > ${since} AND Tests.failed = 1 AND project LIKE '${repo}'
+                            WHERE TestRuns.time > ${since} AND Tests.failed = 1 AND project LIKE '${repo}' AND testname NOT LIKE '%selenium%'
```
martinpitt, line 279: Can you please add a commit message (or comment) like "The selenium tests have a bug where skipped tests show up as failures. They are deprecated and not very flaky, so not interesting to track."

martinpitt (changes requested):
Superb, dakujem! Looks a lot nicer :grin:

store-tests
marusak: Right, I want to keep also checks for already existing, mostly for sanity and also don't want to duplicate insertion.
I'll add comment.

store-tests
marusak: Oh right, I just saw listing of states so I added it without reading what it is really about. Removed it.

marusak:
Thanks, updated

martinpitt (approved):
:+1:

PR #1670, opened 2021-02-12
Weather 100th time

marusak:
https://mmarusak.fedorapeople.org/tests.html
can be compared with https://images-frontdoor.apps.ocp.ci.centos.org/tests.html as it is using the same DB.


martinpitt (approved):
Cheers!

PR #1687, opened 2021-02-16
Add publish-queue, add inspect-queue --secrets-dir

martinpitt:
This CLI tool will allow us to write workflows that interact with our
tests, for example submit `store-tests` command to the statistics queue
for daily master tests.

-----
This is a prerequisite for both [daily master test runs](https://issues.redhat.com/browse/COCKPIT-751), as well as queueing integration tests after the "build-dist" workflow.

The two working together:
```
$ echo 'echo hello' | ./publish-queue -q statistics --secrets-dir ~/redhat/ci-secrets/webhook/

$ ./inspect-queue --secrets-dir ~/redhat/ci-secrets/webhook/ 
public queue:
queue public is empty
rhel queue:
queue rhel is empty
statistics queue:
{"command": "./store-tests --db /cache/images/test-results.db --repo cockpit-project/cockpit 8592582f455a4c67b0a431440b64649c43df12a3"}
{"command": "echo hello"}
```

The store-tests is of course from some test; the publish-queue entry is from the command above.

 - [x] Fix run-local.sh to not rely on internal default `--amqp`: https://github.com/cockpit-project/cockpituous/pull/386

### DIFF
--- a/task/distributed_queue.py
+++ b/task/distributed_queue.py
@@ -33,6 +33,7 @@ logging.getLogger("pika").propagate = False
 __all__ = (
     'DistributedQueue',
     'DEFAULT_SECRETS_DIR',
+    'DEFAULT_AMQP_SERVER',
     'BASELINE_PRIORITY',
     'MAX_PRIORITY',
     'no_amqp',
@@ -42,6 +43,7 @@ BASELINE_PRIORITY = 5
 MAX_PRIORITY = 9
 # see https://github.com/cockpit-project/cockpituous/blob/master/tasks/cockpit-tasks-webhook.yaml
 DEFAULT_SECRETS_DIR = '/run/secrets/webhook'
+DEFAULT_AMQP_SERVER = 'amqp-frontdoor.apps.ocp.ci.centos.org:443'
 
 arguments = {
     'rhel': {
--- a/task/distributed_queue.py
+++ b/task/distributed_queue.py
@@ -18,8 +18,9 @@
 # Shared GitHub code. When run as a script, we print out info about
 # our GitHub interacition.
 
-import ssl
 import logging
+import os
+import ssl
 
 no_amqp = False
 try:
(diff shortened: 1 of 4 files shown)


martinpitt:
Submitting this to origin, as this touches inspect-queue, and we should do the full "cockpituous" test glory.

martinpitt:
@Gundersanne FYI -- I guess you also use inspect-queue from time to time, and that makes it a lot more convenient.

martinpitt:
Hmm, we were a bit shy about hardcoding amqp-frontdoor.apps.ocp.ci.centos.org as a default; but "localhost" makes sense almost nowhere, and like that we would have to hardcode it in several cockpit PRs. So let me rework this to have a sensible --amqp default.

martinpitt:
There, that's better. With that we can write a sensible workflow for submitting PR statistics, instead of [this hack](https://github.com/martinpitt/cockpit/blob/tests-trigger-workflow/.github/workflows/pr-statistics.yml#L29).

martinpitt:
Meh, this breaks -- need to call `inspect-queue` with the right local `--amqp`. I sent https://github.com/cockpit-project/cockpituous/pull/386 to fix this.

task/distributed_queue.py
```diff
@@ -72,13 +79,16 @@ def __init__(self, amqp_server, queues, **kwargs):
         if no_amqp:
             raise ImportError('pika is not available')
 
+        if secrets_dir is None:
+            secrets_dir = DEFAULT_SECRETS_DIR
```
marusak, line 85: Why not just use `DEFAULT_SECRETS_DIR` as default value in `__init__` definition?

marusak (approved):
Super nice, thanks!
One tiny nitpick. Approving in the case I missed reasoning behind it.

task/distributed_queue.py
martinpitt: You are right, this is nicer. Fixed.

marusak (approved):
Thanks!

PR #1705, opened 2021-02-21
logs.html: Change to bootstrap4 (fixing progress bar) and fix miscounting retries

marusak:


marusak:
I've posted some before and after shots here: https://github.com/cockpit-project/cockpit/pull/15378#issuecomment-782882615

task/log.html
```diff
@@ -355,7 +347,9 @@ <h4 class="panel-title">
 }
 
 $(function () {
+    /*
```
martinpitt, line 357: Is that a debugging leftover? I find it rather nice that the page self-updates.

task/log.html
```diff
@@ -108,7 +108,7 @@
         <script id="TestingOverview" type="text/template">
             <div id="testing">
                 {{total}} tests, {{passed}} passed, {{failed}} failed,
-                {{skipped}} skipped, {{left}} to go ({{retries}} retries).<br>
+                {{skipped}} skipped, {{left}} to go ({{retries}} failure and {{affected_retries}} affected retries).<br>
```
martinpitt, line 111: That reads a bit weird -- shouldn't it be `{{retries}} failure retries` or just "retries"? Or perhaps "failures", but I'd only count "final" failures as such.

martinpitt (changes requested):
Thanks for fixing this!

task/log.html
marusak: oh it is. It is annoying when you try to work with dom or css in browser and it reloads every few seconds. But surely it belong there in normal circumstances.

task/log.html
marusak: ideas welcomed.  There is `Retry (be robust against unstable tests) and `Retry (test affected tests 3 times)`. There are rather different as one points to a problem, the other not. The first (here marked as `failure`) is also shown in the top list. The others are not, I just kept it as interesting number to see (as you can see whether no tests are affected or it ran hundred tests more then normally PR would). Do you suggest not showing this other number at all?

marusak:
> but 36 and 42 affected retries seem excessive?

It translates to 12 and 14 affected tests. That does not seem so bad. Standard class has around 6 or so tests in it.

marusak:
I just noticed it has lost those icons for screenshot and journal.

marusak:
Icons are back.

iconsareback

task/log.html
marusak: How about `...(retries: N failures, M successes)`?
Or just `...(N failure retries, M success retries)`?

martinpitt:
Oh, of course, I misinterpreted "affected retries", sorry! That makes total sense of course :)

task/log.html
martinpitt: I just meant that "(3 failure and 26 affected retries)" reads weird -- at least it should be failure*s*. Does it count "final" failures (did not succeed even after 3 tries) and "intermediate" failures (but retry succeeded) separately?

marusak: The full line is:
```
N tests, M passed, K failed, L skipped, O to go (P retries of failures, R retries of successes)
```
- K is number of "final" failures (did not succeed even after 3 tries | failed while was affected)
- P is number of "intermediate" failures. But does not say anything about the whole test. It is just test that was `not ok` and also had `RETRY` (that is different from `test affected tests 3 times`)
- R is number of tests, that were `ok` but we retry them anyway because they were affected. If affected test fails, it is not retried at all.

We don't show if tests succeeded but had retry. But that can be easily seen from the overview as it lists tests - the ones with white background and retries (just the P ones) and with red background are failed tests that were not retried. So if there is `testName` with two white bakgrounds followed by one red one, it means it was retried twice and even the third run failed.

marusak:
affectedsuccesses

task/log.html
martinpitt: I see you fixed the failure*s* now, thanks!

martinpitt (approved):
Thanks!

PR #1701, opened 2021-02-22

cockpituous:

### DIFF
--- a/images/debian-testing
+++ b/images/debian-testing
@@ -1,1 +1,1 @@
-debian-testing-5dacff95c7d497af389d8204d436acf121bae7faf541f794cf8c384fc4917cd4.qcow2
+debian-testing-fec9816680deb8e39e82667c4fe2742a0334373120d2434246cf1bfbf75a621e.qcow2
--- a/naughty/debian-testing/1733-udisks2-fails-to-format-vfat
+++ b/naughty/debian-testing/1733-udisks2-fails-to-format-vfat
@@ -1,0 +1,1 @@
+warning: Error synchronizing after formatting with type `vfat': Timed out waiting for object


martinpitt:
[timeout during image-prepare](https://logs.cockpit-project.org/logs/pull-1701-20210222-075430-11a4e4dc-debian-testing-cockpit-project-cockpit/log.html). Retrying once to see if it's an image regression or a random flake.

martinpitt:
` test/verify/check-storage-format TestStorageFormat.testFormatTypes` [regression](https://logs.cockpit-project.org/logs/pull-1701-20210223-061941-11a4e4dc-debian-testing-cockpit-project-cockpit/log.html#212), needs work.

allisonkarlitskaya:
downgrading `dosfstools` to `4.1-2` fixes the problem

allisonkarlitskaya:
This is really not at all a cockpit issue, because this command on the old `dosfsutils`:

```
root@debian:~# gdbus call --system --dest org.freedesktop.UDisks2 --object-path /org/freedesktop/UDisks2/block_devices/sda --method org.freedesktop.UDisks2.Block.Format vfat '{}'
()
```

and again on the new:

```
root@debian:~# gdbus call --system --dest org.freedesktop.UDisks2 --object-path /org/freedesktop/UDisks2/block_devices/sda --method org.freedesktop.UDisks2.Block.Format vfat '{}'
Error: GDBus.Error:org.freedesktop.UDisks2.Error.Failed: Error synchronizing after initial wipe: Timed out waiting for object
```

allisonkarlitskaya:
oh, one more thing to mention: the output of the two commands was different.

```
root@debian:~# mkfs.vfat -I /dev/sda
mkfs.fat 4.1 (2017-01-24)
attribute "partition" not found
```

vs

```
root@debian:~# mkfs.vfat -I /dev/sda
mkfs.fat 4.2 (2021-01-31)
```

and, indeed, if we try to format `/dev/sda1` instead of `/dev/sda` then we don't have this problem on either version:

```
root@debian:~# gdbus call --system --dest org.freedesktop.UDisks2 --object-path /org/freedesktop/UDisks2/block_devices/sda1 --method org.freedesktop.UDisks2.Block.Format vfat '{}'
()
```

martinpitt:
@allisonkarlitskaya : Right, on image refreshes many/most test failures are OS regressions. In this case it's not clear whether to file the bug against dosfstools or udisks2. It feels like 4.2 actually "fixed" making file systems on raw devices, but perhaps udisks was expecting this error? Either way, this sounds like bug report + naughty. Thanks for investigating!

martinpitt (approved):
Thanks! Seems github workflows are stuck right now. I'm going to land that anyway, as the "bots" test doesn't touch the changes here.

PR #1721, opened 2021-02-25
testmap: remove rhel-8-4-distropkg from master

allisonkarlitskaya:
We're about to introduce a change to cockpit which adds a variable
expansion feature to manifest parsing, and we want to use this feature
right away from the manifests in the same version.  This means that
the -distropkg variant is going to be broken until that new version gets
packaged.

Meanwhile, we're about to move over to RHEL 8.5 anyway.

It seems the easiest thing for now is to just drop rhel-8-4-distropkg
from the testmap.

### DIFF
--- a/task/testmap.py
+++ b/task/testmap.py
@@ -36,7 +36,6 @@ REPO_BRANCH_CONTEXT = {
             'fedora-coreos',
             'fedora-32/firefox',
             'rhel-8-4',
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Confirming. We now have a rhel-8.4 branch and are not going to update 8.4 from master any more; we are past the development freeze. So we don't require this particular test from master any more.

PR #1725, opened 2021-02-26
Image refresh for fedora-32

cockpituous:
Image refresh for fedora-32
 * [x] image-refresh fedora-32


martinpitt:
lorax test has been broken for a while, so this is ok.

martinpitt (approved):

PR #1727, opened 2021-02-27
image-diff: Avoid "command not found" in output

martinpitt:
Explicitly check if `dpkg-query` exists. Otherwise, on RHEL/CentOS 7
the first line of output is `bash: dpkg-query: command not found`
which breaks the parser.

----

Spotted in the rhel-7-9 image refresh in issue #1726, which has a bit of debugging information. I tested this on centos-7, rhel-8-4, fedora-34, and debian-testing.


marusak (approved):
Thanks :)

PR #1765, opened 2021-03-09
add xdg_*_home helpers and move some files to ~/.config/cockpit-dev/

allisonkarlitskaya:
Two major things going on here:

First, add some xdg_*_home() helpers.  These functions consolidate the usual logic for finding the user's xdg directory and forming a path based on it
    
 - check the relevant XDG_ environment variable
    
 - default to the usual place relative to ~
    
 - os.path.join() the result with the provided arguments
    
 - ability to provide an environment variable to override the entire thing with a hardcoded path

We also port a few existing users of files in `~/.config/` and `~/.cache/` (with their homegrown handling) to the new APIs.

Second, we move the `github-token` and `image-stores` files out of `~/.config` and into `~/.config/cockpit-dev`.  For `github-token` we provide backwards compatibility with a deprecation warning.

 * [ ] #1766

### DIFF
--- a/machine/machine_core/directories.py
+++ b/machine/machine_core/directories.py
@@ -64,7 +64,6 @@ def get_images_data_dir():
         _images_data_dir = get_git_config('--type=path', 'cockpit.bots.images-data-dir')
 
         if _images_data_dir is None:
-            _images_data_dir = os.path.join(os.getenv('XDG_CACHE_HOME', os.path.expanduser("~/.cache")),
-                                            "cockpit-images")
+            _images_data_dir = xdg_cache_home('cockpit-images')
 
     return _images_data_dir
--- a/machine/machine_core/directories.py
+++ b/machine/machine_core/directories.py
@@ -24,6 +24,26 @@ _images_data_dir = None
 _temp_dir = None
 
 
+def xdg_home(subdir, envvar, *components, override=None):
+    path = override and os.getenv(override)
+
+    if not path:
+        directory = os.getenv(envvar)
+        if not directory:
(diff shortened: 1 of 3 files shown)


machine/machine_core/directories.py
```diff
@@ -24,6 +24,26 @@
...
+
+    return path
+
+
+def xdg_config_home(*components, envvar=None):
+    return xdg_home('.config', 'XDG_CONFIG_HOME', *components, override=envvar)
+
+
+def xdg_cache_home(*components, envvar=None):
+    return xdg_home('.config', 'XDG_CACHE_HOME', *components, override=envvar)
```
martinpitt, line 44: s/config/cache/

task/github.py
```diff
@@ -129,17 +130,23 @@ def __init__(self, base=None, cacher=None, repo=None):
...
                 self.token = f.read().strip()
         except FileNotFoundError:
-            # fall back to GitHub's CLI token
             try:
-                with open(xdg_config_home("gh/config.yml")) as f:
-                    match = re.search(r'oauth_token:\s*(\S+)', f.read())
-                if match:
-                    self.token = match.group(1)
+                with open(xdg_config_home('github-token'), "r") as f:
+                    print('~/.config/github-token is deprecated.  ' +
```
martinpitt, line 138: No need for the `+` -- the parser will just treat several concatenated strings as one, similar to C.

martinpitt (changes requested):
Some small nitpicks. But mostly: Please resubmit that as a branch against origin, so that the full test suite can run. Thanks!

allisonkarlitskaya:
→ #1767

PR #1766, opened 2021-03-09
add a new toplevel lib/ directory

allisonkarlitskaya:
Add a new toplevel `lib/` directory for the many things that don't fit comfortably into either `machine/` or `task/`.


machine/machine_core/constants.py
```diff
@@ -1,39 +1,10 @@
...
-
-IMAGES_DIR = os.path.join(BOTS_DIR, "images")
-SCRIPTS_DIR = os.path.join(IMAGES_DIR, "scripts")
-
-DEFAULT_IDENTITY_FILE = os.path.join(MACHINE_DIR, "identity")
-
-TEST_OS_DEFAULT = "fedora-33"
-DEFAULT_IMAGE = os.environ.get("TEST_OS", TEST_OS_DEFAULT)
+# import sys
+# print('machine_core.constants is deprecated.  Use lib.constants', file=sys.stderr)
```
martinpitt, line 10: Please drop these comments, or better yet, throw a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning)

machine/machine_core/directories.py
```diff
@@ -1,50 +1,8 @@
...
-    if _images_data_dir is None:
-        _images_data_dir = get_git_config('--type=path', 'cockpit.bots.images-data-dir')
-
-        if _images_data_dir is None:
-            _images_data_dir = os.path.join(os.getenv('XDG_CACHE_HOME', os.path.expanduser("~/.cache")),
-                                            "cockpit-images")
-
-    return _images_data_dir
+# import sys
+# print('machine_core.directories is deprecated.  Use lib.directories', file=sys.stderr)
```
martinpitt, line 8: dito

image-download
```diff
@@ -45,8 +45,8 @@ import urllib.parse
 
 from lib.constants import IMAGES_DIR
 from lib.directories import get_images_data_dir
+from lib.stores import PUBLIC_STORES, REDHAT_STORES, CA_PEM
```
martinpitt, line 48: Let's drop the `As far as I know..." from the commit message. I checked `for d in *; do git -C $d grep STORES; done` in my ~/upstream/ dir (same for `CA_PEM`), and it's indeed only being used in bots itself.

lib/stores.py
```diff
@@ -0,0 +1,66 @@
...
+
+from lib.constants import IMAGES_DIR
+
+# Cockpit image/log server CA
+CA_PEM = os.getenv("COCKPIT_CA_PEM", os.path.join(IMAGES_DIR, "files", "ca.pem"))
+
+# Servers which have public images
+PUBLIC_STORES = [
+    "https://images-frontdoor.apps.ocp.ci.centos.org/",
+    "https://images-cockpit.apps.ci.centos.org/",
```
martinpitt, line 33: While you are at it, please add a commit that drops this. The OCP3 cluster got shut down, and trying to ping this store hangs, and is just in vain.

martinpitt (dismissed):
Thanks! A few small things, but I like it!

martinpitt (approved):
Cheers! (Please wait for green, though)

martinpitt (changes requested):
Breaks project tests, sorry..

martinpitt (approved):
Danke!

martinpitt (approved):

PR #1767, opened 2021-03-09
add xdg_*_home helpers and move some files to ~/.config/cockpit-dev/

allisonkarlitskaya:
Two major things going on here:

First, add some xdg_*_home() helpers.  These functions consolidate the usual logic for finding the user's xdg directory and forming a path based on it
    
 - check the relevant XDG_ environment variable
    
 - default to the usual place relative to ~
    
 - os.path.join() the result with the provided arguments
    
 - ability to provide an environment variable to override the entire thing with a hardcoded path

We also port a few existing users of files in `~/.config/` and `~/.cache/` (with their homegrown handling) to the new APIs.

Second, we move the `github-token` and `image-stores` files out of `~/.config` and into `~/.config/cockpit-dev`.  For `github-token` we provide backwards compatibility with a deprecation warning.

 * [x] #1766 
 * [x] fix selenium tests in cockpit, in case we need more bots changes or a revert for #1766

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -23,11 +23,11 @@ import http.client
 from http import HTTPStatus
 import json
 import os
+import re
 import socket
+import subprocess
 import time
 import urllib.parse
-import subprocess
-import re
 
 from lib.testmap import known_context
 from lib.directories import xdg_config_home, xdg_cache_home
@@ -121,17 +121,21 @@ class GitHub(object):
         self.token = None
         self.debug = False
         try:
-            with open(xdg_config_home('github-token'), "r") as f:
+            with open(xdg_config_home('cockpit-dev', 'github-token'), "r") as f:
                 self.token = f.read().strip()
         except FileNotFoundError:
-            # fall back to GitHub's CLI token
             try:
-                with open(xdg_config_home("gh/config.yml")) as f:
-                    match = re.search(r'oauth_token:\s*(\S+)', f.read())
-                if match:
-                    self.token = match.group(1)
+                with open(xdg_config_home('github-token'), "r") as f:
+                    self.token = f.read().strip()
             except FileNotFoundError:
-                pass
+                # fall back to GitHub's CLI token
+                try:
+                    with open(xdg_config_home("gh/config.yml")) as f:
+                        match = re.search(r'oauth_token:\s*(\S+)', f.read())
+                    if match:
+                        self.token = match.group(1)
+                except FileNotFoundError:
+                    pass
 
         # default cache directory
         if not cacher:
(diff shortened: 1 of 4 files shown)


task/github.py
```diff
@@ -121,17 +122,24 @@ def __init__(self, base=None, cacher=None, repo=None):
         self.token = None
         self.debug = False
         try:
-            with open(xdg_config_home('github-token'), "r") as f:
+            with open(xdg_config_home('cockpit-dev', 'github-token'), "r") as f:
```
martinpitt, line 125: I have to do some bikeshedding here: There is actually nothing cockpit specific about this token. It represents *you* in GitHub, not any particular project. It equally applies when you are hacking on subscription-manager, lorax, starter-kit and all other projects that use our bots project; and you could use it in your own scripts just as well. So unlike image-stores, I don't consider this particularly bad namespace trampling honestly.

If you go through with that, please adjust README and .github/workflows/daily.yml alongside. After that we should then also gradually adjust the workflows and scripts that use it in cockpit, starter-kit, cockpit-podman, etc.

task/github.py
```diff
@@ -121,17 +122,24 @@ def __init__(self, base=None, cacher=None, repo=None):
...
         except FileNotFoundError:
-            # fall back to GitHub's CLI token
             try:
-                with open(xdg_config_home("gh/config.yml")) as f:
-                    match = re.search(r'oauth_token:\s*(\S+)', f.read())
-                if match:
-                    self.token = match.group(1)
+                with open(xdg_config_home('github-token'), "r") as f:
+                    warnings.warn('~/.config/github-token is deprecated.  '
+                                  'Please move to ~/.config/cockpit-dev/github-token.',
```
martinpitt, line 131: Please drop the final period, so that this can be copy&pasted, and there is no ambiguity what is part of the file name and what is merely part of the sentence.

martinpitt (changes requested):
Thanks! The first three commits are fine, the fourth I have some more trouble with. Other opinions?

task/github.py
allisonkarlitskaya: This is not 100% true.

You can create any number of tokens in github.  Back in the day, before I even understood any of this stuff, I created mine with exactly the permissions that were required for cockpit, and named the token "cockpit bots".  At least in my case, it's very cockpit-specific.

My `gh` has a token with more capabilities.

task/github.py
allisonkarlitskaya: Sure.  But let's chat tomorrow about if we just drop this commit or not.

allisonkarlitskaya:
Compromise: support both paths, without warnings.

I updated the README anyway.

martinpitt (approved):
Nice, green now. Thanks!

PR #1771, opened 2021-03-10
image-download cleanup

allisonkarlitskaya:
Here's a pair of commits to make the "find the server" logic in image-download easier to understand

### DIFF
--- a/image-download
+++ b/image-download
@@ -60,11 +60,30 @@ else:
     SYMBOLS = {'present': ' >', 'absent': ' x', 'selected': '=>'}
 
 
+def show_status(quiet, status, *args, prefix=''):
+    if not quiet:
+        message = ' '.join(args)
+        sys.stderr.write(f'{prefix}{SYMBOLS[status]} {message}\n')
+
+
+def curl_cmd(args):
+    return ['curl', '--connect-timeout', '10', '--fail'] + args
+
+
+def check_curl_args(args):
+    head_args = ['--silent', '--head']  # only used for this check
+
+    try:
+        return subprocess.check_output(curl_cmd(args + head_args), universal_newlines=True)
+    except subprocess.CalledProcessError:
+        return None
+
+
 def find(name, stores, latest, quiet):
     found = []
 
     for store in stores:
-        url = urllib.parse.urlparse(store)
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
After trying to rebase the S3 work on top of this I'm not 100% happy about one little detail.  I'm going to improve this before continuing.

image-download
```diff
@@ -85,6 +85,17 @@ def find(name, stores, latest, quiet):
     for store in stores:
         url = urllib.parse.urlparse(urllib.parse.urljoin(store, name))
 
+        # First, try the easy case: for OpenShift proxied stores which send
+        # their own SSL cert initially host name has to match that
```
martinpitt, line 89: This is a bit confusing -- the OpenShift hosted image server has exactly the same cockpit cert than our e2e and aws ones; we use TLS passthrough, as unfortunately TLS endpoint termination doesn't work for large files (it's a pity, as otherwise we'd have a real certificate on these). It's just that its DNS name is predictable and fixed, so we don't have to use the `--resolve` trick.

image-download
```diff
@@ -85,6 +85,17 @@ def find(name, stores, latest, quiet):
...
+        args = ['--cacert', CA_PEM, url.geturl()]
+        output = check_curl_args(args)
+        if output:
+            show_status(quiet, 'present', store)
+            found.append((args, output, store))
+            continue
+
+        # Otherwise, the server is expecting the hostname "cockpit-tests".
```
martinpitt, line 97: The server couldn't care less, it's just that our [certificate](https://github.com/cockpit-project/cockpituous/blob/master/tasks/credentials/openssl.cnf#L54) has `cockpit-tests` as a valid subjectAltName for image servers which don't have a predictable DNS name (such as we used to have with our AWS instance until a few weeks ago). I'm not actually sure if we still need this fallback case, now that all our image stores have a predictable name.

However, we can clean this up with moving to S3, when we can hopefully just drop this entire hackery. :grin:

martinpitt (approved):
Thanks! Not entirely happy about the comments, but then again this file is already so confusingly hacky that it hardly matters now, and you are already working on getting rid of our image stores.

PR #1772, opened 2021-03-10
lib: split stores into a text file

allisonkarlitskaya:
Break the list of the stores out of lib/stores.py and into a normal text
file at lib/stores.  Each line contains two fields, according to python
str.split(): the first is the "scope", which should either be "public"
or "redhat", and the second is the url.

This change will benefit cimg[1] which will no longer need to maintain
its own list.

[1] https://github.com/allisonkarlitskaya/cimg


martinpitt:
Triggering an image refresh to validate that this works.

cockpituous:
image-refresh fedora-coreos done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-coreos-20210310-100404

martinpitt (approved):
That worked, thanks!

PR #1774, opened 2021-03-10
add a new S3 image store in Frankfurt (Linode)

allisonkarlitskaya:
This adds support for S3 image stores and adds one hosted at Linode in Frankfurt.

This server is *substantially* faster for me than the American ones, and should also be nice for most other members of our predominantly-central-European developer team.

Let's test some uploads from the bots!

 * [x] image-refresh cirros
 * [x] image-refresh rhel-8-4

### DIFF
--- a/image-download
+++ b/image-download
@@ -45,9 +45,9 @@ import urllib.parse
 
 from lib.constants import IMAGES_DIR
 from lib.directories import get_images_data_dir, xdg_config_home
-from lib.stores import PUBLIC_STORES, REDHAT_STORES, CA_PEM
-
+from lib.stores import HYBRID_STORES, PUBLIC_STORES, REDHAT_STORES, CA_PEM
 from lib.testmap import get_test_image
+from lib import s3
 
 DEVNULL = open("/dev/null", "r+")
 EPOCH = "Thu, 1 Jan 1970 00:00:00 GMT"
@@ -83,8 +83,23 @@ def find(name, stores, latest, quiet):
     for store in stores:
         url = urllib.parse.urlparse(urllib.parse.urljoin(store, name))
 
-        # First, try the easy case: for OpenShift proxied stores which send
-        # their own SSL cert initially host name has to match that
+        # First, check if this is an S3 store for which we have a key
+        if s3.is_key_present(url):
+            output = check_curl_args([s3.sign_url(url, verb='HEAD')])
+            if output:
+                show_status(quiet, 'present', store, '(authenticated)')
+                found.append(([s3.sign_url(url)], output, store))  # GET
+                continue
+
+        # Next, try to access the URL directly, without further help
+        args = [url.geturl()]
+        output = check_curl_args(args)
+        if output:
+            show_status(quiet, 'present', store)
+            found.append((args, output, store))
+            continue
+
+        # If that didn't work, try using our CA_PEM
         args = ['--cacert', CA_PEM, url.geturl()]
         output = check_curl_args(args)
         if output:
@@ -140,6 +155,7 @@ def download(dest, force, state, quiet, stores):
                 stores = fp.read().strip().split("\n")
         except FileNotFoundError:
             stores = []
+        stores += HYBRID_STORES
         # skip testing public stores for private images
         if not name.startswith("rhel") and not name.startswith("windows"):
             stores += PUBLIC_STORES
--- a/image-upload
+++ b/image-upload
@@ -27,19 +27,20 @@ import urllib.parse
 
 from lib.constants import IMAGES_DIR
 from lib.directories import get_images_data_dir
-from lib.stores import PUBLIC_STORES, REDHAT_STORES, CA_PEM
+from lib.stores import HYBRID_STORES, PUBLIC_STORES, REDHAT_STORES, CA_PEM
+from lib import s3
 
 from task import api
 
 
-def upload(store, source):
-    url = urllib.parse.urlparse(store)
+def upload(dest, source, public):
+    url = urllib.parse.urlparse(dest)
 
     # Start building the command
-    cmd = ["curl", "--progress-bar", "--cacert", CA_PEM, "--fail", "--upload-file", source]
+    cmd = ["curl", "--progress-bar", "--fail", "--upload-file", source]
 
     def try_curl(cmd):
-        print("Uploading to", cmd[-1])
+        print("Uploading to", cmd[-1].split('?')[0])  # hide authentication info
         # Passing through a non terminal stdout is necessary to make progress work
         curl = subprocess.Popen(cmd, stdout=subprocess.PIPE)
         cat = subprocess.Popen(["cat"], stdin=curl.stdout)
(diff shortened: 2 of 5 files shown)


cockpituous:
Task failed: https://logs.cockpit-project.org/logs/image-refresh-1774-20210310-135959/

cockpituous:
image-refresh in progress on 3-cockpit-10.
Log: https://logs.cockpit-project.org/logs/image-refresh-1774-20210310-140505/

cockpituous:
Task failed: https://logs.cockpit-project.org/logs/image-refresh-1774-20210310-140505/

cockpituous:
image-refresh in progress on 2-cockpit-11.
Log: https://logs.cockpit-project.org/logs/image-refresh-1774-20210310-142626/

cockpituous:
image-refresh fedora-coreos done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-coreos-20210310-143333

martinpitt:
[This CI failure](https://github.com/cockpit-project/bots/pull/1774/checks?check_run_id=2078209175#step:6:195) is definitively real:
```
  File "/work/bots/./image-upload", line 130, in <module>
    sys.exit(main())
  File "/work/bots/./image-upload", line 121, in main
    ret = upload(store + basename, source, public)
UnboundLocalError: local variable 'basename' referenced before assignment
```

cockpituous:
image-refresh fedora-coreos done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-coreos-20210310-162424

allisonkarlitskaya:
It's problematic that the signed URL is visible in the logs...

cockpituous:
image-refresh in progress on 2-ci-srv-05.
Log: https://logs.cockpit-project.org/logs/image-refresh-1774-20210310-223737/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20210310-223737

cockpituous:
image-refresh rhel-8-4 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-8-4-20210310-225555

allisonkarlitskaya:
The cirros image can be directly downloaded from the URL in the log.

The rhel-8-4 image can't be, but if you have the correct key installed, you can do this:

```
python -m lib.s3 get https://cockpit-images.eu-central-1.linodeobjects.com/rhel-8-4-3d83ed1971f608c65a0be1c6ce34132764ef633f6b8c027b2d8e5daa0fe7413e.qcow2
```

and you'll get a URL which does work.


In short: this is working as expected.

image-download
```diff
@@ -307,20 +323,16 @@ def download_images(image_list, force, quiet, state, store):
...
-            # don't download the same thing multiple times in parallel
-            wait_lock(target)
+        # don't download the same thing multiple times in parallel
+        wait_lock(target)
 
-            if force or state or not os.path.exists(target):
-                download(target, force, state, quiet, store)
-        except Exception as ex:
-            success = False
-            sys.stderr.write("image-download: {0}\n".format(str(ex)))
```
martinpitt, line 323: Admittedly I don't quite like these bare exception handlers. But this changes behaviour, and thus it should at least be in a separate commit and justified carefully. Possibly it's just a leftover from intermediate debugging where you did want to see the full tracebacks?

image-download
```diff
@@ -83,8 +83,23 @@ def find(name, stores, latest, quiet):
     for store in stores:
         url = urllib.parse.urlparse(urllib.parse.urljoin(store, name))
 
-        # First, try the easy case: for OpenShift proxied stores which send
-        # their own SSL cert initially host name has to match that
+        # First, check if this is an S3 store for which we have a key
+        if s3.is_key_present(url):
+            output = check_curl_args([s3.sign_url(url, verb='HEAD')])
+            if output:
```
martinpitt, line 89: Is that guaranteed to indicate success? Or could it also be a non-empty error message?

lib/s3.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/python3
+
+# This file is public domain (CC0-1.0)
+
+# Adapted from examples in
+# http://s3.amazonaws.com/doc/s3-developer-guide/RESTAuthentication.html
```
martinpitt, line 6: https:// s'il vous plaît

martinpitt (changes requested):
Nice work! Some small issues, but I'm excited about this direction. We have wanted to move to cloud storage for some time already.

Some general issues:

This currently makes the linode store the primary image store, as image-upload prefers uploading there. This is too early, as it will lock out any developer from downloading RHEL images who does not have an s3 key yet. Also, this is you personal Linode account.

Thus, for the time being, I'd say it should upload to linode plus the first public/private store (i.e. CentOS CI or cockpit-11). This can be a hacky special case, as it won't last long -- I suppose we'll soon figure out how to get an S3 bucket in our Red Hat AWS account and replace cockpit-11 with an S3 bucket in the US (so that our CI still has a close mirror).

Image expiry needs to be sorted out -- at least there needs to be a PoC that it can be done. But it needs to be implemented soon, as images pile up quickly, and will soon exceed your 150 GB quota, and there's no point in keeping old ones around for too long.

image-download
allisonkarlitskaya: This was unintentional yes.   I'll remove it.

But I also don't like it.  Separate issue, though :)

image-download
allisonkarlitskaya: the `check` in `subprocess.check_output()` means that non-zero returns are raised as exceptions.  In that case we return `None`.  the `--fail` to `curl` means that non-success will return non-zero exit status.

This is the same basic mechanism we relied on before, except that we did the work inside of the `try` branch.  With the reorg, the only way this function returns non-`None` is the exact case where we were inside of the `try` branch before.

lib/s3.py
allisonkarlitskaya: OK.  But I was at the `http` one :)

cockpituous:
image-refresh in progress on 2-cockpit-11.
Log: https://logs.cockpit-project.org/logs/image-refresh-1774-20210311-082020/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20210311-082020

image-upload
```diff
@@ -95,21 +109,28 @@ def main():
...
 
+        success = False
         for store in stores:
-            ret = upload(store, source)
+            dest = urllib.parse.urljoin(store, basename)
+            ret = upload(dest, source, public)
             if ret == 0:
-                return ret
-        else:
+                success = True
```
martinpitt, line 131: This subtle change has the effect that image-upload will  now upoad to *all* stores. I am a bit afraid that this will take rather long for our AWS mirror. But we want to replace these "legacy stores" anyway as soon as the dust settles on this, so until then this is a good enough quick fix. Better that than accidentally locking ourselves out of the image stores.

martinpitt (approved):
Yay!

PR #1775, opened 2021-03-10
image-refresh: don't run image-diff on all images

allisonkarlitskaya:
Some of our images can't be accessed with the machine API, so avoid
attempting to automatically run image-diff on those.

A good indicator for which machines should work is the presence of a
.setup file, so check for that before running image-diff.

 * [x] image-refresh cirros

### DIFF
--- a/image-refresh
+++ b/image-refresh
@@ -24,7 +24,7 @@ import sys
 import task
 from task import github
 
-from lib.constants import BOTS_DIR
+from lib.constants import BOTS_DIR, SCRIPTS_DIR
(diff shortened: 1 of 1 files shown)


martinpitt:
Looks good, thanks! Testing this with "cirros", which is the one that failed previously

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20210310-162727

martinpitt (approved):
Yay! Worked great with cirros.

PR #1781, opened 2021-03-11
naughty: Fix patterns with unsafe "*" lines

martinpitt:
Our naughties commonly use this pattern:

some expected log message
*
Traceback..

But this does not match if there are *no* lines between the log message
and the Traceback. This is because the fnmatch still considers the (not
quite obvious) \n line break, and thus a log like

blabla
some expected log message
Traceback foo..
        File ...

does *not* match the above pattern. The solution is to put the asterisk
at the end of the previous line:

some expected log message*
Traceback..

which will then correctly match zero or more lines in between.
    
Add a unit test to illustrate and validate this.

Fix all affected naughty patterns accordingly.

This will fix [this test failure](https://logs.cockpit-project.org/logs/pull-15504-20210311-122809-4ac70864-fedora-34/log.html#285-2).

### DIFF
--- a/naughty/debian-stable/1372-libvirt-crashes-on-test-teardown
+++ b/naughty/debian-stable/1372-libvirt-crashes-on-test-teardown
@@ -1,8 +1,6 @@
 *TestMachines*
 *
-Process * (libvirtd) of user * dumped core.
-*
+Process * (libvirtd) of user * dumped core.*
 #1  * virQEMUDriverGetConfig (libvirt_driver_qemu.so)
 *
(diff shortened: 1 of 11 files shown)


martinpitt:
Oops, collided with fedora-33 refresh, so the [tests fail a lot](https://logs.cockpit-project.org/logs/pull-1781-20210311-155925-5b0ad9d8-fedora-33-cockpit-project-cockpit/log.html). Rebasing.

KKoukiou (approved):
Oh I have written the patterns the wrong way so many times! Thanks for adding the test as well!

PR #1782, opened 2021-03-11
README: add note about S3 keys

allisonkarlitskaya:


README.md
```diff
@@ -90,6 +90,26 @@ the [GitHub CLI](https://cli.github.com/) configuration in
...
+If you'd like to download Red Hat-only internal images from S3, you'll
+need to create a key file in `~/.config/cockpit-dev/s3-keys/[endpoint]`.
+The contents of this file should be a single line containing the "access
+key" and the "secret key" separated by whitespace.
+
+For the currently configured mirrors this means that you'd expect to
+have the following files:
+
+- `~/.config/cockpit-dev/s3-keys/eu-central-1.linodeobjects.com`
+- `~/.config/cockpit-dev/s3-keys/us-east-1.linodeobjects.com`
```
martinpitt, line 102: Having one is just enough, or making one a symlink to the other would work as well, I figure.

martinpitt (approved):
Thanks! I followed this and it works fine.

README.md
allisonkarlitskaya: This is a matter of policy.

I created your key with access to both endpoints, but we could just as well make them separately (as we did for upload).  Indeed, now that your key already exists, if we were to add a new endpoint, I'd need to create a separate key for that (since existing ones can't be modified).  Also: it goes without saying that if we do this on AWS, that will also be a different key.

PR #1783, opened 2021-03-11
images: Get Openshift from quay.io

mvollmer:
* [ ] FAIL: image-refresh openshift

### DIFF
--- a/images/scripts/openshift.setup
+++ b/images/scripts/openshift.setup
@@ -18,7 +18,9 @@ function docker_images_has() {
 }
 
 function docker_pull() {
-    docker pull $1
+    # we get images from our mirror on quay to work around dockerhub rate limits, see sync-quay
+    mirror_name=$(echo $1 | sed 's,/,-,')
+    docker pull "quay.io/cockpit/$mirror_name" && docker tag "quay.io/cockpit/$mirror_name" "docker.io/$1"
     echo "$1" >> /tmp/pulledImages
     docker_images_has $1
 }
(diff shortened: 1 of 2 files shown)


cockpituous:
image-refresh in progress on centosci-tasks-ztr72.
Log: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/image-refresh-1783-20210311-174848/

mvollmer:
Oho, I think my changes are all wrong and we still pull a lot from dockerhub, but we didn't seem to have run into any rate limits this time...

cockpituous:
image-refresh openshift done: https://github.com/cockpit-project/bots/commits/image-refresh-openshift-20210311-180505

cockpituous:
image-refresh in progress on 1-cockpit-8.
Log: https://logs.cockpit-project.org/logs/image-refresh-1783-20210311-181919/

cockpituous:
Task failed: https://logs.cockpit-project.org/logs/image-refresh-1783-20210311-181919/

martinpitt:
Oh cool, so the origin-base quay.io container images *are* a working substitute? Nice!

martinpitt:
> we didn't seem to have run into any rate limits this time...

It's really a matter of day time/luck..

mvollmer:
> Oh cool, so the origin-base quay.io container images _are_ a working substitute?

Looks like it.  Tbh, I don't really know what's going on in openshift.setup...

mvollmer:
> > we didn't seem to have run into any rate limits this time...
> 
> It's really a matter of day time/luck..

Yeah, the second attempt with more quay.io prefixes failed on `docker.io/library/registry`, which is Docker's own stuff. It might not be on Quay, but we should be able to mirror it.

Then there is busybox and maybe other images that are used for testing, I guess.  We can use anything we find on quay, I guess, preferably something that doesn't need changes to the tests themselves.

I am thinking right now that we push the image created by my partial changes as is (when tests are green), conceptually broken as they are (maybe with a comment to the effect), and then I try a bit more to make openshift.setup cleaner and not use docker.io at all.

mvollmer:
> > Oh cool, so the origin-base quay.io container images _are_ a working substitute?
> 
> Looks like it. Tbh, I don't really know what's going on in openshift.setup...

No, it doesn't work:

```
+ docker run --rm --entrypoint tar quay.io/openshift/origin-base:v3.11.0 -C /usr/bin -c openshift oc kubectl
tar: openshift: Cannot stat: No such file or directory
tar: oc: Cannot stat: No such file or directory
tar: kubectl: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
```

`docker run` didn't propagate the error, of course.

In my first attempt, I forgot to update this line, and we were still using the image from docker.io.

mvollmer:
There is `quay.io/openshift/origin`, but it is empty.

mvollmer:
What about just dropping the openshift image completely? It's only used by check-openshift on the rhel-7.9 branch.

mvollmer:
> I am thinking right now that we push the image created by my partial changes as is

On scond thought, let's not. We are fine with a old version for a while.

mvollmer:
Okay, I'll just mirror the whole bunch.

mvollmer:
As @martinpitt points out, the openshift containers on docker.io seem to no longer be updated. We would have to find out what the current way of installing openshift is, which doesn't seem to be worth it. So let's keep using our existing openshift image without any further updates until we drop support for rhel-7-9.

PR #1787, opened 2021-03-12
machine_core: Write SSHConnection messages to stderr

martinpitt:
On CI, test's stdout and err are combined into a single log. As print()
has internal buffering, that often leads to garbled output where
messages like "qemu-img create..." appear in the middle of stdout lines.

Write the messages to stderr instead. Cockpit's testlib.py writes its
messages there also, and so does Python's standard logging module.

This also avoids potentially breaking the TAP protocol on stdout with
debug/progress messages.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -53,9 +53,8 @@ class SSHConnection(object):
 
     def message(self, *args):
         """Prints args if in verbose mode"""
-        if not self.verbose:
-            return
-        print(" ".join(args))
+        if self.verbose:
+            sys.stderr.write(" ".join(args) + '\n')
 
     # wait until we can execute something on the machine. ie: wait for ssh
     def wait_execute(self, timeout_sec=120):


martinpitt:
Noticed in https://github.com/osbuild/cockpit-composer/pull/1262 . I ran the tests against this bots PR there.

martinpitt:
This works *much* better, see https://logs.cockpit-project.org/logs/pull-1262-20210312-071456-9dd70cbb-fedora-33-firefox-bots-1787/log

martinpitt:
The [composer run here](https://logs.cockpit-project.org/logs/pull-1787-20210312-072555-995d8ee5-fedora-33-osbuild-cockpit-composer/log) also looks much better -- the two `qemu-img` messages appear right at the start of the tests where they belong.

martinpitt:
And finally, this [failing composer](https://logs.cockpit-project.org/logs/pull-1262-20210312-071456-9dd70cbb-fedora-33-firefox-bots-1787/log) run has a few "restarting global machine" failures, which now also look correct.

machine/machine_core/ssh_connection.py
```diff
@@ -56,6 +56,13 @@ def message(self, *args):
         if not self.verbose:
             return
         print(" ".join(args))
+        # need to flush stdout immediately to avoid delaying the message with stderr output when writing to log files
+        while True:
```
marusak, line 60: Is it possible that this would get stuck indefinitely? If there is slight chance then maybe do this waiting N times and then give up?

martinpitt: That's the code that we've been using in run-tests for a long time (I copied it from there)

Doesn't make much of a difference, though: If the log sink is broken (stdout indefinitely blocked), then the test log will be truncated and corrupted either way. There is no point in running the tests further. The usual result is that the task container aborts the run because the sink's ssh exits with 255; but even if not, we still have the global timeout to make sure that nothing gets stuck indefinitely.

We could "optimize" the timeout by trying to flush for up to, say, 30 seconds, and then raising the exception. I can send a follow-up for that (also for run-tests) if you want.

martinpitt:
I can reproduce this locally quite easily:

    TEST_OS=fedora-34 test/common/run-tests --test-dir test/verify -v TestLogin.testBasic  > /tmp/o 2>&1

gives:
```
[...]
> warning: transport closed: disconnected
# Result testBasic (__main__.TestLogin) succeeded
# 1 TEST PASSED [28s on toolbox]qemu-img create -q -f qcow2 -o backing_file=/var/home/martin/upstream/cockpit/test/images/fedora-34,backing_fmt=qcow2 /tmp/x2ak69ms.qcow2

ok 1 test/verify/check-static-login TestLogin.testBasic [ND@0]

# TESTS PASSED [30s on toolbox, 1 serial tests: 0: 29s]

```

martinpitt:
@marusak: I reworked this to use stderr, which altogether makes more sense anyway I think. With the local reproducer that now works fine, and I restarted the tests here and in https://github.com/osbuild/cockpit-composer/pull/1262

marusak (approved):
Nice, I like this much more :)

PR #1794, opened 2021-03-15
testmap: Drop selenium tests for stable cockpit branches

martinpitt:
We dropped the selenium tests for cockpit/master in commit 6a7256489.
We will drop them from downstream RHEL 8.5 if/when we rebase, and 8.4
and 7.9 won't change any more.

Thus it does not make sense any more to keep them alive for the stable
branches, they just keep blocking image refreshes.


marusak (approved):
Agree, thanks!

PR #1795, opened 2021-03-15
Clean up fedora-32 image

martinpitt:


martinpitt:
Note: There are still some "fedora-32" special cases in fedora.setup. I will clean these up separately, as that will require at least a fedora-33 image refresh.

martinpitt:
Bwah, this ancient issue [haunts us again](https://logs.cockpit-project.org/logs/pull-1795-20210315-114319-1f649ead-fedora-33-container-bastion-cockpit-project-cockpit-rhel-7.9/log.html) -- seems this does need a fedora-33 image rebuild right away *sigh* -- so I'll do that here instead.

marusak (approved):
Thanks!

PR #1799, opened 2021-03-15
image-download: add "preferred image store" support

allisonkarlitskaya:
image-download now looks in ~/.config/cockpit-dev/image-store-preference
on each download.  If this file exists, it should contain a substring
which matches the name of a single mirror (for example "eu-central-1").
In that case, this mirror will be tried before the other mirrors, and if
the image is present on it, the other mirrors will be completely
ignored.

bonus: drop the "cockpit-tests" DNS hack

### DIFF
--- a/image-download
+++ b/image-download
@@ -41,6 +41,7 @@ import tempfile
 import time
 import fcntl
 import urllib.parse
+import warnings
 
 from lib.constants import IMAGES_DIR
 from lib.directories import get_images_data_dir, xdg_config_home
@@ -79,6 +80,20 @@ def check_curl_args(args):
 def find(name, stores, latest, quiet):
     found = []
 
+    try:
+        with open(xdg_config_home('cockpit-dev', 'image-store-preference'), 'r') as fp:
+            pattern = fp.read().strip()
+            matches = [store for store in stores if pattern in store]
+            try:
+                # move it to the front of the list
+                preferred, = matches
+                stores.insert(0, preferred)
+            except ValueError:
+                n_matches = len(matches)
+                warnings.warn(f'image-store-preference matched {n_matches} stores (expecting 1).  ignoring.')
+    except FileNotFoundError:
+        preferred = None
+
     for store in stores:
         url = urllib.parse.urlparse(urllib.parse.urljoin(store, name))
 
@@ -88,6 +103,8 @@ def find(name, stores, latest, quiet):
             if output:
                 show_status(quiet, 'present', store, '(authenticated)')
                 found.append(([s3.sign_url(url)], output, store))  # GET
+                if store == preferred:
+                    break
                 continue
 
         # Next, try to access the URL directly, without further help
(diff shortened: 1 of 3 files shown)


allisonkarlitskaya:
Oops.

allisonkarlitskaya:
More of this: `image-prune: Error accessing /repos/cockpit-project/bots/pulls?page=1&per_page=100&state=open&sort=created&direction=desc`

image-download
```diff
@@ -105,28 +104,11 @@ def find(name, stores, latest, quiet):
         if output:
             show_status(quiet, 'present', store)
             found.append((args, output, store))
+            if store == preferred:
+                break
```
martinpitt, line 127: This bit belongs into the other commit.

image-download
```diff
@@ -79,6 +80,20 @@ def check_curl_args(args):
 def find(name, stores, latest, quiet):
     found = []
 
+    try:
+        with open(xdg_config_home('cockpit-dev', 'image-store-preference'), 'r') as fp:
+            pattern = fp.read().strip()
+            matches = [store for store in stores if pattern in store]
```
martinpitt, line 86: Tiniest style issue: This should already be outside of the `with` again, as there is no further action on the file.

image-download
```diff
@@ -79,6 +80,20 @@ def check_curl_args(args):
...
+            try:
+                # move it to the front of the list
+                preferred, = matches
```
martinpitt, line 89: That's not that easy to read, and the comment is odd - here nothing gets "moved", it just picks out the single value out of the list.

Why so strict? Would it hurt to be able to specify "e2e" or "centos" or "linode" and make it select a subset? Then this code here would become simpler, and below you'd just write `if store in preferred` instead of `==`.

martinpitt (changes requested):
Thanks! Dropping the multiple downloads and going back to latency check will come separately? (It conceptually fits in here quite nicely).

I'm debugging the tasks failure in https://github.com/cockpit-project/cockpituous/pull/404 , this is a subtle new race condition from the recent container rebuild. If at all practical, I'd like to fix this first to not lose CI.

image-download
allisonkarlitskaya: I had a `remove()` here that disappeared somewhere...

I'll think about what to do.

image-download
allisonkarlitskaya: Done

image-download
allisonkarlitskaya: Good catch.  Done.

allisonkarlitskaya:
Martin doesn't like this idea.  I'll salvage the `--resolve cockpit-tests` commit in a separate PR.

martinpitt:
FTR, if you need this for your purposes, I'm not vetoing it. I'm just not totally excited by it as image servers come and go. We may move to AWS or another provider, getting rid of obsolete/wrong config files is hard (you need a mode where you ignore wrong matches if the image servers are invalid), and generally that seems like the sort of thing that image-download should figure out by itself.

allisonkarlitskaya:
I think the removal of the AWS mirror plus the return to "fastest ping" support coming in #1801 will get us most of the way to "correct by default", so I'm happy enough to live without this.

PR #1800, opened 2021-03-15
image-{up,down}load: remove "cockpit-tests" hack

allisonkarlitskaya:
We no longer have any mirror that relies on the "cockpit-tests" DNS
trickery.  Remove it.

PR #1801, opened 2021-03-15
image-download: bring back "fastest ping" selection

allisonkarlitskaya:
Functional revert of 2efccf29fab8b237847a0918dc163ba7e4751cd2.

Now that our best mirrors are of similar configuration, switch to the older and much more simple approach of choosing a mirror based on the fastest reply to HEAD, in essence, "fastest ping".

* [x] #1800

### DIFF
--- a/image-download
+++ b/image-download
@@ -71,7 +71,10 @@ def check_curl_args(args):
     head_args = ['--silent', '--head']  # only used for this check
 
     try:
-        return subprocess.check_output(curl_cmd(args + head_args), universal_newlines=True)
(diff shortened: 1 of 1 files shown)


martinpitt (dismissed):
Thanks! #1800 landed. I wonder if you'll need to rebase, or if the github button will be clever enough to DTRT..

martinpitt:
Ah no, conflict-y..

martinpitt (approved):
Thanks! Works great here, it still prefers the linode one.

PR #1814, opened 2021-03-17
s3: add support for v4 request signing

allisonkarlitskaya:
Add support in lib/s3 for "version 4" request signing.  Provide a nice
curl wrapper, since image-upload and image-download are both using curl
to do the work.

Move image-download and image-upload to using the new mechanism.  This
gives us an extra nice advantage of having the server ensure the
checksum of the uploaded file for us.

Port some of the "cli" tools to use v4 signing, plus add one for
performing a "DELETE" request (which isn't possible under the v2
url signing).  We leave "PUT" on the old mechanism, to avoid having to
determine the checksum.

While we're at it, expand the possible filenames in the s3-keys
directories.  This will allow people to have a common key for everything
at linodeobjects.com, but will also allow per-bucket keys as well, both
of which are nice possibilities based on lessons learned through
experimentation since the first iteration of this work.

 * [x] image-refresh cirros

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,17 +91,24 @@ When generating a new personal access token, the scope only needs to
 encompass `public_repo` (or `repo` if you're accessing a private repo).
 
 If you'd like to download Red Hat-only internal images from S3, you'll
-need to create a key file in `~/.config/cockpit-dev/s3-keys/[endpoint]`.
-The contents of this file should be a single line containing the "access
-key" and the "secret key" separated by whitespace.
+need to create a key file in `~/.config/cockpit-dev/s3-keys/[domain]`.
+The `[domain]` can be any non-toplevel domain which contains the S3 URL
+in question.  The contents of this file should be a single line
+containing the "access key" and the "secret key" separated by
+whitespace.
 
-For the currently configured mirrors this means that you'd expect to
-have the following files:
+For the currently configured mirrors this means that you'd likely have the
+following file:
 
+- `~/.config/cockpit-dev/s3-keys/linodeobjects.com`
+
+For more control, you could also use the following:
+
+- `~/.config/cockpit-dev/s3-keys/cockpit-images.eu-central-1.linodeobjects.com`
 - `~/.config/cockpit-dev/s3-keys/eu-central-1.linodeobjects.com`
-- `~/.config/cockpit-dev/s3-keys/us-east-1.linodeobjects.com`
+- either of the above, with `us-east` instead of `eu-central`
 
-each each file would be a single line which looks like
+each file would be a single line which looks like
 
 ```
 EEVIDIDFSOQ0ABJ2LGTT    009rKOypIoqO44Q3VQGRyYPfugi84zANHF0pOW9f
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -4,22 +4,34 @@
 
 # Adapted from examples in
 # https://s3.amazonaws.com/doc/s3-developer-guide/RESTAuthentication.html
+# and
+# https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html
 
 import base64
+import hashlib
 import hmac
 import os.path
 import sys
 import time
 import urllib.parse
+from typing import Dict, List
+
 from .directories import xdg_config_home
 
 __all__ = (
+    "ACL",
     "ACL_PUBLIC",
+    "PUBLIC",
     "is_key_present",
+    "sign_curl",
+    "sign_request",
     "sign_url",
 )
 
-ACL_PUBLIC = 'x-amz-acl:public-read'
+ACL = 'x-amz-acl'
+PUBLIC = 'public-read'
+ACL_PUBLIC = f'{ACL}:{PUBLIC}'
+SHA256_NIL = hashlib.sha256(b'').hexdigest()
 
 
 def get_key(hostname):
@@ -43,6 +55,44 @@ def get_key(hostname):
 def is_key_present(url: urllib.parse.ParseResult) -> bool:
     """Checks if an S3 key is available for the given url"""
     return get_key(url.hostname) is not None
+
+
+def sign_request(url: urllib.parse.ParseResult, method='GET', checksum=SHA256_NIL, headers={}) -> Dict[str, str]:
+    """Signs an AWS request using the AWS4-HMAC-SHA256 algorithm
+
+    Returns a dictionary of extra headers which need to be sent along with the request.
+    If the method is PUT then the checksum of the data to be uploaded must be provided.
+    @headers, if given, are a dict of additional headers to be signed (eg: `x-amz-acl`)
+    """
+    access_key, secret_key = get_key(url.hostname)
+
+    amzdate = time.strftime('%Y%m%dT%H%M%SZ', time.gmtime())
+
+    headers = headers.copy()  # don't modify the user's copy, or the default
+    headers.update({'host': url.hostname, 'x-amz-content-sha256': checksum, 'x-amz-date': amzdate})
+    headers_str = ''.join(f'{k}:{v}\n' for k, v in sorted(headers.items()))
+    headers_list = ';'.join(sorted(headers))
+
+    credential_scope = f'{amzdate[:8]}/any/s3/aws4_request'
+    signing_key = f'AWS4{secret_key}'.encode('ascii')
+    for item in credential_scope.split('/'):
+        signing_key = hmac.new(signing_key, item.encode('ascii'), hashlib.sha256).digest()
+
+    algorithm = 'AWS4-HMAC-SHA256'
+    canonical_request = f'{method}\n{url.path}\n{url.query}\n{headers_str}\n{headers_list}\n{checksum}'
+    request_hash = hashlib.sha256(canonical_request.encode('ascii')).hexdigest()
+    string_to_sign = f'{algorithm}\n{amzdate}\n{credential_scope}\n{request_hash}'
+    signature = hmac.new(signing_key, string_to_sign.encode('ascii'), hashlib.sha256).hexdigest()
+    headers['Authorization'] = f'{algorithm} Credential={access_key}/{credential_scope},' \
+        f'SignedHeaders={headers_list},Signature={signature}'
+
+    return headers
+
+
+def sign_curl(url: urllib.parse.ParseResult, method='GET', checksum=SHA256_NIL, headers={}) -> List[str]:
+    """Same as sign_request() but formats the result as an argument list for curl, including the url"""
+    headers = sign_request(url, method=method, checksum=checksum, headers=headers)
+    return [f'-H{key}:{value}' for key, value in headers.items()] + [url.geturl()]
 
 
 def sign_url(url: urllib.parse.ParseResult, method='GET', headers=[], duration=12 * 60 * 60) -> str:
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -26,7 +26,6 @@ __all__ = (
     "is_key_present",
     "sign_curl",
     "sign_request",
-    "sign_url",
 )
 
 ACL = 'x-amz-acl'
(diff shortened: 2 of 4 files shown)


cockpituous:
image-refresh in progress on 2-cockpit-7.
Log: https://logs.cockpit-project.org/logs/image-refresh-1814-20210317-101313/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20210317-101414

cockpituous:
Log: https://logs.cockpit-project.org/logs/image-refresh-1814-20210317-101515/

cockpituous:
image-refresh rhel-atomic done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-atomic-20210317-103030

cockpituous:
image-refresh in progress on 1-cockpit-11.
Log: https://logs.cockpit-project.org/logs/image-refresh-1814-20210317-104444/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20210317-104444

cockpituous:
image-refresh in progress on 1-cockpit-11.
Log: https://logs.cockpit-project.org/logs/image-refresh-1814-20210317-104545/

cockpituous:
image-refresh rhel-atomic done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-atomic-20210317-105757

cockpituous:
image-refresh in progress on 1-cockpit-11.
Log: https://logs.cockpit-project.org/logs/image-refresh-1814-20210317-105757/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20210317-105858

README.md
```diff
@@ -91,17 +91,21 @@ When generating a new personal access token, the scope only needs to
...
-For the currently configured mirrors this means that you'd expect to
-have the following files:
+For the currently configured mirrors this means that you could have any
+of the following files:
 
+- `~/.config/cockpit-dev/s3-keys/cockpit-images.eu-central-1.linodeobjects.com`
 - `~/.config/cockpit-dev/s3-keys/eu-central-1.linodeobjects.com`
-- `~/.config/cockpit-dev/s3-keys/us-east-1.linodeobjects.com`
+- (either of the above, with `us-east` instead of `eu-central`); or
+- `~/.config/cockpit-dev/s3-keys/linodeobjects.com`
```
martinpitt, line 106: This is really nice! Is there any reason to not make this the recommended variant? I.e. put it first perhaps?

lib/s3.py
```diff
@@ -22,28 +22,27 @@
...
+        except FileNotFoundError:
+            _, _, hostname = hostname.partition('.')  # strip a leading component
 
-def get_key_filename(endpoint):
-    s3_key_dir = xdg_config_home('cockpit-dev/s3-keys', envvar='COCKPIT_S3_KEY_DIR')
-    return os.path.join(s3_key_dir, endpoint)
+    return None
 
 
 def is_key_present(url):
```
martinpitt, line 49: This function is now redundant -- you can call get_key() directly and interpret the value as boolean.

lib/s3.py
```diff
@@ -40,6 +50,37 @@ def is_key_present(url):
     return get_key(url.hostname) is not None
 
 
+def sign_request(url, method='GET', checksum=SHA256_NIL, headers={}, region='any', service='S3'):
```
martinpitt, line 53: This could do with a docstring. Particularly, what does the checksum mean (it's the checksum of what exactly?) Does it really make sense to default to NIL here, instead of making it explicit in the caller what's going on?

What other "service" values are supported? Given that this is an "S3" library, perhaps we should start with not  having this parameter until we actually need it?

Plus, document the return value.

lib/s3.py
```diff
@@ -40,6 +50,37 @@ def is_key_present(url):
...
+    access_key, secret_key = get_key(url.hostname)
```
martinpitt, line 54: This would `ValueError` if get_key returns None, i.e. there is no key. I suppose that's ok, but the docstring should maybe point out that this needs to be asserted before? In the spirit of robustness, perhaps raise a more useful error than "not enough values to unpack"?

lib/s3.py
```diff
@@ -40,6 +50,37 @@ def is_key_present(url):
...
+    headers_str = ''.join(f'{k}:{v}\n' for k, v in sorted(headers.items()))
+    headers_list = ';'.join(sorted(headers))
+
+    credential_scope = f'{amzdate[:8]}/{region}/{service}/aws4_request'
+    signing_key = f'AWS4{secret_key}'.encode('ascii')
+    for item in credential_scope.split('/'):
+        signing_key = hmac.new(signing_key, item.encode('ascii'), hashlib.sha256).digest()
+
+    algorithm = 'AWS4-HMAC-SHA256'
+    canonical_request = f'{method}\n{url.path}\n{url.query}\n{headers_str}\n{headers_list}\n{checksum}'
```
martinpitt, line 69: Not your "fault" of course, but can you think of any reason why the list of header names appears here? It's already contained in the full headers_str, after all. It's mentioned in [the docs](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html), but not without much rationale.

lib/s3.py
```diff
@@ -40,6 +50,37 @@ def is_key_present(url):
...
+    request_hash = hashlib.sha256(canonical_request.encode('ascii')).hexdigest()
+    string_to_sign = f'{algorithm}\n{amzdate}\n{credential_scope}\n{request_hash}'
+    signature = hmac.new(signing_key, string_to_sign.encode('ascii'), hashlib.sha256).hexdigest()
+    headers['Authorization'] = f'{algorithm} Credential={access_key}/{credential_scope},' \
+        f'SignedHeaders={headers_list},Signature={signature}'
+
+    return headers
+
+
+def sign_curl(url, **kwargs):
```
martinpitt, line 79: This also needs a docstring, and perhaps a better name. This returns a partial CLI argv, and `url` is not really an URL string, it's a parsed object. kwargs needs to be documented, too, and I think it'd be safer to explicitly enumerate these.

lib/s3.py
```diff
@@ -4,8 +4,11 @@
 
 # Adapted from examples in
 # https://s3.amazonaws.com/doc/s3-developer-guide/RESTAuthentication.html
+# and
+# https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html
```
martinpitt, line 8: This is all rather involved, and took you quite some time to figure out. Did you look into existing libraries like [s3](https://pypi.org/project/s3/) for that?

lib/s3.py
```diff
@@ -97,18 +98,24 @@ def sign_url(url, method='GET', headers=[], duration=12 * 60 * 60):
...
 
     url = urllib.parse.urlparse(uri)
     if cmd == 'get':
-        print(sign_url(url))
+        args = sign_curl(url)
+    elif cmd == 'rm':
+        args = ['-XDELETE'] + sign_curl(url, method='DELETE')
     elif cmd == 'put':
-        print(sign_url(url, method='PUT'))
+        args = [sign_url(url, method='PUT')]
```
martinpitt, line 109: How important is this CLI? Keeping a whole second and outdated signature method in the lib just for this seems not very desirable?

martinpitt (changes requested):
Nice, thanks! That should pave the way for image pruning.

README.md
allisonkarlitskaya: It would make the ordering weird with respect to the "either of the above".  I'll figure out a way to rewrite it.

lib/s3.py
allisonkarlitskaya: It's true, but that exposes the key on the API.  I'd prefer not to do that...

(I know there's no real protection here, and that this is not really an "API" in any meaningful sense, but still...)

lib/s3.py
allisonkarlitskaya: The plan is that this would only ever be called after we checked for a key.  I'll document that precondition, rather than change the code here.

On the other hand, this will indeed fail very ungracefully for an invalidly-formatted key file.  I'll fix that.

lib/s3.py
allisonkarlitskaya: Yes: The headers_str is part of what gets hashed, which means that it arrives at the server completely opaque.  The headers_list is sent in the plain to help the server decide which headers it should hash on its side, to ensure that we get the same result on both ends.

There are some parts of v4 signing which I find obnoxious (like the needlessly-complicated key derivation stuff) but this particular feature of making the list of headers explicit is something I consider to be a dramatic improvement.

lib/s3.py
allisonkarlitskaya: OK.

lib/s3.py
allisonkarlitskaya: In the end I ended up adding some debugging output to s3cmd to get some hints about what I was doing wrong, just to see how the hashed values were ending up different.  That was already at the "minor adjustments" stage, though, and ended up finding some really really dumb problems (like I was spelling "linodeobjets.com" instead of "linodeobjects.com").

lib/s3.py
allisonkarlitskaya: I've been finding it pretty useful to be honest, for the purposes of testing.  The main thing that it has over other s3 tools is that it works with our key files in the same way as the other tools.

I'd prefer to keep it.  I will probably eventually port the "put" stuff to include checksums, but right now it's not on the critical path for me...

README.md
martinpitt: I meant, would you actually recommend that "we" (humans + CI) use this  particular name? It seems intuitive to me to name it by provider, as the credentials are valid for all regions.

allisonkarlitskaya: It's not always true.  Our upload credentials are tied to the region, for example.

allisonkarlitskaya:
I'd appreciate doing another round of review (and possible changes) before doing another round of testing with the bots and image refreshes.

lib/s3.py
```diff
@@ -50,11 +51,18 @@ def get_key(hostname):
...
 
 
-def is_key_present(url):
+def is_key_present(url: urllib.parse.ParseResult):
+    """Checks if an S3 key is available for the given url"""
     return get_key(url.hostname) is not None
 
 
-def sign_request(url, method='GET', checksum=SHA256_NIL, headers={}, region='any', service='S3'):
+def sign_request(url: urllib.parse.ParseResult, method='GET', checksum=SHA256_NIL, headers={}) -> Dict[str, str]:
```
martinpitt, line 59: Please squash the "s3: annotate and document the APIs" commit. It looks rather odd otherwise.  Also, changing "S3" to "s3" seems unrelated.

martinpitt (changes requested):
Cheers! LGTM now, with some squashing. I'd say, do another round with cirros and rhel-atomic, and then let's land this.

lib/s3.py
allisonkarlitskaya: Sure.

To be honest, I feel like Linode is completely ignoring both the region and the service values.  I changed this to "s3" just because I found an official example which contained it as lowercase...

cockpituous:
image-refresh in progress on 4-ci-srv-02.
Log: https://logs.cockpit-project.org/logs/image-refresh-1814-20210318-105555/

cockpituous:
image-refresh in progress on 3-ci-srv-02.
Log: https://logs.cockpit-project.org/logs/image-refresh-1814-20210318-105555/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20210318-105555

cockpituous:
image-refresh in progress on 4-ci-srv-05.
Log: https://logs.cockpit-project.org/logs/image-refresh-1814-20210318-105555/

martinpitt (approved):
Danke!

cockpituous:
image-refresh rhel-atomic done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-atomic-20210318-113333

allisonkarlitskaya:
- wget works with the cirros image
 - wget doesn't work with the rhel image
 - `python3 -m lib.s3 get ...` works with the rhel image

PR #1817, opened 2021-03-18
Image prune cleanup

allisonkarlitskaya:


image-prune
```diff
@@ -152,34 +140,74 @@ def prune_images(force, dryrun, quiet=False, open_pull_requests=True, offline=Fa
...
-            continue
-        if entry.is_file() and \
-                any(entry.name.endswith(ext) for ext in ['.qcow2', '.partial']) and \
-                entry.name not in targets:
-            if not quiet or dryrun:
-                sys.stderr.write("Pruning {0}\n".format(entry.name))
+    return targets
+
+
+class ImageCache:
```
martinpitt, line 146: This class should have abstract methods that are implemented by subclasses, such as `def delete_images()`, which raise a `NotImplementedError`. That will clarify and document what concrete subclasses need to implement.

Honestly, I find this whole OOification rather excessive here, but if you want to go through with it, let's do it properly.

image-prune
```diff
@@ -152,34 +140,74 @@ def prune_images(force, dryrun, quiet=False, open_pull_requests=True, offline=Fa
...
-                os.unlink(entry.path)
+                self.delete_image(image)
+
+        self.cleanup(quiet, dryrun)
+
+
+class LocalImageDirectory(ImageCache):
+    def __init__(self, directory=None):
+        self.directory = directory or get_images_data_dir()
+        self.space_threshold = float(os.environ.get("PRUNE_THRESHOLD_G", 15))
```
martinpitt, line 175: btw, if it helps to simplify things for S3: Feel free to remove the entire "check free disk space" and that env var. These days, the solution for "we don't have enough disk space for our purposes" is "turn the diskspace knob".

image-prune
```diff
@@ -152,34 +140,74 @@ def prune_images(force, dryrun, quiet=False, open_pull_requests=True, offline=Fa
...
+    def delete_file(self, filename):
+        os.unlink(os.path.join(self.directory, filename))
+
+    def enough_space(self):
+        st = os.statvfs(self.directory)
+        free = st.f_bavail * st.f_frsize / (1024 * 1024 * 1024)
+        return free >= self.space_threshold
+
+    def cleanup(self, quiet, dryrun):
+        # clean any broken symlinks in IMAGES_DIR
```
martinpitt, line 191: I'm not sure why we need that code still -- the image cache dir *never* gets symlinks. The links are maintained in the bots.git. The task containers clean up their bots checkout for each iteration anyway, and if you want to clean up your local checkout, then `git clean -fdx` is a much better way IMHO. So from my POV, feel free to toss this.

martinpitt (changes requested):
Thanks!

image-prune
```diff
@@ -150,7 +150,7 @@ def prune_images(force, dryrun, quiet=False, open_pull_requests=True, offline=Fa
         path = os.path.join(IMAGES_DIR, filename)
 
         # only consider original image entries as trustworthy sources and ignore non-links
-        if path.endswith(".qcow2") or path.endswith(".partial") or not os.path.islink(path):
+        if '.' in path or not os.path.islink(path):
```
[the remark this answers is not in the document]
allisonkarlitskaya: We do have links with dots: the ones that link fedora-33-abcdef123456.qcow2 to the same filename in the cache directory.

That's what the business about "original images" means.

image-prune
allisonkarlitskaya: I read about this in depth last night and it seems that opinion is a bit split on the correct thing to do here.  If you have a preference, then I'm OK to do that.

As for OO: I was implementing the S3 stuff as a skeleton with this at the same time and it turned into a bunch of ugly cases of if S3: do_this() else: do_that() for the listing, the space determination and the deleting.  This screams for virtual methods.

image-prune
allisonkarlitskaya: I actually want to implement the disk space thing for S3: it can help us make sure the stores stay under a given fixed size.

image-prune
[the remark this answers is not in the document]
martinpitt: Ah sorry, right -- we don't commit these, but they are still generated by image-download (?).

image-prune
allisonkarlitskaya: This is indeed about the local working directory, but it's kinda incomplete: for example, it won't clean up other worktrees, or checkouts in other places.  Even as I was doing this I was thinking "this needs a separate tool...".

I'll drop it for now.

image-prune
martinpitt: Right, as you have the non-OO comparison with the S3 part, I trust that this looks nicer.

image-prune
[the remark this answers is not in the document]
allisonkarlitskaya: Correct.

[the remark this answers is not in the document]
martinpitt: So that is probably the reason why we can't have an image like "rhel-8.5" :-)

martinpitt (approved):
Looks great to me now. I tested this locally in various scenarios, and it did correctly remove old photos from ancient class mates whom I never liked anyway.

PR #1819, opened 2021-03-18
testmap: Create a new entry for cockpit-machines repository

KKoukiou:


lib/testmap.py
```diff
@@ -84,6 +84,24 @@
...
+            'ubuntu-stable',
+            'fedora-33',
+            'fedora-34',
+            'fedora-33/firefox',
+            'rhel-8-4',
+            'centos-8-stream',
+        ],
+        '_manual': [
+            'fedora-testing',
+            'fedora-testing/dnf-copr',
```
martinpitt, line 101: I suppose you won't need that (but it doesn't hurt either, I guess).

martinpitt (commented):
Seems fine to enable them for "master" right away, as this isn't blocking anything yet.

On that note, are you sure you don't want to start with "main" right away? I thought that was the new default for new projects.

martinpitt (approved):

PR #1820, opened 2021-03-18
image-prune S3 support

allisonkarlitskaya:
* [x] #1817

### DIFF
--- a/image-prune
+++ b/image-prune
@@ -34,8 +34,6 @@ from task import github
 from lib.constants import IMAGES_DIR
 from lib.directories import get_images_data_dir
 
-# threshold in G below which unreferenced qcow2 images will be pruned, even if they aren't old
-
 
 def get_refs(open_pull_requests=True, offline=False):
     """Return dictionary for available refs of the format {'rhel-7.4': 'ad50328990e44c22501bd5e454746d4b5e561b7c'}
--- a/image-prune
+++ b/image-prune
@@ -21,11 +21,13 @@
 IMAGE_EXPIRE = 14
 
 import argparse
+import datetime
 import logging
 import os
 import subprocess
 import sys
 import time
+import urllib.parse
 from typing import Iterable, Tuple
 from logging import debug
 
@@ -33,6 +35,7 @@ from task import github
 
 from lib.constants import IMAGES_DIR
 from lib.directories import get_images_data_dir
+from lib import s3
 
 
 def get_refs(open_pull_requests=True, offline=False):
@@ -197,6 +200,32 @@ class LocalImageDirectory(ImageCache):
         return free >= self.space_threshold
 
 
+class S3ImageStore(ImageCache):
+    def __init__(self, url):
+        self.url = url
+        # A bit magic: we have 2 buckets and 250GB quota.  Try to keep each bucket below ca. 80GB.
+        self.max_bytes = float(os.environ.get("S3_IMAGES_MAX_GB", 80)) * 1000 * 1000 * 1000
+
+    def list_files(self):
+        result = s3.list_bucket(self.url)
+
+        # make sure this gets done before we start iterating
+        self.sizes = dict(s3.parse_list(result, "Key", "Size"))
+
+        RFC3339ish = '%Y-%m-%dT%H:%M:%S.%f%z'
+        for name, stamp in s3.parse_list(result, "Key", "LastModified"):
+            yield name, datetime.datetime.strptime(stamp, RFC3339ish).timestamp()
+
+    def delete_file(self, filename):
+        with s3.urlopen(self.url._replace(path='/' + filename), method='DELETE'):
+            pass
+        del self.sizes[filename]
+
+    def enough_space(self):
+        total = sum(int(bytes) for bytes in self.sizes.values())
+        return total < self.max_bytes
+
+
 def main():
     parser = argparse.ArgumentParser(description='Prune downloaded images')
     parser.add_argument("--debug", action="store_true", help="Enable debugging output")
@@ -210,12 +239,27 @@ def main():
                         help="Consider neither pull requests on GitHub nor branches, only look at the current checkout")
     parser.add_argument("-o", "--offline", dest="offline", action="store_true",
                         help="Don't access external sources such as GitHub")
+    cache_arg = parser.add_argument_group(title='Cache location',
+                                          description='default: --directory ~/.cache/cockpit-images')
+    cache_arg = cache_arg.add_mutually_exclusive_group()
+    cache_arg.add_argument('--directory', metavar='DIR', help="images cache directory to prune")
+    cache_arg.add_argument('--s3', metavar='URL', help="url of S3 store to prune", type=urllib.parse.urlparse)
     args = parser.parse_args()
 
     if args.debug:
         logging.basicConfig(level=logging.DEBUG)
 
-    collection = LocalImageDirectory()
+    if args.s3:
+        if args.s3.scheme != 'https':
+            sys.exit("S3 args must be https")
+
+        if not s3.is_key_present(args.s3):
+            sys.exit("No key found for {args.s3.hostname}")
+
+        collection = S3ImageStore(args.s3)
+
+    else:
+        collection = LocalImageDirectory(args.directory)
 
     keepers = get_keepers(quiet=args.quiet,
                           open_pull_requests=(not args.branches_only),
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -15,7 +15,9 @@ import shlex
 import sys
 import time
 import urllib.parse
-from typing import Dict, List
+import urllib.request
+from typing import Dict, List, Iterable, IO
+import xml.etree.ElementTree as ET
 
 from .directories import xdg_config_home
 
@@ -24,8 +26,11 @@ __all__ = (
     "ACL_PUBLIC",
     "PUBLIC",
     "is_key_present",
+    "list_bucket",
+    "parse_list",
     "sign_curl",
     "sign_request",
+    "urlopen",
 )
 
 ACL = 'x-amz-acl'
@@ -95,6 +100,27 @@ def sign_curl(url: urllib.parse.ParseResult, method='GET', checksum=SHA256_NIL, 
     return [f'-H{key}:{value}' for key, value in headers.items()] + [url.geturl()]
 
 
+def urlopen(url: urllib.parse.ParseResult, method='GET') -> IO:
+    """Same as sign_request() but calls urlopen() on the result"""
+    headers = sign_request(url, method=method)
+    request = urllib.request.Request(url.geturl(), headers=headers, method=method)
+    return urllib.request.urlopen(request)
+
+
+def list_bucket(url: urllib.parse.ParseResult) -> ET:
+    """Get the ListBucketResult as a xml.etree.ElementTree"""
+    with urlopen(url) as response:
+        return ET.fromstring(response.read())
+
+
+def parse_list(result: ET, *keys) -> Iterable[Iterable[str]]:
+    """For each item in the bucket, return the given keys"""
+    # 'http' url is API: see https://doc.s3.amazonaws.com/2006-03-01/AmazonS3.wsdl
+    xmlns = {'s3': 'http://s3.amazonaws.com/doc/2006-03-01/'}
+    for child in result.findall('s3:Contents', xmlns):
+        yield (child.find(f's3:{key}', xmlns).text for key in keys)
+
+
 def sign_url(url: urllib.parse.ParseResult, method='GET', headers=[], duration=12 * 60 * 60) -> str:
     """Returns a "pre-signed" url for the given method and headers"""
     access, secret = get_key(url.hostname)
@@ -118,6 +144,10 @@ def main():
     url = urllib.parse.urlparse(uri)
     if cmd == 'get':
         args = sign_curl(url)
+    elif cmd == 'ls':
+        for items in parse_list(list_bucket(url), "Size", "LastModified", "Key"):
+            print('\t'.join(items))
+        sys.exit(0)
     elif cmd == 'rm':
         args = ['-XDELETE'] + sign_curl(url, method='DELETE')
     elif cmd == 'put':
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -142,6 +142,10 @@ def main():
     prognam, cmd, uri = sys.argv
 
     url = urllib.parse.urlparse(uri)
+
+    if not is_key_present(url):
+        sys.exit(f'no key is available for {url.hostname}')
+
     if cmd == 'get':
         args = sign_curl(url)
     elif cmd == 'ls':


lib/s3.py
```diff
@@ -95,6 +100,26 @@ def sign_curl(url: urllib.parse.ParseResult, method='GET', checksum=SHA256_NIL,
...
 
 
+def urlopen(url: urllib.parse.ParseResult, method='GET') -> IO:
+    """Same as sign_request() but calls urlopen() on the result"""
+    headers = sign_request(url, method=method)
+    request = urllib.request.Request(url.geturl(), headers=headers, method=method)
+    return urllib.request.urlopen(request)
+
+
+def list_bucket(url: urllib.parse.ParseResult) -> ET:
```
martinpitt, line 110: The code looks ok, but from the perspective of this commit it's dead code. I suggest adding a "list bucket contents" mode to main() here? That will both exercise the code, and moreover will be very useful for maintaining/inspecting the bucket.

lib/s3.py
```diff
@@ -95,6 +100,26 @@ def sign_curl(url: urllib.parse.ParseResult, method='GET', checksum=SHA256_NIL,
...
+    """Get the ListBucketResult as a xml.etree.ElementTree"""
+    with urlopen(url) as response:
+        return ET.fromstring(response.read())
+
+
+def parse_list(result: ET, *keys) -> Iterable[Iterable[str]]:
+    """For each item in the bucket, return the given keys"""
+    ns = {'s3': 'http://s3.amazonaws.com/doc/2006-03-01/'}
```
martinpitt, line 118: That.. needs an explanation. Without authentication, this URL is "access denied", and if it's authenticated then the http:// isn't exactly inspiring trust. Can you please add a comment?

image-prune
```diff
@@ -199,6 +202,30 @@ class LocalImageDirectory(ImageCache):
         return free >= self.space_threshold
 
 
+class S3ImageStore(ImageCache):
+    def __init__(self, url, max_gb=80):
```
martinpitt, line 206: TBH this feels a bit too arbitrary and too well hidden in the code. Can you please at least put it as a `s3.IMAGES_MAX_GB` constant at the top of the file or into lib/s3.py with a comment? This makes assumptions about how many images we have at any given time. It should be fine, but I'd like to make it a bit more explicit.

I doesn't really impedance-match what the local stores are doing with the "remaining space pressure", but as S3 works differently I dont' have a better idea either.

image-prune
```diff
@@ -199,6 +202,30 @@ class LocalImageDirectory(ImageCache):
...
+
+    def list_files(self):
+        result = s3.list_bucket(self.url)
+
+        # make sure this gets done before we start iterating
+        self.sizes = dict(s3.parse_list(result, "Key", "Size"))
+
+        RFC3339ish = '%Y-%m-%dT%H:%M:%S.%f%z'
+        for name, stamp in s3.parse_list(result, "Key", "LastModified"):
+            yield name, datetime.datetime.strptime(stamp, RFC3339ish).timestamp()
```
martinpitt, line 218: Does it has to be that specific, or could it just use datetime.isoformat()?

image-prune
```diff
@@ -199,6 +202,30 @@ class LocalImageDirectory(ImageCache):
...
+
+    def delete_file(self, filename):
+        with s3.urlopen(self.url._replace(path=filename), method='DELETE') as _:
+            del self.sizes[filename]
```
martinpitt, line 222: That's certainly a rather curious syntax. I take it you take the `with` just to get an automatic closing? I suppose putting `pass` into the `with` and having the `del` separate is a bit cleaner. But feel free to ignore, that's just bikeshedding.

image-prune
```diff
@@ -212,12 +239,27 @@ def main():
...
+    cache_arg.add_argument('--s3', metavar='URL', help="url of S3 store to prune", type=urllib.parse.urlparse)
     args = parser.parse_args()
 
     if args.debug:
         logging.basicConfig(level=logging.DEBUG)
 
-    collection = LocalImageDirectory()
+    if args.s3:
+        if args.s3.scheme != 'https':
+            sys.exit("S3 args must be https")
```
martinpitt, line 254: TIL :+1:

martinpitt (changes requested):
Thanks! 

FTR, we can start hooking this into https://github.com/cockpit-project/cockpituous/blob/master/tasks/cockpit-tasks#L67 . This avoids handing over the s3 secrets to github. I realize that at some point we may want to do that anyway, but as long as we still have the "classic" image stores it fits better into the task containers.

lib/s3.py
allisonkarlitskaya: That's just an XML namespace.  I'll add a comment.

image-prune
allisonkarlitskaya: `datetime.isoformat()` is only for printing.  I did quite some searching and there's no good python standard library function for ISO8601/3339 *parsing*.

image-prune
allisonkarlitskaya: I'm happy to do that.

Also: I learned, meanwhile, that `as _` is optional, and if you exclude it, then the expected closing will happen.

image-prune
allisonkarlitskaya: Sure.

The issue here is that S3 has no reasonable upper bounds on storage space: we use more we pay more.  Therefore "free space" is kinda meaningless.  It gets even more confusing when we consider that the 250GB included in the base price is split between all of our configured buckets, but we only have a view on a single bucket at a time.

The chosen value of 80 corresponds to the fact that we have two buckets, and keeps us very comfortably below our 250GB "included in base price" quota.  We could probably even make it 100, but that's bikeshedding.

image-prune
martinpitt: Argh, "p", not "f", sorry!

image-prune
allisonkarlitskaya: I changed it around to match what we do with the local prune threshold (envvar) and added a comment explaining the logic.  I'd prefer not to add a constant to the top of the file because it's S3-specific, and I'd prefer not to add a constant to `lib.s3` because it's prune-specific.

martinpitt: Right, that was just thinking aloud. That's what I meant with "S3 works differently". The approach is fine, it's just a bit hard to see when you look through the code. Personally I prefer having such magic tweakable constants at the top of the files, and function default parameters shoudl just be something like `=None` or `={}`.

martinpitt:
Code changes look good. Some nitpicks. I only installed a key for eu-central. If I try this with a nonexisting key/region, I get an unsightly error:

```
❱❱❱ python3 -m lib.s3 ls https://cockpit-images.us-east-1.linodeobjects.com/
  File "/usr/lib64/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib64/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/var/home/martin/upstream/bots/lib/s3.py", line 165, in <module>
    main()
  File "/var/home/martin/upstream/bots/lib/s3.py", line 148, in main
    for items in parse_list(list_bucket(url), "Size", "LastModified", "Key"):
  File "/var/home/martin/upstream/bots/lib/s3.py", line 112, in list_bucket
    with urlopen(url) as response:
  File "/var/home/martin/upstream/bots/lib/s3.py", line 105, in urlopen
    headers = sign_request(url, method=method)
  File "/var/home/martin/upstream/bots/lib/s3.py", line 72, in sign_request
    access_key, secret_key = get_key(url.hostname)
TypeError: cannot unpack non-iterable NoneType object
```

(Dir listing works fine for eu-central).

`./image-prune -n --s3 https://cockpit-images.eu-central-1.linodeobjects.com/` does not want to prune anything, which seems right given that it's so new. With `--force` it looks reasonable:

```
./image-prune --force -n --s3 https://cockpit-images.eu-central-1.linodeobjects.com/ 
From https://github.com/cockpit-project/bots
 * branch            feb4fdd46749e7ce247bdadaf20209765a9184d9 -> FETCH_HEAD
From https://github.com/cockpit-project/bots
Considering images from pull request #1790 (images: Mirror openshift containers on quay.io) (feb4fdd46749e7ce247bdadaf20209765a9184d9)
Considering images from /image-refresh-openshift-20210311-180505 (a0adca70e2e7ee27c0eb32a92884c9f89475f90a)
Considering images from /images-openshifting-to-quay (feb4fdd46749e7ce247bdadaf20209765a9184d9)
Considering images from /master (65d8544f8591c69d169cfb1f788e194641f31408)
Considering images from /pixel-testing (6ab1ecd2307dddd2144c16f4ca8abf9fd8b97e1a)
Considering images from /prune-s3 (0831d177d69899fed4c97c849a70f4d14f985bb0)
Pruning cirros-50c4862e81d69f561f0aabeebe8ffcaf9b10aa4f3b37803f74026d223d3ea55a.qcow2
Pruning cirros-c881ff7195b5ab063fecabf1d2fd8102c0d9ca8d6169938c766ba679c9e6fb08.qcow2
Pruning cirros-cd6ec0f961356a1b888a08e028dee689db60050f4f4c1e2d0739c0c3b7cc8533.qcow2
Pruning cirros-d3d4efd92822be4259d4c9973dc1f964c8397bc4b0f5a9685c510accb8e9464b.qcow2
Pruning fedora-33-4c7b0a3a0bc74da766e4d36d526c225504042d27bb00a7318e92dad95dfd98fc.qcow2
Pruning fedora-testing-1862f1f40d9fb0cbd12a379696954edd60e01eb19c1dd1d61c1f8fe743ca5849.qcow2
Pruning rhel-atomic-1d0c2fc676a11e52b5d932e6a5c00618f11b6d4144617949d15db2e7bf3ce54d.qcow2
Pruning rhel-atomic-4a9cf1e46f85a33ef09f0aa89491363e2f129e371352af5981c03d70f117159a.qcow2
Pruning rhel-atomic-6c3a883a6535cf1e538e87cab4a9abc2de9a14326ecdaec0ea2c9a944c583b1a.qcow2
Pruning services-60e27ce47478d6d34aae96f766d1cc423f3c56e9ca3e3e36c58019c0d8cfb270.qcow2
```
These are indeed all obsolete images which never got committed or were obsoleted by a refresh. So I took the plunge, removed `-n`, and ran into this:
```
m https://github.com/cockpit-project/bots
 * branch            feb4fdd46749e7ce247bdadaf20209765a9184d9 -> FETCH_HEAD
From https://github.com/cockpit-project/bots
Considering images from pull request #1790 (images: Mirror openshift containers on quay.io) (feb4fdd46749e7ce247bdadaf20209765a9184d9)
Considering images from /image-refresh-openshift-20210311-180505 (a0adca70e2e7ee27c0eb32a92884c9f89475f90a)
Considering images from /images-openshifting-to-quay (feb4fdd46749e7ce247bdadaf20209765a9184d9)
Considering images from /master (65d8544f8591c69d169cfb1f788e194641f31408)
Considering images from /pixel-testing (6ab1ecd2307dddd2144c16f4ca8abf9fd8b97e1a)
Considering images from /prune-s3 (0831d177d69899fed4c97c849a70f4d14f985bb0)
Pruning cirros-50c4862e81d69f561f0aabeebe8ffcaf9b10aa4f3b37803f74026d223d3ea55a.qcow2
  File "/var/home/martin/upstream/bots/./image-prune", line 278, in <module>
    sys.exit(main())
  File "/var/home/martin/upstream/bots/./image-prune", line 269, in main
    collection.prune(keepers,
  File "/var/home/martin/upstream/bots/./image-prune", line 181, in prune
    self.delete_file(image)
  File "/var/home/martin/upstream/bots/./image-prune", line 220, in delete_file
    with s3.urlopen(self.url._replace(path=filename), method='DELETE'):
  File "/var/home/martin/upstream/bots/lib/s3.py", line 107, in urlopen
    return urllib.request.urlopen(request)
  File "/usr/lib64/python3.9/urllib/request.py", line 214, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python3.9/urllib/request.py", line 523, in open
    response = meth(req, response)
  File "/usr/lib64/python3.9/urllib/request.py", line 632, in http_response
    response = self.parent.error(
  File "/usr/lib64/python3.9/urllib/request.py", line 561, in error
    return self._call_chain(*args)
  File "/usr/lib64/python3.9/urllib/request.py", line 494, in _call_chain
    result = func(*args)
  File "/usr/lib64/python3.9/urllib/request.py", line 641, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 403: Forbidden
```
That may be intended -- my key has a too weak fu to be able to prune images.

So I went ahead and tried that in a tasks container with their credentials. That fails with the same  error.

So I think in order to actually deploy/test this, we need to make the bots' s3 token :heavy_plus_sign: :muscle:  ?

allisonkarlitskaya:
The problem was more subtle: I was performing a dubious operation: replacing the path with one that didn't start with the leading slash.  I checked, and from `.geturl()` we still got the slash correctly inserted, but when accessing the `.path` directly from the signing function, it didn't get the slash included.  That was enough to mess up the signature, and cause the 403.

It should be working now.

martinpitt:
Thanks! Confirming that the message for missing key now works correctly. I also ran the `--force` on a bot, it pruned a correctly-looking set of images:

```
Pruning cirros-c881ff7195b5ab063fecabf1d2fd8102c0d9ca8d6169938c766ba679c9e6fb08.qcow2
Pruning cirros-cd6ec0f961356a1b888a08e028dee689db60050f4f4c1e2d0739c0c3b7cc8533.qcow2
Pruning cirros-d3d4efd92822be4259d4c9973dc1f964c8397bc4b0f5a9685c510accb8e9464b.qcow2
Pruning fedora-33-4c7b0a3a0bc74da766e4d36d526c225504042d27bb00a7318e92dad95dfd98fc.qcow2
Pruning fedora-testing-1862f1f40d9fb0cbd12a379696954edd60e01eb19c1dd1d61c1f8fe743ca5849.qcow2
Pruning rhel-atomic-1d0c2fc676a11e52b5d932e6a5c00618f11b6d4144617949d15db2e7bf3ce54d.qcow2
Pruning rhel-atomic-4a9cf1e46f85a33ef09f0aa89491363e2f129e371352af5981c03d70f117159a.qcow2
Pruning rhel-atomic-6c3a883a6535cf1e538e87cab4a9abc2de9a14326ecdaec0ea2c9a944c583b1a.qcow2
Pruning services-60e27ce47478d6d34aae96f766d1cc423f3c56e9ca3e3e36c58019c0d8cfb270.qcow2
```

and a before/after diff of `python3 -m lib.s3 ls https://cockpit-images.eu-central-1.linodeobjects.com/` looks as expected:
```
--- /tmp/prev	2021-03-19 06:03:39.070462326 +0100
+++ /tmp/after	2021-03-19 06:05:22.046896122 +0100
@@ -5,3 +4,0 @@
-17118208	2021-03-17T10:14:03.992Z	cirros-c881ff7195b5ab063fecabf1d2fd8102c0d9ca8d6169938c766ba679c9e6fb08.qcow2
-17118208	2021-03-18T10:55:23.140Z	cirros-cd6ec0f961356a1b888a08e028dee689db60050f4f4c1e2d0739c0c3b7cc8533.qcow2
-17118208	2021-03-17T10:58:05.717Z	cirros-d3d4efd92822be4259d4c9973dc1f964c8397bc4b0f5a9685c510accb8e9464b.qcow2
@@ -14 +10,0 @@
-1766358016	2021-03-11T14:49:04.550Z	fedora-33-4c7b0a3a0bc74da766e4d36d526c225504042d27bb00a7318e92dad95dfd98fc.qcow2
@@ -19 +14,0 @@
-1603394048	2021-03-12T08:51:13.120Z	fedora-testing-1862f1f40d9fb0cbd12a379696954edd60e01eb19c1dd1d61c1f8fe743ca5849.qcow2
@@ -29,2 +23,0 @@
-1991392768	2021-03-18T11:30:21.951Z	rhel-atomic-1d0c2fc676a11e52b5d932e6a5c00618f11b6d4144617949d15db2e7bf3ce54d.qcow2
-1991576064	2021-03-17T10:27:15.429Z	rhel-atomic-4a9cf1e46f85a33ef09f0aa89491363e2f129e371352af5981c03d70f117159a.qcow2
@@ -32 +24,0 @@
-1991772672	2021-03-17T10:55:03.993Z	rhel-atomic-6c3a883a6535cf1e538e87cab4a9abc2de9a14326ecdaec0ea2c9a944c583b1a.qcow2
@@ -35 +26,0 @@
-2653189632	2021-03-16T18:27:41.822Z	services-60e27ce47478d6d34aae96f766d1cc423f3c56e9ca3e3e36c58019c0d8cfb270.qcow2
```

Note that I left the US mirror as-is, so that still has the obsolete images.

Thanks!

martinpitt (approved):

PR #1824, opened 2021-03-19
prometheus-stats: Export queue time histogram of last 12 hours

martinpitt:
Looking back just one hour would give us zeros most of the time, because
we usually don't land several PRs within one hour, and there is
apparently some time zone shift. So extend the lookback time to 12
hours.

Also, we are not really interested in the average queue time, but how
many tests (absolute and percentage) have to wait longer than 5 mins
(our SLO) or 1 hour (large congestion), or longer (infrastructure is
down). So export the times in three buckets as a histogram.

PR #1831, opened 2021-03-22
Fix prometheus-stats

martinpitt:
I was wondering why our stats [did not get imported](http://prometheus.osci.redhat.com/graph?g0.range_input=1h&g0.expr=%7Bjob%3D~%22cockpit%22%7D&g0.tab=0), and noticed a stray `"` in https://images-frontdoor.apps.ocp.ci.centos.org/prometheus .

This hopefully fixes it.


marusak (approved):
Dammit I saw that stray `"` in https://github.com/cockpit-project/bots/pull/1824 already but went to read about how histograms should be exported and completely forgot to mention that in the review.

PR #1833, opened 2021-03-22
prometheus-stats: Export unstable tests

martinpitt:
Add a new top_failures_pct{project,test} metric for the topmost tests
that fail more than 10% of the time. These represent SLO violations, and
we want to expose them on Grafana.

This is still a bit experimental -- in particular, this does not yet
expose example URLs for further investigation.

---

```
❱❱❱ ./prometheus-stats --db test-results.db 
# HELP queue_time_wait_seconds histogram of queue wait times in the last 12 hours
# TYPE queue_time_wait_seconds histogram
queue_time_wait_seconds_bucket{le="300"} 2
queue_time_wait_seconds_bucket{le="3600"} 2
queue_time_wait_seconds_bucket{le="+Inf"} 2
queue_time_wait_seconds_sum 30
queue_time_wait_seconds_count 2

# HELP top_failures_pct Most unstable tests in the last 7 days (more than 10% failure rate)
# TYPE top_failures_pct gauge
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testRegister (check_subscriptions.TestSubscriptions)"} 47.36842105263158
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testSubAndInAndFail (check_subscriptions.TestSubscriptions)"} 35.294117647058826
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testNotSupported (check_realms.TestRealms)"} 35.294117647058826
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testNegotiate (check_realms.TestKerberos)"} 35.294117647058826
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testMissingPackages (check_subscriptions.TestSubscriptionsPackages)"} 35.294117647058826
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testIpaUnqualifiedUsers (check_realms.TestRealms)"} 35.294117647058826
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testIpa (check_realms.TestRealms)"} 35.294117647058826
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testInsights (check_subscriptions.TestSubscriptions)"} 35.294117647058826
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testNoUpdates (check_packagekit.TestUpdatesSubscriptions)"} 20.0
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testAvailableUpdates (check_packagekit.TestUpdatesSubscriptions)"} 20.0
top_failures_pct{project="cockpit-project/cockpit/rhel-7.9",test="testPackageKitCrash (check_packagekit.TestUpdates)"} 13.333333333333334
```
This coincides well with the [weather report](https://images-frontdoor.apps.ocp.ci.centos.org/tests.html?days=7).

### DIFF
--- a/prometheus-stats
+++ b/prometheus-stats
@@ -67,14 +67,14 @@ def main():
             FROM TestRuns
             WHERE time > ? AND wait_seconds <= 3600; """, (since, )).fetchone()[0]
 
-    output += """# HELP queue_time_wait_seconds histogram of queue wait times in the last %i hours
+    output += f"""# HELP queue_time_wait_seconds histogram of queue wait times in the last {opts.hours} hours
 # TYPE queue_time_wait_seconds histogram
-queue_time_wait_seconds_bucket{le="300"} %i
-queue_time_wait_seconds_bucket{le="3600"} %i
-queue_time_wait_seconds_bucket{le="+Inf"} %i
-queue_time_wait_seconds_sum %i
-queue_time_wait_seconds_count %i
-""" % (opts.hours, test_runs_wait_5m, test_runs_wait_1h, test_runs, test_runs_queue_wait_sum, test_runs)
+queue_time_wait_seconds_bucket{{le="300"}} {test_runs_wait_5m}
+queue_time_wait_seconds_bucket{{le="3600"}} {test_runs_wait_1h}
+queue_time_wait_seconds_bucket{{le="+Inf"}} {test_runs}
+queue_time_wait_seconds_sum {test_runs_queue_wait_sum}
+queue_time_wait_seconds_count {test_runs}
+"""
 
     print(output)
 
--- a/prometheus-stats
+++ b/prometheus-stats
@@ -74,7 +74,32 @@ queue_time_wait_seconds_bucket{{le="3600"}} {test_runs_wait_1h}
 queue_time_wait_seconds_bucket{{le="+Inf"}} {test_runs}
 queue_time_wait_seconds_sum {test_runs_queue_wait_sum}
 queue_time_wait_seconds_count {test_runs}
+
 """
+
+    # failures
+    top_failures = cursor.execute("""
+            SELECT TestRuns.project, t1.testname, (t2.failed * 100.0) / COUNT(run) AS percent
+            FROM Tests AS t1
+            JOIN TestRuns ON t1.run = TestRuns.id
+            JOIN (
+                SELECT testname, COUNT(run) as failed
+                FROM Tests
+                JOIN TestRuns ON Tests.run = TestRuns.id
+                WHERE TestRuns.time > strftime('%s', 'now', '-7 days') AND Tests.failed = 1
+                GROUP BY testname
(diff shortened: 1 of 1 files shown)


prometheus-stats
```diff
@@ -74,7 +74,32 @@ queue_time_wait_seconds_bucket{{le="3600"}} {test_runs_wait_1h}
...
+            JOIN (
+                SELECT testname, COUNT(run) as failed
+                FROM Tests
+                JOIN TestRuns ON Tests.run = TestRuns.id
+                WHERE TestRuns.time > strftime('%s', 'now', '-7 days') AND Tests.failed = 1
+                GROUP BY testname
+            ) AS t2 ON t1.testname = t2.testname
+            WHERE TestRuns.time > strftime('%s', 'now', '-7 days') AND t2.failed > 1
+            GROUP BY t1.testname
+            ORDER BY percent DESC
```
marusak, line 94: cannot we also do `WHERE percent > 10`? and then we don't need to limit it to 50 and subsequently check if failure rate is over 10%.

Or we cannot use `percent` in WHERE clause?

prometheus-stats
```diff
@@ -67,14 +67,39 @@ def main():
...
+            LIMIT 50""").fetchall()
+
+    output += """# HELP top_failures_pct Most unstable tests in the last 7 days (more than 10% failure rate)
```
marusak, line 97: I am wondering if we are not doing this wrongly as one can query in graphana stats for lets say 24 hours and we will show "Most unstable tests in last 7 days from point of last 24 hours". (or we will show for every hour how it looks 7 days back?)

prometheus-stats
martinpitt: I tried that, but unfortunately WHERE can't be used with aggregates (`SUM`). `HAVING` only works for aggregates. I did not find a way to do this in one select. Possibly we could wrap this into yet another sub-select, do the full table first, and then only select the ones which are > 10%. I'll try that in a bit.

prometheus-stats
martinpitt: I would not represent these as a graph, but as a table. It may be possible to get a selector for this like "how did that table look 3 days ago", but  at least initially the current list seems fine (what we also have on the weather report).

But admittedly I'm not at all versed in Grafana yet, so maybe there is some impedance mismatch here.

prometheus-stats
marusak: Right, was afraid of that.
If it will take another subquery, lets maybe not do that as it will most likely become too difficult to read and maintain.

prometheus-stats
marusak: ack, lets go with this then and see how it looks in graphana and how it works. If we realize we need something else, we can just change it.

marusak (approved):
Danke!

PR #1834, opened 2021-03-23
Image refresh for fedora-33

cockpituous:
Image refresh for fedora-33
 * [x] image-refresh fedora-33

### DIFF
--- a/images/fedora-33
+++ b/images/fedora-33
@@ -1,1 +1,1 @@
-fedora-33-4308e5846b8c93b2e5eb005dcdfab2efd546bca27e51c1dabc64c6054e4e8f68.qcow2
+fedora-33-2a495eb818bdac9cfdc15dab82f4db67dca015df870f4ebaaba31ede23d895d2.qcow2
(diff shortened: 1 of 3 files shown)


skobyda:
Looking at `https://logs.cockpit-project.org/logs/pull-1834-20210323-090932-0f28e52f-fedora-33-cockpit-project-cockpit/log.html#242-2`, it seems like a common failure. Retrigerring.

martinpitt:
`TestStorageResize.testGrowShrinkEncryptedHelp` is broken; regression, needswork.

skobyda:
So I tried to debug it a bit and found the problem is that the after resizing the volume (clicking on the "Grow" button which triggers lvol_grow at pkg/storaged/lvol-tabs.jsx), the volume gets unmounted automatically. This unmount happens on fedora-testing and fedora-33 but doesn't happen on fedora-34 for example.
I got lost futher and could not find the problem any deeper. Perhaps @mvollmer know what could be the bug there? Is it normal that sometimes volume gets unmounted after resizing it? If so, then perhaps test needs to be changed to account for that. If not then I guess the problem is with resizing function.

marusak:
@skobyda you triggered only two contexts here. There were 13 more to be triggered. Please use `image:fedora-33` pattern on such PRs. Thanks!

martinpitt (approved):
All green, nice! Thanks @skobyda !

PR #1840, opened 2021-03-24
Introduce test-failure-policy

marusak:
This is simplified and cleaned up version of tests-policy.
We cannot directly replace this script as specific version of run-tests
depends on it.

Testing for now.

### DIFF
--- a/task/test-failure-policy
+++ b/task/test-failure-policy
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2021 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+import os
+import subprocess
+import sys
+import unittest
+
+BASE = os.path.dirname(__file__)
+sys.path.insert(1, os.path.join(BASE, ".."))
+
+from lib.constants import BOTS_DIR
+
+
+PCP_CRASH = """
+# ----------------------------------------------------------------------
+# testFrameNavigation (check_multi_machine.TestMultiMachine)
+#
+Unexpected journal message '/usr/libexec/cockpit-pcp: bridge was killed: 11'
+not ok 110 testFrameNavigation (check_multi_machine.TestMultiMachine)
+Traceback (most recent call last):
+  File "test/verify/check-multi-machine", line 202, in tearDown
+    MachineCase.tearDown(self)
+  File "/home/martin/upstream/cockpit/test/common/testlib.py", line 533, in tearDown
+    self.check_journal_messages()
+  File "/home/martin/upstream/cockpit/test/common/testlib.py", line 689, in check_journal_messages
+    raise Error(first)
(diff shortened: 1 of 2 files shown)


martinpitt:
From the peanut gallery: This does not really "validate" a test, honestly I don't like the name. How about "test-failure-policy"? (Deliberately not calling it "tests", as this handles a single result at a time). You could even call it "test-policy" (no no, that wouldn't be confusing *at all* :grin: )

marusak:
heh right, naming was the the part where I had no good idea :)
I can call it `test-failure-policy`

marusak:
renamed + made commit message a bit more verbose.

task/test-failure-policy
```diff
@@ -0,0 +1,89 @@
...
+Journal extracted to TestMultiMachine-testFrameNavigation-fedora-i386-127.0.0.2-2502-FAIL.log
+"""
+
+
+class TestPolicy(unittest.TestCase):
+    def testKnownIssue(self):
+        script = os.path.join(BOTS_DIR, "test-failure-policy")
+        cmd = [script, "--offline", "example"]
+        proc = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, universal_newlines=True)
+        (output, unused) = proc.communicate(PCP_CRASH)
```
martinpitt, line 59: Either s/unused/_/ or output = ...[0], or use `subprocess.run()`, which is now available everywhere and a bit more comfortable.

task/test-failure-policy
```diff
@@ -0,0 +1,89 @@
...
+        self.assertEqual(output, "Known issue #9876\n")
```
martinpitt, line 60: Please also assert the returncode 77 here.

task/test-failure-policy
```diff
@@ -0,0 +1,89 @@
...
+
+        proc = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, universal_newlines=True)
+        (output, unused) = proc.communicate("""
+> log: phase coils are misaligned by 0.34 micron
+Traceback (most recent call last):
+  File "test/verify/check-warp-drive", line 34, in testCoils
+     self.assertTrue(all_in_order)
+not ok 1 test/verify/check-warp-drive TestDrive.testCoils
+""")
+        self.assertEqual(output, "Known issue #123\n")
```
martinpitt, line 85: Please also add test cases for "do nothing" (one for a passing, one for a failing test), with empty output and exit code 0; and a "retry" case with code 1.

test-failure-policy
```diff
@@ -0,0 +1,261 @@
...
+from lib.constants import BOTS_DIR
+from lib.testmap import get_test_image
+
+
+def main():
+    script = os.path.basename(__file__)
+    parser = argparse.ArgumentParser(description='Check a traceback for a known issue')
+    parser.add_argument('-o', "--offline", action='store_true',
+                        help="Work offline, don't fetch new data or contact servers")
+    parser.add_argument('image', help="The image to check against")
```
martinpitt, line 41: Was it a conscious decision to replace context with image? I wonder if that makes the caller's job harder -- right now you can just use `$TEST_OS`, and tests-policy would strip off the scenario.

martinpitt (changes requested):
Thanks!

test-failure-policy
marusak: It was. 

- It was really only interested in image
- Tests only pass image name, never scenario
- The parser in `tests-policy` is not correct (it works because it is called only with image name)
- Tests-policy assumes that default contex looks like `verify/<OS>` which hasn't been true for a while now

So I thought that rather than doing the whole parsing of context etc. just so we can call it manually a bit differently felt a bit pointless so I simplified it as much as possible. You can just do `./test-failure-policy fedora-33` instead of `./test-failure-policy fedora-33/firefox`

task/test-failure-policy
marusak: done

test-failure-policy
martinpitt: If the tests only pass the image name, and that's not much harder than passing `$TEST_OS`, then agreed. Thanks!

martinpitt (approved):
Thanks! As-is this is safe to land as it only introduces a new script.

PR #1846, opened 2021-03-25
testmap: Declare cockpit-podman fedora-34/rawhide [alternative to #1847]

martinpitt:


martinpitt:
I ran `TEST_SCENARIO=rawhide TEST_OS=fedora-34 make check` in current podman master and that was successful.

However, now I'm not actually sure about this any more -- we already run the integration tests in packit now, and that seems reasonably reliable. So maybe we don't need this at all any more? I'll send an alternative PR that just drops it.

marusak:
We do run packit, but only on PRs. If there is no activity in podman (which is often) then we are likely not to test it for weeks.
This way we check it at least once a week on image-refresh.

marusak (approved):
Thanks!

martinpitt:
Good point!

PR #1828, opened 2021-03-21
Image refresh for fedora-testing

cockpituous:
Image refresh for fedora-testing
 * [x] image-refresh fedora-testing
 * [ ] ~copr nightly regression: https://bugzilla.redhat.com/show_bug.cgi?id=1944517~
 * [x] Fix scenario setup to not install recommends on upgrade: https://github.com/cockpit-project/cockpit/pull/15630

PR #1849, opened 2021-03-25
naughty: systemd regression spilled from fedora-34 to fedora-33 and image-refresh fedora-33

skobyda:
Copy the existing fedora-33 naughty to fedora-34

Known issue #1739

### DIFF
--- a/images/fedora-33
+++ b/images/fedora-33
@@ -1,1 +1,1 @@
-fedora-33-4308e5846b8c93b2e5eb005dcdfab2efd546bca27e51c1dabc64c6054e4e8f68.qcow2
+fedora-33-2a495eb818bdac9cfdc15dab82f4db67dca015df870f4ebaaba31ede23d895d2.qcow2
--- a/naughty/fedora-33/1739-cryptsetup-systemd-umount-fs
+++ b/naughty/fedora-33/1739-cryptsetup-systemd-umount-fs
@@ -1,0 +1,4 @@
+umount: /dev/mapper/luks-*: not mounted.
+*
+  File "*test/verify/check-storage-resize", line *, in testGrowShrinkEncryptedHelp
+    self.shrink_extfs(fs_dev, "200M")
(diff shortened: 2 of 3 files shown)


skobyda:
Moved the fedora-33 image refresh from #1834 to here to see if moving naughty files fixes failures

martinpitt:
@skobyda : It's much easier to commit the naughty updates etc. straight into the image-refresh-* branch and thus the existing PR. Then you don't need to duplicate PRs and split the information. It's not even possible to handle more than one naughty update or image build script fix at a time with this "separate PR" approach.

marusak:
Tests are not green, I suggest rather then retrying to just push that one commit into  #1828 and closing this.

marusak:
> naughty: systemd regression spilled from fedora-34 to fedora-33

Maybe also please rename the commit message as well. There is no process in which changes from F34 trickle down down to F33 (unless F34 would be current rawhide and F33 would be the next branched out release, but that is not the case). It is just that F33 now has new systemd that landed in F34 sooner.

mvollmer:
I am not yet comfortable calling this a "systemd regression".  In my understanding right now, the udev rules were always set up to produce this kind of behavior (suspended device mapper devices are treated by systemd as unplugged), but we got lucky that the udev rules never ran while a dm device was suspended until now.

skobyda:
> I am not yet comfortable calling this a "systemd regression". In my understanding right now, the udev rules were always set up to produce this kind of behavior (suspended device mapper devices are treated by systemd as unplugged), but we got lucky that the udev rules never ran while a dm device was suspended until now.

Im just copying the name from existing fedora-34 naughties.
But sure, reading at https://bugzilla.redhat.com/show_bug.cgi?id=1934567 I'm seeing it's either cryptsetup failt? we could rename it to something else (on both fedora-34 and fedora-33), like `udev-rules-unmount-fs` ? WDYT @mvollmer

skobyda:
So I added the commit to #1834 . Closing this PR

PR #1853, opened 2021-03-26
machine: Allow arbitrary matches with journal_matches

mvollmer:
So that we can also look at messages with GLIB_DOMAIN=cockpit-bridge,
say.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -18,6 +18,7 @@
 import os
 import errno
 import subprocess
+import re
 
 from lib.constants import DEFAULT_IDENTITY_FILE, ATOMIC_IMAGES, OSTREE_IMAGES, TEST_DIR, BOTS_DIR
 from . import ssh_connection
@@ -161,7 +162,7 @@ class Machine(ssh_connection.SSHConnection):
         """
         return self.execute("journalctl --show-cursor -n0 -o cat | sed 's/^.*cursor: *//'")
 
-    def journal_messages(self, syslog_ids, log_level, cursor=None):
+    def journal_messages(self, matches, log_level, cursor=None):
         """Return interesting journal messages"""
 
         # give the OS some time to write pending log messages, to make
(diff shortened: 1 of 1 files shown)


machine/machine_core/machine.py
```diff
@@ -187,7 +187,9 @@ def journal_messages(self, syslog_ids, log_level, cursor=None):
             cursor_arg = ""
 
         cmd = "journalctl 2>&1 %s -o cat -p %d %s || true" % (cursor_arg, log_level, matches)
+        print("J", cmd)
         messages = self.execute(cmd).splitlines()
+        print("M", messages)
```
martinpitt, line 192: Debugging leftover? If not, this shoudl be a bit more verbose.

martinpitt:
Running two tests against this to make sure it doesn't break.

martinpitt (changes requested):
See above

machine/machine_core/machine.py
mvollmer: Yes, oops.

mvollmer: Fixed.

mvollmer:
What about changing the meaning of the syslog_ids argument of  `journal_messages`?  We can re-define it to be an array of match expressions like "_PID=1234" and for compatibility translate strings without a "FIELD=" into "SYSLOG_IDENTIFIER=...". I think that would be way cleaner.

mvollmer:
> What about changing the meaning of the syslog_ids argument of `journal_messages`?

I'll do this.

martinpitt:
@mvollmer: That would feel rather hackish IMHO -- isn't that exactly what `match` is for?

mvollmer:
> @mvollmer: That would feel rather hackish IMHO -- isn't that exactly what `match` is for?

Yes, we would remove `match`.  I think it is awkward to have two arguments that do the same thing.

martinpitt:
@mvollmer : But they don't do the same thing -- certainly the new `match` is more powerful, so ISTM that we can port stuff over to `match` and then drop the old `syslog_ids` parameter once nothing uses it any more. I guess there's not actually that many consumers, so maybe you can just do it in one go.

mvollmer:
Tests have run twice.

martinpitt (approved):
Ack, thanks! Still triggering two integration tests to make sure these work.

PR #1855, opened 2021-03-27
Image refresh for rhel-8-3

cockpituous:
Image refresh for rhel-8-3
 * [x] image-refresh rhel-8-3


marusak:
I guess we can drop 8-3 unless subscription-manager needs it.

marusak (approved):

martinpitt:
@marusak: They still need it. There is also a grave bug on 8.4 which currently skips almost all of their tests on 8.4: https://bugzilla.redhat.com/show_bug.cgi?id=1930037 So I'd really like to keep 8.3 running until 8.4 gets fixed.

Issue #1453, opened 2020-12-07
fsadm crashes with "unbound variable"


mvollmer:
https://bugzilla.redhat.com/show_bug.cgi?id=1905072

mvollmer:
verify/rhel-7-9
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testVdo (__main__.TestStorage)
#

DevTools listening on ws://127.0.0.1:9562/devtools/browser/83bef601-4ded-4854-bc77-44cf6bd181eb
[1208/123536.875812:ERROR:gl_initializer_linux_x11.cc(122)] eglGetProcAddress not found.
[1208/123536.889750:ERROR:viz_main_impl.cc(150)] Exiting GPU process due to errors during initialization
[1208/123536.895885:WARNING:vaapi_wrapper.cc(523)] VAAPI video acceleration not available for disabled
> info: %cDownload the React DevTools for a better development experience: https://fb.me/react-devtools font-weight:bold
> warning: grep: /sys/class/dmi/id/power/autosuspend_delay_ms: Input/output error

> warning: fsadm exited with code 1
Wrote screenshot to TestStorage-testVdo-rhel-7-9-127.0.0.2-2201-FAIL.png
Wrote HTML dump to TestStorage-testVdo-rhel-7-9-127.0.0.2-2201-FAIL.html
Wrote JS log to TestStorage-testVdo-rhel-7-9-127.0.0.2-2201-FAIL.js.log
Journal extracted to TestStorage-testVdo-rhel-7-9-127.0.0.2-2201-FAIL.log
Traceback (most recent call last):
  File "/home/mvo/work/cockpit/./test/verify/check-storage-vdo", line 102, in testVdo
    self.dialog({"lsize": 10 * 1024})
  File "/home/mvo/work/cockpit/test/verify/storagelib.py", line 312, in dialog
    self.dialog_wait_close()
  File "/home/mvo/work/cockpit/test/verify/storagelib.py", line 276, in dialog_wait_close
    self.browser.wait_not_present('#dialog')
  File "/home/mvo/work/cockpit/test/common/testlib.py", line 344, in wait_not_present
    self.wait_js_func('!ph_is_present', selector)
  File "/home/mvo/work/cockpit/test/common/testlib.py", line 335, in wait_js_func
    self.wait_js_cond("%s(%s)" % (func, ','.join(map(jsquote, args))))
[...]

PR #1864, opened 2021-03-30
naughty: Drop #1453 for rhel-7-9 and rhel-8-4

mvollmer:
It got fixed there, but it still happens on centos-7.

### DIFF
--- a/naughty/rhel-7/1453-fsadm-unbound-variable
+++ b/naughty/rhel-7/1453-fsadm-unbound-variable
@@ -1,1 +1,0 @@
-/usr/sbin/fsadm: line *: $3: unbound variable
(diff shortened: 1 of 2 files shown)


naughty/rhel-7/1453-fsadm-unbound-variable
```diff
@@ -1 +0,0 @@
-/usr/sbin/fsadm: line *: $3: unbound variable
```
martinpitt, line 1: This doesn't work -- naughty/centos-7 is a symlink to rhel-7. So at this point you have to copy the naughty dir, or we just don't care and leave rhel 7 alone.

martinpitt (changes requested):

naughty/rhel-7/1453-fsadm-unbound-variable
mvollmer: Yeah, let's not bother.

martinpitt:
I'd still like to drop the naughty for rhel-8..

PR #1867, opened 2021-03-31

cockpituous:


martinpitt:
This and ubuntu-stable failed on c-machines as download-dist seems to get confused: apparently there's some naming mismatching if current HEAD is a tagged release? Then the tarball will look like cockpit-machines-241.tar.gz instead of 240.12.git12345.tar.gz? @KKoukiou , does that ring a bell?

martinpitt (approved):

PR #1872, opened 2021-03-31
task: Allow external link patterns for log.html

mvollmer:
The idea is that testlib.py attaches an appropriate link-patterns.json
file for all the kinds of files that it attaches.  This will be used
by the pixel tests to make links to the pixels comparison pages.

Also, the default pattern for a screenshot is more specific so that we
don't bogusly pick up all the PNGs from the "initialisation" segment.

### DIFF
--- a/task/log.html
+++ b/task/log.html
@@ -50,16 +50,10 @@
                 {{#tests}} {{{html}}} {{/tests}}
             </div>
         </script>
-        <script id="ScreenshotLink" type="text/template">
-            <a href="./{{screenshot}}" title="{{screenshot}}">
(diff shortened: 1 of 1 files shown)


mvollmer:
> Also, the default pattern for a screenshot is more specific so that we
don't bogusly pick up all the PNGs from the "initialisation" segment.

This is debatable, I guess.  It's a backwards incompatible change, and the only reason to have any default link patterns at all is to stay backwards compatible... so maybe we should just keeps things as they were regexp wise.

martinpitt (approved):
Thank you! I like the generalization of this concept, and the cleaning up of the "initialization" screenshots. I tested this with current logs, and it looks fine, so it's backwards compatible.

PR #1877, opened 2021-04-01
testmap: Test rhel-9 by default

marusak:


KKoukiou (approved):

martinpitt:
Hmm, I haven't seen this [TestAccounts failure](https://logs.cockpit-project.org/logs/pull-1877-20210401-073603-7aef75c1-rhel-9-0-cockpit-project-cockpit/log.html#322-2) at all so far. It could be vaguely related to https://github.com/cockpit-project/cockpit/pull/15613  , which was not tested (with CI) on RHEL 9.0. Of course I *developed* the fix on RHEL 9.0, as we got the bug report there.. But I figure this needs to be looked at.

marusak:
I actually saw this issue on some different OS today as well in https://github.com/cockpit-project/cockpit/pull/15626 so not strictly related to rhel-9

marusak:
This is broken due to today being 1.4 :) Fixed in https://github.com/cockpit-project/cockpit/pull/15636

martinpitt:
Cockpit got fixed, retrying

martinpitt:
@mvollmer: [This failure here](https://logs.cockpit-project.org/logs/pull-1877-20210401-184359-7aef75c1-rhel-9-0-cockpit-project-cockpit/log.html#245-2) smells very much like https://github.com/cockpit-project/bots/issues/1739 , but it does not time out, but fails on `assert (size > 300000)`; This is a rather bad assertion, as it does not tell us the actual value. But supposedly it's the same bug, just hitting at a slightly different place?

marusak:
Ok, now the assert is better:
```
AssertionError: 268624 not greater than 300000
```

marusak:
First observation - when we `Mount` it then the resize works as expected.

marusak:
Looking into this in https://github.com/cockpit-project/cockpit/pull/15657

martinpitt:
Ugh, a [lot of test failures](https://logs.cockpit-project.org/logs/pull-1877-20210406-163013-a6c14423-rhel-9-0-cockpit-project-cockpit/log.html#65) .. Maybe let's drop -machines first?

Retrying once, anyway, to compare.

martinpitt (approved):
Thanks! (Assuming this goes green)

marusak:
Ok, this is annoying. `TestMachinesSnapshots.testSnapshots` is dumping `qemu-kvm`.
This is nothing new (weather page knows about it) yet it seems strongly persistent.
As we are moving machines away this will solve itself. We will likely hit it again when we will try to enable rhel 9 on machines tests though.

martinpitt:
> We will likely hit it again when we will try to enable rhel 9 on machines tests though.

c-machines already has rhel-9 enabled.

PR #1878, opened 2021-04-01
naughty: Improve patterns for #1739

mvollmer:
- Narrow it down to only match when the filesystem is unexpectedly not
  mounted.
- Match both call sites of shrink_extfs
- Don't insist on "testlib.Error: timeout" as Firefox produces some
  weird RuntimeError here (which is probably a bug in itself).

### DIFF
--- a/naughty/fedora-33/1739-cryptsetup-systemd-umount-fs
+++ b/naughty/fedora-33/1739-cryptsetup-systemd-umount-fs
@@ -1,4 +1,4 @@
 umount: /dev/mapper/luks-*: not mounted.
 *
   File "*test/verify/check-storage-resize", line *, in testGrowShrinkEncryptedHelp
-    self.shrink_extfs(fs_dev, "200M")
+    self.shrink_extfs(fs_dev, *)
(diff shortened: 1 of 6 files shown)


mvollmer:
This should take care of the recent TestStorageResize.testResizeLuks flake on fedora-33/firefox.

But we have a lot of "testlib.Error: timeout" expectations in our naughty patterns, and they will all fail with Firefox. It would be really good to figure out why Firefox doesn't produce the expected testlib.Error. -> #1879

marusak:
[rhel-9](https://logs.cockpit-project.org/logs/pull-1878-20210401-182240-c3fdc999-rhel-9-0-cockpit-project-cockpit/log.html#245-2) failed, I'll retry but it looks suspicious.

martinpitt:
Failed again in the same way -- but it does so in #1877 also. The problem is that it fails there a bit later on -- on the (badly written) `assert (size > 300000)`. This for sure needs fixing also, but at least this is some incremental improvement.

martinpitt (approved):
Danke!

PR #1902, opened 2021-04-08
prometheus-stats: add S3 disk usage stats

allisonkarlitskaya:
Iterate over the list of stores, and for each one the user has an S3 key
for, query how much disk space is being used and report that as a gauge.

### DIFF
--- a/prometheus-stats
+++ b/prometheus-stats
@@ -17,11 +17,15 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
-import sys
-import time
-import sqlite3
 import argparse
 import logging
+import sqlite3
+import sys
+import time
+import urllib.parse
+
+from lib import stores, s3
+
 
 # Seconds in an hour
 HOUR = 3600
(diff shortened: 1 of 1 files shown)


marusak:
what, why `unit-tests` points to `http://localhost:8080/logs/pull-1902-20210408-122512-c4dbd773-unit-tests/log.html`?
Have you posted that status manually @allisonkarlitskaya ?

prometheus-stats
```diff
@@ -124,6 +128,21 @@ pr_retries_sum {acc}
...
+    # S3 space usage
+    space_usage = {}
+    for uri in stores.HYBRID_STORES:
+        url = urllib.parse.urlparse(uri)
+
+        if s3.is_key_present(url):
+            space_usage[uri] = sum(int(size) for size, in s3.parse_list(s3.list_bucket(url), "Size"))
+    if space_usage:
+        output += """
+# HELP s3_usage_bytes gauge of how much disk space an S3 store is using
```
martinpitt, line 140: The "gauge of" is fairly redundant, but this is also entirely unimportant.

martinpitt (approved):
Thanks! I'll add that to Grafana once it gets exported.

prometheus-stats
allisonkarlitskaya: I copied it from the "histogram" of example above

PR #1903, opened 2021-04-08
prometheus-stats: fixup to S3 disk usage

allisonkarlitskaya:
- remove the = sign.  This is not the valid syntax.

 - reword the awkward help text


martinpitt (approved):
Danke!

PR #1925, opened 2021-04-14
testmap: test rhel-8-4 for subscription-manager/1.28

ptoscano:
Test the subscription-manager-1.28 branch of subscription-manager using
rhel-8-4, as that branch is the current for RHEL 8.4.

 - [x] Land https://github.com/candlepin/subscription-manager/pull/2582

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -155,6 +155,9 @@ REPO_BRANCH_CONTEXT = {
         'master': [
             'rhel-8-4',
         ],
+        'subscription-manager-1.28': [
+            'rhel-8-4',
+        ],
         '_manual': [
             'rhel-8-5',
             'rhel-9-0',


marusak:
Do you have some PR where you tested this?

ptoscano:
> Do you have some PR where you tested this?

Hmm not sure what you mean, can you please explain? My goal is to test PRs for the `subscription-manager-1.28` branch using the cockpit jobs, as right now they are not run on that branch.

marusak:
Right, so the change you send in this PR means: "When there will be PR opened or updated (force pushed) in `subscription-manager-1.28` branch, it will automatically trigger `rhel-8-4` context (unless `no-test` label given). This branch will be also tested whenever we refresh `rhel-8-4` image, which we do ever 7 days."

Therefore we have `_manual` branch, which contains contexts that we want to trigger manually to make sure context works before we auto-enable given context (you auto-enabled it). So normal workflow is:
1. Add new context into `_manual`. We land it right away.
2. Open PR in your project against given branch, manually trigger this context.
3. When it succeeds, send here another PR where you move it from `_manual` to given context to auto run it.

You did the last step so I just want to know if you already tested it. Mostly so our image refresh does not trip over it not working but also so you have it set up correctly.

(Note: Every context must be defined for every "project" it wants to run in. Thus you don't need to define `rhel-8-4` in `_manual` since you already have it in `master`, so you can just take PR that is opened against `subscription-manager-1.28` branch, do `./bots/tests-trigger --repo candlepin/subscription-manager <PR_number> rhel-8-4`.)

ptoscano:
> Right, so the change you send in this PR means: "When there will be PR opened or updated (force pushed) in `subscription-manager-1.28` branch, it will automatically trigger `rhel-8-4` context (unless `no-test` label given). This branch will be also tested whenever we refresh `rhel-8-4` image, which we do ever 7 days."

Exactly what I'd like, yes :)

> Therefore we have `_manual` branch, which contains contexts that we want to trigger manually to make sure context works before we auto-enable given context (you auto-enabled it). So normal workflow is:
> 
>     1. Add new context into `_manual`. We land it right away.

So, theoretically speaking, something like:
```diff
diff --git a/lib/testmap.py b/lib/testmap.py
index 00b0af4..8362221 100644
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -156,6 +156,7 @@ REPO_BRANCH_CONTEXT = {
             'rhel-8-4',
         ],
         '_manual': [
+            'rhel-8-4',
             'rhel-8-5',
             'rhel-9-0',
         ],
```
Given what you wrote later on, I understand it is not needed in the case of subscription-manager.

>     2. Open PR in your project against given branch, manually trigger this context.

Done: https://github.com/candlepin/subscription-manager/pull/2582
I had to backport few fixes that @martinpitt did to the CI scripts (thanks!), but otherwise it seems to work fine.

> (Note: Every context must be defined for every "project" it wants to run in. Thus you don't need to define `rhel-8-4` in `_manual` since you already have it in `master`, so you can just take PR that is opened against `subscription-manager-1.28` branch, do `./bots/tests-trigger --repo candlepin/subscription-manager <PR_number> rhel-8-4`.)

This worked... after I passed `-a` too, as I'm not a member of the cockpit-project organization, and thus tried to debug for half an hour the following error:
```
task.github.GitHubError: Error accessing /orgs/cockpit-project/teams
  Status: 403
  Reason: Forbidden
  Response: {"message":"Must have admin rights to Repository.","documentation_url":"https://docs.github.com/rest/reference/teams#list-teams"}
```

martinpitt:
Great, thanks @ptoscano ! As soon as https://github.com/candlepin/subscription-manager/pull/2582 we'll land this.

martinpitt:
https://github.com/candlepin/subscription-manager/pull/2582 landed, but in parallel with the rhel-8-4 refresh. Triggering the test here again, to make double-sure it works with the current image.

martinpitt (approved):

PR #1938, opened 2021-04-19
testmap: Validate service image refresh against subscription-manager

martinpitt:
subscription-manager's tests run against candlepin and thus are affected
by behaviour changes.


marusak (approved):
Indeed, thanks!

PR #1940, opened 2021-04-19
lib: Make redhat_network() independent from our image stores

martinpitt:
We will eventually drop our custom image stores and only use the hybrid
S3 ones. If/when our cockpit-11 store is down, none of the RH-internal
tasks runs any more, even though they still could. The most important
resource for internal tests is the internal RHEL repository, so ping
that instead of the image stores.

### DIFF
--- a/lib/stores.py
+++ b/lib/stores.py
@@ -20,7 +20,6 @@
 
 import os
 import socket
-import urllib
 
 from lib.constants import IMAGES_DIR, LIB_DIR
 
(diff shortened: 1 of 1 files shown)


marusak (approved):
Makes sense!

allisonkarlitskaya:
I'm sorry I didn't see this in time.  I'd have suggested moving the code out of this file (now that it has nothing to do with stores) and also to have tried to switch to using a standard urllib method instead of poking around with sockets for ourselves...

martinpitt:
I thought about that as well. I sent PR #1939, and once that lands, the only consumer will be run-queue, and then we can simplify it. I mostly wanted to get this out of the door with a simple change, as this broke RHEL tests over most of the weekend as the image server on cockpit-11 didn't start (argh docker)

PR #1941, opened 2021-04-19
testmap: Allow "pixeldiffs" context for starterkit

mvollmer:
This is used for reviewing changes to pixel tests.  We want to "abuse"
our testing machinery to produce a nice web page.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -66,6 +66,7 @@ REPO_BRANCH_CONTEXT = {
         ],
         '_manual': [
             'fedora-34/firefox',
+            'pixeldiffs',
         ],
     },
     'cockpit-project/cockpit-ostree': {


marusak:
This one seems strange. The context pattern is `OS[/scenarion][@project[/branch]]`.
So this would treat `pixeldiffs` as operating system.
Is this just running on host and therefore does not need OS?

mvollmer:
> Is this just running on host and therefore does not need OS?

Exactly.  The job of "pixeldiffs" is to work on the source code alone and come up with some report. This is independent of the OS. Much like a static code analyzer, say, or spell checker.

Using our testing machinery for this is a hack. If we find a way to implement this with a Github workflow, that would be better, but it's not obvious to me how.

I want the "Details" link of a check result to go straight to a web page with the image comparisons. Our testing machinery can do this pretty easily, but Github workflow artifacts don't get anywhere near that, as far as I can see.

martinpitt:
If the extra pixel diff tests should just run on one OS, I suggest to conditionalize these to `TEST_OS_DEFAULT`, like test/common/testlib.py already does for `checkRunAxe()`:
```
def checkRunAxe():
    # only run this on the default OS test, that's enough
    if os.getenv("TEST_OS") not in [None, testvm.TEST_OS_DEFAULT]:
        return False
```
If this really does do something entirely different than running a test, then perhaps this should be run in test/run (if it depends on test results) or a workflow (if it's independent from any test results)? What exactly will this do?

martinpitt:
@mvollmer: github workflows can do that as well. But either way, I have no objection at all against using our own test machinery for this -- we can move it someplace else later. For the first steps, just use the path of least resistance. I just wanted to make sure that this really behaves like a fake "OS". But how would that actually work? If you'd trigger a pixeltest in any of our projects, it would just try to run the normal integration tests with `TEST_OS=pixeltest`.

mvollmer:
> If you'd trigger a pixeltest in any of our projects, it would just try to run the normal integration tests with `TEST_OS=pixeltest`.

Yes, and we will catch that in test/run and run test/comm/review-pixeldiffs instead of making a VM and running test/common/run-tests.

A more proper way would probably be to code this into tests-scan or tests-invoke somehow.

One special thing about this is that there is no pass/fail judgement at the end.  We _always_ have to look at the pixeldiffs.  This is different from the AXE tests, I believe, but I'll check those out as well.

mvollmer:
> github workflows can do that as well.

Oh, how?

martinpitt:
@mvollmer: They can just create/update their own statuses (or the more modern [checks API](https://docs.github.com/en/developers/apps/creating-ci-tests-with-the-checks-api). But as I said, we don't need to do it "pretty", let's do it "fast" first. If test/run recognizes this as a "magic" TEST_OS, then this PR makes sense. Thanks!

martinpitt (approved):

PR #1913, opened 2021-04-11
images: Build centos-7 image from official qcow image

martinpitt:
Starting from the cloud image is faster, more robust, and more realistic
(using an official product) than our own virt-install bootstrap.

CentOS 7 offers xz-compressed qcow images (254 MB xz vs. 819 MB
uncompressed, so well worth it). Download these and teach
cloudimage.bootstrap to handle these as well.

 * [x] image-refresh centos-7
 * [x] fix cockpit rhel-7.9 branch for current bots: https://github.com/cockpit-project/cockpit/pull/15675

PR #1945, opened 2021-04-21
Image refresh for centos-7

cockpituous:
Image refresh for centos-7
 * [x] image-refresh centos-7


martinpitt (approved):

martinpitt:
Reverted in #1947, as this image wreaks havoc. The package diff is very small:
```
Removed:

Added:

Changed:
  nettle (2.7.1-8.el7 -> 2.7.1-9.el7_9)
```

That change is [here](https://git.centos.org/rpms/nettle/c/796117145aea371151dc9eadccd12a7ccdb82903?branch=c7), some hardening of ECDSA validation. It's possible that this affects the booting somehow, but this rather smells like some bad image build. We still use virt-install for that image -- I tried to convert that to a cloud image in PR #1913, but that requires a lot of test adjustments by itself.

So I propose let's just try a new build and see how that behaves, before wasting hours on debugging this.

PR #1961, opened 2021-04-24
images: Always update selinux-policy

marusak:
* [x] image-refresh rhel-9-0

If this helps, then we will need it in fedora images as well, but testing it first here.


cockpituous:
image-refresh in progress on 2-ci-srv-06.
Log: https://logs.cockpit-project.org/logs/image-refresh-1961-20210424-045757/

cockpituous:
image-refresh rhel-9-0 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-9-0-20210424-050808

marusak:
So actually this new command did nothing:
```
+ dnf update -y selinux-policy
Last metadata expiration check: 0:01:37 ago on Sat Apr 24 00:57:29 2021.
Dependencies resolved.
Nothing to do.
Complete!
```

But the image has been updates since yesterday afternoon and it now includes this new version. But this is race:
1. Image is composed
2. New version of selinux-policy is released
3. We do image-refresh where we use image with older version but in mock we have already newer one

So I think this commit makes sense in a long run.

images/scripts/rhel.setup
```diff
@@ -184,6 +184,12 @@ fi
 
 dnf install -y $TEST_PACKAGES $COCKPIT_DEPS $IPA_CLIENT_PACKAGES
 
+# We are going to install the newest selinux-policy into mock and we need at least that version
+# Always update to prevent case where image has older version than mock
+if [ "${IMAGE#rhel-9*}" != "$IMAGE" ]; then
```
martinpitt, line 189: Why is this specific to RHEL 9? Doesn't it in principle apply to RHEL 8/CentOS (we just don't build the policy there for now) and Fedora (fedora.setup) as well?

More importantly, I think we should always run `dnf update -y` on all RHEL/Fedora images to make sure we don't fall behind if images are not built often enough.

martinpitt (approved):
Thanks! Let's do this as a quick fix, and I add it to my TODO to do the more general "update everything" for all images. We already do that for Debian/Ubuntu images.

images/scripts/rhel.setup
marusak: > Why is this specific to RHEL 9?

As I wrote above, this was test (and quick fix). We don't need it on rhel-8/centos.

> More importantly, I think we should always run dnf update -y on all RHEL/Fedora images to make sure we don't fall behind if images are not built often enough.

I was thinking if we should not update those images in any case, but  did not want to introduce this kind of change in this PR.

martinpitt: Fully agreed, this was just thinking aloud.

martinpitt:
Hmm, rhel.setup actually *already* [updates the entire system](https://github.com/cockpit-project/bots/blob/bd6ff6470a8b02b4fea4475e2b665246e8b66321/images/scripts/rhel.setup#L93), but before installing the test dependencies.  But in the log here it says:
```
+ dnf --nogpgcheck -y update
baseos                                          6.7 MB/s | 955 kB     00:00    
appstream                                        25 MB/s | 5.7 MB     00:00    
Dependencies resolved.
Nothing to do.
Complete!
[...]
+ dnf install -y boom-boot clevis-luks [...]
Package selinux-policy-targeted-34.1.2-1.el9.noarch is already installed.
```

and then later on, the fix in this PR is also a no-op:
```
+ dnf update -y selinux-policy
Last metadata expiration check: 0:01:37 ago on Sat Apr 24 00:57:29 2021.
Dependencies resolved.
Nothing to do.
```

later on, the mock install has the same version:
```
DEBUG:  selinux-policy                noarch  34.1.2-1.el9                build   83 k
DEBUG:  selinux-policy-devel          noarch  34.1.2-1.el9                build  832 k
```

Comparing this with the [log of the previous refresh](https://logs.cockpit-project.org/logs/image-refresh-1955-20210423-045353/log) which had the bug:
```
+ dnf --nogpgcheck -y update
baseos                                          7.2 MB/s | 955 kB     00:00    
appstream                                        19 MB/s | 5.8 MB     00:00    
Dependencies resolved.
Nothing to do.
Complete!
[...]
+ dnf install -y boom-boot clevis-luks [...]
Package selinux-policy-targeted-34-1.el9.noarch is already installed.
[...]
# mock installation
DEBUG:  selinux-policy                noarch  34.1.2-1.el9                build   83 k
DEBUG:  selinux-policy-devel          noarch  34.1.2-1.el9                build  832 k
```

... which shows that this was apparently a bad compose. The brewroot compose already had the newer version 34.1.2-1, while BaseOS still had 34-1.

At first I thought this was another instance of the [dnf installs outdated packages bug](https://bugzilla.redhat.com/show_bug.cgi?id=1946975), but there is no evidence here.

Instead, this is a problem with inconsistent composes, and the "fix" here is really a no-op.

martinpitt:
FTR, we also upgrade the Fedora images already, I just searched for "dnf update" before, but it's spelled "dnf upgrade" (the two are synonyms, AFAIK)

PR #1978, opened 2021-05-02
Image refresh for rhel-9-0 (broken compose)

cockpituous:
Image refresh for rhel-9-0
 * [x] image-refresh rhel-9-0
 * [ ] wait until [compose](http://download.eng.bos.redhat.com/rhel-9/nightly/RHEL-9-Beta/latest-RHEL-9/compose/BaseOS/x86_64/os/Packages/) has selinux-policy 34.1.3-1


martinpitt:
This is another compose failure again:
```
error: Failed dependencies:
	(selinux-policy >= 34.1.3 if selinux-policy-targeted) is needed by cockpit-ws-243.21.gded7977a4-1.wip.el9.x86_64
```

The buildroot has a newer version 34.1.3-1.el9 while BaseOS still has 34.1.2-1.el9. So this image is useless, and we need to wait for a fixed compose again.

@vmojzis how do you deal with that? That breaks the version comparison all the time, and surely must affect all packages that ship a custom policy? That's already the second failed refresh in two weeks, so 50% failure rate..

martinpitt:
One idea: We could figure out how to convince mock to install selinux-policy* packages from the BaseOS repo instead of buildroot. That's a sad workaround for a major bug that will hit RHEL 9 hard, but at least it would unblock us.

marusak:
> One idea: We could figure out how to convince mock to install selinux-policy* packages from the BaseOS repo instead of buildroot. That's a sad workaround for a major bug that will hit RHEL 9 hard, but at least it would unblock us.

Or install the exact same version that we have in image itself. In rhel all versions should be available. Still hack though.

martinpitt:
I asked about this on rhel-devel@. The RHEL 9 nightly BaseOS compose seems just plain broken, it hasn't updated since Apr 28. So on second thought, let's just wait for that to get fixed, instead of introducing gross hacks on our side.

martinpitt:
I got a reply, and composes were indeed broken. This was *not* a conscious policy decision, nor is it a systematic error. So let's just wait for a new compose (should happen today), and then close/rebuild this one.

martinpitt:
> This was not a conscious policy decision, nor is it a systematic error

I got a follow-up, and it turns out that it *is* a systematic error after all. buildroot is regularly ahead of BaseOS, which is just plain broken.

martinpitt:
Anyway, compose updated today, and selinux-policy is current. Closing and let the bots build a new one.

PR #1984, opened 2021-05-04
scripts: update RHEL 8.4 distro url

jkozol:
RHEL 8.4 is no longer under the `nightly` subdir in `download.devel`. Instead, it is under `rhel-eng`.

 * [x] image-refresh rhel-8-4


cockpituous:
image-refresh in progress on 2-cockpit-9.
Log: https://logs.cockpit-project.org/logs/image-refresh-1984-20210504-153939/

cockpituous:
image-refresh rhel-8-4 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-8-4-20210504-155959

marusak:
Nice!
You need to pick the refreshed image into this PR. While pushing again, can you also add `Fixes #1977`?
And then trigger tests with `image:rhel-8-4`.

images/scripts/rhel-8-4.bootstrap
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -ex
 
-URL=http://download.devel.redhat.com/rhel-8/nightly/RHEL-8/latest-RHEL-8.4/compose/BaseOS/x86_64/images/
+URL=http://download.devel.redhat.com/rhel-8/rel-eng/RHEL-8/latest-RHEL-8.4/compose/BaseOS/x86_64/images/
```
marusak, line 4: Is this going to get updates - like 8.4.1?

jkozol: I am unsure. Other than 8.0.1 (which is currently a nightly) I could not find any z stream releases in download.devel. It would make sense to me that this would get updates like 8.4.1 since it is labeled as 1latest-RHEL-8.4` and not `latest-RHEL-8.4.0`.

jkozol:
@marusak I do not have permissions to trigger the `image:rhel-8-4` test

marusak:
> I do not have permissions to trigger the image:rhel-8-4 test

Interesting, I though that if you could do refresh you also could trigger tests. But anyway, I've triggered them.

images/scripts/rhel-8-4.bootstrap
martinpitt: FTR, in the recently removed rhel-8-3 image (commit 654fa074ceb8965b87b1c0fd37aaa4e2d989b334) we used http://download.devel.redhat.com/released/rhel-8/RHEL-8/8.3.0/BaseOS/x86_64 . But annoyingly this does not exist yet for /8.4.0. But indeed this URL seems better, like a "floating" 8.4.z one. Thanks!

martinpitt:
This refresh completely breaks cockpit-composer tests, with "Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory" (They are all still running due to the retries, so it takes ages). Does that ring a bell, @jkozol or @henrywang ?

jkozol:
@martinpitt I am perplexed. Investigating these failures now.

jkozol:
https://github.com/osbuild/cockpit-composer/pull/1290 This PR should fix these failures as explained here: https://github.com/osbuild/cockpit-composer/pull/1290#issuecomment-832640768

martinpitt:
Hah nice, thanks @jkozol ! I'm glad that it was the same root cause after all.

martinpitt:
So indeed the composer tests fail in exactly the  same way on the current image already, so the failures above are not a regression and not a blocker. Still waiting for my extra triggered tests in https://github.com/osbuild/cockpit-composer/pull/1290 though.

martinpitt (approved):
Danke!

martinpitt:
https://github.com/osbuild/cockpit-composer/pull/1290 landed, re-running tests.

martinpitt:
Actually, no point -- [it just succeeded](https://logs.cockpit-project.org/logs/pull-1290-20210505-123834-77a80803-rhel-8-4-bots-1984/log.html).

PR #2006, opened 2021-05-12

cockpituous:
 * [x] cockpit-project/cockpit#15826


martinpitt:
Seems the tang host key algorithm or encoding changed? Pleeease let it be moving away from SHA-1 :grin:

martinpitt:
clevis didn't change recently, but [tang did](https://koji.fedoraproject.org/koji/buildinfo?buildID=1744121). The [build log](https://logs.cockpit-project.org/logs/image-refresh-2006-20210512-023131/log) does not show tang as updated, and indeed our [previous refresh from 4 days ago](https://logs.cockpit-project.org/logs/image-refresh-2000-20210508-022525/log) already had the new tang.

So the only plausible change is [jose](https://bodhi.fedoraproject.org/updates/FEDORA-2021-1843061abf)

    jose (10-8.fc33 -> 11-1.fc33)

The [new release description](https://github.com/latchset/jose/releases/tag/v11) indeed says "Change default JWK thumbprint algorithm to SHA-256" :partying_face:

martinpitt:
So we need to adjust [crypto_keyslots()](https://github.com/cockpit-project/cockpit/blob/master/pkg/storaged/crypto-keyslots.jsx#L85) to do some version/feature detection of jose, and use sha256 in that case. @mvollmer , is that something you want to look into, or is that pilot matter?

allisonkarlitskaya:
This has been kinda broken for a while, actually, since `tang-show-keys` looks like this:

```
THP_DEFAULT_HASH=S256    # SHA-256.
echo $adv \
    | jose fmt -j- -g payload -y -o- \
    | jose jwk use -i- -r -u verify -o- \
    | jose jwk thp -i- -a "${THP_DEFAULT_HASH}"
```

(ie: we're telling the user to check the fingerprint against a tool that will produce a different result).

I'm going to do the following:
 - pass `-a S256` to `jose` unconditionally in the test
 - switch to SHA256 in `crypto-keyslots.jsx`

martinpitt:
jose fix landed in cockpit, re-running.

marusak (approved):

PR #2008, opened 2021-05-13

cockpituous:
 * [x] https://github.com/cockpit-project/cockpit/pull/15837
 * [x] add naughty for new SELinux systemd-hostnam violation

### DIFF
--- a/images/fedora-coreos
+++ b/images/fedora-coreos
@@ -1,1 +1,1 @@
-fedora-coreos-1c1d1022e253ab4e85f24e378a85710f3aae4dc999dd42b964fd548177d01e6e.qcow2
+fedora-coreos-293d7217c4cc24202deb059fb9bfa987dce238c82cc35ad7a4de074da38a50e0.qcow2
--- a/naughty/fedora-coreos/2031-selinux-systemd-hostnam
+++ b/naughty/fedora-coreos/2031-selinux-systemd-hostnam
@@ -1,0 +1,2 @@
+testlib.Error: FAIL: Test completed, but found unexpected journal messages:
+audit*avc:  denied  { write } for * comm="systemd-hostnam" name="systemd" dev="tmpfs" *


martinpitt:
Most tests fail with "invalid or unusable locale: en_US.UTF-8", so some change in the image lost locales? If that was on purpose, we need to  switch our default locale to C.UTF-8 now. This won't work on RHEL 7, but we don't test that from master any more, so we can finally do that.

allisonkarlitskaya:
I mean, that's kinda obvious:

```
Removed:
  glibc-all-langpacks (2.32-4.fc33)
```

We could just add back `glibc-langpack-en`.  That's what I did in my toolboxes, anyway...

allisonkarlitskaya:
To be honest, I'd almost rather add "invalid or unusable locale" to the allowed journal messages, or to stop printing this message entirely.  All that it means is that the browser sent a language header for which we have no system locale installed.  That's not really an error, but we always treat it as if it were.

Otherwise, the locale is set according to the brower's language string, and I don't know of a way to force a browser to send "C" as its language, so I think the only other fix is indeed to add the langpack.

allisonkarlitskaya:
Third and forth option, assuming the change was introduced accidentally upstream:
 - naughty all the things
 - wait

martinpitt:
> We could just add back glibc-langpack-en

Please not -- we want to test the actual image, not a "hacked" one specifically for cockpit.

> I'd almost rather add "invalid or unusable locale" to the allowed journal messages

That's my preferred solution. We could start with conditionally adding it on fcos.

martinpitt:
After https://github.com/cockpit-project/cockpit/pull/15837 there are [six remaining failures](https://logs.cockpit-project.org/logs/pull-15837-20210518-110048-a1255a1d-fedora-coreos-bots-2008/log.html) due to a new SELinux violation:

 > audit: type=1400 audit(1621337254.332:217): avc:  denied  { write } for  pid=672 comm="systemd-hostnam" name="systemd" dev="tmpfs" ino=2 scontext=system_u:system_r:systemd_hostnamed_t:s0 tcontext=system_u:object_r:init_var_run_t:s0 tclass=dir permissive=0

This looks new, and we don't have a naughty for it yet, but it's already reported downstream: https://bugzilla.redhat.com/show_bug.cgi?id=1953060

martinpitt (approved):
Thanks! Assuming green -- you still need to trigger the tests.

allisonkarlitskaya:
> Thanks! Assuming green -- you still need to trigger the tests.

already did that :)

martinpitt:
Interesting, this PR moved the FCOS image from Fedora 33 to 34. That means that we should also switch the build image.

martinpitt:
Done in PR #2033

PR #2007, opened 2021-05-13
Image refresh for fedora-34

cockpituous:
Image refresh for fedora-34
 * [x] image-refresh fedora-34


martinpitt:
podman is kaputt

allisonkarlitskaya:
I think the correct solution here is to wait:

https://bodhi.fedoraproject.org/updates/FEDORA-2021-51708c90bc

martinpitt:
Agreed, changing to blocked. It already has enough karma.

marusak:
It has been in stable for two days. Closing so tomorrow we will get a new update

PR #2033, opened 2021-05-19
naughty: Add pattern for Fedora CoreOS agetty.reload SELinux violation

martinpitt:
Fixes failures like [this](https://logs.cockpit-project.org/logs/pull-15841-20210519-040829-bcbe6949-fedora-coreos/log.html)


martinpitt:
Ah, moving the build image [does not work yet](https://logs.cockpit-project.org/logs/pull-2033-20210519-052347-41dcf3de-fedora-coreos-cockpit-project-cockpit/log.html), we first need to move our cockpit/ws container to Fedora 34. This is another maintenance thing we are lagging behind.. I back this out for now, as the naughty is urgent.

marusak (approved):

PR #2039, opened 2021-05-22

cockpituous:


allisonkarlitskaya:
The fail is semi-reproducible and seems to be a timing issue.  Using `TEST_SHOW_BROWSER` or inserting a `sit()` or `time.sleep(1)` is enough to make the fail disappear.  Even having a hot CPU on my laptop at the start of the test (ie: throttling) seems to change the outcome.

The fail happens here:

 ```
        b.set_input_text('#accounts-create-password-pw1', good_password)
        b.set_input_text('#accounts-create-password-pw2', good_password)
        b.click('#accounts-create-dialog button.apply')
```

..but that comes just after another section which successfully `b.click()`ed the exact same button.  On visual inspection, the button isn't disappearing or greying out the button on validation failures, and even if it was, `b.click()` seems to wait for the button to reappear anyway.

Finally, I think it's unrelated to this PR anyway, because as per the log:

```
Uploading...
Removed:

Added:

Changed:
```

Do we have (or want to have) a policy for non-op image refreshes?  Should we even bother landing this?  Just close it and save everyone a bit of bandwidth?

martinpitt:
@allisonkarlitskaya I don't mind just closing it as a no-op.

PR #2057, opened 2021-05-28
Image refresh for fedora-34

cockpituous:
Image refresh for fedora-34
 * [x] image-refresh fedora-34


allisonkarlitskaya:
```
error: Failed dependencies:
	(selinux-policy >= 34.8-1.fc34 if selinux-policy-targeted) is needed by cockpit-ws-245.1.g00da54e0c-1.wip.fc34.x86_64
```

allisonkarlitskaya:
So this addition from cockpit-project/cockpit#15707

```
%define selinux_policy_version %(rpm --quiet -q selinux-policy && rpm -q --queryformat "%{V}-%{R}" selinux-policy || echo 1)
```

causes the built cockpit package to depend on the version of selinux-policy that was available at build time.  It seems that it has happened a couple of times now that the version that we get in our `mock` is newer than the version that's on the image.

This is the build in question: https://bodhi.fedoraproject.org/updates/FEDORA-2021-558e78822f

The crux of this issue seems to be that we build cockpit packages against a mock environment which is created at `image-prepare` time, but we expect those packages to install (via `rpm`, not `dnf`) against a VM image which can be a bit older.

We either need to figure out a way to:
 - get the selinux-policy version from the system into the mock that builds cockpit
 - upgrade selinux-policy before installing cockpit, possibly by installing cockpit with `dnf`
 - relax this requirement a bit

@marusak I guess this is yours to look at?

marusak:
Indeed, this hit us a few times already. The problem is that builds are available in mock earlier than in repos.

> relax this requirement a bit

That is unfortunately not a possiblity

> upgrade selinux-policy before installing cockpit, possibly by installing cockpit with dnf

We already update all the packages. The problem is with availability

> get the selinux-policy version from the system into the mock that builds cockpit

Indeed, that is the way we were thinking about how to fix this. It is downgrade though, which does not work always in Fedora, but hopefully in this case it should work.

allisonkarlitskaya:
I misunderstood the situation: I thought we were pulling in all the packages for `mock` at `image-prepare` time, but it seems that they are pre-seeded during `image-create`.

But this probably doesn't help:

```
if [ "${IMAGE%-testing}" != "$IMAGE" ]; then
    # Enable updates-testing in mock
    echo "config_opts['dnf.conf'] += '[updates-testing]\nenabled=1'" >>/etc/mock/default.cfg
fi
```

allisonkarlitskaya:
→ #2059

PR #2077, opened 2021-06-04
images: Use only first link when looking for a bootstrap image

mvollmer:
* [x] image-refresh fedora-34

Some mirrors return directory listings that have two links for each
entry, like this:

```
   <tr class="even">
    <td class="indexcolicon">
     <a href="Fedora-Cloud-Base-34-1.2.x86_64.qcow2"><img src="/icons2/drive-harddisk.png" alt="[VM]"></a>
    </td>
    <td class="indexcolname">
     <a href="Fedora-Cloud-Base-34-1.2.x86_64.qcow2">Fedora-Cloud-Base-34-1.2.x86_64.qcow2</a>
    </td>
    <td class="indexcollastmod">2021-04-23 12:59  </td><td class="indexcolsize">249M</td>
   </tr>
```

Let's make sure there is only one word in the IMAGES variable in this case.

### DIFF
--- a/images/scripts/fedora-33.bootstrap
+++ b/images/scripts/fedora-33.bootstrap
@@ -20,7 +20,7 @@
 set -eux
 
 URL='https://download.fedoraproject.org/pub/fedora/linux/releases/33/Cloud/x86_64/images/'
-IMAGE="$(curl -L -s "$URL" | grep -o '"Fedora-Cloud-Base-[^"]*.qcow2"' | tr -d '"')"
+IMAGE="$(curl -L -s "$URL" | grep -o '"Fedora-Cloud-Base-[^"]*.qcow2"' | tr -d '"' | head -n1)"
 [ -n "$IMAGE" ]
 
 exec $(dirname $0)/lib/cloudimage.bootstrap "$1" "$URL/$IMAGE"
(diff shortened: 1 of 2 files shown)


cockpituous:
Log: https://logs.cockpit-project.org/logs/image-refresh-2077-20210604-114848/

mvollmer:
Let's make a fedora-34 image as proof that this works.

cockpituous:
image-refresh fedora-34 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-34-20210604-120606

images/scripts/fedora-33.bootstrap
```diff
@@ -20,7 +20,7 @@
 set -eux
 
 URL='https://download.fedoraproject.org/pub/fedora/linux/releases/33/Cloud/x86_64/images/'
-IMAGE="$(curl -L -s "$URL" | grep -o '"Fedora-Cloud-Base-[^"]*.qcow2"' | tr -d '"')"
```
allisonkarlitskaya, line 23: how about `grep -m 1`?

Bikeshedding at this point, really, but maybe you didn't know about it.

mvollmer: Oh, I didn't know about -m, indeed. But it doesn't seem to work with multiple matches in a single line. (The example HTML in the description is in fact all in a single line. I had reformatted it to make it easier to read, but I probably shouldn't have...)

```
$ echo abcabc | grep a -o -m 1
a
a
```

martinpitt (approved):
Right, `-o` and `-m` don't work well together. So let's land this for now. Thanks!

PR #2086, opened 2021-06-07
Image refresh for rhel-9-0

cockpituous:
Image refresh for rhel-9-0
 * [x] image-refresh rhel-9-0
 * [x] https://github.com/cockpit-project/cockpit-machines/pull/198


mvollmer:
Most of the issues are https://bugzilla.redhat.com/show_bug.cgi?id=1965743... Hmm, I am starting to think we should indeed just allow the selinux denial message and carry on. The bug is making its way into all the rhelish/fedoralic images...

mvollmer:
The TestSystemInfo.testCPUSecurityMitigationsEnable might be interesting. I take a look.

mvollmer:
> The TestSystemInfo.testCPUSecurityMitigationsEnable might be interesting. I take a look.

Ok, quick summary:

The "command line options" for a new entry in /boot/loader/entries come from /etc/kernel/cmdline, or from /proc/cmdline if the former does not exist.

Previous rhel-9-0 images do not have /etc/kernel/cmdline. Thus we get this sequence:

- `grubby --args=nosmt` adds "nosmt" to all entries.
- A reboot puts nosmt into /proc/cmdline
- kernel-install copies /proc/cmdline into the new entry in /boot/loader/entries.
- The test finds it there.

The new rhel-9-0 image here does have /etc/kernel/cmdline (no idea who creates it), thus we get this:

- `grubby --args=nosmt` adds "nosmt" to all entries, but not to /etc/kernel/cmdline.
- A reboot puts nosmt into /proc/cmdline
- kernel-install copies /etc/kernel/cmdline into the new entry in /boot/loader/entries, without "nosmt"
- The test doesn't find it.

Maybe we need to update our `kernelopt` script to also update /etc/kernel/cmdline.

Or maybe mgrubby should do that.

Or maybe using kernel-install is all wrong now and we should use something else. *shrug*

mvollmer:
test/verify/check-selinux TestSelinux.testTroubleshootAlerts needs https://github.com/cockpit-project/cockpit/pull/15908

mvollmer:
test/verify/check-storage-lvm2 TestStorageLvm2.testLvm fails because lvm2 doesn't tolerate a loopback device that had the name of its backing file removed (inode still there, of course). This used to work just fine.

mvollmer:
> test/verify/check-storage-lvm2 TestStorageLvm2.testLvm fails because lvm2 doesn't tolerate a loopback device that had the name of its backing file removed (inode still there, of course). This used to work just fine.

https://bugzilla.redhat.com/show_bug.cgi?id=1969408

mvollmer:
Blocked on https://github.com/cockpit-project/cockpit/pull/15910 for the lvm2 bug.

KKoukiou:
Regarding the machines failures:
https://github.com/cockpit-project/cockpit-machines/pull/198

KKoukiou:
For the podman failure: https://bugzilla.redhat.com/show_bug.cgi?id=1969996 and i will create a naughty.

martinpitt:
AttributeError: 'VirtMachine' object has no attribute 'allowed_messages'

Oops! This needs rebasing then. Pushed.

martinpitt (approved):

PR #2088, opened 2021-06-07
npm-update: write "/npm install" into the PR

allisonkarlitskaya:
When opening a PR, make sure we write "/npm install" there to trigger
the rebuild.


martinpitt (approved):
As discussed on IRC, that should actually work, thanks!

PR #2089, opened 2021-06-08
machine: Add allowed_messages method

mvollmer:
For returning image specific messages that should be ignored.  These
messages are mostly updated when refreshing images, so it is
convenient to have them in the same repo as the images.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -203,6 +203,35 @@ class Machine(ssh_connection.SSHConnection):
             messages = []
         return messages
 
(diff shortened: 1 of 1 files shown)


mvollmer:
Corresponding change in `testlib.py`: https://github.com/cockpit-project/cockpit/pull/15906

Proof that this all works: https://logs.cockpit-project.org/logs/pull-15906-20210608-085119-919197ff-fedora-34-bots-2089/log.html

martinpitt (approved):
Danke!

PR #2092, opened 2021-06-08
task: add ~/.config/cockpit-dev/allowed-users

allisonkarlitskaya:
Add a local file which can be used as an alternative to querying
permitted users for test runs from GitHub (which requires a fairly
powerful token).

Putting your own GitHub login in your allowed-users file, for example,
will allow you to use tests-scan or tests-trigger on your own pull
requests, with a token that has only the repo:status scope.

The permitted users must be separated by whitespace, eg. one per line.

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -392,6 +392,15 @@ class GitHub(object):
         raise KeyError("Team {0} not found".format(name))
 
     def is_user_allowed(self, user):
+        # First, check the local userlist, if it exists:
+
+        try:
+            with open(xdg_config_home('cockpit-dev', 'allowed-users'), "r") as f:
+                if user in f.read().split():
+                    return True
+        except FileNotFoundError:
+            pass
+
         # Firstly check if user has push access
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
For the record: I considered using (and very much wanted to use) the `author_association` field in the pull request. but it seems to be pretty unreliable.  Sometimes it's contributor, sometimes member.  Either way, someone who contributed one commit shouldn't be treated on the same level as a team member, ...

martinpitt:
One would assume that `read:org` permissions would be sufficient, but annoyingly they aren't: "Must have push access to view collaborator permission."  But this is because `is_user_allowed()` checks the project specific collaborators first, and the intended fallback is broken.

I really don't like this approach -- it spreads that kind of user ACL list to local machines which makes it redundant and outdated. I sent PR #2093 with a fix which hopefully also works for you -- want to have a look?

allisonkarlitskaya:
I also don't like this approach.  It's lame.

PR #2093, opened 2021-06-09
task: Fix is_user_allowed() for low-privileged tokens

martinpitt:
If the token does not have `public_repo` permissions, reading the
project collaborators will fail. This caused a `GitHubError` crash
and prevented the intended fallback to self.allowlist(). Turn that
exception into a warning instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,9 @@ the [GitHub CLI](https://cli.github.com/) configuration in
 
     https://github.com/settings/tokens
 
-When generating a new personal access token, the scope only needs to
-encompass `public_repo` (or `repo` if you're accessing a private repo).
+When generating a new personal access token, the scope should only contain
+`repo:status` and `read:org`.  Note in particular, that `repo` and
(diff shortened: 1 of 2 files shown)


martinpitt:
I removed `public_repo` from my ~/.config/github-token permissions, and with this fix I can now run tests-trigger.

marusak (approved):
Indeed, I saw this exception before as well. Thanks!

allisonkarlitskaya (changes requested):
The concept is good, and it works for me, but two things are missing here:
 - the error output is really too much for an expected failure in the default configuration.  I'd like to reverse the order of the checks.
 - we need to document the expected token permissions as being `repo:status` and `read:org` now

martinpitt:
@allisonkarlitskaya : LGTM and WFM :+1: (I can't self-approve, please do it in my stead)

allisonkarlitskaya (approved):

PR #2095, opened 2021-06-09
npm-update: only write "/npm install" for gitlink node_modules

allisonkarlitskaya:
Check if `node_modules/` is a gitlink before writing "/npm install" into
PRs on npm updates.

### DIFF
--- a/npm-update
+++ b/npm-update
@@ -94,6 +94,9 @@ def run(specified_package, verbose=False, **kwargs):
                     if task.verbose:
                         sys.stderr.write("Ignoring '{0}' as there is pending PR #{1}\n".format(pkg, issue["number"]))
 
+    # The only time git will checkout an empty subdir is if it's a gitlink
+    is_gitlink = not subprocess.call(['rmdir', 'node_modules'])
+
     # Force all current dependencies in place
     execute("npm", "install")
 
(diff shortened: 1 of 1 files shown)


npm-update
```diff
@@ -94,6 +94,9 @@ def run(specified_package, verbose=False, **kwargs):
                     if task.verbose:
                         sys.stderr.write("Ignoring '{0}' as there is pending PR #{1}\n".format(pkg, issue["number"]))
 
+    # The only time git will checkout an empty subdir is if it's a gitlink
+    is_gitlink = not subprocess.call(['rmdir', 'node_modules'])
```
martinpitt, line 98: Seems a bit strange to not use [os.rmdir()](https://docs.python.org/3/library/os.html#os.rmdir) instead (and catch `OSError`), but sure.

martinpitt (approved):
Thanks! That's indeed much more elegant than hardcoding project names.

npm-update
allisonkarlitskaya: Eh.  I didn't feel like dealing with the try/catch nonsense and also avoiding a blanket catch by catching both the "Directory is full" and "Directory doesn't exist" cases... etc etc.  I started coding that, and it sucked.

I also tried to query the information directly out of git but that's not so elegant either.

I think this is fine.

PR #2096, opened 2021-06-09
task: Adjust webhook IP

martinpitt:
I botched the setup of the previous public IP, it was not actually
stable. This one is.


allisonkarlitskaya (approved):
OK :)

PR #2099, opened 2021-06-10
Image refresh for fedora-34

cockpituous:
Image refresh for fedora-34
 * [x] image-refresh fedora-34
 * [x] crun fix publishes into rawhide: https://bodhi.fedoraproject.org/updates/FEDORA-2021-3526832553
 * [x] update crun to have a higher version number: https://bugzilla.redhat.com/show_bug.cgi?id=1970740


marusak:
podman is busted, see https://github.com/containers/podman/issues/10623

This is supposed to be fixed in `crun-0.20.1.1-0.13.git7adf4d0.fc35` but previous release of crun was `crun-0.20.3-0.12.git8d6a8b5.fc35 ` which has higher release, doesn't it? (In any case, currently prepared rawhide installed `crun-0.20.3-0.12.git8d6a8b5.fc35.x86_64` either because it has higher version or because the new release is not yet visible in mirrors.

martinpitt:
Confirming that podman/rawhide does not fail with the current fedora-34 image, as we don't reboot the VM after installing the rawhide kernel. That older kernel does not yet have the change that breaks podman. So keeping blocked.

martinpitt:
I filed https://bugzilla.redhat.com/show_bug.cgi?id=1970740 about the broken version number.

martinpitt (approved):
podman/rawhide works fine now, the previous crun version now apparently got sufficiently "unpublished".

PR #2107, opened 2021-06-11
Revert: task: Move default AMQP server to AWS fallback

martinpitt:
CentOS CI is back online, and the webhook works much more reliably
there.

This reverts commit 836457afd43b1971, but keeps the fallback commented
out for the future.


martinpitt:
Self-approval alert. But I'd like to do this before everyone else comes online, and it's a trivial ops.

PR #2109, opened 2021-06-11
machine: Loop until all data has been written to stderr

mvollmer:


mvollmer:
Hmm, technically, we should also check the return value to deal with partial writes, no...?

machine/machine_core/ssh_connection.py
```diff
@@ -336,7 +336,13 @@ def execute(self, command=None, script=None, input=None, environment={},
                             rset.remove(stderr_fd)
                             proc.stderr.close()
                         elif not quiet or self.verbose:
-                            os.write(sys.__stderr__.fileno(), data)
+                            while data:
+                                try:
+                                    os.write(sys.__stderr__.fileno(), data)
+                                    break
+                                except BlockingIOError as e:
+                                    data = data[e.characters_written:]
```
martinpitt, line 344: Sorry, I just sent #2110 and we collided.

Curious how the exception could know the number of bytes written? os.write() *either* returns the number of bytes written for a short write, *or* with -1/EAGAIN for blocking I/O. How could these happen both at the same time?

Either way, if that's a thing, it's very wise of course to take that into account. However, let's combine this with my approach of providing a new function, and use that in all three places?

mvollmer: Good point, I cargo culted this from run-tests.

mvollmer:
Let's go with #2110.

machine/machine_core/ssh_connection.py
martinpitt: If we use that, we need to use `__hasattr__` or so, as the [documentation](https://docs.python.org/3/library/exceptions.html#BlockingIOError) says that it's not always present.

martinpitt:
> should also check the return value to deal with partial writes, no

Yes, agreed. I'll rework my PR then.

PR #2110, opened 2021-06-11
machine_core: Handle EAGAIN from os.write()

martinpitt:
retry, instead of crashing.

----

Recently this started happening a *lot*, e.g. [here](https://logs.cockpit-project.org/logs/pull-15924-20210611-103243-42e6e0c6-fedora-coreos/log.html) or [here](https://logs.cockpit-project.org/logs/pull-15933-20210611-101052-86488262-fedora-33-container-bastion/log.html)

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -30,6 +30,16 @@ from . import exceptions
 from . import timeout as timeoutlib
 
 
+def write_all(fd, data):
+    while len(data) > 0:
+        try:
+            written = os.write(fd, data)
+            data = data[written:]
+        except BlockingIOError:
+            time.sleep(0.1)
+            pass
+
+
 class SSHConnection(object):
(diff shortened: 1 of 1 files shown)


machine/machine_core/ssh_connection.py
```diff
@@ -30,6 +30,16 @@
 from . import timeout as timeoutlib
 
 
+def retry_write(fd, data):
```
mvollmer, line 33: I would call this "write_all" and also make it handle partial writes.

mvollmer (changes requested):

machine/machine_core/ssh_connection.py
martinpitt: Agreed, done. This now incorporates your idea from #2109.

mvollmer (approved):

martinpitt:
Moving to an origin branch to [work around this failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-2110-20210611-114718-9a6f51f2-fedora-33-firefox-cockpit-project-cockpit/log.html) until https://github.com/cockpit-project/cockpit/pull/15936 lands.

PR #2111, opened 2021-06-11
machine_core: Handle EAGAIN from os.write() [sleep version]

martinpitt:
retry, instead of crashing. Also handle partial writes (thanks Marius
for the idea).

----

Recently this started happening a *lot*, e.g. [here](https://logs.cockpit-project.org/logs/pull-15924-20210611-103243-42e6e0c6-fedora-coreos/log.html) or [here](https://logs.cockpit-project.org/logs/pull-15933-20210611-101052-86488262-fedora-33-container-bastion/log.html)

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -30,6 +30,16 @@ from . import exceptions
 from . import timeout as timeoutlib
 
 
+def write_all(fd, data):
+    while len(data) > 0:
+        try:
+            written = os.write(fd, data)
+            data = data[written:]
+        except BlockingIOError:
+            time.sleep(0.1)
+            pass
+
+
 class SSHConnection(object):
     ssh_default_opts = ["-o", "StrictHostKeyChecking=no", "-o", "UserKnownHostsFile=/dev/null", "-o", "BatchMode=yes"]
 
@@ -328,7 +338,7 @@ class SSHConnection(object):
                             proc.stdout.close()
                         else:
                             if self.verbose:
-                                os.write(sys.__stdout__.fileno(), data)
+                                write_all(sys.__stdout__.fileno(), data)
                             output += data.decode('utf-8', 'replace')
                     elif fd == stderr_fd:
                         data = os.read(fd, 1024)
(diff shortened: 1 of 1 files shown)


mvollmer (approved):

machine/machine_core/ssh_connection.py
```diff
@@ -328,19 +338,19 @@ def execute(self, command=None, script=None, input=None, environment={},
...
                             rset.remove(stderr_fd)
                             proc.stderr.close()
                         elif not quiet or self.verbose:
-                            os.write(sys.__stderr__.fileno(), data)
+                            write_all(sys.__stderr__.fileno(), data)
                 for fd in ret[1]:
                     if fd == stdin_fd:
                         if input:
-                            num = os.write(fd, input.encode('utf-8'))
+                            num = write_all(fd, input.encode('utf-8'))
```
mvollmer, line 353: This should not be replaced, as we do expect it to have partial writes.

mvollmer (changes requested):

machine/machine_core/ssh_connection.py
```diff
@@ -30,6 +30,16 @@
 from . import timeout as timeoutlib
 
 
+def write_all(fd, data):
+    while len(data) > 0:
+        try:
+            written = os.write(fd, data)
+            data = data[written:]
+        except BlockingIOError:
+            time.sleep(0.1)
```
allisonkarlitskaya, line 39: :frowning_face:

allisonkarlitskaya (changes requested):

allisonkarlitskaya:
Two alternatives:
 1. just shove the output back to blocking.  if node can mess it up in the first place, then why are we so careful not to put it back?
 1. `poll()`

machine/machine_core/ssh_connection.py
martinpitt: Argh, of course. I fixed it to just retry on BlockingIOError.

machine/machine_core/ssh_connection.py
martinpitt: Sure, it's not nice -- this is a quickfix to unbreak stuff. I am not against rewiring this entire messy function, it's just not something that I want to do under pressure.

allisonkarlitskaya, line 39:
```suggestion
        select.select([], [fd], [])
        written = os.write(fd, data)
        data = data[written:]
```

martinpitt: Thanks! Sent to PR #2112 . If that works (hopefully should), let's land that one. Otherwise this will be ready to land.

PR #2112, opened 2021-06-11
machine_core: Handle EAGAIN from os.write() [select version]

martinpitt:
retry, instead of crashing. Also handle partial writes (thanks Marius
for the idea).

----

Alternative to #2111. If that works, let's land that one.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -30,6 +30,13 @@ from . import exceptions
 from . import timeout as timeoutlib
 
 
+def write_all(fd, data):
+    while len(data) > 0:
+        select.select([], [fd], [])
+        written = os.write(fd, data)
+        data = data[written:]
+
+
 class SSHConnection(object):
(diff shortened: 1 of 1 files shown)


machine/machine_core/ssh_connection.py
```diff
@@ -328,19 +335,22 @@ def execute(self, command=None, script=None, input=None, environment={},
...
                             proc.stderr.close()
                         elif not quiet or self.verbose:
-                            os.write(sys.__stderr__.fileno(), data)
+                            write_all(sys.__stderr__.fileno(), data)
                 for fd in ret[1]:
                     if fd == stdin_fd:
                         if input:
-                            num = os.write(fd, input.encode('utf-8'))
+                            try:
+                                num = os.write(fd, input.encode('utf-8'))
```
allisonkarlitskaya, line 351: this slipped in again?

allisonkarlitskaya (changes requested):

machine/machine_core/ssh_connection.py
allisonkarlitskaya: I don't like this → #2113.

PR #2113, opened 2021-06-11
machine_core: Handle EAGAIN from os.write() [select version 2]

martinpitt:
Writing larger blocks to the sink often causes `EAGAIN`. Handle that by
using select() for waiting for write-ability (thanks Allison).
Also handle partial writes (thanks Marius for the idea).


allisonkarlitskaya (approved):
pikachu, i choose you!

PR #2119, opened 2021-06-15
image-download: Fix locking across containers

### LINKED ISSUE #1473: image-download in separate containers on shared cache directory is failing
In #1472 I made a change and triggered a refresh.  The refresh happened to run on the centos cluster, and the image was available there.  I cherry-picked the refresh commit into the PR and pushed again, then triggered the entire testmap for `image:fedora-32`.

That results in an awful lot of tests spinning up on various containers on the cockpit cluster, all of which are going to start the large download from the centos cluster.  Some of these containers ended up on the same host, sharing cache directories and out of those, some tests failed like this:

```
 > https://images-frontdoor.apps.ocp.ci.centos.org/
 x https://images-cockpit.apps.ci.centos.org/
 x https://cockpit-11.e2e.bos.redhat.com:8493
 x https://10.29.163.169:8493/


=> https://images-frontdoor.apps.ocp.ci.centos.org/fedora-32-9a4530a00b40f734b97cce0dcbfa399b1e2153d9a6103c2e76c540d4c52b23c3.qcow2 is the fastest store
image-download: [Errno 2] No such file or directory: '/cache/images/fedora-32-9a4530a00b40f734b97cce0dcbfa399b1e2153d9a6103c2e76c540d4c52b23c3.qcow2.https:__images-frontdoor.apps.ocp.ci.centos.org_.partial'
image-prepare: unable to download all necessary images
Test run finished, return code: 1
```

For each host where there was a failing container (1-, 4- and 9-cockpit), there was a passing test in a different container on that same host.

That sounds a lot like image-download falling over itself.  Probably we need to up our locking game, or do some other approach.

martinpitt:
Creating the lock in /tmp/.cockpit-test-resources causes multiple
parallel downloads and crashes (due to a race condition with accessing
the .partial file concurrently) in our bots, where several tasks
containers share the same /cache/images volume mount.

Fix that by creating the lock file right next to the downloaded target
file, so that they get shared across containers. Clean it up after
download() succeeds, to avoid cluttering the cache dir.

Fixes #1473

-----

See the issue #1473 for details about the investigation, reproducer commands, and considering alternative implementations.

### DIFF
--- a/image-download
+++ b/image-download
@@ -37,7 +37,6 @@ import shutil
 import stat
 import subprocess
 import sys
-import tempfile
 import time
 import fcntl
 import urllib.parse
@@ -248,8 +247,7 @@ def committed_target(image):
 
 
 def wait_lock(target):
-    lockfile = os.path.join(tempfile.gettempdir(), ".cockpit-test-resources", os.path.basename(target) + ".lock")
-    os.makedirs(os.path.dirname(lockfile), exist_ok=True)
+    lockfile = os.path.join(os.path.dirname(target), ".lock." + os.path.basename(target))
 
     # we need to keep the lock fd open throughout the entire runtime, so remember it in a global-scoped variable
     wait_lock.f = open(lockfile, "w")
@@ -263,6 +261,8 @@ def wait_lock(target):
             time.sleep(10)
     else:
         raise TimeoutError("timed out waiting for concurrent downloads of %s\n" % target)
+
+    return lockfile
 
 
 def download_images(image_list, force, quiet, state, store):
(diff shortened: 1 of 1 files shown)


martinpitt:
The [test/cockpituous](https://github.com/cockpit-project/bots/pull/2119/checks?check_run_id=2826233492) does not work any more. This does not run very often. I noticed that last week already, but did not yet get to fixing that since then. I'll address this in a separate PR.

image-download
```diff
@@ -292,10 +292,16 @@ def download_images(image_list, force, quiet, state, store):
...
             # don't download the same thing multiple times in parallel
-            wait_lock(target)
+            lockfile = wait_lock(target)
 
             if force or state or not os.path.exists(target):
                 download(target, force, state, quiet, store)
+
+            # clean up lock file; parallel runs may remove it already
+            try:
+                os.unlink(lockfile)
```
allisonkarlitskaya, line 302: Okay, so the problem here is that unlinking lockfiles isn't really safe: someone might have just created the file again, and then you `unlink()` it while they think they still hold an fd that refers to it...

In practice, though, eh.... this is gonna happen *so* incredibly rarely, and this patch is a massive improvement compared to the current situation.

allisonkarlitskaya (approved):

PR #2120, opened 2021-06-16
testmap: add new _manual triggers for cockpit

allisonkarlitskaya:
... fedora-34/devel and fedora-34/devel+firefox.
    
Those are part of cockpit-project/cockpit#15959.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -56,6 +56,8 @@ REPO_BRANCH_CONTEXT = {
             'fedora-testing',
             'fedora-testing/dnf-copr',
             'rhel-9-0-distropkg',
+            'fedora-34/devel',
+            'fedora-34/devel+firefox',
         ],
     },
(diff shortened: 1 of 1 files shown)


martinpitt:
Normally we introduce these through `_manual` first, then you can conveniently test the new contexts in some PR with tests-trigger.

But let me do some black magic to trigger them on https://github.com/cockpit-project/cockpit/pull/15959

allisonkarlitskaya:
This is pretty clearly broken.  The firefox scenario finished immediately, with success.  Will do some more local testing first...

martinpitt:
@allisonkarlitskaya : I triggered the real test in https://github.com/cockpit-project/cockpit/pull/15959 . I'm not sure what exactly you tried here with the pull/xxx/head stuff, but I've never seen it before. Our CI just doesn't support this.

martinpitt (approved):

PR #2123, opened 2021-06-17

cockpituous:


machine/machine_core/machine.py
```diff
@@ -244,6 +244,10 @@ def get_admin_group(self):
         else:
             return "wheel"
 
+    def get_cockpit_container(self):
+        tag = "released" if "released" in self.execute("podman ps") else "latest"
```
martinpitt, line 248: I don't understand this -- do we ever actually start the `:released` container in our tests? fedora-coreos.install uses this to back up the original container before refreshing it, but tests are always supposed to use `:latest`.

KKoukiou: Then I misunderstood the failure from https://logs.cockpit-project.org/logs/pull-2123-20210621-094713-4e63e07a-fedora-coreos-cockpit-project-cockpit/log.html#244-2

Will adjust, thanks for the pointer.

martinpitt: NB I'm not saying it's wrong, I was just wondering where that comes from.

KKoukiou: So the install script of fedora-coreos tags the 'released' image as 'latest', so even when we see released in the podman ps it's actually the latest. I just realized it, i was quite confused.

machine/machine_core/machine.py
```diff
@@ -244,6 +244,9 @@ def get_admin_group(self):
         else:
             return "wheel"
 
+    def get_cockpit_container(self):
+        return self.execute("podman ps --quiet --all --filter name=cockpit/ws").strip()
```
martinpitt, line 248: FTR, I originally used ancestor to match on the image name instead of the container name. I was not sure at that time whether we could rely on any particular name, as a simple `podman run <image>` usually creates a non-predictable random name. These days we should use the `RUN` label and a fixed name, so good enough for me if that works.

Conceptually, filtering on `*/cockpit/ws` image name would still be more robust.

machine/machine_core/machine.py
```diff
@@ -287,7 +291,8 @@ def restart_cockpit(self):
         if self.ostree_image:
             # HACK: podman restart is broken (https://bugzilla.redhat.com/show_bug.cgi?id=1780161)
             # self.execute("podman restart `podman ps --quiet --filter ancestor=cockpit/ws`")
-            tls = "--no-tls" not in self.execute("podman inspect `podman ps --quiet --filter ancestor=cockpit/ws`")
+            inspect_res = self.execute("podman inspect `echo {0}`".format(self.get_cockpit_container()))
```
martinpitt, line 294: `echo x` is the same as just x.

machine/machine_core/machine.py
```diff
@@ -298,7 +303,7 @@ def stop_cockpit(self):
         """Stop Cockpit.
         """
         if self.ostree_image:
-            self.execute("podman ps --quiet --all --filter ancestor=cockpit/ws | xargs --no-run-if-empty podman rm -f")
+            self.execute("echo {0} | xargs --no-run-if-empty podman rm -f".format(self.get_cockpit_container()))
```
martinpitt, line 306: Likewise, ther can only ever be one container with a given  name, so a simple `f"podman rm -f {self.get_cockpit_container()}"` would do.

KKoukiou: tests did not succeed - reposting with suggestions

machine/machine_core/machine.py
martinpitt: Note that the name is just `ws`, not `cockpit/ws`.

KKoukiou: Image name filter does not accept regexp unfortunately - according to the manpage.

machine/machine_core/machine.py
KKoukiou: I think the xargs is needed because sometimes podman ps command returns an empty list.

martinpitt: Of course you are right.

machine/machine_core/machine.py
```diff
@@ -257,6 +260,7 @@ def start_cockpit(self, atomic_wait_for_host=None, tls=False):
             if not tls:
                 cmd += " -- --no-tls"
             self.execute(cmd)
+
```
KKoukiou, line 263: Ups sorry for the new line here.

martinpitt:
The [fedora-coreos failure](https://logs.cockpit-project.org/logs/pull-2123-20210629-090741-c54a6342-fedora-coreos-cockpit-project-cockpit/log.html) got fixed in cockpit/main in the meantime.

The change looks a bit brittle (starting to hardcode the `ws` name), but good enough for now for sure, thanks for tracking this down! @KKoukiou, do you mind rebasing/force-pushing this again, to not ruin our test stats any further? (feel free to clean up the empty line, although that's utterly unimportant)

martinpitt (approved):
Danke!

Issue #2141, opened 2021-06-24
avc: denied { unlink } for pid=1 comm="systemd" name="default-hostname" dev="tmpfs"


martinpitt:
Spotted sporadically during fedora-34 test runs. Totally unrelated to the actual test.

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1975811

cockpituous:
fedora-34
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testTracer (__main__.TestUpdates)
rm: cannot remove '/etc/systemd/system/dnf-automatic*': No such file or directory
[0624/140819.633923:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[0624/140819.635753:WARNING:headless_browser_main_parts.cc(83)] Cannot create Pref Service with no user data dir.

DevTools listening on ws://127.0.0.1:9616/devtools/browser/da9e3cf0-1a15-41e8-82f4-f395be3dee20
[0624/140819.650909:ERROR:gpu_init.cc(426)] Passthrough is not supported, GL is egl
[...]

PR #2142, opened 2021-06-24
naughty: Add SELinux message for systemd unlinking /run/systemd/default-hostname

martinpitt:
Known issue #2141
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1975811

----

Observed [here](https://logs.cockpit-project.org/logs/pull-15987-20210624-125257-823fc21a-fedora-34/log.html#101).

### DIFF
--- a/naughty/fedora-34/2141-selinux-unlink-default-hostname
+++ b/naughty/fedora-34/2141-selinux-unlink-default-hostname
@@ -1,0 +1,1 @@
+type=1400 audit(*): avc:  denied  { unlink } for  pid=1 comm="systemd" name="default-hostname" dev="tmpfs"


allisonkarlitskaya (changes requested):
filename: `12141` → `2141`.

allisonkarlitskaya (approved):

martinpitt:
@allisonkarlitskaya : Argh, thanks, fixed. I go and mow the lawn now, I can make fewer typos there :grin:

allisonkarlitskaya:
I don't feel completely awesome about this naughty because we have no idea about how frequently it will pop up, and so it will be difficult to remove it.  But let's let the bots tell us if they think it's stale, and in the meantime, it might save us from seeing some red.

martinpitt:
@allisonkarlitskaya : Right, same here -- but at least tracking it that way will give us some idea how frequently it actually happens. I also wait for the test run to finish and then attach the journal to the downstream bug -- that may have some further insights what actually happened.

allisonkarlitskaya:
I take it back, actually.  We just hit the exact same fail again on the retry in cockpit-project/cockpit#15987, so I guess we're gonna see this one more often...

Maybe it's an selinux violation if the working directory contains instructions about installing PAM modules :wink:

Issue #2153, opened 2021-06-28
rhel-9: libvirt detach device fails


KKoukiou:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1976945

KKoukiou:
This was wrong way of using the API from our side.

PR #2154, opened 2021-06-28
naughty: add pattern for libvirt failing to detach interface on RHEL-9

KKoukiou:
This is supposed to fix the issue on https://logs.cockpit-project.org/logs/pull-224-20210628-162525-03f76900-rhel-9-0/log.html#60

I tried to verify with `./tests-policy -o --simple -v image:rhel-9-0 < /tmp/err` where `/tmp/err` contains the output from the failure in the link above, but it does not work.


KKoukiou:
https://github.com/cockpit-project/bots/issues/2153

PR #2159, opened 2021-06-30
testmap: run firefox scenario on fedora-34 image

KKoukiou:


martinpitt (approved):
Thanks! Fine if/when green.

FYI, you don't need no-test any more in bots.

PR #2167, opened 2021-07-02
naughty: #2141 also happens on Fedora CoreOS

martinpitt:
[example](https://logs.cockpit-project.org/logs/pull-16033-20210701-153132-839e602e-fedora-coreos/log.html#266), seen in https://github.com/cockpit-project/cockpit/pull/16033


allisonkarlitskaya (approved):
I think this is OK, particularly since it's just copying the existing one.

In the future, though, maybe we should approach "Test passed, but with unexpected journal messages" by actually matching on the string that said the test passed.  This would avoid us inadvertently allowing other (actual) failures, if the journal message somehow otherwise ends up in the log.

PR #2160, opened 2021-07-01
Image refresh for centos-8-stream

cockpituous:
Image refresh for centos-8-stream
 * [x] image-refresh centos-8-stream
 * [x] add allowed message for new SELinux regression: PR #2156
 * [x] https://github.com/cockpit-project/cockpit/pull/16050
 * [x] https://github.com/cockpit-project/starter-kit/pull/473

PR #2168, opened 2021-07-02
machine: Don't send invalid DHCP hostname with colons

### LINKED ISSUE #1908: images with hostname "localhost" try to set an invalid hostname
While debugging with @martinpitt the issue with subscription-manager tests with RHEL 8.4 (#1711), I noticed this in the logs:
```
Apr 05 22:15:20 localhost NetworkManager[868]: <warn>  [1617675320.8611] hostname: couldn't set the system hostname to 'rhel-8-4-127.0.0.2\0582201' using hostnamed: GDBus.Error:org.freedesktop.DBus.Error.InvalidArgs: Invalid hostname 'rhel-8-4-127.0.0.2\0582201'
Apr 05 22:15:20 localhost NetworkManager[868]: <warn>  [1617675320.8612] policy: set-hostname: couldn't set the system hostname to 'rhel-8-4-127.0.0.2\0582201': (1) Operation not permitted
Apr 05 22:15:20 localhost NetworkManager[868]: <warn>  [1617675320.8629] policy: set-hostname: you should use hostnamed when systemd hardening is in effect!
```
(from https://logs.cockpit-project.org/logs/pull-2557-20210408-110040-fee291b6-rhel-8-4/log.html )

Indeed `rhel-8-4-127.0.0.2\0582201` is not a valid hostname.

martinpitt:
`control` is "127.0.0.1:<port>". Using that as DHCP-announced host name
is invalid, as `:` must not appear  in a hostname, they caused
NetworkManager/hostnamed to groan:

    policy: set-hostname: current hostname was changed outside NetworkManager: 'localhost'
    policy: set-hostname: set hostname to 'rhel-8-4-127.0.0.2\0582201' (from DHCPv4)
    hostname: couldn't set the system hostname to 'rhel-8-4-127.0.0.2\0582201' using hostnamed:
      GDBus.Error:org.freedesktop.DBus.Error.InvalidArgs: Invalid hostname 'rhel-8-4-127.0.0.2\0582201'

Replace the colon with a dash. Also rename the "name" key to "hostname"
to make it more obvious what this is for.

Fixes #1908

 - [x] https://github.com/cockpit-project/cockpit/pull/16043

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -96,7 +96,7 @@ TEST_DOMAIN_XML = """
     {ethernet}
     <qemu:arg value='-netdev'/>
     <qemu:arg value='user,id=base0,restrict={restrict},net=172.27.0.0/24,""" \
-"""dnssearch=loopback,hostname={name},{forwards}'/>
+"""dnssearch=loopback,hostname={hostname},{forwards}'/>
     <qemu:arg value='-device'/>
     <qemu:arg value='virtio-net-pci,netdev=base0,bus=pci.0,addr=0x0e'/>
   </qemu:commandline>
@@ -346,7 +346,7 @@ class VirtMachine(Machine):
             sys.stderr.write("WARNING: Machine will run about 10-20 times slower\n")
 
         keys.update(self.networking)
-        keys["name"] = "{image}-{control}".format(**keys)
+        keys["hostname"] = keys["image"] + '-' + keys["control"].replace(':', '-').replace('.', '-')
         test_domain_desc = TEST_DOMAIN_XML.format(**keys)
 
         # add the virtual machine
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -95,7 +95,8 @@ TEST_DOMAIN_XML = """
   <qemu:commandline>
     {ethernet}
     <qemu:arg value='-netdev'/>
-    <qemu:arg value='user,id=base0,restrict={restrict},net=172.27.0.0/24,hostname={name},{forwards}'/>
+    <qemu:arg value='user,id=base0,restrict={restrict},net=172.27.0.0/24,""" \
+"""dnssearch=loopback,hostname={name},{forwards}'/>
     <qemu:arg value='-device'/>
     <qemu:arg value='virtio-net-pci,netdev=base0,bus=pci.0,addr=0x0e'/>
   </qemu:commandline>


allisonkarlitskaya (commented):
Couldn't we send something more like an actual hostname?  This doesn't look very useful...

martinpitt:
@allisonkarlitskaya : It matches exactly what you see in `virsh list`. Also, it's not actually that important -- all our images have /etc/hostname, so the net effect of this is almost zero (it only gets rid of the journalctl warnings). See the linked issue for details.

martinpitt:
I can locally reproduce [these rhel-8-5 failures](https://logs.cockpit-project.org/logs/pull-2168-20210702-072811-94c81ca7-rhel-8-5-cockpit-project-cockpit/log.html), so needswork. Somehow the transient host name confuses the tests.

martinpitt:
Indeed rhel-8-5 uses the DHCP provided name, but can't resolve it because it's not in /etc/hosts, and `myhostname` comes after `dns` in `nsswitch.conf`, and our tests are offline so that DNS does not respond. I. e. `getent hosts $(hostname)` hangs. Bummer.. So, choices:

 - Set RHEL image to a fixed host name (similar to Fedora) and add that to /etc/hosts
 - Flip the order of `dns myhostname` in nsswitch
 - Adjust individual tests to write host name to /etc/hosts

These all suck somewhat. We could also continue to ignore the problem and just send `localhost` through DHCP, but that feels even more wrong -- somehow we need to make our images fit for offline testing, so I am currently leaning towards the second one.

allisonkarlitskaya:
I like the second option.  It's pretty lame that `dns` comes before `myhostname` and this arrangement hangs on no network.  It seems almost like a downstream bug to be fixed...

martinpitt:
[testRaid](https://logs.cockpit-project.org/logs/pull-2168-20210702-072811-94c81ca7-rhel-8-5-cockpit-project-cockpit/log.html#183-2) is a harder issue: On current images it shows 

image

and then counts upward to propose "raid1" as new device name. But with a different host name it looks like

image

and thus our page somehow treats it different, even though it's still /dev/md/raid0.

This is an actual bug, let's fix this first.

allisonkarlitskaya:
My thoughts: the DNS being stuck isn't really an OS fail exactly: it's caused by qemu sending a DHCP config which provides:
 - no default route; but
 - a local network; and
 - a DNS server reachable in that network; and
 - the DNS server ignores requests

That is, indeed, a very broken network config.  We need to figure out how to get the (non-existent) DNS server out of the DHCP config, or to send one which is at least outside of the local subnet, so that we'll declare it unrouteable and fail fast.

martinpitt:
Fully agreed. I played around with various `-netdev user` options. Unfortunately `dns=`, `dns=0.0.0.0` etc. fails with `DNS doesn't belong to network`, so it's not that easy to disable. `dns=172.27.0.99` is accepted, but has no actual effect, resolv.conf still says `nameserver 172.27.0.3`.

There's another interesting possibility: I was also bothered by the bogus `search 0.0.2\0582201` option in /etc/resolv.conf. `dnssearch=loopback` generates a resolv.conf with just `search=loopback` and *no* `nameserver` option at all. `getent hosts $(hostname)` returns immediately, so this fixes the bug. I'll run some tests with that.

martinpitt:
Fully agreed. I played around with various `-netdev user` options. Unfortunately `dns=`, `dns=0.0.0.0` etc. fails with `DNS doesn't belong to network`, so it's not that easy to disable. `dns=172.27.0.99` is accepted, but has no actual effect, resolv.conf still says `nameserver 172.27.0.3`.

There's another interesting possibility: I was also bothered by the bogus `search 0.0.2\0582201` option in /etc/resolv.conf. `dnssearch=loopback` generates a resolv.conf with just `search=loopback` and *no* `nameserver` option at all. `getent hosts $(hostname)` returns immediately, so this fixes the bug. I'll run some tests with that.

martinpitt:
I now have some solution which is reasonably "nice and correct" (even though I don't fully understand why it works, that's some qemu internal). The [testRaid failure](https://logs.cockpit-project.org/logs/pull-2168-20210702-072811-94c81ca7-rhel-8-5-cockpit-project-cockpit/log.html#183-2) still stands (very very likely), that's something to fix on the cockpit side. But retrying the other tests to make sure this doesn't break anything else.

allisonkarlitskaya:
https://libvirt.org/formatnetwork.html#elementsAddress

```
dns
    The dns element of a network contains configuration information for the virtual network's DNS server Since 0.9.3.

    The dns element can have an optional enable attribute Since 2.2.0. If enable is "no", then no DNS server will be setup by libvirt for this network (and any other configuration in <dns> will be ignored). If enable is "yes" or unspecified (including the complete absence of any <dns> element) then a DNS server will be setup by libvirt to listen on all IP addresses specified in the network's configuration. 
```

Maybe this?

martinpitt:
@allisonkarlitskaya : We use QEMU socket networking, we can't (and don't) create a libvirt network on user libvirt. This requires bridges, and thus privileges, and none of that works in containers.

allisonkarlitskaya:
> @allisonkarlitskaya : We use QEMU socket networking, we can't (and don't) create a libvirt network on user libvirt. This requires bridges, and thus privileges, and none of that works in containers.

You mean the slirp usermode stuff, ya?

I didn't realise that this docs section referred only to bridge-mode...

Aside: we still have some vestiges of our bridge configuration setup in `vm-run`, for example (surely among other places).  Is there any reason for that anymore?  The hacks there, particularly with the qemu-bridge-helper, are impressively awful.  It might be nice to drop that?

martinpitt:
@allisonkarlitskaya : Yes, I think by now nobody uses/relies on the bridge stuff any more. We used to for a while for the Windows image, but first I fixed that, and second that windows image is gone. :grin:

martinpitt:
@allisonkarlitskaya : I sent PR #2178 to drop the bridge mode.

martinpitt:
The cockpit RAID fix landed, rebased and re-running.

martinpitt:
@jkozol , @allisonkarlitskaya : Argh, someone renamed cockpit-composer from master to main without updating the testmap -- [tests now fail](https://logs.cockpit-project.org/logs/pull-2168-20210705-144421-fdd174c3-rhel-8-4-osbuild-cockpit-composer/log.html). So again back to blocked, I'll send a PR.

martinpitt:
Yay, all green at last :partying_face:

allisonkarlitskaya (approved):
Still not thrilled about the hostname, but the "matches `virsh list`" argument is a pretty good one :)

allisonkarlitskaya:
Possible regression (via https://github.com/cockpit-project/bots/pull/2160): https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-2160-20210705-173357-7091a173-centos-8-stream-cockpit-project-starter-kit/log.html

PR #2169, opened 2021-07-02
github-upload-action-secrets: download right pubkey

allisonkarlitskaya:
We need to download the public-key specific to the endpoint where we'll
be uploading our secrets.  Before, there was only one endpoint (the
project) but now it might be an environment.

Download the public-key from the right place, based on the already-set
`secrets_path`.

### DIFF
--- a/github-upload-action-secrets
+++ b/github-upload-action-secrets
@@ -82,8 +82,6 @@ def main():
         # organization
         resource = '/orgs/' + opts.receiver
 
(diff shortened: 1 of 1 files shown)


martinpitt:
@allisonkarlitskaya already ran this to update the c-podman env. I re-ran https://github.com/cockpit-project/cockpit-podman/actions/runs/992946934 to ensure that this works now. It's past the "set up secrets" phase, and it's now full of :star2: !

martinpitt (approved):

PR #2178, opened 2021-07-05
Drop support for bridged networking

martinpitt:
We have not used this in a long time, all our project tests use QEMU
socket networking. We also don't want to, as  bridged networking
requires privileges (which we don't have in a container).


allisonkarlitskaya (approved):
:+2:

PR #2183, opened 2021-07-06
images: Build samba-domain container locally

martinpitt:
https://hub.docker.com/r/nowsci/samba-domain has not been updated in
over a year. To pick up OS updates, build it locally instead,  which
does not take long.

This also gets rid of the sync-quay machinery and usage of the
`QUAY_TOKEN` secret in the bots repo.

 * [x] image-refresh services
 * [x] https://github.com/cockpit-project/cockpit/pull/16052


cockpituous:
image-refresh in progress on 2-ci-srv-02.
Log: https://logs.cockpit-project.org/logs/image-refresh-2183-20210706-054949/

cockpituous:
Task failed: https://logs.cockpit-project.org/logs/image-refresh-2183-20210706-054949/

martinpitt:
Urgh, of course that container also builds on top of the ubuntu image published on dockerhub.. [log](https://logs.cockpit-project.org/logs/image-refresh-2183-20210706-054949/log)

cockpituous:
image-refresh in progress on 3-ci-srv-02.
Log: https://logs.cockpit-project.org/logs/image-refresh-2183-20210706-061818/

cockpituous:
Command '('git', 'push', 'https://github.com/cockpit-project/bots', '+HEADNone/heads/image-refresh-services-20210706-065555')' returned non-zero exit status 1.: https://logs.cockpit-project.org/logs/image-refresh-2183-20210706-061818/

martinpitt:
FTR, I just manually updated the image symlink in my branch. Not sure what went wrong with the git push, nothing changed in that regard. Let's hope for a temporary network glitch or so.

martinpitt:
The image refresh [breaks the cockpit Grafana test](https://logs.cockpit-project.org/logs/pull-2183-20210706-072952-7646a1fc-rhel-8-4-cockpit-project-cockpit/log.html#51-2)

martinpitt:
Grafana 8 fix landed, rebasing.

images/scripts/services.setup
```diff
@@ -71,8 +66,13 @@ poweroff
 #
 #############
 
-# See https://hub.docker.com/r/nowsci/samba-domain
-docker_pull nowsci samba-domain
+# See https://github.com/Fmstrat/samba-domain
+# image on dockerhub is too old, and not published on quay or anywhere else
+git clone --depth=1 https://github.com/Fmstrat/samba-domain /tmp/samba-domain
```
allisonkarlitskaya, line 71: I know that this is basically what we were doing before with pulling the latest version of whatever container image happened to be there, but for some reason, with git, this looks a lot worse.  Probably because the Dockerfile can also do stuff on the "host"?  (even though the "host" is, itself, just a VM)...

Just wanted to ask if you considered (and dismissed) the idea to pin a specific SHA here...

allisonkarlitskaya (approved):
This is a pretty cool simplification.  See one comment below about a possible improvement (but you may disagree).

One big question I have, though: why can't we just use samba from the host?  Or from a normal OS container with some (probably small?) config file injections?

images/scripts/services.setup
martinpitt: `podman build` also happens inside of a container. You mean that a Dockerfile can somehow break out of that?

I didn't consider a specific SHA, as with our test images we want the "latest" stuff for everything, and there is nothing interesting/secret in our test images. In general we could of course start from scratch from a Fedora or Ubuntu container, and inject some stuff -- but (1) this is not actually trivial, and (2) it's pretty much exactly what this repo does. In essence, it's "someone else already did that work".

martinpitt: > You mean that a Dockerfile can somehow break out of that?

If that was possible, how does dockerhub, quay, other services which build containers for you not hopelessly explode?

allisonkarlitskaya: so a Dockerfile, for example, can easily copy files from the host.  I'm not well-versed in all of the features of what you can do from there, but it's clear that there is at least some interaction with the host environment...

martinpitt: That way around, yes -- after all, the Dockerfile itself comes from the host. I'm just not aware of a container build being able to change anything on the host, and the VM host has no secrets.

allisonkarlitskaya: OK.  WMF.

PR #2187, opened 2021-07-07
workflows: Only run tests on pull requests

martinpitt:
Running them on both push and pull_request runs the integration test
twice in parallel, which collides on the "unit-tests" status -- only one
of these two is going to actually run the test and win the race, the
other will fail.

cockpituous does the same [1], and it works reliably there.

[1] https://github.com/cockpit-project/cockpituous/blob/main/.github/workflows/tests.yml

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,5 @@
 name: test
-on: [push, pull_request]
+on: [pull_request]
 jobs:
   bots:
     runs-on: ubuntu-20.04


allisonkarlitskaya (commented):
What if we limited `push` to `main`?  That seems to have been the original intent here, and should also be OK, no?

martinpitt:
@allisonkarlitskaya : I think the intent was maybe to run the tests for experimentation on some branch of your's *before* sending a PR. Limiting to main would be fine, I think -- just not sure anyone looks at these. (The intent was more about *non*-main branches)

allisonkarlitskaya (approved):
Makes sense, I guess.  Sounds fine.

PR #2192, opened 2021-07-08
github-upload-action-secrets: support for ssh-keygen and deploy keys

allisonkarlitskaya:


allisonkarlitskaya (commented):
Note: I considered to do things in this order:
1. upload new deploy key
2. update secret
3. delete old deploy keys

So that any uses in the middle would work, but this is silly, and also pointless, because if a workflow is already running, it will keep using the old key...

One outstanding thing I want to make very very sure of is that the method that we're using to generate the key is getting high-quality random data.

allisonkarlitskaya:
Here's a working use of the automatically-generated/uploaded node-cache key: https://github.com/cockpit-project/cockpit/pull/16068

github-upload-action-secrets
```diff
@@ -45,10 +51,24 @@ def encrypt(public_key: str, secret_value: str) -> str:
...
     public_key = public.PublicKey(public_key.encode("utf-8"), encoding.Base64Encoder())
     sealed_box = public.SealedBox(public_key)
-    encrypted = sealed_box.encrypt(secret_value.encode("utf-8"))
+    encrypted = sealed_box.encrypt(secret_value)
     return b64encode(encrypted).decode("utf-8")
 
 
+def generate_ssh_key(pubkey):
+    """Generates a Ed25519 keypair and outputs public/private in OpenSSH format"""
+    key = Ed25519PrivateKey.generate()
```
allisonkarlitskaya, line 60: so, internally, this is using OpenSSL.

and inside of that, it's calling `getrandom()`.

github-upload-action-secrets
```diff
@@ -46,10 +51,24 @@ def encrypt(public_key: str, secret_value: str) -> str:
...
+    public = key.public_key().public_bytes(Encoding.OpenSSH, PublicFormat.OpenSSH)
+    private = key.private_bytes(Encoding.PEM, PrivateFormat.OpenSSH, NoEncryption())
+
+    return public, encrypt(pubkey["key"], private)
```
martinpitt, line 64: The [documentation](https://pynacl.readthedocs.io/en/latest/public/#nacl.public.Box.encrypt) makes a big point about the nonce being a nonce, not simply the private key. This should at least get some documentation what it does, and why this is safe.

Frankly, I'd feel a lot better about not reimplementing this, but call `ssh-keygen` to write files into tmpfs (`$XDG_RUNTIME_DIR`). There is even [ways to write to stdout](https://gist.github.com/kraftb/9918106).

github-upload-action-secrets
```diff
@@ -46,10 +51,24 @@ def encrypt(public_key: str, secret_value: str) -> str:
...
+
+
+def read_encrypted(filename, pubkey):
```
martinpitt, line 67: I find this naming a bit confusing -- it does *not* read an encrypted file, it reads a plaintext file and encrypts it. `read_and_encrypt()` or just `read_encrypt()`?

github-upload-action-secrets
```diff
@@ -63,12 +82,21 @@ def main():
                         help="Only show which secrets would get uploaded where")
     parser.add_argument('-v', '--verbose', action="store_true", default=False,
                         help="Print verbose information")
-    parser.add_argument("secrets_dir", help="directory with one file per secret")
+
+    megr = parser.add_mutually_exclusive_group()
```
martinpitt, line 86: TIL! this is much nicer than explicitly testing this.

github-upload-action-secrets
```diff
@@ -63,12 +82,21 @@ def main():
...
+    megr.add_argument('--ssh-keygen', metavar="SECRET_NAME",
+                      help="Generate an SSH key, upload private, public to stdout")
+
     opts = parser.parse_args()
 
     NAME_RE = r'[a-z][-0-9a-z]*'
     REPO_RE = f'{NAME_RE}/{NAME_RE}'
 
+    if not opts.directory and not opts.ssh_keygen:
+        parser.error('Must give exactly one of --directory or --ssh-keygen')
```
martinpitt, line 100: [add_mutually_exclusive_group()](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) has a `required` argument which you can set to `True`, and drop this.

github-upload-action-secrets
```diff
@@ -88,6 +88,8 @@ def main():
                       help="Upload given directory with one file per secret")
     megr.add_argument('--ssh-keygen', metavar="SECRET_NAME",
                       help="Generate an SSH key, upload private, public to stdout")
+    parser.add_argument('--deploy-to', metavar="OWNER/REPO",
```
martinpitt, line 91: At first sight I thought "wait, this wrong", and only then I saw that this is to `parser`, not `megr`. Could you move this up, or leave a new line or so?

github-upload-action-secrets
```diff
@@ -146,7 +154,28 @@ def main():
         secret_name = opts.ssh_keygen
         public, private = generate_ssh_key(pubkey)
         upload(secrets_path, secret_name, private)
-        sys.stdout.buffer.write(public + b'\n')
+
+        if opts.deploy_to:
+            title = f"{secrets_path}/{secret_name}"
+            keys_path = f"/repos/{opts.deploy_to}/keys"
```
martinpitt, line 160: The deploy code is sufficiently large to better be moved into a separate function.

martinpitt (changes requested):
Nice! With such a script and then collecting all the concrete cases in an Ansible playbook in cockpituous, my "codify/automate knowledge" concern will be completely addressed. Thanks!

github-upload-action-secrets
```diff
@@ -65,6 +66,12 @@ def main():
     parser.add_argument("secrets_dir", help="directory with one file per secret")
     opts = parser.parse_args()
 
+    NAME_RE = r'[a-z][-0-9a-z]*'
+    REPO_RE = f'{NAME_RE}/{NAME_RE}'
+
+    if opts.receiver and not re.fullmatch(f'{NAME_RE}|{REPO_RE}', opts.receiver, re.I):
+        parser.error('--receiver specifies an invalid org or repository name')
```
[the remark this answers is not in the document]
allisonkarlitskaya: I'm happy to expand the expression.  I guess I'm just looking to avoid random junk ending up in a URL?

github-upload-action-secrets
allisonkarlitskaya: > The [documentation](https://pynacl.readthedocs.io/en/latest/public/#nacl.public.Box.encrypt) makes a big point about the nonce being a nonce, not simply the private key. This should at least get some documentation what it does, and why this is safe.

You're confusing the `.encrypt()` function on the box with our own `encrypt()` function which calls it like this:

```python
    encrypted = sealed_box.encrypt(secret_value)
```

ie: without providing a nonce.

In any case, this needs to be a secure operation because the public-key that you download from GitHub is the same each time.

> Frankly, I'd feel a lot better about not reimplementing this, but call `ssh-keygen` to write files into tmpfs (`$XDG_RUNTIME_DIR`). There is even [ways to write to stdout](https://gist.github.com/kraftb/9918106).

I really strongly disagree with this.  Those are some awful ugly hacks you link.  The only one that I might even be vaguely inclined to implement is the openssl one...

What's your exact concern here?  Insufficient entropy?  Degenerate keys?  Sidechannel leaks?

github-upload-action-secrets
allisonkarlitskaya: Sure.

github-upload-action-secrets
allisonkarlitskaya: TIL!  Thanks, in return :)

github-upload-action-secrets
allisonkarlitskaya: Sure.

github-upload-action-secrets
allisonkarlitskaya: Ya..... The problem is the `opts` thing.  I might just bump that to be global...

This is the same reason `upload()` is nested and not a proper function of its own.

github-upload-action-secrets
martinpitt: Whoops! Sorry about mixing up the two encrypt(), of course you are right.

> What's your exact concern here?

Mostly that it wouldn't be the first case of someone being burned by reimplementing crypto, i.e. standard paranoia. I trust that ssh-keygen knows what it needs to do, I am not concerned about writing the key to tmpfs, and everything else I need to properly understand. :grin: So no concrete "this is unsafe".

github-upload-action-secrets
martinpitt: But you could just pass the necessary opts.* values to the function?

allisonkarlitskaya:
Ok, so, choose your poison:

```python
    key = Ed25519PrivateKey.generate()
    public = key.public_key().public_bytes(Encoding.OpenSSH, PublicFormat.OpenSSH)
    private = key.private_bytes(Encoding.PEM, PrivateFormat.OpenSSH, NoEncryption())
```

or

```python
with tempfile.TemporaryDirectory() as dir:
    key=f'{dir}/link-to-fd3-and-fd4'

    yes_r, yes_w = os.pipe()
    priv_r, priv_w = os.pipe()
    publ_r, publ_w = os.pipe()

    try:
        os.symlink(f'/proc/self/fd/{priv_w}', f'{key}')
        os.symlink(f'/proc/self/fd/{publ_w}', f'{key}.pub')

        ssh_keygen = subprocess.Popen(
            ['ssh-keygen', '-t', 'ed25519', '-N', '', '-q', '-f', key],
            stdin=yes_r, stdout=open('/dev/null', 'w'),
            pass_fds = (publ_w, priv_w))
            
        os.write(yes_w, b'y\n')
        public = os.read(publ_r, 1024)
        private = os.read(priv_r, 1024)

    finally:
        for fd in (yes_r, yes_w, priv_r, priv_w, publ_r, publ_w):
            os.close(fd)
```

or

```python
    with tempfile.TemporaryDirectory() as dir:
        key = f'{dir}/link-to-fd3-and-fd4'

        os.symlink(f'/proc/self/fd/3', f'{key}')
        os.symlink(f'/proc/self/fd/4', f'{key}.pub')
        process = subprocess.Popen(f'ssh-keygen -t ed25519 -N "" -q -f {key} 3>&1 4>&2 &>/dev/null', shell=True,
                                   stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
        private, public = process.communicate(b'y\n')
```


My personal preference is the pure library approach, but I understand that you have reservations about that.  Second place is the "shell abuse" approach, I guess?  Lucky that we have public/private and also stdout/stderr.  Seems that the number 2 is popular!

martinpitt:
@allisonkarlitskaya Thanks for working these out! If nothing else, that's a nice reference to point to in the future. Undoubtedly 1 is the nicest (my previous review was blurred by that encrypt() mis-reading), 2 "OMG please no", and I find 3 actually quite clever. Not too big to understand/read, and doesn't require thinking about crypto or keeping up with SSH key format changes (which [did happen not too long ago](https://lwn.net/Articles/763444/). So I feel like 2 > 1 > 3, but I would not veto 1 either -- after all, these are not long-lived, so if they stop working because reasons, we can always adjust the implementation.

github-upload-action-secrets
allisonkarlitskaya: Replaced with "shell abuse" variant from below.

github-upload-action-secrets
allisonkarlitskaya: - added `upload_deploy_key()` as function
 - renamed `upload()` to `upload_secret()` and made it a proper function
 - added `api`, `pubkey`, `opts`, and about 12 other arguments ;)

Starting to think this might benefit from a small utility class....

github-upload-action-secrets
```diff
@@ -46,10 +48,51 @@ def encrypt(public_key: str, secret_value: str) -> str:
...
+    how awkward it is: ssh-keygen knows how to make good SSH keys.  The word
+    "hazmat" in the module name is enough of a warning here about the dangers
+    of getting this stuff wrong.
+
+    See discussion in https://github.com/cockpit-project/bots/pull/2192
+    """
+    with tempfile.TemporaryDirectory() as dir:
+        key = f'{dir}/link-to-fd3-and-fd4'
+
+        os.symlink('/proc/self/fd/3', f'{key}')
```
martinpitt, line 73: This is clearly overdoing f-strings :grin:

github-upload-action-secrets
```diff
@@ -63,10 +106,16 @@ def main():
...
+    megr = parser.add_mutually_exclusive_group(required=True)
+    megr.add_argument('--directory',
+                      help="Upload given directory with one file per secret")
+    megr.add_argument('--ssh-keygen', metavar="SECRET_NAME",
+                      help="Generate an SSH key, upload private, public to stdout")
+
     opts = parser.parse_args()
 
-    NAME_RE = r'[a-z][-0-9a-z]*'
+    NAME_RE = r'[a-z][-0-9a-z._]*'
```
martinpitt: Rebase error, this should be in the previous commit that introduces the name checking.

martinpitt (changes requested):
One tiny rebase error, but this looks good. Many thanks for figuring this out! :100:

github-upload-action-secrets
allisonkarlitskaya: So, obviously, visual uniformity with the line below.

I thought about

```
  , key)
  , key + '.pub')
```

but you know me: I'm a sucker for f-strings :)

github-upload-action-secrets
allisonkarlitskaya: Fixed.  Thanks for the careful checking.

martinpitt (approved):

PR #2193, opened 2021-07-08
task: Drop scanning of /orgs/cockpit-project/teams

martinpitt:
It is not necessary to map a team name to an ID; one can directly look
up a team by name:
https://docs.github.com/en/rest/reference/teams#get-a-team-by-name

Do that and drop the teamIdFromName() helper.

-----

Tested with `./tests-trigger` and  making sure that it can still correctly determine team membership (with `read:org` permission, that does not change).

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -57,8 +57,6 @@ NOT_TESTED_DIRECT = "Not yet tested (direct trigger)"
 
 ISSUE_TITLE_IMAGE_REFRESH = "Image refresh for {0}"
 
-TEAM_CONTRIBUTORS = "Contributors"
-
 
 class Logger(object):
     def __init__(self, directory):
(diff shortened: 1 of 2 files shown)


task/github.py
```diff
@@ -381,22 +381,15 @@ def allowlist(self):
         users = {"github-actions[bot]", "candlepin", "cockpit-project", "osbuild"}
 
         # individual persons from https://github.com/orgs/cockpit-project/teams/contributors/members
-        teamId = self.teamIdFromName(TEAM_CONTRIBUTORS)
```
marusak, line 384: You can also remove then "TEAM_CONTRIBUTORS" constant.

marusak (changes requested):
Just a small cleanup

task/github.py
martinpitt: Thanks for spotting! Fixed.

marusak (approved):
Thanks!

PR #2194, opened 2021-07-08
workflows: Run deployment tests with default GitHub token

martinpitt:
This avoids another usage of our cockpituous token.

Also completely lock down the "bots" job, which just runs the unit tests.

----

 - [x] Builds on top of PR #2193

Similar to https://github.com/cockpit-project/cockpituous/pull/431 , this is mostly an experiment for now. This test run is the only usage of the cockpituous token by cockpituous.git and bots.git, so it'll make it easier to move this token into an environment once we can lose it from these two repos.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,6 +3,8 @@ on: [pull_request]
 jobs:
   bots:
     runs-on: ubuntu-20.04
+    permissions:
+      pull-requests: none
     steps:
       - name: Clone repository
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


martinpitt:
[Nope](https://github.com/cockpit-project/bots/pull/2194/checks?check_run_id=3019737485)

martinpitt:
Will reopen once I fix the crash, but it's not that easy to reproduce locally, as a PAT just doesn't exhibit that behaviour. You can't take away org read enough for it to crash like the GITHUB_TOKEN does.

allisonkarlitskaya (approved):

PR #2202, opened 2021-07-13
Image refresh for fedora-33

cockpituous:
Image refresh for fedora-33
 * [x] image-refresh fedora-33


marusak:
Osbuild has some issues. It is being fixed in https://github.com/osbuild/cockpit-composer/pull/1326 by @jkozol

marusak (approved):
Actually, this is not a regression. See for example currently opened PRs - https://github.com/osbuild/cockpit-composer/pull/1329 https://github.com/osbuild/cockpit-composer/pull/1328 or https://github.com/osbuild/cockpit-composer/pull/1327

In all of those these contexts fail the same. So I think we can merge this. @martinpitt do you agree?

martinpitt (approved):
Indeed, thanks for checking!

PR #2210, opened 2021-07-16
Image refresh for fedora-34

cockpituous:
Image refresh for fedora-34
 * [x] image-refresh fedora-34
 * [x] https://github.com/cockpit-project/cockpit/pull/16109


marusak:
`TestStorageNfs.testNfsBusy` failed in all contexts, same as in https://github.com/cockpit-project/bots/pull/2211
Also podman failed in both contexts the same way, need to be looked at as well.

marusak:
ah, it just says 'of User root' instead of 'of user root'. Let's see where this capitalization comes from.

marusak:
Ok, so it comes from dbus directly:

```
$ busctl get-property org.freedesktop.systemd1 /org/freedesktop/systemd1/unit/session_2d3_2escope org.freedesktop.systemd1.Unit Description
s "Session 3 of User root"
```

Lets see if I can pinpoint this to a specific package and a specific commit or if we will just work around it.

marusak:
Downgrading systemd does the trick.

marusak:
Found the commit: https://github.com/systemd/systemd/commit/0a59216caebdf0488d1200cd818cb46361d1f17b

TBH it feels incorrect. But at the same time, they seem to do this very consciously so I think we just need to adjust our side. I have not yet decided if I should just hack the test or if I should de-capitalize it before we show it in the UI.

martinpitt:
It reads rather weird, even for English. I [asked on the systemd change](https://github.com/systemd/systemd/commit/0a59216caebdf0488d1200cd818cb46361d1f17b#r53563654), but I suppose for now we can just accept both variants from the test. This is a bit awkward with our current API, maybe you can do multiple smaller substring tests, like "Session" and "ser root"?

marusak:
For cockpit sent https://github.com/cockpit-project/cockpit/pull/16109 (also triggered there `@bots#` context)

martinpitt:
https://github.com/cockpit-project/cockpit/pull/16109 landed, retrying the cockpit tests. podman tests are still broken, leaving them (for the final run we should rebase this to avoid ruining test stats).

marusak:
Rebased

marusak (approved):

PR #2213, opened 2021-07-16
Image refresh for rhel-9-0

cockpituous:
Image refresh for rhel-9-0
 * [x] image-refresh rhel-9-0
 * [x] https://github.com/cockpit-project/cockpit-machines/pull/262


martinpitt:
@KKoukiou : Failures look very similar to what we saw downstream. It'll just be some "fun" to write patterns for all of these..

martinpitt:
I pushed naughties for the NFS related failures (issue #2214). There are two remaining c-machines failures:

 * [TestMachinesFilesystems.testBasic](https://logs.cockpit-project.org/logs/pull-2213-20210716-151016-062cd17e-rhel-9-0-cockpit-project-cockpit-machines/log.html#25-2) explicitly tests buggy/undesirable behaviour of libvirt. @KKoukiou recently filed https://gitlab.com/libvirt/libvirt/-/issues/178 and this got fixed upstream now -- it plausibly went into RHEL 9 (libvirt-daemon 7.4.0-1.el9 -> 7.5.0-1.el9). Supposedly this should show an error somewhere, but the [screenshot shows none](https://logs.cockpit-project.org/logs/pull-2213-20210716-151016-062cd17e-rhel-9-0-cockpit-project-cockpit-machines/TestMachinesFilesystems-testBasic-rhel-9-0-127.0.0.2-2201-FAIL.png). @KKoukiou , can you please have a look at that one?

 * [TestMachinesStoragePools.testStoragePoolsCreate](https://logs.cockpit-project.org/logs/pull-2213-20210716-151016-062cd17e-rhel-9-0-cockpit-project-cockpit-machines/log.html#61-2) fails because of my_dir_pool_two being in "Failed" state. It's exactly the same error as in the downstream test. Again possibly a regression from libvirt 7.5? This effortlessly reproduces locally.

martinpitt:
I took a closer look at the remaining machines failure: The dialog closes immediately when trying to add a duplicate mount ID. The only hint is in the journal, it now says

    libvirtd[1438]: unsupported configuration: filesystem target 'dir1' specified twice

and it does not get added any more, so the test times out on waiting for the third shared file system which never appears. So either the libvirt API call fails and we don't show the error properly, or it doesn't fail and needs to be fixed further.

martinpitt:
I sent https://github.com/cockpit-project/cockpit-machines/pull/260 to fix the error propagation. This will *not* yet fix the failure here, as the test still expects the duplicate ID to be added. I want to discuss with @KKoukiou how we can fix the test to not require a lockstep update for each image refresh that fixes this bug. My current idea is:

 - Adjust the test for the new and fixed behaviour, but don't run it anywhere
 - Land rhel-9-0 image refresh
 - Enable the test on rhel-9-0, so that we assert the behaviour on at least one OS
 - Over time we can get rid of the special case (but we'll forget) or we track it as a naughty (might be better, but then let's reorganize it so that this is the last check -- otherwise we'll skip all the bits after it)

@KKoukiou WDYT?

martinpitt:
@KKoukiou : I adjusted the test for the duplicate shared mount ID issue. But now it fails on the last step: It tries to start the VM and expects the "Add shared directory" button to get disabled (as that can't be done while the VM is running). However, the VM does not run:

image

Does that ring a bell? ISTR that virtiofs support status on RHEL 9 was unclear, maybe it was removed now?

martinpitt:
On current RHEL 9 image:
```
# rpm -qf /usr/libexec/virtiofsd
qemu-kvm-common-6.0.0-7.el9.x86_64
```
This is no different on this refreshed image, so apparently libvirt 7.5 requires something newer. I'll file a bugzilla and handle this through a naughty. I sent https://github.com/cockpit-project/cockpit-machines/pull/262 to fix the first issue, blocking on that now.

KKoukiou:
> @KKoukiou : I adjusted the test for the duplicate shared mount ID issue. But now it fails on the last step: It tries to start the VM and expects the "Add shared directory" button to get disabled (as that can't be done while the VM is running). However, the VM does not run:
> 
> image
> 
> Does that ring a bell? ISTR that virtiofs support status on RHEL 9 was unclear, maybe it was removed now?

Yes, should be this https://bugzilla.redhat.com/show_bug.cgi?id=1966580

martinpitt:
c-machines update landed.

martinpitt (approved):
Yay green! :four_leaf_clover:

PR #2226, opened 2021-07-20
Adjust pattern for crashing libvirt and copy to other images were it applies

KKoukiou:
This applies also for images without libvirt0-dbgsym installed, so let's
match the test failure line instead of the backtrace from journal.


martinpitt (approved):
These patterns are now *very* generic, but without changes to c-machines itself (like, adding some logging to the service proxy that yells if libvirtd.service goes into state "failed") we can't do better. Approving this in case you want to go this route. Due to Debian's long freeze, debian testing and ubuntu-stable are still stuck at libvirt 7.0.0 unfortunately.

KKoukiou:
> These patterns are now _very_ generic, but without changes to c-machines itself (like, adding some logging to the service proxy that yells if libvirtd.service goes into state "failed") we can't do better. Approving this in case you want to go this route. Due to Debian's long freeze, debian testing and ubuntu-stable are still stuck at libvirt 7.0.0 unfortunately.

I tried showing the libvirt messages but they don't see to work. In addition I can't reproduce this locally on ubuntu images. So I decided I will move this ahead, as it works for other Oses fine, and maybe iterate on it at some other point.

PR #2235, opened 2021-07-23
naughty: Add another pattern variant for #1739

martinpitt:
Similar to pattern 3, but happens in testGrowShrinkEncryptedHelp instead
of testResize().

----

[example](https://logs.cockpit-project.org/logs/pull-16144-20210723-081920-db112798-ubuntu-stable/log.html#198-2)


KKoukiou:
Ack the screenshot in #16144 clearly shows that the filesystem in question is not mounted after at the point of the grow operation.

KKoukiou (approved):

martinpitt:
Right, that's why the size check fails -- `df /run/foo` just shows the free space on /run, as /run/foo is not mounted any more. Exactly like in pattern 3.

PR #1786, opened 2021-03-12
naughty: Add two Fedora 34 reportd crashes

martinpitt:
Known issue #1784
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1938085

Known issue #1785
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=1938086

Who watches the watchers?

Issue #1785, opened 2021-03-12
fedora-34: reportd crashes


martinpitt:
[example](https://logs.cockpit-project.org/logs/pull-1781-20210311-170606-182404e6-fedora-34-cockpit-project-cockpit/log.html#279-2)

Downstream reports:
https://bugzilla.redhat.com/show_bug.cgi?id=1938086
https://bugzilla.redhat.com/show_bug.cgi?id=1966950

cockpituous:
verify/fedora-34
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testAbrtReport (__main__.TestJournal)
Job for systemd-journald-dev-log.socket canceled.
Job for systemd-journald.socket canceled.
/usr/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
[0312/143536.464891:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory

DevTools listening on ws://127.0.0.1:9321/devtools/browser/dc63d836-01aa-48f3-a9ca-c98958a2db6f
[...]

Issue #1784, opened 2021-03-11
fedora: reportd crashes in reportd_task_start


martinpitt:
Today we have seen several fedora-34 test runs where reportd crashes:
```
Process 70488 (reportd) of user 0 dumped core.
Stack trace of thread 70581:
#0  0x00007f2e033d4fab g_cancellable_is_cancelled (libgio-2.0.so.0 + 0x4afab)
#1  0x00007f2e033d6e63 g_cancellable_set_error_if_cancelled (libgio-2.0.so.0 + 0x4ce63)
#2  0x00005633732ae1f2 reportd_task_start (reportd + 0xe1f2)
#3  0x00007f2e0343808f g_task_thread_pool_thread (libgio-2.0.so.0 + 0xae08f)
#4  0x00007f2e0327abd4 g_thread_pool_thread_proxy.lto_priv.0 (libglib-2.0.so.0 + 0x86bd4)
#5  0x00007f2e03277ce2 g_thread_proxy (libglib-2.0.so.0 + 0x83ce2)
#6  0x00007f2e02f0a269 start_thread (libpthread.so.0 + 0x9269)
#7  0x00007f2e030cd653 __clone (libc.so.6 + 0x100653)
Stack trace of thread 70492:
#0  0x00007f2e030c29af __poll (libc.so.6 + 0xf59af)
#1  0x00007f2e0329d4ac g_main_context_iterate.constprop.0 (libglib-2.0.so.0 + 0xa94ac)
#2  0x00007f2e03246b13 g_main_context_iteration (libglib-2.0.so.0 + 0x52b13)
#3  0x00007f2e03246b61 glib_worker_main (libglib-2.0.so.0 + 0x52b61)
#4  0x00007f2e03277ce2 g_thread_proxy (libglib-2.0.so.0 + 0x83ce2)
#5  0x00007f2e02f0a269 start_thread (libpthread.so.0 + 0x9269)
#6  0x00007f2e030cd653 __clone (libc.so.6 + 0x100653)
Stack trace of thread 70493:
#0  0x00007f2e030c29af __poll (libc.so.6 + 0xf59af)
#1  0x00007f2e0329d4ac g_main_context_iterate.constprop.0 (libglib-2.0.so.0 + 0xa94ac)
#2  0x00007f2e032489a3 g_main_loop_run (libglib-2.0.so.0 + 0x549a3)
[...]

PR #2244, opened 2021-07-26
naughty: Add two Fedora 34 reportd crashes

marusak:
The cause is the same, it just differes when reportd crashes and whether
it makes the test fail or it just sees the journal message.

Re-introduces: #1786

Known issues: #1784 and #1785
Downstream reports: https://bugzilla.redhat.com/show_bug.cgi?id=1938086

This is our [top flakes and the only unexpected message](https://images-frontdoor.apps.ocp.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit)


martinpitt (approved):
Dakujem!

Issue #2248, opened 2021-07-27
unspecific systemd-resolved crash on Ubuntu 20.04


martinpitt:
We sometimes [see random failures](https://logs.cockpit-project.org/logs/pull-16151-20210727-043808-47a25c67-ubuntu-2004/log.html#47) due to unspecific systemd-resolved crashes:
```
Jul 27 04:49:52 ubuntu kernel: systemd-resolve[651]: segfault at 228 ip 00005589171cf61a sp 00007ffe46befbc0 error 6 in systemd-resolved[5589171a2000+40000]
Jul 27 04:49:52 ubuntu kernel: Code: 48 85 c0 74 0e 48 8b 8d 00 01 00 00 48 89 88 00 01 00 00 48 8b 85 00 01 00 00 48 85 c0 0f 84 1d 01 00 00 48 8b 95 f8 00 00 00 <48> 89 90 f8 00 00 00 48 c7 85 00 01 00 00 00 00 00 00 48 c7 85 f8
Jul 27 04:49:52 ubuntu systemd[1]: Created slice system-systemd\x2dcoredump.slice.
Jul 27 04:49:52 ubuntu systemd[1]: Started Process Core Dump (PID 29972/UID 0).
Jul 27 04:49:52 ubuntu systemd[1]: systemd-resolved.service: Main process exited, code=dumped, status=11/SEGV
Jul 27 04:49:52 ubuntu systemd[1]: systemd-resolved.service: Failed with result 'core-dump'.
Jul 27 04:49:52 ubuntu systemd[1]: systemd-resolved.service: Scheduled restart job, restart counter is at 1.
Jul 27 04:49:52 ubuntu systemd[1]: Stopped Network Name Resolution.
Jul 27 04:49:52 ubuntu systemd[1]: Starting Network Name Resolution...
Jul 27 04:49:53 ubuntu systemd-coredump[29973]: Process 651 (systemd-resolve) of user 101 dumped core.
                                                
                                                Stack trace of thread 651:
                                                #0  0x00005589171cf61a n/a (systemd-resolved + 0x3761a)
                                                #1  0x00005589171cf592 n/a (systemd-resolved + 0x37592)
                                                #2  0x00005589171b6148 n/a (systemd-resolved + 0x1e148)
                                                #3  0x00005589171cb99f n/a (systemd-resolved + 0x3399f)
                                                #4  0x00007fd228f50b96 n/a (libsystemd-shared-245.so + 0x76b96)
[...]

PR #2249, opened 2021-07-27
naughty: Add pattern for unspecific systemd-resolved segfault

martinpitt:
This happens unrelated to any particular test, and seems to not happen
any more with current upstream and Ubuntu versions. As the stack trace
has no useful information, and none of our tests directly interact with
resolved, let's just ignore these.

Known issue #2248

----

[example](https://logs.cockpit-project.org/logs/pull-16151-20210727-043808-47a25c67-ubuntu-2004/log.html#47)


marusak (approved):
Thanks!

PR #2256, opened 2021-07-29
naughty: Fix pattern for #2248

martinpitt:
The initial commit left a hardcoded thread number in it. Also, the stack
trace sometimes starts with `__GI_abort`, so so the first n/a frame is
not #0.

---

Noticed in [this failure](https://logs.cockpit-project.org/logs/pull-16172-20210729-124605-44c1dc7e-ubuntu-2004/log.html#47). Validated locally that the pattern now matches.


allisonkarlitskaya (approved):
sure

PR #2266, opened 2021-08-02
tests-invoke: Avoid test failure due to GitHub API timeout during collision detection

martinpitt:
During times when the GitHub API is a bit flaky (like right now), a lot
of tests fail with "TimeoutError: [Errno 110] Connection timed out" in
detect_collisions(). This is awkward -- it's not that bad if two tests
run in parallel for a little longer when skipping a collision check or
two. So ignore connection timeouts for that particular check.

Note that when the GitHub API is permanently down, test invocation and
result reporting will fail anyway, so it's not completely hidden.

---

[example](https://logs.cockpit-project.org/logs/pull-16000-20210802-070918-81f4f244-centos-8-stream/log), I've retried PRs at least 5 times this morning already. /me draws a line into the sand

### DIFF
--- a/tests-invoke
+++ b/tests-invoke
@@ -40,7 +40,13 @@ def main():
     def detect_collisions(opts):
         api = github.GitHub(repo=opts.repo)
 
-        pull = api.get("pulls/{0}".format(opts.pull_number))
+        try:
+            pull = api.get("pulls/{0}".format(opts.pull_number))
+        except TimeoutError as e:
+            sys.stderr.write('Warning: collision detection failed, skipping: %s' % e)
+            # GitHub API is sometimes flaky, don't break test just because it sometimes does not respond
+            return None
+
         if pull:
             if pull["head"]["sha"] != opts.revision:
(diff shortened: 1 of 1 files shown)


martinpitt:
Meh, seems [tests-invoke is being used from main](https://logs.cockpit-project.org/logs/pull-2266-20210802-074029-7e03acde-fedora-coreos-cockpit-project-cockpit/log), not from the checked out branch. The test *does* check out the correct bots ref, but calls /work/bots/make-checkout-workdir/../tests-invoke which is the parent bots checkout.

Also, it seems our [integration tests got broken](https://github.com/cockpit-project/bots/pull/2266/checks?check_run_id=3218448273) *sigh*

martinpitt:
> Meh, seems tests-invoke is being used from main, not from the checked out branch

That was a conscious decision, from commit 5588e4d264845bc7d65c014f6a8eab610db420ed.

I'll investigate the integration tests separately, for now let's just get this fixed as it breaks *everything* right now.

PR #2272, opened 2021-08-03
Set IdentitiesOnly for ssh connections

jelly:
With IdentitiesOnly option enabled, ssh only tries the configured
IdentitiesFile for authentication. This is useful for when ssh offers to
many different identities causing an authentication failure.


martinpitt (approved):
Thanks! This is a nice bug fix, hard to see at first, and certainly someone else would have fallen into that trap already.

Triggering a few tests across our projects to make double sure that this works as expected.

PR #2275, opened 2021-08-03
README: Various updates

martinpitt:
- bots does not contain any release-y bits, that's between projects and
   cockpituous.
 - Use a consistent itemization style
 - Fix formatting of the image names, the `*` were interpreted as
   italics before
 - Update services on the "services" image
 - Break overly long lines
 - Stop recommending `read:org` privilege for ones' personal GitHub
   access token; not necessary any more since commit 558d0e688a332e
 - Drop paragraph about updating bots in "main feature" PRs -- this has
   not been a thing since we split off bots from cockpit.git two years
   ago

Preview: https://github.com/martinpitt/bots/blob/doc-updates/README.md

PR #2276, opened 2021-08-03
Fix integration tests

martinpitt:
Yesterday the self-deployment integration tests failed. That could have been due to the GitHub DNS troubles, or something on our side. Using this PR to investigate.

### DIFF
--- a/publish-wrapper
+++ b/publish-wrapper
@@ -2,7 +2,7 @@
 
 # This file is part of Cockpit.
 #
-# Copyright (C) 2020 Red Hat, Inc.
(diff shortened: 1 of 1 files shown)


martinpitt:
Bah, the [DNS troubles are still ongoing](https://github.com/cockpit-project/bots/pull/2275/checks?check_run_id=3232690339), so this will most probably have to wait a bit.

martinpitt:
... so the complicated one works, but the simple one in PR #2275 doesn't, "fun". Anyway, glad that it's not a regression on our side!

PR #2282, opened 2021-08-06
machines: Remove image file in rem_disk

mvollmer:
Otherwise ./tmp/run will grow without bounds.


martinpitt:
Thanks! Triggering a few tests.

martinpitt (approved):

martinpitt (changes requested):
@mvollmer this branch needs rebasing, several missing images.

martinpitt:
Actually, coudl we remove the file right away after starting the VM? I don't think we'll need the name afterwards, and like that the fs driver will clean up the anonymous inode automatically once the process goes away

martinpitt:
> I don't think we'll need the name afterwards,

Hmm, the self._domain.detachDeviceFlags() looks like it would need it.

I just rebased this for now.

martinpitt (approved):

PR #2284, opened 2021-08-07
Image refresh for rhel-8-5

cockpituous:
Image refresh for rhel-8-5
 * [x] image-refresh rhel-8-5


martinpitt:
[this failure](https://logs.cockpit-project.org/logs/pull-2284-20210807-023327-c580b9a5-rhel-8-5-cockpit-project-cockpit/log.html#259-2) is a bug in the test when they run over midnight. I'll look at it, but this is not an image regression.

martinpitt (approved):

PR #2285, opened 2021-08-07
Image refresh for services

cockpituous:
Image refresh for services
 * [x] image-refresh services


martinpitt:
We should not trigger rhel-8-4 on cockpit main any more, we don't officially support that any more. Moved to rhel-9-0.

martinpitt (approved):

Issue #2291, opened 2021-08-09
avc:  denied  { add_name } for  pid=8164 comm="systemd-hostnam" 


jelly:
Downstream report for Fedora 34: https://bugzilla.redhat.com/show_bug.cgi?id=1991444

jelly:
This does not appear to affect our test which uses `hostnamectl set-hostname`

cockpituous:
fedora-34
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testBasic (__main__.TestHostSwitching)
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.
[0811/162104.063068:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[0811/162104.064333:WARNING:headless_browser_main_parts.cc(83)] Cannot create Pref Service with no user data dir.

DevTools listening on ws://127.0.0.1:10139/devtools/browser/6035f610-4e2a-4e98-b813-0f5203c3450e
[0811/162104.067812:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 96268: Permission denied (13)
[0811/162104.073560:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 96271: Permission denied (13)
[0811/162104.075073:ERROR:gpu_init.cc(426)] Passthrough is not supported, GL is egl
[0811/162104.319110:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 96368: Permission denied (13)
[...]

PR #2292, opened 2021-08-09
naughty: Add pattern for SELinux hostnamectl denial

jelly:
Known issue #2291
Downstream report for RHEL 34: https://bugzilla.redhat.com/show_bug.cgi?id=1991444

### DIFF
--- a/naughty/fedora-34/2291-selinux-hostnamectl
+++ b/naughty/fedora-34/2291-selinux-hostnamectl
@@ -1,0 +1,1 @@
+avc:  denied  { add_name } for  * comm="systemd-hostnam" name="*" scontext=system_u:system_r:systemd_hostnamed_t:s0 tcontext=system_u:object_r:init_var_run_t:s0 tclass=dir permissive=0


jelly:
I'm not 100% sure if this is correct, is there a way for me to locally test this?

naughty/fedora-34/2291-selinux-hostnamectl
```diff
@@ -0,0 +1 @@
+avc:  denied  { add_name } for  * comm="systemd-hostnam" name="*" scontext=system_u:system_r:systemd_hostnamed_t:s0 tcontext=system_u:object_r:init_var_run_t:s0 tclass=dir permissive=0
```
marusak, line 1: Two things some to mind:
1. If you look at other avcs we have, it is always a good idea to stick to the pattern. So start with `* type=1400 audit(*): avc:  denied`... (Back in the day this hit us, as some pattern did not work - so might this as well, as it does not start with `*`)
2. You can test your pattern with `tests-policy`

How to use test policy:
1. Put output from test into file
2. `./tests-policy -o context < file_name` where context is for example `fedora-34` and `file_name` is the one, where you put output of the test

martinpitt:
https://github.com/cockpit-project/cockpit/pull/16213 does not seem to actually need this, it went green. Filing the bugzilla was good of course, it is a bug -- it just doesn't seem to manifest in that particular test setup? So should we just close this?

jelly:
Yes, I'll close it as tests succeed.

marusak:
We see it now for example [here](https://logs.cockpit-project.org/logs/pull-16213-20210811-102834-7b4744c3-fedora-34/log.html#274-2), reopenning.

marusak (approved):
The pattern works. I wanted to complain about it not following the same pattern as rest of avc naughties , but then noticed, they already don't follow one pattern :)
Thanks!

marusak:
I triggered `fedora-34@bots#2292` context in https://github.com/cockpit-project/cockpit/pull/16213 for testing, hopefully it hits it.

@jelly FYI - we can test also across projects, so we can do something like `fedora-34@cockpit-project/cockpit` for example in bots PR, and it will test current `main` in `cockpit-project/cockpit` with bots from the given PR. We can also do it the other way around, which is what I just did - testing bots PR in other project PR.

PR #2297, opened 2021-08-11
naughty: adjust podman test expected failure

KKoukiou:
For https://github.com/cockpit-project/cockpit-podman/pull/743

### DIFF
--- a/naughty/rhel-9/2158-podman-api-call-loses-data-2
+++ b/naughty/rhel-9/2158-podman-api-call-loses-data-2
@@ -1,0 +1,6 @@
+Traceback (most recent call last):
+*
+  File "test/check-application", line *, in testDownloadImage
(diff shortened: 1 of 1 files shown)


marusak (changes requested):
Hmm, this is one of those annoying ones, where we need to land this on the right time - which still won't be the right time as downstream still needs the old pattern.
How about creating a new pattern for the same issue? It is fine to have two patterns and naughty-prune can remove both of them, once it is fixed.

mvollmer:
I'll trigger a test run in https://github.com/cockpit-project/cockpit-podman/pull/743

mvollmer:
Test in https://github.com/cockpit-project/cockpit-podman/pull/743 is green, thanks!

mvollmer (approved):

PR #2319, opened 2021-08-18
Github token doc/tooling improvements

jelly:
When setting up `./tests-trigger` locally I encountered the missing scope and lack of error message when no token is configured. I've applied a bit of a catch all solution by raising an exception, alternatively I could implement a check in every consumer of the `Github` class if required (as one cannot return without an exception from `__init__`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,9 +83,9 @@ the [GitHub CLI](https://cli.github.com/) configuration in
 ~/.config/gh/config.yml.  You can create one for your account at
 [Developer Settings → Personal access tokens](https://github.com/settings/tokens).
 
-When generating a new personal access token, the scope should only contain
-`repo:status`. Note in particular, that `repo` and `public_repo` scopes each
-grant full push access, and should not be used.
+When generating a new personal access token, the scopes should contain
+`repo:status` and `read:org`. Note in particular, that `repo` and
+`public_repo` scopes each grant full push access, and should not be used.
 
 If you'd like to download Red Hat-only internal images from S3, you'll
 need to create a key file in `~/.config/cockpit-dev/s3-keys/[domain]`.


README.md
```diff
@@ -83,9 +83,9 @@ the [GitHub CLI](https://cli.github.com/) configuration in
 ~/.config/gh/config.yml.  You can create one for your account at
 [Developer Settings → Personal access tokens](https://github.com/settings/tokens).
 
-When generating a new personal access token, the scope should only contain
-`repo:status`. Note in particular, that `repo` and `public_repo` scopes each
-grant full push access, and should not be used.
+When generating a new personal access token, the scopes should contain
+`repo:status` and `read:org`. Note in particular, that `repo` and
```
marusak, line 87: So, not that long ago this `read:org` was added in https://github.com/cockpit-project/bots/commit/558d0e688a332e8b4381f481b5159f0a98767976

Then in https://github.com/cockpit-project/bots/commit/fb2d79e3b17f5ed7b93ad60ccf60addc24e93ec5 it was dropped stating it is not needed since the commit, that only introduced it.

From this it seems to me, that it is actually needed. Have you tried token without and with this privilege?

But from code:
```
    def allowlist(self):                                                        
        # bots and organizations which are allowed to use our CI (these use branches within the main repo for PRs)
        users = {"github-actions[bot]", "candlepin", "cockpit-project", "osbuild"}
                                                                                
        # individual persons from https://github.com/orgs/cockpit-project/teams/contributors/members
        # this requires a token with read:org permissions                                                                  
```

So I think this is actually correct.

marusak:
I think you need to open this PR from origin. Can you please do so, so tests can properly run?

task/github.py
```diff
@@ -142,6 +142,9 @@ def __init__(self, base=None, cacher=None, repo=None):
                 except FileNotFoundError:
                     pass
 
+        if self.token is None:
```
allisonkarlitskaya, line 145: I'm pretty sure there *are* things that can be done without a token.  `tests-scan` for example.  This would break that.

README.md
allisonkarlitskaya: This change is definitely right.  Happy to see that one go in.  I'm pretty sure I wrote that, and I thought I had added both.

But indeed, #2275 seems very intentional about removing this.  I feel like this must have been a mistake...

allisonkarlitskaya (changes requested):

README.md
jelly: Without having `read:org` it returns an empty list, which will always fail the permission check.

task/github.py
jelly: Yeah, I just nuked the commit for now. A better approach would be to not return `None` when we receive a 404, this would require the callers to be updated and handle a possible `GithubStatus` error like in  https://github.com/cockpit-project/bots/blob/main/push-rewrite#L80

marusak (approved):

martinpitt:
Indeed `read:org` seems a bit unpredictable right now -- I *did* remove it from my PAT, and am happily using tests-trigger in most cases -- just sometimes it acts up, and I need to use `--allow`. So, OK from me to put it back into the documentation, `read:org` is not too scary.

PR #2356, opened 2021-08-30
testmap: Move to new cockpit rhel-8.5 branch

martinpitt:
I just created https://github.com/cockpit-project/cockpit/tree/rhel-8.5 , let's set up CI for it from day one.


martinpitt:
@marusak: Somehow we need to convince https://github.com/cockpit-project/cockpit-weblate/tree/rhel8 to pull from cockpit's rhel-8.5 branch instead of -8.4 now. Where is that configured?

marusak (approved):
Thanks!

PR #2361, opened 2021-08-31
machine: add machine.libexecdir

allisonkarlitskaya:
This points to /usr/lib/cockpit on Debian and /usr/libexec elsewhere.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -85,6 +85,11 @@ class Machine(ssh_connection.SSHConnection):
         else:
             self.label = "{}@{}:{}".format(self.ssh_user, self.ssh_address, self.ssh_port)
 
+        if "debian" in self.image or "ubuntu" in self.image:
+            self.libexecdir = "/usr/lib/cockpit"
+        else:
+            self.libexecdir = "/usr/libexec"
+
         # The Linux kernel boot_id
         self.boot_id = None
 


martinpitt (changes requested):
Can we please keep this in cockpit.git? This does not really belong into bots or the Machines API, especially not because it's specific to cockpit, and that is also used by lorax and other projects.

`test/common/testlib.py MachineCase` is still a bit too generic and low-level, but it feels less misplaced there. If possible/practical, rather generalize the setUp() code in test/verify/check-connection.

martinpitt:
Indeed https://github.com/cockpit-project/cockpit/pull/16281/files#diff-0a021f2b189c0bc16d23651fc96f367571fd719cdda25c443086f0b6caf270c0 is also check-connection, so keeping it in its setUp() seems perfectly fine -- all certificate tests should be in there anyway.

allisonkarlitskaya:
I had the same thought and considered adding this as a function like

```python
def libexecdir(pkg):
    if debian...:
        return f'/usr/lib/{pkg}'...
```

but I thought that'd be excessive.  There are quite some other hard references to cockpit in this file.

Also: it doesn't exactly belong in the test object either, because it's definitively an object of the machine.  We might imagine having two machines provisioned, with different libexecdir.  That's kinda why it actually felt like a hack to have it there.

Thoughts?

allisonkarlitskaya:
It's also worth noting that there are quite some places where this can be used from various testcases.  I've cleaned up 7 distinct sites across 3 files, *before* any remotectl/certificate-ensure changes.

martinpitt:
But we define`--libexecdir` in the projects -- it has nothing to do with the generic bots machinery. The cockpit references in machine.py are for starting cockpit's web server (which we need for every project when they explicitly call `.start_cockpit()`). I'd accept such a generic `libexecdir(pkg)` helper, but honestly I find it both excessive and inelegant -- it's much nicer IMHO to keep this in check-connection's setUp() function, where it is specific and targeted.

allisonkarlitskaya:
`check-connection`'s `setUp()` function isn't sufficient.  Take a look at cockpit-project/cockpit#16282 for the extent to which I plan to use this.

martinpitt:
So, `test/common/testlib.py MachineCase` then? It still won't apply to other projects, but at least keeps the code inside cockpit.git.

allisonkarlitskaya:
Okay.  Sure.

PR #2363, opened 2021-08-31
testmap: Drop rhel-8-4 and add fedora-35

marusak:
- Declare manual fedora-35 tests for projects that didn't have any
- Stop testing on rhel-8-4
- cockpit-composer won't be rebasing `rhel-8` brnach anymore, so it is
  "rhel-8-4"

Next I'll try to move f-35 into `main`.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -68,6 +68,7 @@ REPO_BRANCH_CONTEXT = {
             'centos-8-stream',
         ],
         '_manual': [
+            'fedora-35',
             'fedora-34/firefox',
         ],
     },
@@ -80,7 +81,6 @@ REPO_BRANCH_CONTEXT = {
     },
     'cockpit-project/cockpit-podman': {
         'main': [
-            'rhel-8-4',
             'rhel-8-5',
             'rhel-9-0',
             'fedora-33',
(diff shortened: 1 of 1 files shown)


lib/testmap.py
```diff
@@ -143,13 +142,10 @@
             'rhel-8-5',
             'rhel-9-0'
         ],
-        'rhel-8': [
-            'rhel-8-4',
-            'rhel-8-4/firefox',
-        ],
```
martinpitt, line 149: Did you check that with @henrywang and @teg ? Shouldn't that be moved to 8.5?

marusak: I checked with @jkozol : 

> I would like to create a 8.5 and a 9 branch instead of using rhel-8

jkozol: Yep. There is now a specific branch for 8.5: https://github.com/osbuild/cockpit-composer/tree/rhel-8-5. The `rhel-8` branch will remain at 8.4. But, then I think instead of removing this test shouldn't it be changed to `rhel-8-5`?

marusak: >  But, then I think instead of removing this test shouldn't it be changed to rhel-8-5?

Happy to add them of course. I'll first try to trigger them to check they work and then add them to testmap.

martinpitt (approved):

PR #2364, opened 2021-08-31
testmap: Run more contexts by default

marusak:
Testing here

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -65,10 +65,10 @@ REPO_BRANCH_CONTEXT = {
         'main': [
             'fedora-33',
             'fedora-34',
+            'fedora-35',
             'centos-8-stream',
         ],
         '_manual': [
-            'fedora-35',
             'fedora-34/firefox',
         ],
     },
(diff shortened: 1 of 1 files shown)


lib/testmap.py
```diff
@@ -139,12 +139,16 @@
         'main': [
             'fedora-33',
             'fedora-33/firefox',
+            'fedora-34',
             'rhel-8-5',
             'rhel-9-0'
         ],
+        'rhel-8-5': [
+            'rhel-8-5',
```
martinpitt, line 147: This ran in #2363 already.

lib/testmap.py
```diff
@@ -139,12 +139,16 @@
...
+            'rhel-8-5/firefox',
```
martinpitt, line 148: This did not run (unknown context), but indeed the chance is very high that it works, too. If not, we'll have to pay the price on the next image refresh :grin:

martinpitt (approved):
Thanks! Waiting for the final composer test (they are always a bit hard to get green..)

marusak:
> Waiting for the final composer test (they are always a bit hard to get green..)

f-34 is green, f-35 is gonna fail, but I did not move that away from manual as I had hunch it is not just gonna work. Was just trying out my luck here :)

marusak:
@jkozol FYI [fedora-35](https://logs.cockpit-project.org/logs/pull-2364-20210831-160644-6f7b31ae-fedora-35-osbuild-cockpit-composer/log.html) fails on main. I also enabled fedora-34 by default on main.

PR #2367, opened 2021-09-01
naughty: Relax pattern for systemd-tty-askpass watch_reads

martinpitt:
It also happens for path="/dev/ttyS0" and possibly others.

[example](https://logs.cockpit-project.org/logs/pull-16281-20210831-184227-dbe34f3f-fedora-34-firefox/log.html#151)


marusak (approved):
Thanks!

PR #2371, opened 2021-09-01
Add mobile manual scenario for fedora-34 (default OS) for cockpit and cockpit-machines

KKoukiou:


KKoukiou (commented):
Do we want to make the name more parameterizable for other screensizes? 
For example fedora-34/ws-xs etc.? Or even fedora-34/ws-414x736 (ws -> window size)

martinpitt (approved):
Fine for me to start with just "mobile" -- we can change it in the future if we want to, it's just between testmap and cockpit's test/run anyway.

PR #2384, opened 2021-09-06
testmap: run mobile fixes tests automatically

KKoukiou:


KKoukiou:
Let/s not break them now that they are in. :)

martinpitt (approved):
Thanks!

PR #2388, opened 2021-09-07
testmap: Add new cockpit rhel-9.0 branch

martinpitt:
Only test on rhel-9-0 image. We don't build any containers from this
branch.


martinpitt:
- PO refresh run: https://github.com/cockpit-project/cockpit/runs/3530124179?check_suite_focus=true
 - PO refresh PR: https://github.com/cockpit-project/cockpit/pull/16314

marusak (approved):
Thanks!

PR #2389, opened 2021-09-07
task: Use $GITHUB_REF as PR base branch if available

martinpitt:
This is set by GitHub action runs for "push", "schedule",
"workflow_dispatch", and similar non-PR events [1]. Similarly how our
release workflows already use `$GITHUB_REF` to determine which branch to
act on, use this variable as a fallback in `task.pull()` when not giving
an explicit `base` branch.

This fixes bots like po-refresh or npm-update when running them on
stable branches. Previously, their generated PRs were targetting main.

[1] https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables

----

https://github.com/cockpit-project/cockpit/pull/16313 is an example PR which was generated by current bots from me triggering "po-refresh" manually on the rhel-9.0 branch. Admittedly I did not test this locally, as it's quite involved to set up everything. After this lands, I'll do another rhel-8.5 translation update (which we need to do anyway), and test this in real life. I'll also trigger a run on main to be sure it still works there as well.

### DIFF
--- a/task/__init__.py
+++ b/task/__init__.py
@@ -373,9 +373,13 @@ def pull(branch, body=None, issue=None, base=None, labels=['bot'], run_tests=Tru
     if "pull" in kwargs:
         return kwargs["pull"]
 
+    # $GITHUB_REF is set when running from workflows
+    if not base:
+        base = os.path.basename(os.getenv("GITHUB_REF", default_branch()))
+
     data = {
         "head": branch,
-        "base": default_branch() if base is None else base,
+        "base": base,
         "maintainer_can_modify": True
     }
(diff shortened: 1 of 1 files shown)


marusak (approved):
Thanks! It makes sense and should be safe. I havn't tested this, but checked po-refresh and task/__init__.py code and seems like it should work.

martinpitt:
I [manually ran po-refresh against the rhel-9.0 branch](https://github.com/cockpit-project/cockpit/runs/3541620749?check_suite_focus=true), and lo and behold, it created a correct PR: #16323

PR #2391, opened 2021-09-08
po-refresh: Drop changed() heuristics

martinpitt:
It is error prone: There are translation fixes/updates which happen in
continuation lines instead of the initial `msgstr` line, and they
currently get ignored.

It is also obsolete with weblate, as that only commits files which got
actual changes.

### DIFF
--- a/po-refresh
+++ b/po-refresh
@@ -49,16 +49,6 @@ def run(context, verbose=False, **kwargs):
             sys.stderr.write("+ " + " ".join(args) + "\n")
         subprocess.check_call(args, cwd=cwd)
 
-    def changed(filename):
-        lines = output("git", "diff", "--", filename).split("\n")
-        msgstr = False
-        for line in lines:
-            if line.startswith("+msgstr ") and not line.endswith('""'):
-                if verbose:
-                    sys.stderr.write("{0}: {1}\n".format(filename, line[8:]))
-                msgstr = True
-        return msgstr
-
     # Make a tree with source and run "update-po" in it... outputs working directory
     line = output("bots/make-source", "update-po")
     work = os.path.abspath(line.strip())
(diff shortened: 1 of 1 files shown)


martinpitt:
Seen in https://github.com/cockpit-project/cockpit/pull/16323

po-refresh
```diff
@@ -159,16 +149,6 @@ def run(context, verbose=False, **kwargs):
                 current_linguas.sort()
             local_branch = add_and_commit_lang(name, locale[0], "Add")
 
-    # Here we have logic to only include files that actually
-    # changed translations, and reset all the remaining ones
-    files = output("git", "ls-files", "--modified", "po/")
-    for name in files.splitlines():
-        if name.endswith(".po"):
-            if changed(name):
-                execute("git", "add", "--", name)
```
marusak, line 168: Don't we need to `git add *.po` then?

marusak (changes requested):
Have you tested this locally? Just from the logic of the previous code seems that we should `git add` all the .po files?

po-refresh
martinpitt: No, we don't need to do that.`task.branch()` already `git add`s all files here: https://github.com/cockpit-project/bots/blob/main/task/__init__.py#L352 

po-refresh calls it with `pathspec="po/"`.

I did test it locally:
```
cp tmp/weblate-repo/*.po /var/home/martin/upstream/cockpit/po/
HEAD is now at 92f03a398 po: Update from Fedora Weblate
M	po/fr.po
M	po/pl.po
M	po/zh_CN.po
```
and the resulting commit had all three fr/pl/zh_CN files.

marusak (approved):
Ah, nice! Thanks :)

Issue #2392, opened 2021-09-08
Error mounting system-managed device /dev/dm-0: unknown filesystem type 'ntfs'


skobyda:
Reported: https://bugzilla.redhat.com/show_bug.cgi?id=2001755

cockpituous:
fedora-testing
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testResizeNtfs (__main__.TestStorageResize)
[0908/141611.495993:ERROR:bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[0908/141611.497764:WARNING:headless_browser_main_parts.cc(83)] Cannot create Pref Service with no user data dir.

DevTools listening on ws://127.0.0.1:9736/devtools/browser/f877f7eb-66af-4cc7-9cc3-0bb333a452c0
[0908/141611.508347:ERROR:gpu_init.cc(426)] Passthrough is not supported, GL is egl
[...]

PR #2380, opened 2021-09-03
Image refresh for fedora-testing

cockpituous:
Image refresh for fedora-testing
 * [x] image-refresh fedora-testing

PR #2393, opened 2021-09-08
naughty: Add pattern for ntfs-3g bug preventing a mount for ntfs file…

skobyda:
…system

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2001755
Known issue #2392


martinpitt:
Thanks! Can you please add this commit into #2380 instead, so that it can test itself?

PR #2431, opened 2021-09-21
Image refresh for rhel-7-9

cockpituous:
Image refresh for rhel-7-9
 * [x] image-refresh rhel-7-9
 * [x] Fix `check-menu` in cockpit on rhel-7.9 branch: https://github.com/cockpit-project/cockpit/pull/16373


martinpitt:
This [fails the same way](https://logs.cockpit-project.org/logs/pull-2431-20210921-023749-80bec242-rhel-7-9-cockpit-project-cockpit-rhel-7.9/log.html) as in the centos-7 refresh in #2428, and doesn't look like a flake. Not sure how it came about, possibly due to the tasks container refresh? needswork

martinpitt (approved):

PR #2437, opened 2021-09-22
testmap: promote rhel-8-6 for cockpit-podman from _manual

KKoukiou:


marusak:
`AssertionError: '0.00%' != 'n/a'` seems like 8-6 supports cgroupv2? This need to be adjusted in podman tests

KKoukiou:
> `AssertionError: '0.00%' != 'n/a'` seems like 8-6 supports cgroupv2? This need to be adjusted in podman tests

There is not cgroupv2 support but there is not an error any more. I will look into that.

martinpitt (approved):
Hah, thanks! I remember that naughty..

PR #2434, opened 2021-09-22
Image refresh for fedora-coreos

cockpituous:
Image refresh for fedora-coreos
 * [x] image-refresh fedora-coreos

PR #2400, opened 2021-09-10
Image refresh for fedora-34

cockpituous:
Image refresh for fedora-34
 * [x] image-refresh fedora-34
 * [ ] Blocked on https://github.com/cockpit-project/bots/issues/2434

PR #2439, opened 2021-09-23
Image refresh for fedora-34

cockpituous:
Image refresh for fedora-34
 * [x] image-refresh fedora-34


marusak:
```
$ rpm-ostree install --cache-only /var/tmp/build-results/cockpit-ws-*.rpm
error: Could not depsolve transaction; 1 problem detected:
 Problem: package cockpit-ws-253.41.gb19254ae3-1.wip.fc34.x86_64 requires (selinux-policy >= 34.20-1.fc34 if selinux-policy-targeted), but none of the providers can be installed
  - conflicting requests
```

oh no, not this again

martinpitt:
Still the same as in #2400 then. I was hoping that the Fedora CoreOS refresh in #2434 would update to a more recent version. But apparently this needs some deeper investigation, possibly the package is not installed at all any more or so?

allisonkarlitskaya:
→ #2455

PR #2460, opened 2021-09-29
two tweaks to po-refresh

allisonkarlitskaya:
See commits.

See also `po-refresh-via-deploy-key` branch on `cockpit-project/cockpit` where I'm testing this stuff out.


task/__init__.py
```diff
@@ -321,7 +321,8 @@ def censored(text):
         if verbose:
             sys.stderr.write("! " + str(e))
         raise
-    sys.stderr.write(censored(output))
+    if output:
```
martinpitt, line 324: How can this not  exist? the `except` handler re-raises the exception, so it should never get here?

allisonkarlitskaya: This is true.  And indeed, it didn't fix my problem that we're getting 

```
  File "/home/runner/work/cockpit/cockpit/src/bots/task/__init__.py", line 325, in execute
    sys.stderr.write(censored(output))
  File "/home/runner/work/cockpit/cockpit/src/bots/task/__init__.py", line 312, in censored
    return text.replace(api.token, "CENSORED")
TypeError: replace() argument 1 must be str, not None
```

so it must be something else.

allisonkarlitskaya: Oh.  It's the `api.token` that's NULL.  Duh.

martinpitt (approved):
Hah, and now it's actually *two* tweaks again :grin:

allisonkarlitskaya:
https://github.com/cockpit-project/cockpit/pull/16421 was opened by https://github.com/cockpit-project/cockpit/runs/3745189572?check_suite_focus=true

This seems to work very nicely now, complete with automatic runs of the workflows.  The "extra push over SSH" wasn't required in the end: our force-push to include the "Closes" line now happens via SSH (with the deploy key) and gets the workflows running as a side effect.

martinpitt (approved):
Cheers!

Developer pro tip: Always title your PRs "some tweaks" instead of a concrete number :grin:  SCNR

PR #2469, opened 2021-10-01
npm-update: use `npm outdated` instead of `upgrade`

allisonkarlitskaya:
There's a much easier way to figure out which versions are available:
`npm` now features an `outdated` command which will tell us which
versions we'd end up with for a given version string.  Now we prefix all
the versions, run the tool to get the upgrades, and then apply our logic
for grouping and removing already-pending updates.

### DIFF
--- a/npm-update
+++ b/npm-update
@@ -35,9 +35,9 @@ GROUP = [
 import collections
 import json
 import os
-import random
-import sys
 import subprocess
+import sys
+import tempfile
 
 sys.dont_write_bytecode = True
 
@@ -45,30 +45,23 @@ import task
 from lib.constants import BASE_DIR
 
 
-def package_json(data=None, package=None, version=None):
-    package_path = os.path.join(BASE_DIR, "package.json")
-    if data is None:
-        with open(package_path, "r") as f:
-            return json.load(f, object_pairs_hook=collections.OrderedDict)
-    else:
-        if package:
-            data = dict(data, dependencies=dict(data["dependencies"], **{package: version}))
-        with open(package_path, "w") as f:
-            json.dump(data, f, indent=2, separators=(',', ': '))
-            f.write("\n")
+def read_package_json(package_json_path):
+    with open(package_json_path, "r") as f:
+        return json.load(f, object_pairs_hook=collections.OrderedDict)
 
 
-def map_dict(dependencies, function):
-    items = []
-    for (name, value) in dependencies.items():
-        items.append((name, function(name, value)))
-    return collections.OrderedDict(items)
+def write_package_json(package_json_path, data):
+    with open(package_json_path, "w") as f:
+        json.dump(data, f, indent=2, separators=(',', ': '))
+        f.write("\n")
 
 
-def execute(*args):
-    if task.verbose:
-        sys.stderr.write("+ " + " ".join(args) + "\n")
-    subprocess.check_call(args, cwd=BASE_DIR)
+def prefix(name, version):
+    if not version[0].isdigit():
+        return version
+    if name not in FRAGILE:
+        return "^" + version
+    return "~" + version
 
 
 def output(*args):
@@ -77,9 +70,8 @@ def output(*args):
     return subprocess.check_output(args, cwd=BASE_DIR, universal_newlines=True)
 
 
-def run(specified_package, verbose=False, **kwargs):
+def run(context, verbose=False, **kwargs):
     pending_updates = []
-    updated_packages = []
 
     if not kwargs["dry"]:
         api = task.github.GitHub()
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
The main advantages here are two:
 - this is a lot faster and doesn't involve thrashing `node_modules` around
 - it's less of a security risk than installing a bunch of npm packages (with possible postinst scripts and so on)

npm-update
```diff
@@ -97,27 +89,23 @@ def run(specified_package, verbose=False, **kwargs):
...
-
-    def prefix(name, version):
-        if not version[0].isdigit():
-            return version
-        if name == specified_package or name not in FRAGILE:
-            return "^" + version
-        return "~" + version
+        if task.verbose:
+            sys.stderr.write("+ npm outdated --json")
+        result = subprocess.run(['npm', 'outdated', '--json'], cwd=tmpdir, stdout=subprocess.PIPE)
```
martinpitt, line 103: This covers both devDependencies and dependencies. Is there some way to filter only the latter? I didn't find something useful in `--help`, I don't think the command can do it by itself? But I also don't see code here which would do that filtering?

allisonkarlitskaya: As far as I know this only does the "dependencies".  I've looked at this output a lot of times and I also vaguely remember reading something in the docs that you need to give another flag to also do the dev depends.  I've also tested it and it's not doing anything like promoting all the dev depends to the "dependencies" section...

martinpitt: In current starter-kit for me:
```
❱❱❱ npm outdated 
Package                            Current        Wanted  Latest  Location
chrome-remote-interface             0.28.2        0.28.2  0.31.0  starter-kit
compression-webpack-plugin           6.1.1         6.1.1   9.0.0  starter-kit
copy-webpack-plugin                  6.4.1         6.4.1   9.0.1  starter-kit
core-js                             3.18.0        3.18.0  3.18.1  starter-kit
css-loader                           5.2.7         5.2.7   6.3.0  starter-kit
eslint-config-standard              14.1.1        14.1.1  16.0.3  starter-kit
eslint-config-standard-react         9.2.0         9.2.0  11.0.1  starter-kit
eslint-plugin-flowtype              5.10.0        5.10.0   6.1.0  starter-kit
eslint-plugin-promise                4.3.1         4.3.1   5.1.0  starter-kit
eslint-plugin-react                 7.26.0        7.26.1  7.26.1  starter-kit
eslint-plugin-standard               4.1.0         4.1.0   5.0.0  starter-kit
eslint-webpack-plugin                2.5.4         2.5.4   3.0.1  starter-kit
mini-css-extract-plugin             0.11.3        0.11.3   2.3.0  starter-kit
po2json                       1.0.0-beta-3  1.0.0-beta-3   0.4.5  starter-kit
react                              16.14.0       16.14.0  17.0.2  starter-kit
react-dom                          16.14.0       16.14.0  17.0.2  starter-kit
sass                                1.42.0        1.42.1  1.42.1  starter-kit
webpack                             5.53.0        5.56.1  5.56.1  starter-kit
```
which clearly includes both -- core-js and react are runtime dependencies, but e.g. webpack and eslint are development.

allisonkarlitskaya: weird.  I'll check this again.

allisonkarlitskaya:
```
⬢[lis@f34 starter-kit]$ npm outdated
Package                 Current   Wanted   Latest  Location
@patternfly/patternfly  MISSING  4.115.2  4.135.2  starter-kit
@patternfly/react-core  MISSING  4.135.0  4.157.3  starter-kit
core-js                 MISSING   3.15.2   3.18.1  starter-kit
react                   MISSING  16.14.0   17.0.2  starter-kit
react-dom               MISSING  16.14.0   17.0.2  starter-kit
```

I wonder if we have different npm configs around?

I just found this, for example: https://github.com/npm/npm/issues/6567

allisonkarlitskaya: In any case, it would be easy enough to filter the output from `npm outdated` to consider only the packages that we know are in `dependencies`.  I'll do that.

martinpitt: FTR, I have no ~/.npmrc, and validated that it does the same after `rm -rf ~/.npm` (which is just a download cache). Might be npm 6 vs. 7? Fedora 34 (and thus the cockpit/tasks container, and thus my toolbox) has npm 6.14.15.

allisonkarlitskaya:
Working nicely here: 

https://github.com/allisoninc/cockpit/actions/runs/1304037342 → https://github.com/allisoninc/cockpit/pull/6

martinpitt (approved):
Lovely, thanks! Tested locally against starter-kit (bumped core-js) and cockpit (bumped the patternfly group), and you already [tested it here](https://github.com/allisoninc/cockpit/pull/6/files)

PR #2480, opened 2021-10-06
Image refresh for fedora-33

cockpituous:
Image refresh for fedora-33
 * [x] image-refresh fedora-33


martinpitt:
For some reason the image is gone. Let's try this again.

PR #2499, opened 2021-10-08
test: Skip cockpituous test on testmap changes

martinpitt:
These are often sent from developer forks (where the cockpituous test
does not work), don't affect the deployment, and are self-validating.

----

Seen in e.g. PR #2498 which always get that unsightly cockpituous test failure.


marusak (approved):
Indeed, thanks!

PR #2517, opened 2021-10-14
Add support for pacman in image-diff

jelly:
Arch Linux uses pacman as it's package manager which does not output the
name and version seperated by a tab but a space.


image-diff
```diff
@@ -23,14 +23,16 @@ from machine import testvm
...
     # If both are missing, then the command will fail.
     #
     # We'd ideally like to get source packages everywhere, but it's a
     # bit more difficult on RPM. (TODO)
     pkgcmd = """if type dpkg-query > /dev/null 2>&1; then
                     dpkg-query -W -f='${source:Package}\t${version}\n' 2>/dev/null;
-                else rpm -qa --qf '%{NAME}\t%{EVR}\n' 2>/dev/null; fi"""
+                elif type rpm > /dev/null 2>&1; then
+                    rpm -qa --qf '%{NAME}\t%{EVR}\n' 2>/dev/null;
+                else pacman -Q | sed 's/ /\t/' 2>/dev/null; fi"""
```
martinpitt, line 35: FTR, it's just as well to use a space separator for dpkg and rpm, and change the split in line 38 accordingly. But this is fine, of course.

martinpitt (approved):
Thanks! Now please run `./tests-trigger 2517 image:arch` -- this will trigger all tests which use that image. There is only one right now, but it's a good habit to pick up, as e.g. `fedora-34` will trigger a dozen or more.

PR #2523, opened 2021-10-15
testmap: Drop cockpit rhel-9.0 branch again

martinpitt:
We do not need this for the time being.


marusak (approved):
Because this was beta but now we are rebasing GA from main, right?

martinpitt:
@marusak : Correct. We now have agreement between QA and other front door teams to do that.

PR #2536, opened 2021-10-18
images: copy the Candlepin certs out

ptoscano:
Copy the Candlepin SSL certificates/keys out of the container; this way
it will be possible for tests to validate the SSL connection to
Candlepin.

 * [x] image-refresh services


cockpituous:
image-refresh services done: https://github.com/cockpit-project/bots/commits/image-refresh-services-20211019-125151

PR #2582, opened 2021-11-01
images: Add missing bits for creating an arch package

jelly:
Implement the correct commands for generating an Arch Linux cockpit
package.

### DIFF
--- a/images/scripts/arch.install
+++ b/images/scripts/arch.install
@@ -2,7 +2,6 @@
 
 set -ex
 
-
 do_build=
 do_install=
 args=$(getopt -o "vqs:" -l "verbose,quick,skip:,build,install" -- "$@")
@@ -10,10 +9,8 @@ eval set -- "$args"
 while [ $# -gt 0 ]; do
 	case $1 in
 	    -v|--verbose)
-                shift
 		;;
 	    -q|--quick)
-                shift
 		;;
             -s|--skip)
                 skip="$skip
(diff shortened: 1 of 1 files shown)


images/scripts/arch.install
```diff
@@ -0,0 +1,5 @@
+#! /bin/bash
+
+set -e
+
+/var/lib/testvm/arch.install "$@"
```
martinpitt, line 5: Arch does not have multiple releases, and there is no parameterization here. Please just move arch.install from scripts/lib/ to scripts and avoid the indirection.

images/scripts/lib/arch.install
```diff
@@ -42,9 +42,20 @@ if [ -n "$do_build" ]; then
...
+    upstream_ver=$(echo "$dist_tar" | sed 's/^.*-//; s/.tar\..z//' | head -n1)
+
+    rm -rf cockpit-*/
+    tar -xf $dist_tar
+    ( cd $resultdir/
+      cp ../cockpit-*/tools/arch/* .
+      sed -i "s/SOURCE/$dist_tar/" PKGBUILD
+      sed -i "s/VERSION/$upstream_ver/" PKGBUILD
+      # Source needs to be next to the PKGBUILD
+      cp ../$dist_tar .
```
martinpitt, line 54: Would a `mv` be enough?

martinpitt (changes requested):
Thanks!

martinpitt (dismissed):
Danke! Feel free to land if you tested `test/image-prepare arch` locally, otherwise run an `arch@cockpit-project/cockpit` test here. It'll most likely fail, but it should at least get past image-prepare.

jelly:
> Danke! Feel free to land if you tested `test/image-prepare arch` locally, otherwise run an `arch@cockpit-project/cockpit` test here. It'll most likely fail, but it should at least get past image-prepare.

Tested it locally, still needs a PR against cockpit for the missing packaging bits. I'll open a PR there and then trigger the test.

jelly:
> > Danke! Feel free to land if you tested `test/image-prepare arch` locally, otherwise run an `arch@cockpit-project/cockpit` test here. It'll most likely fail, but it should at least get past image-prepare.
> 
> Tested it locally, still needs a PR against cockpit for the missing packaging bits. I'll open a PR there and then trigger the test.

This PR https://github.com/cockpit-project/cockpit/pull/16547 now proves the tests can run with the changes in this PR.
https://logs.cockpit-project.org/logs/pull-16547-20211102-100749-e0a2e91d-arch-bots-2582/log.html

martinpitt (approved):
Nice, thanks! (and ugh, so many failures)

PR #2592, opened 2021-11-03
testmap: Add Arch Linux to cockpit-machines main testset

jelly:


martinpitt (approved):
Thanks! Please trigger the test here again, to make sure it still works.

PR #2594, opened 2021-11-04
RHEL: make insights-client non-verbose

### LINKED ISSUE #2563: [no-test] Image refresh for rhel-9-0
Image refresh for rhel-9-0
 * [ ] image-refresh rhel-9-0

ptoscano:
Apparently, newer versions of the insights-client in RHEL 9 log the HTTP traffic, which includes binary data (the egg file). Since we are not really interested in it, and that the log file is available locally anyway, simply make the `--status` run as non-verbose.

Also, print the RPM version of the `insights-client` package, to help debugging problems.

Fixes #2563

 - [ ] image-refresh rhel-9-0

### DIFF
--- a/images/scripts/rhel-7-9.setup
+++ b/images/scripts/rhel-7-9.setup
@@ -195,6 +195,7 @@ fi
 # last_stable.egg makes that work as expected, too.
 
 if [ -x /usr/bin/insights-client ]; then
+    rpm -q insights-client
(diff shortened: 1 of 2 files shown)


martinpitt:
Sorry @ptoscano , I copied that as PR #2595 as you can't refresh images as a non-member.

PR #2600, opened 2021-11-05
machine: disable PKCS11Provider to avoid errors

allisonkarlitskaya:
Due to our use of BatchMode=yes and the fact that I use PKCS11 for my
SSH key, I see a bunch of superfluous logging output from ssh every time
I start a VM:

    $ bots/vm-run -q fedora-35
    pin required
    C_FindObjectsInit failed: 179
    C_FindObjectsFinal failed: 179
    C_FindObjectsInit failed: 179
    C_FindObjectsFinal failed: 179

    SSH ACCESS
      $ ssh -p 2201 -i bots/machine/identity root@127.0.0.2

    COCKPIT
      http://127.0.0.2:9091

    [ ^C to terminate ]

That's caused by https://bugzilla.mindrot.org/show_bug.cgi?id=3359.

Let's increase the LogLevel by one notch to FATAL to avoid these
messages (which are issued at the "ERROR" level).

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -39,7 +39,7 @@ def write_all(fd, data):
 
 class SSHConnection(object):
     ssh_default_opts = ["-o", "StrictHostKeyChecking=no", "-o", "UserKnownHostsFile=/dev/null",
-                        "-o", "IdentitiesOnly=yes", "-o", "BatchMode=yes"]
+                        "-o", "IdentitiesOnly=yes", "-o", "BatchMode=yes", "-o", "PKCS11Provider=none"]
 
     def __init__(self, user, address, ssh_port, identity_file, verbose=False):
         self.verbose = verbose


allisonkarlitskaya:
This is one possible solution, but after a bit of testing, I don't like it, since it also causes messages like this one to disappear:

```
ssh: connect to host 127.0.0.2 port 2201: Connection refused
```

Other ideas:

 - drop `BatchMode=yes` and possibly replace it with `PasswordAuthentication=no`
 - use `-F /dev/null` to cancel out any of the user's config options
 - try to disable PKCS11 mode.  This turns out to be difficult, though, because:
   - there doesn't seem to be a way to *unset* an `ssh_config` variable from the commandline
   - `-o PKCS11Provider=` gives "no argument after keyword "
   - `-o PKCS11Provider=/dev/null` just trades one error for another (`dlopen` error)
 - I could try and cook up some local hack on my system and save the rest of you the trouble, although this impacts anyone who is using PKCS11 with ssh (which is a really great idea)

allisonkarlitskaya:
I figured it out: the correct way to unset a parameter in SSH is to set it to the magic value of `none`.  I'll redo the patch with `-o PKCS11Provider=none`.

martinpitt (approved):
Nice! Triggering a starter-kit test to make double-sure, but that ought to be fine.

allisonkarlitskaya:
I also noticed something else kinda random, but probably not useful: my kids got this toy whistle included for free in a box of Captain Crunch that I bought last week.  If I hold it near the microphone of my computer and blow in it while attempting to make an SSH connection then I can also connect without authentication.  Cool!

I guess that's not a practical workaround for most people, though.  So let's stick with the hardwired identity file.

martinpitt:
@allisonkarlitskaya : Do hold on to the whistle, though! It also unlocks Kerberos, and even works for ice cream shop doors.

PR #2610, opened 2021-11-08
images: Drop fedora-33 image

KKoukiou:
And the relevants tests from the projects which were still running
against this image.


martinpitt (changes requested):
Yay! Can you please drop naughty, the install/bootstrap scripts and the actual image symlink as well? See `find -name fedora-33*`

martinpitt (changes requested):
Thanks! can you please also remove images/fedora-33 itself?

Triggering the new composer test in the meantime -- last time it failed, not sure if that was a flake.

If it still fails, I suggest to move it to _manual instead, and ask @jkozol / @henrywang to look at it separately.

martinpitt (approved):
Cheers! Test succeeded in the previous round, I updated the comment to capture the log URL.

PR #2612, opened 2021-11-08
tests-trigger: support specifying '-' as PR number

allisonkarlitskaya:
In case '-' is given as the PR number to trigger, look for the upstream
branch of the current branch, and trigger the tests on the corresponding
HEAD commit.

This will do the "right thing" most of the time for when you want to
trigger tests for the feature branch you're hacking on and just pushed
changes to.

This skips the "is the PR author trusted?" checks because we don't look
at the PR.  That's fine: this is the code you're working on locally.

### DIFF
--- a/tests-trigger
+++ b/tests-trigger
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import argparse
+import subprocess
 import sys
 
 sys.dont_write_bytecode = True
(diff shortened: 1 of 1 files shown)


martinpitt:
I tried to test this with my branch layout, where origin == c-project/cockpit, and my own repo origin name is "my". That doesn't work:
```
❱❱❱ bots/tests-trigger - rhel-9-0
fatal: no upstream configured for branch 'sepolicy-rhel8'
[...]
subprocess.CalledProcessError: Command '['git', 'rev-parse', '@{u}']' returned non-zero exit status 128.
```
supposedly this only works for a different origin setup. That said, if it works for you, it's a nice trick!

tests-trigger
```diff
@@ -10,20 +11,26 @@ from task import github
...
+        # nor isn't in the 'Contributors' group, but issue a warning in case of an oversight
+        login = pull["head"]["user"]["login"]
+        if not opts.allow and not api.is_user_allowed(login):
+            sys.stderr.write("Pull request author '{0}' isn't allowed. Override with --allow.\n".format(login))
+            return 1
+
+        revision = pull['head']['sha']
+        target = f'pull request {opts.pull}'
+    else:
+        revision = subprocess.check_output(['git', 'rev-parse', '@{u}'], encoding='utf-8').strip()
```
martinpitt, line 31: I suggest to use the long name `{upstream}` to make this a bit easier to read.

martinpitt (changes requested):
Can you please add this to the help of the "pull" argument as well? Thanks!

allisonkarlitskaya:
> I tried to test this with my branch layout, where origin == c-project/cockpit, and my own repo origin name is "my". That doesn't work:
> 
> ```
> ❱❱❱ bots/tests-trigger - rhel-9-0
> fatal: no upstream configured for branch 'sepolicy-rhel8'
> [...]
> subprocess.CalledProcessError: Command '['git', 'rev-parse', '@{u}']' returned non-zero exit status 128.
> ```
> 
> supposedly this only works for a different origin setup. That said, if it works for you, it's a nice trick!

You need to push with `git push -u` to set the upstream branch.

I always do like

```
main$ git worktree add ../feature
main$ cd ../feature
feature$
...work...
feature$ git push -u lis HEAD
```

which pushes `feature` to my `lis` remote and sets it as the upstream branch.

tests-trigger
allisonkarlitskaya: done, thanks

allisonkarlitskaya:
Made the suggested changes, plus changed the output to display the branch name instead of the raw revision.  Introduce a `git` helper function to DRY things up a bit.

It might also be interesting to support github PR URLs here from which both the `--repo` and target could be parsed out.

Going in a different direction, it might also be nice to turn the target into an explicit argument like `--pr` or `--url` (for the above suggestion) and default to the `-` case in the future, if no argument is given.  Let's see how useful this feature ends up being before we do that, though.

martinpitt (approved):
Thanks!

PR #2613, opened 2021-11-09
Image refresh for arch

cockpituous:
Image refresh for arch
 * [x] image-refresh arch


martinpitt:
[c-machines regression](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-2613-20211109-022124-12e8d9cc-arch-cockpit-project-cockpit-machines/log.html#34-2) -- there is a running VM even though the test expects none. Some broken cleanup? But the arch test has worked quite reliably recently, so that might be an actual regression. Retrying once for comparison.

jelly (approved):

PR #2616, opened 2021-11-09
testmap: add Fedora branches for subscription-manager/main

ptoscano:
The 'main' branch of subscription-manager supports Fedora now in the
cockpit CI; hence, add the currently supported Fedora versions for it.

The Fedora fixes were done in https://github.com/candlepin/subscription-manager/pull/2861


martinpitt (approved):
Awesome, grazie!

PR #2617, opened 2021-11-09
naughty: Copy pmlogger naughty to Arch

jelly:
Arch Linux has the latest pcp which has the same issue of not starting
when not enabled.


jelly:
Should show up in here https://logs.cockpit-project.org/logs/pull-16547-20211109-175024-2746809d-arch/log.html

martinpitt (approved):

PR #2649, opened 2021-11-20
Image refresh for rhel-9-0

cockpituous:
Image refresh for rhel-9-0
 * [x] image-refresh rhel-9-0


marusak:
This fails because of https://bugzilla.redhat.com/show_bug.cgi?id=2023579 and with cockpit 258 this should work fine. Let refresh this in couple of days when we have cockpit-258 in rhel.

martinpitt:
RHEL 9 nightly got a new compose with cockpit 258 now.

PR #2648, opened 2021-11-20
Image refresh for rhel-8-6

cockpituous:
Image refresh for rhel-8-6
 * [x] image-refresh rhel-8-6


marusak:
This fails because of https://bugzilla.redhat.com/show_bug.cgi?id=2023579 and with cockpit 258 this should work fine. Let refresh this in couple of days when we have cockpit-258 in rhel.

martinpitt:
FTR, RHEL 8.6 nightly still did not get a newer compose, blocked for longer.

marusak:
> Update (Dec 1, 06:30 UTC): latest compose is still RHEL-8.6.0-20211129.1, still blocked

This compose if from Monday and it has cockpit-258 already present.

PR #2665, opened 2021-11-26
testmap: update cockpit-composer testing

jkozol:
cockpit-composer will only release into RHEL 8.6 and 9.0 as well as fedora 34. Eventually, we will also release into fedora 35 so enable the firefox test for fedora 35.


jkozol:
Tests passed here: https://github.com/osbuild/cockpit-composer/pull/1373

jkozol:
Updated my token and repushed. Hopefully passes test now :crossed_fingers:

marusak (approved):
Thanks!

PR #2669, opened 2021-11-27
Image refresh for rhel-9-0

cockpituous:
Image refresh for rhel-9-0
 * [x] image-refresh rhel-9-0
 * [x] https://github.com/cockpit-project/cockpit-podman/pull/838
 * [x] https://github.com/cockpit-project/cockpit-machines/pull/461
 * [x] https://github.com/osbuild/cockpit-composer/pull/1377


marusak:
There are no new composes until Monday, so we need to wait with this for a couple of days

marusak:
> There are no new composes until Monday, so we need to wait with this for a couple of days

Ignore me, I am taking this back. There was indeed new compose. The problem is that 258 has new shell - we can bump test/common to 258, but that will break OSes that still have 257. I will make update to test/common to make it work for each version.

martinpitt:
I filed PRs for c-{podman,machines,composer}, added as blockers to description. However, [sub-man](https://logs.cockpit-project.org/logs/pull-2669-20211127-022620-f250b10b-rhel-9-0-candlepin-subscription-manager/log.html) fails due to some package uninstallability, and it is *not* affected by the .logout() change (it doesn't use that API).

martinpitt:
The projects were fixed for cockpit 258. It's still blocked by the sub-man package uninstallability, so keeping needswork.

martinpitt:
I tried `dnf install -y cockpit-ws python3-devel openssl-devel intltool gcc` on that refreshed image, and that works now. Possibly it was fixed in the rhel 9 nightly composes. Retrying that one, if that works, I'll rebase to get clean test runs.

martinpitt:
Cool, [worked](https://logs.cockpit-project.org/logs/pull-2669-20211129-142509-f250b10b-rhel-9-0-candlepin-subscription-manager/log.html)!

martinpitt (approved):

PR #2677, opened 2021-11-30
lib: Drop CentOS CI image store

martinpitt:
Linode stores have proven to be very reliable, and we still have
redundancy on cockpit-11. Let's tear down the image server on CentOS CI
to reduce usage of our custom cert/key and upload token, and save some
bandwidth/time on image refreshes (which upload to all stores).


allisonkarlitskaya (approved):
:+1: thanks

PR #2688, opened 2021-12-03
images: Set up firewalld correctly for cockpit on Arch Linux

jelly:
Some firewalld tests remove all public ports which means that port 9090
is removed, as we always want cockpit to be available enable it via a
service. Also move the firewalld enabling to .install as the other
images.

 * [x] image-refresh arch


martinpitt:
Looks good, thanks! Needs an image refresh now to pick up the .setup change

cockpituous:
image-refresh in progress on 4-cockpit-11.
Log: https://logs.cockpit-project.org/logs/image-refresh-2688-20211203-081818/

cockpituous:
image-refresh arch done: https://github.com/cockpit-project/bots/commits/image-refresh-arch-20211203-082929

martinpitt (approved):
Triggering the arch image tests. If/when they go green, and it helps to advance your cockpit tests, please land.

images/scripts/arch.install
```diff
@@ -59,6 +59,8 @@ if [ -n "$do_install" ]; then
     packages=$(find build-results -name "*.pkg.tar.zst")
     pacman -U --noconfirm $packages
 
+    firewall-cmd --add-service=cockpit --permanent
```
martinpitt, line 62: FTR, it might be nicer to move the `systemctl enable --now firewalld` to the .install script, or drop it at all. Our other images also don't do this by default, and it's always nice to keep the pristine image as close to a default install as possible.

PR #2705, opened 2021-12-09
constants, testmap: Set TEST_DEFAULT_OS to fedora-35

mvollmer:


mvollmer:
Hmm, I guess `fedora-35/mobile` needs to be in testmap on main in order to be triggerable here, right?  But `fedora-35` should run the pixeltests now, let's see.

lib/testmap.py
```diff
@@ -27,16 +27,16 @@
...
             'debian-stable',
             'debian-testing',
             'ubuntu-2004',
             'ubuntu-stable',
             'fedora-34',
-            'fedora-34/devel',
-            'fedora-34/firefox',
-            'fedora-34/mobile',
-            'fedora-35',
+            TEST_OS_DEFAULT,
```
allisonkarlitskaya, line 36: I'd maybe avoid this particular one and just leave it open.

lib/testmap.py
```diff
@@ -27,16 +27,16 @@
...
+            TEST_OS_DEFAULT + '/devel',
+            TEST_OS_DEFAULT + '/firefox',
+            TEST_OS_DEFAULT + '/mobile',
```
allisonkarlitskaya, line 39: ...and for these three, please use f-strings.

lib/testmap.py
```diff
@@ -52,28 +52,25 @@
             'rhel-8-5',
             'rhel-8-5-distropkg',
             'centos-8-stream',
-            'fedora-34/container-bastion',
+            TEST_OS_DEFAULT + '/container-bastion',
```
allisonkarlitskaya, line 55: ditto.

(and all the others under)

allisonkarlitskaya (changes requested):

lib/testmap.py
mvollmer: Ok, done.

lib/testmap.py
mvollmer: Done.

lib/testmap.py
mvollmer: Done.

mvollmer:
Summary of pixel test failures with Fedora 35:
- virbr0 isn't listed anymore on the networking page
- setroubleshoot text has changed, affecting sort order of solutions
- zram is offered for creating vgroups on and stuff
- /dev/vda4 is offered as well, empty partition on root disk?
- system info says "fedora-35" instead 34 :-)

I check the storage things, ideally zram0 and vda4 wouldn't be offered.

mvollmer:
> Summary of pixel test failures with Fedora 35:

Also, fedora-35 has a failed service in the health card.  I check that as well.

mvollmer:
> Also, fedora-35 has a failed service in the health card. I check that as well.

This is rhsm.service, due to https://bugzilla.redhat.com/show_bug.cgi?id=1995465.  It should be ignored in the pixel test, along with the rest of the health card... hmm.

mvollmer:
> I check the storage things, ideally zram0 and vda4 wouldn't be offered.

Let's just accept them. vda4 is flagged as "prep", which is some powerpc boot thingy, zram should be ignored in general, but right now it isn't so let's live with it.

mvollmer:
Nice, machines and podman don't need any changes to pixeltests.

mvollmer:
Updated pixels in https://github.com/cockpit-project/cockpit/pull/16716

mvollmer:
fedora-35/devel@cockpit-project/cockpit failure needs investigation: https://logs.cockpit-project.org/logs/pull-2705-20211210-130557-c6eac50e-fedora-35-devel-cockpit-project-cockpit/log.html#109-2

mvollmer:
> fedora-35/devel@cockpit-project/cockpit failure needs investigation:

TestUpdatesSubscriptions.testNoSubOsRepo fails on */devel because packagekit is not pre-loaded there and thus the health card stays empty. I'll fix that by explicitly loading the `/updates` page in the test.

mvollmer:
> > fedora-35/devel@cockpit-project/cockpit failure needs investigation:
> 
> TestUpdatesSubscriptions.testNoSubOsRepo fails on */devel because packagekit is not pre-loaded there and thus the health card stays empty. I'll fix that by explicitly loading the `/updates` page in the test.

We agreed to just skip all of TestUpdatesSubscriptions on fedora-35, https://github.com/cockpit-project/cockpit/pull/16720

allisonkarlitskaya (approved):
Looks fine to me, if it lands in lockstep with the pixel test changes on cockpit repo.

PR #2707, opened 2021-12-09
testmap: Add Arch Linux as cockpit test requirement

jelly:
Cockpit now supports testing on Arch Linux.


martinpitt (approved):
Thanks! Triggering arch test run to double-check, please merge after green.

PR #2718, opened 2021-12-13
vm-run: add -c option for Cockpit Client

allisonkarlitskaya:
Add a `-c` commandline argument to `vm-run` to open Cockpit Client on
the VM once it's started.

In order for this to work nicely, you should have followed the ssh
configuration suggestions in `test/README.md` in the Cockpit repository.
You also need a (very new) version of Cockpit Client, which understands
hostname:port syntax.

### DIFF
--- a/vm-run
+++ b/vm-run
@@ -31,6 +31,7 @@ parser.add_argument('-m', '--maintain', action='store_true', help='Changes are p
 parser.add_argument('-M', '--memory', default=None, type=int, help='Memory (in MiB) of the target machine')
 parser.add_argument('-G', '--graphics', action='store_true', help='Display a graphics console')
 parser.add_argument('-q', '--quiet', action='store_true', help="Don't connect text console")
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
I want to add a `vm-run --page` (`-p`) here as well, that opens to a specific page, but I'm having difficulty getting it working.

martinpitt:
This depends on https://github.com/cockpit-project/cockpit/pull/16721

martinpitt (approved):
Cool! That's what you meant with "yay D-Bus activation" (although of course a simple CLI arg would do just as well)

allisonkarlitskaya:
> Cool! That's what you meant with "yay D-Bus activation" (although of course a simple CLI arg would do just as well)

The win here is more about not needing to figure out how to launch the app, potentially already running, potentially from inside of a toolbox, etc.

PR #2734, opened 2021-12-19
Image refresh for fedora-35

cockpituous:
Image refresh for fedora-35
 * [x] image-refresh fedora-35

Needs to land together with:
 - https://github.com/cockpit-project/cockpit-podman/pull/874
 - https://github.com/cockpit-project/cockpit-machines/pull/501


martinpitt:
Whoa, pixel test failure galore -- what on earth happened here? Some just look like a different sizing, but e.g. [this one](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-2734-20211219-024407-eeced981-fedora-35-rawhide-cockpit-project-cockpit-podman/pixeldiff.html#TestApplication-testRunImageSystem-integration) actually changed spacing.

jelly:
> Whoa, pixel test failure galore -- what on earth happened here? Some just look like a different sizing, but e.g. [this one](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-2734-20211219-024407-eeced981-fedora-35-rawhide-cockpit-project-cockpit-podman/pixeldiff.html#TestApplication-testRunImageSystem-integration) actually changed spacing.

In the updated packages it lists cockpit going from 258 => 259, could it be that a new cockpit changed something which influences the css/styling of podman/containers.

mvollmer:
> Whoa, pixel test failure galore -- what on earth happened here?

The content iframe has changed size between Cockpit 258 and 259. The height has decreased a bit but the width is unchanged.

If the width would have changed as well, I think we would have seen a lot more pixel test failures.

(The pixel text results are kinda confusing because pixeldeiff.html is sh*t at showing differences between images of different size. So it looks like some margins have changed while in reality pixeldiff.html is cropping the larger image down to the size of the smaller one.)

So, what can we do about this?

I think we need to guarantee a fixed iframe size in the pixel tests. Right now we set a fixed browser window size depending on the scenario (desktop vs mobile), but I think we should set a fixed iframe size.

mvollmer:
> I think we need to guarantee a fixed iframe size in the pixel tests.

Here is my attempt: https://github.com/cockpit-project/cockpit/pull/16770

This will fix the size to what current Cockpit main produces, which is different from what cockpit-machines and cockpit-podman expect. So we need to update their pixel tests anyway and we can do that before landing #16770 and switching c-m and c-p over to it.

martinpitt (approved):
Thanks @mvollmer! Moving to blocked, waiting for the two PRs to finish testing.

PR #2733, opened 2021-12-18
Image refresh for arch

cockpituous:
Image refresh for arch
 * [x] image-refresh arch

PR #2754, opened 2021-12-25
Image refresh for rhel-9-0

cockpituous:
Image refresh for rhel-9-0
 * [x] image-refresh rhel-9-0
 * [x] sort out qemu 6.1 → 6.2 NIC delete regression in #2733


skobyda:
At first glance the c-machines failure looks like the same as here https://github.com/cockpit-project/bots/pull/2733

martinpitt:
The [cockpit-composer failures](https://logs.cockpit-project.org/logs/pull-2754-20211225-062911-0ac9cd4a-rhel-9-0-osbuild-cockpit-composer/log.html) happen on the current image as well, this spontaneously broke because the test uses the real nightly archive. See e.g. https://github.com/osbuild/cockpit-composer/pull/1385#issuecomment-1003884825 . So we an ignore these here.

martinpitt:
Right, this image also updated QEMU from 6.1 to 6.2. Moving "needswork" to "blocked", let's figure this out in #2733 and add naughties for all affected OSes there.

martinpitt:
Rebased to pick up naughty for the virt-xml regression.

martinpitt (approved):

PR #2761, opened 2022-01-01

cockpituous:
 * [ ] https://github.com/cockpit-project/cockpit/pull/16767

### DIFF
--- a/images/centos-8-stream
+++ b/images/centos-8-stream
@@ -1,1 +1,1 @@
-centos-8-stream-526ba9d9fd02ee4f6e299eeb4ca318cc44f13f49faa2dd53e11cb71e7a20d1b6.qcow2
+centos-8-stream-17765dbca3d79e34bef7b6f0870fe4dcec901c33c76c8343be58afbd6614ed8f.qcow2


martinpitt:
This [testPlots failure](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-2761-20220101-081017-c533283e-centos-8-stream-cockpit-project-cockpit/log.html#117-2) consistently happens on CentOS CI, also on current image. So this is not a regression, and once e2e is back it'll probably succeed. But at some point we should investigate this anyway.

martinpitt:
Calculating back from `wait_js_cond(ph_plot_data_plateau("direct",459072102.4,507395481.6,15,"mem"))` it means that MemAvailable was 483233792, or 460.85 MB. The [plot](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-2761-20220101-081017-c533283e-centos-8-stream-cockpit-project-cockpit/TestPages-testPlots-centos-8-stream-127.0.0.2-2301-FAIL.png) seems okay-ish, but it is hard to see the exact number because it is really small. 

I can perfectly reproduce this locally as well. There is no bubble help or so, and when I log into the VM a second time, the available memory goes down dramatically. So IMHO the test is just overly picky.

I filed https://github.com/cockpit-project/cockpit/pull/16767 to relax this.

martinpitt:
I accidentally removed the bots label. #2773 is a newer refresh

PR #2785, opened 2022-01-10

cockpituous:


martinpitt:
There's some SELinux change for restricted users which may or may not be intented.

martinpitt:
There is also [another new SELinux violation](https://logs.cockpit-project.org/logs/pull-16809-20220111-055612-4e8fec6a-fedora-testing-bots-2785/log.html#249):

    avc:  denied  { unlink } for  pid=3663 comm="systemd-user-ru" name="bus" dev="tmpfs" ino=15 scontext=system_u:system_r:systemd_logind_t:s0 tcontext=unconfined_u:object_r:session_dbusd_tmp_t:s0 tclass=sock_file permissive=0

Sounds related to https://bugzilla.redhat.com/show_bug.cgi?id=2020977 . There is nothing to block in bodhi, both systemd and selinux-policy are already in stable. So back to needswork, this needs to be reported and naughtied.

martinpitt:
Filed known issue #2799, pushed naughty. After this lands, we should move f-testing to F35, it's past time.

marusak (approved):

PR #2795, opened 2022-01-11
rename pr-tests to tests-status, adding features

allisonkarlitskaya:
- we now support the `-` syntax for "upstream of current branch", as per `tests-trigger`
 - add a new `--wait` option to block until one test is red or all tests are green

### DIFF
--- a/pr-tests
+++ b/pr-tests
@@ -16,11 +16,11 @@
 #
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
 import argparse
+import ssl
 import sys
 import urllib.request
-import ssl
-import json
 
 from lib.stores import CA_PEM
 import task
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
On the topic of triggering statuses on non-PRs: if the PR is triggered on a commit that exists on a branch that's part of the main cockpit repo (ie: if you pushed to origin, not your fork) then we get an event on the status webhook.  I've stolen the data from the "deliveries" tab and injected it via curl into the webhook running locally.  

That gets at least as far as to the point where it would inject the job into the queue, but I'm not sure what happens past that point, since I don't have an AMQP server set up locally.

I'm not sure if this is working on the real webhook, though, because I've spammed inspect-queue like crazy while posting the status, and I've never seen it show up there.

martinpitt (approved):
The changes look good to me, and this is a tool that hardly anyone uses right now. I still think it's easier all around to just file draft PRs than "detached" statuses (also a nice place to keep notes, TODOs, and old URL links and such), but the changes here are useful for PRs as ewll. Thanks!

PR #2797, opened 2022-01-12

cockpituous:


martinpitt:
Still seems to have trouble booting, now for second boot? Although not all tests are slow, just a few.

martinpitt (approved):

PR #2802, opened 2022-01-12
testmap: let anaconda tests run automatically

KKoukiou:
They are very reliable, so i think this is the way to go.


martinpitt (changes requested):
Not much point to *also* have that in `_manual` -- can you clean that up, please?

KKoukiou:
@martinpitt I need it in order to be able to trigger tests in other branches if needed.

martinpitt (approved):
Right, but you can do that -- the context just needs to be known, it can then be manually triggered on any other branch.

PR #2808, opened 2022-01-13
test: Really really skip cockpituous test on testmap changes

marusak:
Commit e3c5f32bef tried to fix this but we still were running tests on
testmap changes. The issue is that there was trailing `/` for each of
the locations, but testmap.py is not a directory.


martinpitt (approved):
Whoops, thanks!

PR #2818, opened 2022-01-18
fedora.install: make --quick quicker

allisonkarlitskaya:
In case --quick is specified, give `--nodigest --nosignature` to the
`rpm` invocation (which speeds things up by almost a second) and skip
the "file installed by multiple packages" check.

Also, instead of implementing the duplicates and "is this a normal
file?" logic in a for loop in shell, use `uniq -d` and `find -type f`,
respectively.  This simplifies the test and makes it considerably faster
(as well as substantially less chatty).

I also entertained adding `nosync` to the image and setting it as an
LD_PRELOAD at the top of the script, but some quick experiments didn't
show it to make much of a difference, either way.


martinpitt (approved):
Nice, thanks! I triggered a cockpit test to be sure that this doesn't break.

PR #2820, opened 2022-01-18
Image refresh for fedora-35

cockpituous:
Image refresh for fedora-35
 * [x] image-refresh fedora-35


martinpitt:
selinux-policy is still too old. Testing Farm already sees the new version. Let's retry this in a day or two.

PR #2823, opened 2022-01-19

cockpituous:


marusak:
[one](https://logs.cockpit-project.org/logs/pull-2823-20220119-075933-d17da974-ubuntu-stable-cockpit-project-cockpit/log.html#261-2) failure, let's retry

martinpitt:
This failure happens all the time now, also on the current image. I'm investigating. My first instinct was that we need to wait a little for the cockpit-delegation rule to propagate to the client. But that's not it -- even with a sit() and waiting for a minute or so, running `sudo` with the kerberos ticket doesn't work.

martinpitt:
See above, pushed naughty.

martinpitt:
Good news: The issue above is fixed in 22.04 (current devel series), so over time this will fix itself.

marusak (approved):
THanks!

Issue #2775, opened 2022-01-03
"virt-xml --remove-device --network" fails with QEMU 6.2 update


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2036562

cockpituous:
arch
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testNICDelete (__main__.TestMachinesNICs)

DevTools listening on ws://127.0.0.1:9318/devtools/browser/898b7a5e-fa03-419e-af34-f53e557bbd19
[0103/072719.590858:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: cannot open shared object file: No such file or directory
[0103/072719.591938:ERROR:viz_main_impl.cc(162)] Exiting GPU process due to errors during initialization
[0103/072719.595457:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is disabled
> error: Modal: Specify at least one of: title, aria-label, aria-labelledby.
> error: You are currently using minified code outside of NODE_ENV === "production". This means that you are running a slower development build of Redux. You can use loose-envify (https://github.com/zertosh/loose-envify) for browserify or setting mode to production in webpack (https://webpack.js.org/concepts/mode/) to ensure you have the correct code for your production build.
> error: Scrollbar test exception: TypeError: Cannot read properties of null (reading 'appendChild')
> warning: grep: /sys/class/dmi/id/power/autosuspend_delay_ms: Input/output error

> warning: sh: warning: setlocale: LC_ALL: cannot change locale (C.UTF-8)

Traceback (most recent call last):
  File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/work/bots/make-checkout-workdir/test/check-machines-nics", line 141, in testNICDelete
    b.wait_not_present("#vm-subVmTest1-network-1-mac")
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 424, in wait_not_present
    self.wait_js_func('!ph_is_present', selector)
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 415, in wait_js_func
    self.wait_js_cond("%s(%s)" % (func, ','.join(map(jsquote, args))))
[...]

PR #2828, opened 2022-01-20

cockpituous:

### DIFF
--- a/images/debian-testing
+++ b/images/debian-testing
@@ -1,1 +1,1 @@
-debian-testing-e67f0e19b0e64f0078a90034854e20199b94d232b7ecbb96309a871c03ec9b1f.qcow2
+debian-testing-16d83256c7cab0e62380c68acabf8fcce80c5cddfcba579bfa5dc0bd017ed633.qcow2
--- a/naughty/debian-testing/2775-virtxml-remove-net
+++ b/naughty/debian-testing/2775-virtxml-remove-net
@@ -1,0 +1,5 @@
+Traceback (most recent call last):*
+  File "test/check-machines-nics", line *, in testNICDelete
+    b.wait_not_present("#vm-subVmTest1-network-1-mac")
+*
+testlib.Error: timeout


martinpitt:
The c-machines failure looks like known issue #2775, just need to copy the naughty.

martinpitt:
Copied the naughty.

martinpitt:
The remaining failure is different. That looks like a time-dependent failure?

marusak:
So there is something broken there.
If I do `sudo journalctl -r` I can scroll through to previous boot. But if I add `--boot -1` it says `Data from the specified boot (-1) is not available: No such boot ID in journal`. `--boot 0` works fine. Also `--boot -1` works fine if I run this as normal user. This will need reporting and naughty.

marusak:
systemd (249.7-1 -> 250.2-3) was updated, that seems like the real cause.
```
sudo SYSTEMD_LOG_LEVEL=debug journalctl -b -1
Considering root directory '/run/log/journal'.
Root directory /run/log/journal added.
Considering root directory '/var/log/journal'.
Root directory /var/log/journal added.
Considering directory '/var/log/journal/a00284278e9842778876cde8e7368230'.
Directory /var/log/journal/a00284278e9842778876cde8e7368230 added.
Journal effective settings seal=no keyed_hash=yes compress=no compress_threshold_bytes=8B
File /var/log/journal/a00284278e9842778876cde8e7368230/user-1002.journal added.
File /var/log/journal/a00284278e9842778876cde8e7368230/system.journal added.
File /var/log/journal/a00284278e9842778876cde8e7368230/system@0005d58a4e00e157-df6634627734a947.journal~ added.
File /var/log/journal/a00284278e9842778876cde8e7368230/user-1002@550c7000686641298dd1fbf5a5178e55-0000000000000df1-00079fb330b311f6.journal added.
File /var/log/journal/a00284278e9842778876cde8e7368230/system@d5f9d98926094c1db3c9bd1da84de691-0000000000000001-0005d58a4dffa585.journal added.
File /var/log/journal/a00284278e9842778876cde8e7368230/system@1becac4763ef4dc992b0d57ab3be1ec3-00000000000009a4-0005d6183c40dff6.journal added.
Data from the specified boot (-1) is not available: No such boot ID in journal
Directory /var/log/journal/a00284278e9842778876cde8e7368230 removed.
Root directory /run/log/journal removed.
Root directory /var/log/journal removed.
mmap cache statistics: 216 context cache hit, 36 window list hit, 6 miss
```

```$ journalctl --list-boots
Hint: You are currently not seeing messages from other users and the system.
      Users in groups 'adm', 'systemd-journal' can see all messages.
      Pass -q to turn off this notice.
-1 1a107b4c9935429d993f17a1458cc475 Fri 2038-01-01 00:00:01 UTC—Fri 2038-01-01 00:00:04 UTC
 0 608be9aa76604ee088b4cbf0afbffc1d Fri 2038-01-01 00:05:02 UTC—Fri 2038-01-01 00:05:55 UTC

```

This suggests it might be just a corrupted journal file. Or possibly related to that message about user?

martinpitt:
Just some hints: Is there actually a previous boot, i.e. does this make sense? Perhaps the persistent journal was accidentally disabled, and it only logs to /run/log/journal?

marusak:
The issue is with changing of time. If I don't change the time it all works fine. With the time change (to 2038) it cannot find old boots with `--list-boots` or with `--boot -1`. But if I change the time to something more reasonable like 2025 it works again. So my theory is that new systemd does not look for boots that are older than 10 years or something like that.

I thought this might be 2038 related issue, but 14. January 2037 work but 15. January 2037 does not. (both at midnight) This seems rather very specific and very odd number from which it stops to work. All dates before 14th work fine, all after 15 do not. This date is in now way related to the image build time nor to image run time. 

One idea is that maybe it is 15 years since the package was built, but I don;t know where to find info when this systemd was built for debian-testing (I found just [this](https://packages.debian.org/testing/admin/systemd))

In any case, I am not gonna dig into this any more, I just change the date to 2037 instead of 2038 and call it "fine". Not like someone is gonna hit this in next 15 years.

marusak:
Sorry for not satisfactory explanation why this happens - I have no idea. I sent https://github.com/cockpit-project/cockpit/pull/16869 to just work around tests

martinpitt:
Indeed that's an odd number. 2038-01-19 is the timeocalypse, but not a year before that.

marusak (approved):
Should be green now. I've rebeased this PR to cleanup tests.

martinpitt (approved):
Dakujem!

PR #2832, opened 2022-01-20
testlib: Stop testing rhel-8.5 by default

marusak:
RHEL 8.5.1 has been released and 8.6 is closing in. I kept this context
in rhel-8-5 specific branches and for candlepin/subscription-manager as
they have different workflows. (There I also cleaned up _manual as all
of those context are defined in different branches)

Also introduced 8-6 variants for projects that had 8-5 only in _manual.


martinpitt (approved):
Thanks! However, I'd prefer to block this on introducing rhel-8-6-distropkg for cockpit, to make sure we don't break that.

martinpitt:
I had expected this to conflict, as the changes are are just one line apart, but seems git can figure it out?

PR #2833, opened 2022-01-20
fedora.install: be less picky about make-srpms output

allisonkarlitskaya:
fedora.install calls make-srpm and takes its output and pastes it
directly (as a single string) into a shell script fragment which it then
evaluates with `sh -c`.  If the output happened to be newline-separated,
then each new line will run as a separate command.

Capture the output as what it's meant to be: as a list of filenames.
Store that in an array, and expand the array inside the shell fragment.


martinpitt (approved):
Thanks! Please trigger a fedora-35 cockpit test run and wait for it to complete.

allisonkarlitskaya:
I also tested this via a `@bots#2833` trigger for `centos-8-stream` on cockpit-project/cockpit#16853

PR #2847, opened 2022-01-25
test-failure-policy: Never skip test with failed pixel test as known …

marusak:
…issue

This leads to situations when we ignore pixel test failures in tests that
fail with known issues.

Example: https://logs.cockpit-project.org/logs/pull-893-20220125-124748-273163c6-fedora-35/log.html
(notice how it has 'changed pixels' but the test is still green)

One issue that could happen is when known issue would actively break pixel test. In such case we would have to add some check into the test itself to fail before the pixel test.

### DIFF
--- a/test-failure-policy
+++ b/test-failure-policy
@@ -128,6 +128,10 @@ def check_known_issue(api, trace, image):
         return None
 
     trace = normalize_traceback(trace)
+
+    if "Differences in pixel test" in trace:
+        return None
+
     number = None
     for naughty in os.listdir(image_naughty):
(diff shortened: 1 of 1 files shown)


marusak:
Yes, we did a change but since the test had naughty we never updated the screenshot. In this case it was no big deal, but I can imagine that we would break something worse and would release it despite our tests telling us we did something wrong.

I am also not super happy about it but I feel it is tiny bit better than skipping it as I saw this already couple of times.

martinpitt (approved):
Hmmkay. Well, we can always revert it at the bots level if it gets too much in the way.

PR #2853, opened 2022-01-26
Revert "naughty: Close 2703: "tuned-adm recommend" reports non-existing "atomic-guest""

martinpitt:
This still happens in CentOS 9 Stream, like in http://artifacts.dev.testing-farm.io/c64adb27-6c06-41df-b790-b06c2e497c62/.


marusak (approved):
Danke!

PR #2854, opened 2022-01-26
push-rewrite: stop trying to set no-test label

allisonkarlitskaya:
push-rewrite checks to make sure that tests won't be automatically
    triggered before pushing the changes to the remote.  It does this by
    checking the title for "[no-test]" or seeing if the "no-test" label is
    set, and setting it if not.
    
    Unfortunately, the high level of access required to set labels on a PR
    (repo, or public_repo) is incompatible with our recent goals of scaling
    back the scopes we grant to our GitHub tokens.  Worse still: instead of
    a proper error message, we get a cryptic exception.
    
    Instead of attempting to change the label, let's just give a proper
    error message asking the user to set it for themselves.

### DIFF
--- a/push-rewrite
+++ b/push-rewrite
@@ -45,15 +45,10 @@ def find_pr_with_sha(api, sha):
 def main():
     parser = argparse.ArgumentParser(description='Force push after a rewrite')
     parser.add_argument('--repo', help="The GitHub repository to work with", default=None)
-    parser.add_argument('remote', help='The remote to push to')
-    parser.add_argument('branch', nargs='?', help='The branch to push')
     opts = parser.parse_args()
 
-    if not opts.branch:
-        opts.branch = git("rev-parse", "--abbrev-ref", "HEAD")
-
-    local = git("rev-parse", opts.branch)
-    remote = git("rev-parse", opts.remote + "/" + opts.branch)
+    local = git('rev-parse', 'HEAD')
+    remote = git('rev-parse', 'HEAD@{push}')
 
     if local == remote:
         sys.exit('Nothing to push')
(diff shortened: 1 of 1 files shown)


martinpitt:
push-rewrite:20: 'task.label' imported but unused

All hail the [pre-push hook](https://github.com/cockpit-project/bots/blob/main/HACKING.md#contributing-to-bots)! :wink:

push-rewrite
```diff
@@ -67,9 +67,10 @@ def main():
     pull = find_pr_with_sha(api, remote)
     labels = labels_of_pull(pull)
     tests_disabled = "no-test" in labels or "[no-test]" in pull["title"]
-    # add no-test label to prevent tests bring triggered
+    # needs no-test label to prevent tests bring triggered
+    # we cannot set the label for ourselves without `repo` permission
     if not tests_disabled:
-        label(pull, {"labels": labels + ["no-test"]})
+        sys.exit("Please set the 'no-test' label on the PR before trying this")
```
martinpitt, line 73: It could also just try label(), and intercept the IOError (or whatever it is), and fall back to the "please set"?

Just a suggestion, not a veto, I'm fine either way.

martinpitt (changes requested):
Approach looks good, just needs the tests fixed.

push-rewrite
allisonkarlitskaya: I did that first, but it was messy and I wasn't sure what to write for the error message.  I started like "You need to add the `public_repo` scope on your ..." and realized that we are actively discouraging people from doing that these days.  That's when I changed tack.

push-rewrite
```diff
@@ -71,7 +66,7 @@ def main():
     if not tests_disabled:
         label(pull, {"labels": labels + ["no-test"]})
 
-    git("push", "--force-with-lease=" + opts.branch + ":" + remote, opts.remote, opts.branch + ":" + opts.branch)
+    git("push", "--force-with-lease")
```
martinpitt, line 70: There's a small opportunity that this will push to origin, if people configured "origin" to be writable and the actual cockpit-project/cockpit. That's unexpected (but not technically wrong). All in all I think that's fine.

martinpitt (approved):
Ack, thanks!

PR #2858, opened 2022-01-27
Various code cleanups

martinpitt:
I tested them locally a little, but I will trigger a bunch of representative project tests against it.  In particular, cockpit's check-bots-api covers image-customize rather thoroughly.

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -116,8 +116,6 @@ def install_packages(machine_instance, packagelist):
 
 
 def main():
-    os.environ["PATH"] = "{0}:{1}".format(os.environ.get("PATH"), BOTS_DIR)
-
     parser = argparse.ArgumentParser(
         description='Run command inside or install packages into a Cockpit virtual machine',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
@@ -147,7 +145,7 @@ def main():
     args.base_image = testvm.get_test_image(args.base_image)
 
     if '/' not in args.base_image:
-        subprocess.check_call(["image-download", args.base_image])
+        subprocess.check_call([os.path.join(BOTS_DIR, "image-download"), args.base_image])
     network = testvm.VirtNetwork(0, image=args.base_image)
     machine = testvm.VirtMachine(maintain=True,
                                  verbose=args.verbose,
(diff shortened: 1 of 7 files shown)


image-customize
```diff
@@ -144,6 +116,39 @@ def install_packages(machine_instance, packagelist, install_command):
         machine_instance.execute(install_command + " " + ' '.join(allpackages), timeout=1800)
 
 
+#
```
allisonkarlitskaya, line 118: I rather enjoy defining a `main()` function and adding at the very end:

```
if __name__ == '__main__':
    main()
```

martinpitt: Sure! I'll let the first wave of tests run, and then change this.

image-customize
```diff
@@ -143,30 +143,33 @@ parser.add_argument('-n', '--no-network', action='store_true', help='Do not conn
 parser.add_argument('image', help='The image to use (destination name when using --base-image)')
 args = parser.parse_args()
 
+if not (args.run_command or args.install or args.script or args.upload or args.resize):
```
allisonkarlitskaya, line 141: For what it's worth, I've never liked how this works.  I'd sort of prefer if we built a single array and executed the operations in the order they were given, like how `vm-run` does with `-s` and `-e`.  In the case of `vm-run` the implementation isn't beautiful, but it's conceptually better.

allisonkarlitskaya (dismissed):
Everything here looks good, and are nice cleanups (thanks!) but there are a couple of suggestions for even more changes.

image-customize
martinpitt: Right, I can't say I like it much either. There are good reasons to run a command before and after installing packages, and it's totally not obvious why --run-command runs before, but --script runs after. (It usually makes sense in practice, but being explicit is better). argparse does not make this easy, but I'll add it to my todo list to experiment with that.

image-customize
martinpitt: Done. This spotted another little issue.

allisonkarlitskaya (approved):
approved, subject to you promising to unify the list :)

allisonkarlitskaya:
> approved, subject to you promising to unify the list :)

→ #2860

PR #2859, opened 2022-01-27
naughty: copy fedora-coreos iptables naughty to fedora-35

allisonkarlitskaya:
This can fail in any container scenario, and cockpit runs the
container-bastion on fedora-35.

See https://logs.cockpit-project.org/logs/pull-16871-20220127-104105-88bebe63-fedora-35-container-bastion/log.html

### DIFF
--- a/naughty/fedora-35/2787-selinux-ioctl-iptables-containers
+++ b/naughty/fedora-35/2787-selinux-ioctl-iptables-containers
@@ -1,0 +1,1 @@
+avc:  denied  { ioctl } for * comm="iptables" path="/var/lib/containers/storage/*" dev="overlay" * scontext=unconfined_u:system_r:iptables_t:* tcontext=system_u:object_r:container_file_t:s0:*


naughty/fedora-35/2787-selinux-ioctl-iptables-containers
```diff
@@ -0,0 +1 @@
+grep -E 'avc:  denied  { ioctl } for * comm="iptables" path="/var/lib/containers/storage/*" dev="overlay" * scontext=unconfined_u:system_r:iptables_t:* tcontext=system_u:object_r:container_file_t:s0:*'
```
martinpitt, line 1: so.. what's that grep here? that certainly doesn't belong here, and it wouldn't match either

martinpitt (changes requested):

martinpitt (approved):
:rofl:

allisonkarlitskaya:
dis.as.ter :rofl:

PR #2860, opened 2022-01-27
image-customize: Run action arguments in order

martinpitt:
Get rid of the hardcoded, and somewhat arbitrary order of executing the
actions; e.g. it's not obvious that `--run-command` runs before
`--install`, but `--script` runs afterwards.

Instead, run them as they are given on the command line. Our various
projects already give the arguments in a sane order.

Also rename the handlers to `do_` plus the option name, so that we can
replace the `if` ladder with a straight function lookup.

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -18,7 +18,6 @@
 
 import argparse
 import os
-import re
 import sys
 import subprocess
 
@@ -52,23 +51,53 @@ def prepare_install_image(base_image, install_image, resize):
     return install_image
 
 
-def run_command(machine_instance, commandlist):
-    """Run command inside image"""
-    for foo in commandlist:
+class ActionBase(argparse.Action):
+    '''Keep an ordered list of actions'''
+
+    def __call__(self, parser, namespace, value, option_string=None):
+        getattr(namespace, self.dest).append((self.execute, value))
+
+
+class InstallAction(ActionBase):
+    '''Install local rpm or distro package'''
+    @staticmethod
+    def execute(machine_instance, package):
+        # this will fail if neither is available -- exception is clear enough, this is a developer tool
+        out = machine_instance.execute("which dnf || which yum || which apt-get")
+        if 'dnf' in out:
+            install_command = "dnf install -y"
+        elif 'yum' in out:
+            install_command = "yum --setopt=skip_missing_names_on_install=False -y install"
+        else:
+            install_command = "apt-get install -y"
+
+        if os.path.isfile(package):
+            pname = os.path.basename(package)
+            dest = "/var/tmp/" + pname
+            machine_instance.upload([package], dest)
+            package = dest
+        elif '/' in package:
+            sys.stderr.write("Bad package name or path: %s\n" % package)
+
+        machine_instance.execute(f"{install_command} {package}", timeout=1800)
+
+
+class RunCommandAction(ActionBase):
+    @staticmethod
+    def execute(machine_instance, command):
         try:
-            machine_instance.execute(foo, timeout=1800)
+            machine_instance.execute(command, timeout=1800)
         except subprocess.CalledProcessError as e:
             sys.stderr.write("%s\n" % e)
             sys.exit(e.returncode)
 
 
-def run_script(machine_instance, scriptlist):
-    """Run script inside image"""
-    for foo in scriptlist:
-        if os.path.isfile(foo):
-            pname = os.path.basename(foo)
-            uploadpath = "/var/tmp/" + pname
-            machine_instance.upload([os.path.abspath(foo)], uploadpath)
+class ScriptAction(ActionBase):
+    @staticmethod
+    def execute(machine_instance, script):
+        if os.path.isfile(script):
+            uploadpath = "/var/tmp/" + os.path.basename(script)
+            machine_instance.upload([os.path.abspath(script)], uploadpath)
             machine_instance.execute("chmod a+x %s" % uploadpath)
             try:
                 machine_instance.execute(uploadpath, timeout=1800)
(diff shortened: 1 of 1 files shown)


martinpitt:
I tested this with:

```
$ bots/image-customize -v -r 'echo first; ! ls -l /tmp/README.md' -u README.md:/tmp/ -r 'echo second; ls -l /tmp/README.md' -i coreutils fedora-35

02a480ee-0001-4288-97b6-8fbd8b518744
+ echo first; ! ls -l /tmp/README.md
first
ls: cannot access '/tmp/README.md': No such file or directory
Uploading /var/home/martin/upstream/cockpit-podman/README.md
scp -B -r -p -P 2201 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o IdentitiesOnly=yes -o BatchMode=yes -o PKCS11Provider=none -o ControlPath=/tmp/.cockpit-test-resources/ssh-%h-%p-%r-25253 /var/home/martin/upstream/cockpit-podman/README.md root@[127.0.0.2]:/tmp/
README.md                                                                                        100% 1976     5.4MB/s   00:00    
+ echo second; ls -l /tmp/README.md
second
-rw-r--r--. 1 root root 1976 Oct 11 04:47 /tmp/README.md
+ which dnf || which yum || which apt-get
/usr/bin/dnf
+ dnf install -y coreutils
[...] 
Package coreutils-8.32-31.fc35.x86_64 is already installed.
Nothing to do.
killing ssh master process 25290
```

I'll also work on a more specific test case in cockpit.

image-customize
```diff
@@ -154,14 +153,19 @@ def main():
...
     try:
-        if args.upload:
-            upload_files(machine, args.upload)
-        if args.run_command:
-            run_command(machine, args.run_command)
-        if args.install:
-            install_packages(machine, args.install)
-        if args.script:
-            run_script(machine, args.script)
+        for (action, arg) in args.actions:
```
allisonkarlitskaya, line 156: Let's get rid of the if/elif ladder and have different subtypes of `ImageAction` instead, each of which has its own `invoke()` method that does the thing in question.

allisonkarlitskaya (changes requested):

image-customize
martinpitt: subtypes don't work, as there is just one Action instance which mops up all the arguments. However, I did this differently now, let's see how you like it.

martinpitt: Actually no, I guess it *might* instantiate the Action class multiple times.. But anyway, this is not obvious to wire up. Let's keep it simple.

allisonkarlitskaya:
here's a working example

```
#!/usr/bin/python3

import argparse

thelist = []

class ImageAction(argparse.Action):
    def __call__(self, parser, namespace, values, option_string=None):
        thelist.append(lambda: self.invoke(values))

class CopyAction(ImageAction):
    def invoke(self, values):
        print('copy', values)

class InstallAction(ImageAction):
    def invoke(self, values):
        print('install', values)

parser = argparse.ArgumentParser()
parser.add_argument('-c', action=CopyAction)
parser.add_argument('-i', action=InstallAction)
parser.parse_args()

for f in thelist:
    f()
```

image-customize
```diff
@@ -101,42 +95,53 @@ def install_packages(machine_instance, packagelist):
...
+
+
+class ImageAction(argparse.Action):
+    '''Keep an ordered list of actions'''
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        if 'actions' not in namespace:
+            setattr(namespace, 'actions', [])
+        previous = namespace.actions
+        handler = globals()['do_' + self.dest]
```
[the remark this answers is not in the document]
martinpitt: done

allisonkarlitskaya (changes requested):
One idea (optional) might be to skip messing around with the namespace object and to keep the list of actions as a class property on the base action class, and provide a static method there to run the list.

That might be getting to the point of being a bit over-engineered, though.

image-customize
```diff
@@ -52,91 +51,101 @@ def prepare_install_image(base_image, install_image, resize):
...
+        previous = namespace.actions
+        previous.append((self.execute, value))
+        setattr(namespace, 'actions', previous)
+
+
+class InstallAction(ActionBase):
+    """Install package into a test image
+    It could be done via local rpms or normal package installation
+    """
+    @classmethod
```
allisonkarlitskaya, line 69: Why `@classmethod`?  I could see an argument for making this a `@staticmethod`, but class method really makes no sense at all when every class has exactly one instance...

martinpitt:
> skip messing around with the namespace object

well, you have to set *something* to the namespace object to get your data, as what you get from parse_args() is exactly, and only, that namespace object. It could be set to the 0 to 4 action class instances, but this makes stuff unnecessily ugly IMHO.

image-customize
allisonkarlitskaya: Sure, but why not `@staticmethod` in that case, and drop the `cls` parameter?

martinpitt:
Ah, I now see what you mean -- but no, I really don't like that. It's messing around with the APIs and making assumptions about using a single argument parser which we don't need.

image-customize
```diff
@@ -52,91 +51,101 @@ def prepare_install_image(base_image, install_image, resize):
...
 
-def run_command(machine_instance, commandlist):
-    """Run command inside image"""
-    for foo in commandlist:
+class ActionBase(argparse.Action):
+    '''Keep an ordered list of actions'''
+
+    def __call__(self, parser, namespace, value, option_string=None):
+        if 'actions' not in namespace:
+            setattr(namespace, 'actions', [])
```
allisonkarlitskaya, line 59: I feel like this could be avoided by using the `default=[]` and `dest='actions'` parameters to `.add_argument()`.  Then `self.dest` here will be set to `action` and you can just say `getattr(namespace, self.dest).append(...)`

It would also be much neater that this class doesn't hardcode its destination name, and would let you remove the check for making sure something exists at the destination when you go to run the actions.

martinpitt: Right, it avoids the [gs]etattr, but hardcoding the dest name was kind of the point.. I could drop it as you say, but then at least dest= has to be repeated. default= doesn't have to be, it just looks a little asymmetric. But at this point I don't care much either way, changed accordingly.

image-customize
```diff
@@ -52,91 +51,91 @@ def prepare_install_image(base_image, install_image, resize):
...
 
 
-def run_command(machine_instance, commandlist):
-    """Run command inside image"""
-    for foo in commandlist:
+class ActionBase(argparse.Action):
+    '''Keep an ordered list of actions'''
+
+    def __call__(self, parser, namespace, value, option_string=None):
+        namespace.actions.append((self.execute, value))
```
allisonkarlitskaya, line 58: Still a bit gross that `actions` is hardcoded here.

image-customize
```diff
@@ -52,91 +51,91 @@ def prepare_install_image(base_image, install_image, resize):
...
+
+
+class InstallAction(ActionBase):
+    '''Install local rpm or distro package'''
+    @classmethod
```
allisonkarlitskaya, line 63: And that's still an open question.

allisonkarlitskaya (commented):
Just two little thorns left.  Otherwise this looks really good now, and pretty simple in the end.

image-customize
martinpitt: Oops, thanks, forgot about that. Fixed.

image-customize
martinpitt: Ah, I didn't know about `@staticmethod`, that's fine of course. We don't need to access/store class attributes here. Changed.

allisonkarlitskaya (approved):
If the bots are happy, so am I!

Thanks for all of the iteration.  In the end, looks pretty good.  Too bad the Action interface is so weird...

PR #2863, opened 2022-01-27
naughty: Add pattern for podman segv on restore

marusak:
As seen in http://artifacts.dev.testing-farm.io/abe0bdc3-7096-4ea9-81d2-607d956f5e47/ for example

There is not much to go with than just this test failing on this one line :man_shrugging:

### DIFF
--- a/naughty/fedora-35/2862-podman-restore-segv
+++ b/naughty/fedora-35/2862-podman-restore-segv
@@ -1,0 +1,2 @@
+testCheckpointRestoreCGroupsV2
+*b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in 'Running')


martinpitt (approved):
Thanks!

martinpitt:
@marusak Err wait, did you mean to add that to fedora-3**6**? It only affects rawhide so far, doesn't it? [bodhi](https://bodhi.fedoraproject.org/updates/?packages=podman)

martinpitt:
I suppose you added it to fedora-35 for the fedora-35/rawhide scenario?

Seems that [still fails](https://logs.cockpit-project.org/logs/pull-2866-20220128-060828-4426ef06-fedora-35-rawhide-cockpit-project-cockpit-podman/log.html#3-2), but for cgroupsv1. I'll adjusts it in #2866

PR #2875, opened 2022-01-31
image-customize: Add --build command to build/install distro packages

martinpitt:
This option takes a dist tarball argument. Depending on the OS, it finds
the the packaging metadata, builds packages in mock/pbuilder/arch, and
installs all resulting packages into the VM.

This centralizes the repeated code in starter-kit and derivatives, and
also cockpit itself to a large degree (it cannot build/install multiple
srpms yet, which we need for testing cockpit on RHEL 8).

----

This is based on cockpit's *.install files and c-podman's test/vm.install script, with a few generalizations and cleanups. Tested in https://github.com/cockpit-project/cockpit-podman/pull/901 and https://github.com/cockpit-project/cockpit-machines/pull/545

 - [x] Depends on PR #2876

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -18,11 +18,16 @@
 
 import argparse
 import os
+import re
 import sys
 import subprocess
 
 from lib.constants import BOTS_DIR, TEST_DIR
 from machine import testvm
+
+
+opt_quick = False
+opt_verbose = False
 
 
 def prepare_install_image(base_image, install_image, resize):
@@ -82,6 +87,103 @@ class InstallAction(ActionBase):
         machine_instance.execute(f"{install_command} {package}", timeout=1800)
 
 
+class BuildAction(ActionBase):
+    '''Build and install distribution package(s) from dist tarball'''
+
+    @staticmethod
+    def execute(machine_instance, source):
+        # upload the tarball or srpm
+        sourcename = os.path.basename(source)
+        vm_source = os.path.join("/var/tmp", sourcename)
+        # looks like cockpit-foo-123.10.gdbde381.tar.xz
+        m = re.fullmatch(r'.+-([^-]+)\.tar(\..*)?', sourcename)
+        if not m:
+            raise ValueError(f"cannot parse version number from {sourcename}")
+        version = m.group(1)
+        machine_instance.upload([source], vm_source, relative_dir=".")
+
+        machine_instance.execute("rm -rf /var/tmp/build; mkdir -p /var/tmp/build")
+
+        # this will fail if neither is available -- exception is clear enough, this is a developer tool
+        out = machine_instance.execute("(which pbuilder || which mock || which pacman) 2>/dev/null")
+        if 'pbuilder' in out:
+            BuildAction.build_deb(machine_instance, vm_source, version)
+        elif 'mock' in out:
+            BuildAction.build_rpm(machine_instance, vm_source, version)
+        elif 'pacman' in out:
+            BuildAction.build_arch(machine_instance, vm_source, version)
+        else:
+            raise NotImplementedError(f"unknown build platform: {out}")
+
+    def build_deb(machine, vm_source, version):
+        build_opts = 'nocheck' if opt_quick else ''
+
+        # build source packge
+        machine.execute(f"""
+            set -eu
+            tar -C /var/tmp/build -xf '{vm_source}'
+            cd "$(ls -d /var/tmp/build/*)"
+            # find and copy debian packaging directory
+            cp -r "$(dirname $(find -path '*/debian/control'))" .
+            # create orig.tar link for building dsc
+            source=$(awk '/^Source: / {{ print $2 }}' debian/control)
+            ln -s '{vm_source}' ../${{source}}_{version}.orig.tar.xz
+            # fill in the version number
+            sed -i '1 s/([^)-]*/({version}/' debian/changelog
+
+            DEB_BUILD_OPTIONS='{build_opts}' dpkg-buildpackage -S -us -uc -nc""")
+
+        # build binary packages
+        machine.execute("cd /var/tmp/build; pbuilder build --buildresult . *.dsc", timeout=1800)
+
+        # install packages
+        machine.execute("dpkg -i /var/tmp/build/*.deb")
+
+    def build_rpm(machine, vm_source, version):
+        mock_opts = ''
+        if opt_verbose:
+            mock_opts += ' --verbose'
+        if opt_quick:
+            mock_opts += ' --nocheck'
+
+        # build source package
+        machine.execute(f'rpmbuild --define "_srcrpmdir /var/tmp/build" -ts "{vm_source}"')
+
+        # build binary RPMs from srpm; disable all repositorys as mock insists on
+        # calling `dnf builddep`, which insists on a cache; our test VMs don't have a cache,
+        # as the mock is offline and pre-installed
+        machine.execute('chown -R builder:builder /var/tmp/build')
+        machine.execute("su builder -c 'cd /var/tmp/build; mock --no-clean --no-cleanup-after --disablerepo=* "
+                        f"--offline --resultdir . {mock_opts} --rebuild *.src.rpm'",
+                        timeout=1800)
+
+        # install RPMs
+        machine.execute('packages=$(find /var/tmp/build -name "*.rpm" -not -name "*.src.rpm"); '
+                        f'rpm -U --force --verbose {"--nodigest --nosignature" if opt_quick else ""} $packages')
+
+    def build_arch(machine, vm_source, version):
+        # unpack source tree's arch packaging directory (PKGBUILD refers to some files)
+        # and set PKGBUILD variables
+        machine.execute(f"""
+            set -eu
+            tar -C /var/tmp/build -xf '{vm_source}'
+            cd /var/tmp/build/
+            unpackdir="$(ls)"
+            archdir=$(dirname $(find -path '*/arch/PKGBUILD'))
+            cp "$archdir"/* .
+            sed -i 's/VERSION/{version}/; s/SOURCE/{os.path.basename(vm_source)}/' PKGBUILD
+            # tarball must be in same directory as PKGBUILD
+            cp '{vm_source}' /var/tmp/build/
+            """)
+
+        # build binaries
+        machine.execute("chown -R builder:builder /var/tmp/build")
+        machine.execute("cd /var/tmp/build; su builder -c extra-x86_64-build", timeout=1800)
+
+        # install packages
+        machine.execute("pacman -U --noconfirm /var/tmp/build/*.pkg.tar.zst")
+
+
 class RunCommandAction(ActionBase):
     @staticmethod
     def execute(machine_instance, command):
@@ -122,6 +224,8 @@ def main():
     # actions (at least one must be given, executed in order)
     parser.add_argument('-i', '--install', action=InstallAction, metavar="PACKAGE", dest='actions', default=[],
                         help='Install package')
+    parser.add_argument('-b', '--build', action=BuildAction, metavar="DIST-TAR", dest='actions', default=[],
+                        help='Build and install distribution package(s) from dist tarball')
     parser.add_argument('-r', '--run-command', action=RunCommandAction, dest='actions',
                         help='Run command inside virtual machine')
     parser.add_argument('-s', '--script', action=ScriptAction, dest='actions',
@@ -135,6 +239,8 @@ def main():
     parser.add_argument('-n', '--no-network', action='store_true', help='Do not connect the machine to the Internet')
     parser.add_argument('-v', '--verbose', action='store_true',
                         help='Display verbose progress details')
+    parser.add_argument('-q', '--quick', action='store_true',
+                        help='Disable tests during package build with --build')
     parser.add_argument('image', help='The image to use (destination name when using --base-image)')
     args = parser.parse_args()
 
@@ -145,6 +251,10 @@ def main():
         args.base_image = os.path.basename(args.image)
 
     args.base_image = testvm.get_test_image(args.base_image)
+
+    global opt_quick, opt_verbose
+    opt_quick = args.quick
+    opt_verbose = args.verbose
 
     if '/' not in args.base_image:
         subprocess.check_call([os.path.join(BOTS_DIR, "image-download"), args.base_image])


martinpitt:
@allisonkarlitskaya : I made this a little more tasteful now, I think it's ready for a first round of review. This should be good enough to convert starter-kit, c-{machines,podman,certificates}. Possibly composer, but cockpit will need some more work.

martinpitt:
This worked with c-podman in https://github.com/cockpit-project/cockpit-podman/pull/901 , but does not work for building cockpit itself on arch -- just unpacking PKGBUILD from the tarball isn't enough, it also needs some other files:
```
==> Making package: cockpit 261.28.g748cfe360-1 (Tue Feb  1 05:01:29 2022)
==> Retrieving sources...
  -> Found cockpit-261.28.g748cfe360.tar.xz
==> ERROR: cockpit.pam was not found in the build directory and is not a URL.
```

So I [reworked](https://github.com/cockpit-project/bots/compare/63c202e53e3376e264f5be02f51f8123fdcb81fa..5f6c9f4ca3b294f20f968ebe466c4e1b0bc83c69) the arch building a bit, and now it works for cockpit and cockpit-podman.

martinpitt:
I sent https://github.com/cockpit-project/cockpit-machines/pull/545 to test this for c-machines as well. It looks very similar, of course.

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+        if 'pbuilder' in out:
+            BuildAction.build_deb(machine_instance, vm_tar, version)
+        elif 'mock' in out:
+            BuildAction.build_rpm(machine_instance, vm_tar, version)
+        elif 'pacman' in out:
+            BuildAction.build_arch(machine_instance, vm_tar, version)
+        else:
+            raise NotImplementedError(f"unknown build platform: {out}")
+
+        # machine_instance.execute(f"{install_command} {package}", timeout=1800)
```
allisonkarlitskaya, line 117: Oops?

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+
+    def build_deb(machine, vm_tar, version):
+        build_opts = 'nocheck' if BuildAction.quick else ''
```
allisonkarlitskaya, line 120: ya... This `BuildAction.quick` stuff looks pretty odd.

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+
+        # build source packge
+        machine.execute(f"""
```
allisonkarlitskaya, line 123: I'm kinda confused about the `command` vs `script` args in `SSHConnection.execute()`.  If this is a command, then what the heck is a script?  Maybe time to clean that up...

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+        tarname = os.path.basename(tarball)
+        vm_tar = os.path.join("/var/tmp", tarname)
+        # looks like cockpit-foo-123.10.gdbde381.tar.xz
+        m = re.search(r'([^-]+)\.tar(\..*)?$', tarname)
+        if not m:
+            raise ValueError(f"cannot parse version number from tarball {tarname}")
+        version = m.group(1)
+        machine_instance.upload([tarball], vm_tar, relative_dir=".")
+
+        machine_instance.execute("rm -rf /var/tmp/build; mkdir -p /var/tmp/build")
```
allisonkarlitskaya, line 104: Why don't you deal with that from the scripts?

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+            export DEB_BUILD_OPTIONS='{build_opts}'
+            tar -C /var/tmp/build -xf '{vm_tar}'
+            cd "$(ls -d /var/tmp/build/*)"
+            # find and copy debian packaging directory
+            cp -r "$(dirname $(find -path '*/debian/control'))" .
+            # create orig.tar link for building dsc
+            source=$(awk '/^Source: / {{ print $2 }}' debian/control)
+            ln -s '{vm_tar}' ../${{source}}_{version}.orig.tar.xz
+            # fill in the version number
+            sed -i '1 s/([^)-]*/({version}/' debian/changelog
```
allisonkarlitskaya, line 134: I wonder if this is another case of something we could do statically in the tarball itself, to prevent the need for after-the-fact patching, just as we recently did for the spec?  Definitely for another PR, though.

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+
+        # machine_instance.execute(f"{install_command} {package}", timeout=1800)
+
+    def build_deb(machine, vm_tar, version):
+        build_opts = 'nocheck' if BuildAction.quick else ''
+
+        # build source packge
+        machine.execute(f"""
+            set -eu
+            export DEB_BUILD_OPTIONS='{build_opts}'
```
allisonkarlitskaya, line 125: Move that to ...

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+            cd "$(ls -d /var/tmp/build/*)"
+            # find and copy debian packaging directory
+            cp -r "$(dirname $(find -path '*/debian/control'))" .
+            # create orig.tar link for building dsc
+            source=$(awk '/^Source: / {{ print $2 }}' debian/control)
+            ln -s '{vm_tar}' ../${{source}}_{version}.orig.tar.xz
+            # fill in the version number
+            sed -i '1 s/([^)-]*/({version}/' debian/changelog
+
+            dpkg-buildpackage -S -us -uc -nc""")
```
allisonkarlitskaya, line 136: ....here please:

```
DEB_BUILD_OPTIONS='{build_opts}' dpkg-buildpackage ...
```

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+
+        # build binary packages
+        machine.execute("cd /var/tmp/build; pbuilder build --buildresult . *.dsc", timeout=1800)
```
allisonkarlitskaya, line 139: Why isn't that part of the same script?

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+
+        # install packages
+        machine.execute("dpkg -i /var/tmp/build/*.deb")
```
allisonkarlitskaya, line 142: ...ditto.

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+        mock_opts = ''
+        if BuildAction.verbose:
+            mock_opts += ' --verbose'
+        if BuildAction.quick:
+            mock_opts += ' --nocheck'
+
+        # HACK: mock has *_builddep_command config options for all backends *except* dnf
+        # dnf builddep insists on a cache, which we don't need as the mock is offline, pre-installed,
+        # and single-use only
+        machine.execute('''sed -i '/self.builddep_command =/ s_self.command_"/bin/true"_' '''
```
allisonkarlitskaya, line 154: I found that `mock --disablerepo=updates --offline` was enough to make it stop trying to update the selinux packages (which I think was the only issue).

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+                        '/usr/lib/python*/*-packages/mockbuild/package_manager.py')
+
+        # build source package
+        machine.execute(f'rpmbuild --define "_srcrpmdir /var/tmp/build" -ts "{vm_tar}"')
```
allisonkarlitskaya, line 158: Beautiful.  But what about `-appstream.src.rpm`?

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+
+    def build_arch(machine, vm_tar, version):
+        # unpack source tree's arch packaging directory (PKGBUILD refers to some files)
+        # and set PKGBUILD variables
+        machine.execute(f"""
+            set -eu
+            tar -C /var/tmp/build -xf '{vm_tar}'
+            cd /var/tmp/build/
+            unpackdir="$(ls)"
+            archdir=$(dirname $(find "$unpackdir" -path '*/arch/PKGBUILD'))
```
allisonkarlitskaya, line 178: Why not just run `find .` here?

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+            cp "$archdir"/* .
+            sed -i 's/VERSION/{version}/; s/SOURCE/{os.path.basename(vm_tar)}/' PKGBUILD
```
allisonkarlitskaya, line 180: More opportunities for doing things properly when building the tarball...

allisonkarlitskaya (changes requested):
This looks really nice, thanks.

I'm a bit confused about the somewhat-arbitrary way in which some commands are grouped together into the main script but others are run separately.

image-customize
```diff
@@ -146,6 +259,9 @@ def main():
 
     args.base_image = testvm.get_test_image(args.base_image)
 
+    BuildAction.quick = args.quick
```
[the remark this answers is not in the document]
martinpitt: I did consider it, it just makes stuff awfully connected together. These behave more like global flags, they are not specific to any action. I'd be fine to actually *make* them global variables, would that be acceptable for you?

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+    quick = False
+    verbose = False
+
+    @staticmethod
+    def execute(machine_instance, tarball):
+        # upload the tarball
+        tarname = os.path.basename(tarball)
+        vm_tar = os.path.join("/var/tmp", tarname)
+        # looks like cockpit-foo-123.10.gdbde381.tar.xz
+        m = re.search(r'([^-]+)\.tar(\..*)?$', tarname)
```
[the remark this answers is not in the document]
martinpitt: The intention is "give me the last of the - separated components", which is the version number. Indeed that assumes that the version number does not contain `-`, but that assumption is true in the projects for which we use this.

What's the advantage of explicitly matching the `^.+-` prefix? We don't use it for anything, and we can't even assume that it starts with cockpit (starter-kit!)

image-customize
martinpitt: Mostly DRY - because that's common to all backends. If you prefer, I can move it into the backend functions, I don't have a strong opinion about it.

image-customize
martinpitt: That would make rpm vs. spec more consistent; there was some reason why I didn't do that back then, but I can't remember. I added it to my TODO list to try and put that into the dist tarball.

image-customize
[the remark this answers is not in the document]
allisonkarlitskaya: I think the thing that bothers me most is that your argument shows that they have nothing to do with building.  We might choose to make any action more or less verbose, for example.  Maybe hanging them on the base class is the right idea, then?  That seems more correct then assigning them specifically for BuildAction.   As before: I'm trying to think with the mindset that the integration of any particular single new action type should be limited to the definition of a new class, and the addition of the new argument type...

image-customize
martinpitt: I tried to group the machine.execute() blobs into "prepare source tree", "build source package", "build binary package", and "install packages", so that the `image-customize --verbose` output is a bit nicer to follow, and you get more targeted output on failures. It also makes it easier to copy&paste such a shell script into a VM for debugging this step.

The "prepare source tree" steps are rather involved, and all assume that /var/tmp/build is the cwd; these commands are also not independent from each other, so I think it's fine to keep them together as a single shell script.

image-customize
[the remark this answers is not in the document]
allisonkarlitskaya: `fullmatch()` just seems like a better fit than `search()` here.  We're not really *searching* per se.  We have a thing which has a precisely known format.  Let's describe that format.

It also specifically allows to avoid the `^` and `$` anchors, as they become implicit.

At the very least please add a `-` to the front here to show that a `-` should immediately precede the version number.

image-customize
martinpitt: I was really hoping that this bug got fixed in the recent years, but alas it wasn't. Without it:

```
 # /usr/bin/dnf builddep --installroot /var/lib/mock/fedora-35-x86_64/root/ --releasever 35 -C --setopt=deltarpm=False --allowerasing --disableplugin=local --disableplugin=spacewalk --disableplugin=versionlock /var/lib/mock/fedora-35-x86_64/root//builddir/build/SRPMS/cockpit-podman-40.13.ga3acf22-1.fc35.src.rpm --setopt=tsflags=nocontexts
No matches found for the following disable plugin patterns: local, spacewalk, versionlock
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Error: Cache-only enabled but no cache for 'fedora'
```

image-customize
allisonkarlitskaya: Nah.  Fair enough :)

image-customize
martinpitt: See commit message. I want to add `--build foo.src.rpm` (instead of a tarball) as a follow-up, to support this use case. But I'd like to experiment a bit how that looks on the cockpit side.

image-customize
martinpitt: would it help if I turned these back into `@classmethod`s and use `kls.quick`?

image-customize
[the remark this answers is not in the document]
martinpitt: Changed to fullmatch, but honestly I find that *less* readable/comprehensible. (But then again, no strong opinion).

image-customize
martinpitt: Ah, leftover from copy&pasting, removed.

image-customize
martinpitt: done

image-customize
martinpitt: done

image-customize
martinpitt: Not a big difference indeed, dropped it.

image-customize
```diff
@@ -82,6 +83,114 @@ class InstallAction(ActionBase):
...
+            tar -C /var/tmp/build -xf '{vm_tar}'
+            cd /var/tmp/build/
+            unpackdir="$(ls)"
+            archdir=$(dirname $(find "$unpackdir" -path '*/arch/PKGBUILD'))
+            cp "$archdir"/* .
+            sed -i 's/VERSION/{version}/; s/SOURCE/{os.path.basename(vm_tar)}/' PKGBUILD
+            """)
+
+        # tarball must be in same directory as PKGBUILD
+        machine.execute(f"cp '{vm_tar}' /var/tmp/build/")
```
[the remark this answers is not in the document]
martinpitt: I tried to avoid having shell comments in the multi-line scripts, but I agree (wrt. the grouping that I want to do). Moved.

image-customize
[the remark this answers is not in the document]
martinpitt: `--quick` *is* specific to building (while `--verbose` isn't). This indeed is a bit awkward -- we could avoid it with having `--build` vs `--quick-build`, but I can't say that I like that much either. So I am following the --quick option of image-prepare for now.

I can't say I like attaching them to the base class either. These are practically singletons and static. So I changed them to global variables, which is the honest answer to what they really are.

image-customize
martinpitt: Can't -- Debian doesn't have a "builder" system user, pbuilder does that internally.

martinpitt:
@allisonkarlitskaya , I responded to your review comments and fixed some. I re-triggered https://github.com/cockpit-project/cockpit-podman/pull/901for another validation run. Supposedly this needs some more discussion, though.

image-customize
allisonkarlitskaya: Can we also `--disablerepo=fedora` then?  Or just not delete the cache...

image-customize
allisonkarlitskaya: No.

image-customize
allisonkarlitskaya: No, I mean the opposite: that it should be created at the point where we assign the permissions (or not), rather than creating it and assigning the permissions later.

image-customize
allisonkarlitskaya: OK.

image-customize
allisonkarlitskaya: I have some ideas for the cockpit experiments.  Mind if I give them a go?

martinpitt: I have a fairly straightforward plan for that, but I want to finish this first.

image-customize
martinpitt: `--disablerepo=*` works, tested with cockpit-podman and cockpit. That's a bit less gross.

(FTR, that hack originated in commit 847b2043 two years ago, it's far from new)

image-customize
martinpitt: Indeed the difference in [the definition](https://github.com/cockpit-project/bots/blob/main/machine/machine_core/ssh_connection.py#L301) is marginal -- script= does not support stdin in put, but command argvs are reduced to a shell script either way. I made a todo item to clean this up.

image-customize
martinpitt: Sure, the prep and build could happen in a `su` subshell, and we could triplicate the build directory creation. Given how little this chown thing matters in practice (i.e. "not at all"), I prefer this over adding yet another nesting level to these commands. It's how we use our test VMs (root shells). If rewriting it is what it takes to land this PR, I'll do it, but I really don't like making the code more complicated..

martinpitt: Ugh, sorry -- my last sentence was unnecessarily snarky. Please accept my apology! I didn't mean that personally, just that I'd rather like to discuss that again -- for such glue stuff, simplicity and copy&paste-ability are desirable IMHO.

martinpitt:
Pushed an argument renaming from "tarball" to "source", as after this I'll add functionality to build a source rpm as well.

allisonkarlitskaya (approved):
This is definitely overdue for landing by now.  Thanks!

PR #2882, opened 2022-02-02
testmap: Update triggered test for cockpit po refresh

martinpitt:
We recently changed our GitHub branch protection to set "fedora-coreos"
as required test.

---

I saw this in https://github.com/cockpit-project/cockpit/pull/16906 this morning, where fedora-35 was auto-triggered, but fcos was required/missing.


marusak (approved):
Thank you!

PR #2894, opened 2022-02-04
testmap: Remove fedora-35/mobile scenario from c-machines

mvollmer:
The "mobile" scenario is no longer recognized by cockpit-machines and
a "fedora-35/mobile" test run is identical to a "fedora-35" run.


martinpitt (approved):
Cheers!

PR #2905, opened 2022-02-07
workflows: don't try cockpituous tests on non-origin

allisonkarlitskaya:
These tests will fail to run unless the PR was proposed from a branch on the origin, so bomb out early with a meaningful error rather than trying to run them anyway.

plus two other nice changes

 - [x] #2904 for what the failure looks like.
 - [x] #2906 for a contrived test of what a successful PR from a fork looks like.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,6 +42,10 @@ jobs:
               ':!lib/testmap.py' \
           >&2 || echo "::set-output name=changed::true"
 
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
See also https://github.com/cockpit-project/bots/pull/2906 for a test run proposed from non-origin with only irrelevant changes.

martinpitt (approved):
Very cool, thank you!

PR #2909, opened 2022-02-07
image-customize: Add --fresh option, some bug fixes

martinpitt:
By default, multiple image-customize calls are additive. This is useful,
and API by now. But our various `make vm` rules explicitly remove the
overlays, and hardcoding/duplicating overlay names in projects is ugly.
Let's give this a proper option.

---

Once this lands, I'll send starter-kit, c-podman, c-machines PRs to make use of this, and also adjust https://github.com/cockpit-project/cockpit/pull/16935 accordingly. For now I tested this manually.

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -29,7 +29,7 @@ opt_quick = False
 opt_verbose = False
 
 
-def prepare_install_image(base_image, install_image, resize):
+def prepare_install_image(base_image, install_image, resize, fresh):
     '''Create the necessary layered image for the build/install'''
 
     if "/" not in base_image:
@@ -37,12 +37,20 @@ def prepare_install_image(base_image, install_image, resize):
     if "/" not in install_image:
         install_image = os.path.join(os.path.join(TEST_DIR, "images"), os.path.basename(install_image))
 
-    # In vm-customize we don't force recreate images
+    qcow2_image = f"{install_image}.qcow2"
+
+    # Remove existing overlay if --fresh was requested
+    if fresh:
+        for f in [install_image, qcow2_image]:
+            try:
+                os.unlink(f)
+            except FileNotFoundError:
+                pass
+
     if not os.path.exists(install_image):
         install_image_dir = os.path.dirname(install_image)
         os.makedirs(install_image_dir, exist_ok=True)
         base_image = os.path.realpath(base_image)
-        qcow2_image = "{0}.qcow2".format(install_image)
         subprocess.check_call(["qemu-img", "create", "-q", "-f", "qcow2",
                                "-o", "backing_file={0},backing_fmt=qcow2".format(base_image), qcow2_image])
         if os.path.lexists(install_image):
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (commented):
Reset sounds too much like an action, rather than a flag to modify behaviour.  The current documentation doesn't help with this perception.

The way it's written ("remove an existing overlay..."), I might expect to be able to run

```
image-customize --reset
```

as an equivalent to `bots/vm-reset`.

I might also have some idea that it could be used in the middle of a bunch of other actions to somehow reset the state along the way, before making further changes.

I wonder if we could somehow change the name to something more "flaggy" like `--fresh` or `--non-cumulative` or so.  All names that I can think of right now aren't great...

allisonkarlitskaya:
I'd also consider simply abandoning the current "API" behaviour and making this the new default behaviour, possibly adding a switch to turn it off.

martinpitt:
@allisonkarlitskaya : Until a week ago, [cockpit-podman](https://github.com/cockpit-project/cockpit-podman/commit/d50f71cece52c76b08e3dfa3b3d905ad14ced3fe) used two successive image-customize calls. cockpit-session-recording [also does that](https://github.com/Scribery/cockpit-session-recording/blob/da32f4f344bd7b55eaa88ae0a7245244684b402b/Makefile#L132) (although admittedly in a super inefficient way), and I didn't check all our other projects yet.

As to running `image-customize --reset`, that quickly and clearly fails with "Must specify at least one operation", so that part is safe. It also doesn't make sense to reset an image in the middle between actions. How about `--remove-existing`?

martinpitt:
@allisonkarlitskaya : Renamed to `--fresh`, and also did another annoying bug fix.

martinpitt:
@allisonkarlitskaya : You were right yesterday, the [cockpituous test fails](https://github.com/cockpit-project/bots/runs/5104665438?check_suite_focus=true). I'm not sure when it passed last, but I first need to fix it for cockpituous in  https://github.com/cockpit-project/cockpituous/pull/451

image-customize
```diff
@@ -258,7 +268,7 @@ def main():
     machine = testvm.VirtMachine(maintain=True,
                                  verbose=args.verbose,
                                  networking=network.host(restrict=args.no_network),
-                                 image=prepare_install_image(args.base_image, args.image, args.resize))
+                                 image=prepare_install_image(args.base_image, args.image, args.resize, args.fresh))
```
allisonkarlitskaya, line 272: Another good time to mention that I enjoy just passing the entire `args` through to the various action functions... :)

[note: not a change request!]

image-customize
```diff
@@ -231,6 +239,8 @@ def main():
     # options
     parser.add_argument('--base-image',
                         help='Base image name, if "image" does not match a standard Cockpit VM image name')
+    parser.add_argument('--fresh', action='store_true',
+                        help="Remove an existing overlay; by default, multiple image-customize calls are additive")
```
allisonkarlitskaya, line 244: help="Start fresh from the base image; ..."

allisonkarlitskaya (changes requested):
Just one more round of bikeshedding about the help string please, and then this looks good.

image-customize
martinpitt: Heh, matter of style -- I actually prefer function arguments to be explicit, and not couple them to the argparse structure.

image-customize
martinpitt: Done, thanks!

allisonkarlitskaya (approved):
looks good, tyvm.

PR #2912, opened 2022-02-08
store-tests: record individual test time

jelly:
Record the individual test time in the SQLite database to visualize the
10 slowest tests on our Grafana dashboard and compare test durations
over time.

Schema migration to add the new column to record the time of an
individual test:

    ALTER TABLE Tests ADD COLUMN seconds INTEGER NOT NULL DEFAULT 0

### DIFF
--- a/store-tests
+++ b/store-tests
@@ -44,6 +44,7 @@ from task import github
 # not ok 78 testNodeNavigation (check_openshift.TestOpenshift) # duration: 172s
 re_run = re.compile(r"(not )?ok \d+ (\S+) (\S+)")
 re_retry = re.compile(r"# RETRY .* \(([^)]+)\)")
+re_duration = re.compile(r"# \d TEST (FAILED|PASSED) \[(\d+)s")
 
 DATE_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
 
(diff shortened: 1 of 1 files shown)


store-tests
```diff
@@ -44,6 +44,7 @@ from task import github
 # not ok 78 testNodeNavigation (check_openshift.TestOpenshift) # duration: 172s
 re_run = re.compile(r"(not )?ok \d+ (\S+) (\S+)")
 re_retry = re.compile(r"# RETRY .* \(([^)]+)\)")
+re_test = re.compile(r"# \d TEST (FAILED|PASSED) \[(\d+)s")
```
martinpitt, line 47: Perhaps `re_test_duration`?

martinpitt (changes requested):
Please rebase to get over the cockpituous failure

store-tests
martinpitt: Or actually just `re_duration`, everything here is about tests

store-tests
```diff
@@ -44,6 +44,7 @@ from task import github
 # not ok 78 testNodeNavigation (check_openshift.TestOpenshift) # duration: 172s
```
martinpitt, line 44: That comment is not actually true any more -- can you please update it to the current format?

martinpitt (approved):
Danke!

martinpitt (approved):

PR #2913, opened 2022-02-08
workflows: undo committer ID changes for rebase

allisonkarlitskaya:
This (partial) reverts commit a7ac56a027c8a86a900d7bd6dc0ce007983efee1.

The EMAIL='<>' tricks that work in webpack-jumpstart don't work here for
some reason.  That's presumably because we only ever call
`webpack-jumpstart --rebase` from the packit sandcastle, and there we
probably have a gecos field or something that is missing on the GitHub
runner.

Let's just put back what was there before: it was working just fine.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,7 +30,9 @@ jobs:
 
       - name: Rebase to target branch
         run: |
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Cheers! Also good to see that the cockpituous test still actually can succeed.

PR #2914, opened 2022-02-08
machine: add machine.curl()

allisonkarlitskaya:
An awful lot of our tests call `curl` on the VM and inspect the result.
Some variation of `-s -S` is common (~= `--no-progress-meter`) and
passing headers is also done from many locations.

Add a nice wrapper that executes `curl` on the VM with the correct
arguments, with support for passing headers as a dictionary.


martinpitt (approved):
:+1: from me, this is a nice one. (It's still draft, though)

PR #2915, opened 2022-02-08
machine: add machine.write_conf()

allisonkarlitskaya:
An awful lot of our tests have various takes on writing to cockpit.conf
before running a test.

Add a small wrapper to make this easier.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -482,3 +482,40 @@ class SSHConnection(object):
             The pid of the /bin/sh process that executes the command.
         """
         return int(self.execute("{ (%s) >/var/log/%s 2>&1 & }; echo $!" % (shell_cmd, log_id)))
+
+    def write_conf(self, *args, file='/etc/cockpit/cockpit.conf'):
+        '''Writes the keyfile /etc/cockpit/cockpit.conf, or another file
+
+        The fully general form takes a single dictionary and looks like:
+
+            m.write_conf({
+                'WebService': {
+                    'LoginTitle': 'Hello World'
+                }
+            })
+
+        But there are two "simplified" forms when only setting values in a
+        single section, or only setting a single value.  The following two
+        examples are equivalent to the above, and easier to read:
+
+            m.write_conf('WebService', {'LoginTitle': 'Hello World'})
+
+            m.write_conf('WebService', 'LoginTitle', 'Hello World')
+
+        Values can be any value that will expand in an f-string.  No quoting is
+        performed.
+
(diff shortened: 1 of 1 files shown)


martinpitt:
See https://github.com/cockpit-project/cockpit/pull/16951#issuecomment-1032829584

machine/machine_core/ssh_connection.py
```diff
@@ -482,3 +482,40 @@ def spawn(self, shell_cmd, log_id):
...
+
+        Values can be any value that will expand in an f-string.  No quoting is
+        performed.
+
+        The file= kwarg can be provided to write to a different file.
+        /etc/cockpit/cockpit.conf is the default.
+        '''
+        while len(args) > 1:
+            args = args[:-2] + ({args[-2]: args[-1]},)
+        config, = args
```
martinpitt, line 513: If we go ahead with this PR, this wizardry *really* needs a comment. However, honestly I feel this function is too special. write() already has an `append=True` mode. Honestly I never missed the ini-style formatting, our tests rarely add more than two lines of config anyway. The much more annoying thing for writing files is that you always have to `mkdir -p` the parent dir, teaching .write() about that would make me a lot happier :grin:

allisonkarlitskaya: I agree (generally) about the `mkdir -p` thing, although we should be careful about how we handle this side effect in nondestructive tests.

It doesn't apply to this particular case though, since /etc/cockpit always exists (but see https://github.com/cockpit-project/cockpit/issues/16937 about that).

I think my main motivation here, as with the parallel curl feature, is to address two instances of what I consider to be a pattern of somewhat serious issue in our tests: we are often encouraged to write very long and somewhat unreadable strings for doing all kinds of things in the VM.  We have an impressive array of wrappers for doing all kinds of checks in the browser (saving us from writing long JavaScript conditionals), but relatively few primitives for the VM.

The goal here, as with the curl thing, is to provide interfaces that allow these giant strings to be broken up into Python syntax which can then be properly formatted and is way easier to read.

martinpitt: Maybe, but honestly I don't find https://github.com/cockpit-project/cockpit/pull/16951/files to be convincing. It does not make much of a difference in terms of line length to use a multi-line string or a "proper" python object formatting; and with the plain string one sees *exactly* what is going on; with the abstraction this is lost.

allisonkarlitskaya:
I believe that the general direction of reducing blobs is important, but I kinda agree that this case is a more marginal one.  It was always meant as a kind of RFC, so I think we can let this one die.

PR #2918, opened 2022-02-09
machine: work around old curl in machine.curl()

allisonkarlitskaya:
--no-progress-meter isn't available from the version of curl on rhel-8-6
and centos-8-stream.  Detect those OSes and patch it out.

This patch is done in a way that it could either be straight-reverted,
or the relevant section simply deleted, when the time comes.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -358,8 +358,8 @@ class Machine(ssh_connection.SSHConnection):
             self.execute(script=WAIT_COCKPIT_RUNNING)
 
     def curl(self, *args, headers=None):
-        header_args = []
+        cmd = ['curl', '--silent', '--show-error']
         if headers is not None:
             for key, value in headers.items():
-                header_args.extend(['--header', f'{key}: {value}'])
-        return self.execute(['curl', '--no-progress-meter'] + header_args + list(args))
+                cmd.extend(['--header', f'{key}: {value}'])
+        return self.execute(cmd + list(args))


allisonkarlitskaya:
Uh, oops.  The test runs triggered here are meaningless since they're against the main branch, which doesn't have the changes yet.  AFAIK we don't have a syntax for running a new bots PR against an open cockpit repo PR, so I'll do it from the other repo targeting here, instead.

machine/machine_core/machine.py
```diff
@@ -358,8 +358,11 @@ def wait_for_cockpit_running(self, address="localhost", port=9090, seconds=30, t
             self.execute(script=WAIT_COCKPIT_RUNNING)
 
     def curl(self, *args, headers=None):
-        header_args = []
+        cmd = ['curl', '--no-progress-meter']
+        if self.image in ['rhel-8-6', 'centos-8-stream']:
+            cmd.remove('----no-progress-meter')  # not available in old curl
+            cmd.extend(['--silent', '--show-error'])
```
martinpitt, line 364: Ugh -- what's wrong with just always using `silent --show-error`?

allisonkarlitskaya: I'd agree with this except that I specifically want to make this stick out like a sore thumb so that we can remove it when we drop support for RHEL8.

martinpitt: (I'm pecking on this because I don't really like curl to behave differently when testing across different OSes.)

allisonkarlitskaya: Are you making your argument on the basis that `--silent` `--show-error` is exactly equivalent to `--no-progress-meter` or are you making it on the basis that it might not be? :)

martinpitt: It's about "I don't know and I don't want to". Either it is exactly equivalent, then we can use the options which work everywhere. Or it's not, then we'd do differnet things on different OSes. So it doesn't matter, I wouldn't want this conditional either way.

allisonkarlitskaya: Okay.  We can do it your way :)

martinpitt (approved):
Thanks!

PR #2927, opened 2022-02-10
machine_virtual: store transient images in /var/tmp

allisonkarlitskaya:
- assume backing file is qcow2
 - use NamedTemporaryFile for the overlay


machine/machine_core/machine_virtual.py
```diff
@@ -323,13 +323,8 @@ def execute(*args):
         image_to_use = self.image_file
         if not self.maintain:
             (unused, self._transient_image) = tempfile.mkstemp(suffix='.qcow2', prefix='cockpit-', dir=self.run_dir)
-            options = ""
-            # check if it is qcow2 (libvirt complains otherwise)
-            with open(self.image_file, "rb") as f:
-                if f.read(3) == b"QFI":
-                    options += ",backing_fmt=qcow2"
```
martinpitt, line 332: That's what we had in the past. This was specifically introduced in commit d2ced0c18c23c2d to also support raw overlays for testing lorax. Let's please drop this commit for now; if you care much, please check with @atodorov whether it's still needed, but I don't see much reason to stop supporting it.

machine/machine_core/machine_virtual.py
```diff
@@ -320,12 +321,14 @@ def execute(*args):
             self.message(*args)
             return subprocess.check_call(args)
 
-        image_to_use = self.image_file
         if not self.maintain:
-            (unused, self._transient_image) = tempfile.mkstemp(suffix='.qcow2', prefix='cockpit-', dir=self.run_dir)
+            self._transient_image = tempfile.NamedTemporaryFile(suffix='.qcow2', prefix='cockpit-', dir=self.run_dir)
```
martinpitt, line 325: :+1: I remember when we wrote that, NamedTemporaryFile wasn't a thing yet.

martinpitt (changes requested):
Thanks for the cleanup!

machine/machine_core/machine_virtual.py
allisonkarlitskaya: Reworked this a bit to look nicer.

machine/machine_core/machine_virtual.py
allisonkarlitskaya: :)

machine/machine_core/machine_virtual.py
```diff
@@ -316,21 +318,19 @@ def _libvirt_connection(self, hypervisor, read_only=False):
...
             with open(self.image_file, "rb") as f:
                 if f.read(3) == b"QFI":
-                    options += ",backing_fmt=qcow2"
-            execute("qemu-img", "create", "-q", "-f", "qcow2",
-                    "-o", "backing_file=%s%s" % (self.image_file, options), self._transient_image)
-            image_to_use = self._transient_image
+                    cmd.extend(['-F', 'qcow2'])
+            image_to_use = self._transient_image.name
+            cmd.append(image_to_use)
+            self.message(shlex.join(cmd))
```
martinpitt, line 330: Note to self: This is only shown with `-v`, so that's fine.

Issue #2518, opened 2021-10-14
Failing to start pmlogger.service with result 'protocol' 


marusak:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2013937

for Debian 11: https://bugs.debian.org/1005325

cockpituous:
arch
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testLoggerSettings (__main__.TestHistoryMetrics)
Failed to try-restart sssd.service: Unit sssd.service not found.
Created symlink /etc/systemd/system/multi-user.target.wants/pmlogger.service → /usr/lib/systemd/system/pmlogger.service.

DevTools listening on ws://127.0.0.1:10147/devtools/browser/5d1da999-4c1a-4493-b51b-ab4765749d45
[1110/101631.820604:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: cannot open shared object file: No such file or directory
[1110/101631.822590:ERROR:viz_main_impl.cc(162)] Exiting GPU process due to errors during initialization
[1110/101631.825503:ERROR:gpu_init.cc(441)] Passthrough is not supported, GL is disabled
> error: Modal: Specify at least one of: title, aria-label, aria-labelledby.
> warning: grep: /sys/class/dmi/id/power/autosuspend_delay_ms: Input/output error

> log: 
Traceback (most recent call last):
  File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/work/bots/make-checkout-workdir/test/verify/check-metrics", line 513, in testLoggerSettings
    b.wait_not_present("#pcp-settings-modal")
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 423, in wait_not_present
    self.wait_js_func('!ph_is_present', selector)
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 414, in wait_js_func
    self.wait_js_cond("%s(%s)" % (func, ','.join(map(jsquote, args))))
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 411, in wait_js_cond
    self.raise_cdp_exception("timeout\nwait_js_cond", cond, result["exceptionDetails"], trailer)
[...]

PR #2930, opened 2022-02-11
naughty: Add very general pattern for pmlogger.service failure on debian-stable

martinpitt:
This bug is just impossible to work around [1][2], and it can make the
TestHistoryMetrics tests fail in lots of different ways, so precise
patterns are not possible.

Blanket-attribute all PCP failures on Debian 11 on this bug. This is
still a bit better than skipping the tests entirely, as at least someone
could actually fix the bug and then we'd learn about it.

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2013937
Known issue #2518

[1] https://github.com/cockpit-project/cockpit/issues/16974
[2] https://github.com/cockpit-project/cockpit/pull/16979

----

I do realize that this is a last-resort kind of action. But I wasted hours and hours on a better solution already, I am running out of ideas. I updated and reopened https://github.com/cockpit-project/bots/issues/2518 and filed a Debian bug report.

### DIFF
--- a/naughty/debian-stable/2518-pmlogger-protocol-error
+++ b/naughty/debian-stable/2518-pmlogger-protocol-error
@@ -1,0 +1,2 @@
+*TestHistoryMetrics*
+  File "test/verify/check-metrics", line *


marusak (approved):
Fair enough, thanks!

PR #2931, opened 2022-02-11
images: Actually keep original cockpit/ws image on fedora-coreos

martinpitt:
The `podman commit` command did not actually create a new local
cockpit/ws image, but overwrites the quay.io/cockpit/ws tag.  Add an
explicit `localhost/` prefix to work around that.

`podman run cockpit-ws` happily and silently ran the `:released` tag 
implicitly. Avoid falling into this trap by renaming the original
container harder.

With the new container image, one needs to specify an explicit `CMD`. It
defaults to the command used to start the container, instead of the 
command of the original image (argh, unintuitive). There is no way to
actually tell `podman commit` to use the command as given (it always
wraps it into a `sh -c` wrapper), but that's harmless enough for us to
not care.


martinpitt:
I noticed that while working on https://github.com/cockpit-project/cockpit/pull/16935 . Of course I'll use the fixed commands there right away.

PR #2944, opened 2022-02-15
images: stop Arch from updating when building

jelly:
When building a cockpit in a change root with extra-x86_64-build the
latest packages are fetched which means a newer glibc can be used to
build cockpit which fails on an older glibc runtime.

Also start firewalld as it isn't started and therefore fails to add the
cockpit rule.

### DIFF
--- a/images/scripts/arch.install
+++ b/images/scripts/arch.install
@@ -41,6 +41,11 @@ if [ -n "$do_build" ]; then
     dist_tar="$(ls cockpit-*.tar.?z)"
     upstream_ver=$(echo "$dist_tar" | sed 's/^.*-//; s/.tar\..z//' | head -n1)
 
+    # HACK: freeze the repository so we don't pull a newer GLIBC
+    cat << EOF > /etc/pacman.d/mirrorlist
+Server = https://america.archive.pkgbuild.com/repos/2021/12/18/\$repo/os/\$arch
+EOF
+
     rm -rf cockpit-*/
     tar -xf $dist_tar
     ( cd $resultdir/


images/scripts/arch.install
```diff
@@ -59,6 +64,7 @@ if [ -n "$do_install" ]; then
     packages=$(find build-results -name "*.pkg.tar.zst")
     pacman -U --noconfirm $packages
 
+    systemctl start firewalld
```
martinpitt, line 67: Odd, how did that work before? The base image didn't change?

martinpitt:
Seems the archive mirror is rather slow:

     automake-1.16.5-1-any doerror: failed retrieving file 'gdbm-1.22-1-x86_64.pkg.tar.zst' from archive.archlinux.org : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds

[log](https://logs.cockpit-project.org/logs/pull-2944-20220215-120949-c22c546d-arch-cockpit-project-cockpit/log.html)

I'll retry once for comparison, but it seems we may be better off with just (1) fixing the image refresh, and (2) figuring out how to do the build offline inside the image in the future?

jelly:
> Seems the archive mirror is rather slow:
> 
> ```
>  automake-1.16.5-1-any doerror: failed retrieving file 'gdbm-1.22-1-x86_64.pkg.tar.zst' from archive.archlinux.org : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds

It's hosted in germany, so I guess Germany <-> America is bad? We can switch to https://america.archive.pkgbuild.com/ or europe or asia.


> [log](https://logs.cockpit-project.org/logs/pull-2944-20220215-120949-c22c546d-arch-cockpit-project-cockpit/log.html)
> 
> I'll retry once for comparison, but it seems we may be better off with just (1) fixing the image refresh, and (2) figuring out how to do the build offline inside the image in the future?

Agreed.

images/scripts/arch.install
jelly: This is indeed fishy,

images/scripts/arch.setup:systemctl enable --now firewalld

martinpitt (approved):
Thank you! Good enough hack to get us until the real image fix.

PR #2945, opened 2022-02-15
WIP: script: introduce logic for creating ISO files aside from qcow2 images

KKoukiou:
Introduce fedora rawhide ISO to the list of the existing images in
store. This is about to be used for the anconda-webui tests.

This could be one approach for introducing boot.iso files. Another approach could be to create another directory called isos/ as images/ exist. Then I need to teach the test runner classes in anaconda, to look for files in bots/isos/ and not bots/images.


image-create
```diff
@@ -161,15 +166,18 @@ class MachineBuilder:
 
         partial = os.path.join(data_dir, self.machine.image + ".partial")
 
-        # Copy image via convert, to make it sparse again
-        subprocess.check_call(["qemu-img", "convert", "-c", "-O", "qcow2", self.machine.image_file, partial])
+        if not self.iso:
```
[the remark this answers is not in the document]
martinpitt: "partial" is a really confusing name here. `*.partial` files are used in image-download during the curl'ing, so that you can interrupt and resume a download - and only once it finished, it will be renamed to the final file. This avoids broken (interrupted) files.

However, here it is merely some "temporary new" name for "repacking" the just built image into a fresh qcow.

So my vote would be for renaming `partial` to `rebuild`.

image-create
```diff
@@ -161,15 +166,18 @@ class MachineBuilder:
...
+            # Copy image via convert, to make it sparse again
+            subprocess.check_call(["qemu-img", "convert", "-c", "-O", "qcow2", self.machine.image_file, partial])
+        else:
+            subprocess.check_call(["cp", self.machine.image_file, partial])
```
martinpitt, line 173: Copying is pointless and just a waste of time. That whole "rebuild" code path does not need to happen for iso files.

image-create
```diff
@@ -56,10 +56,12 @@ if args.image in ["fedora-coreos", "services", "selenium", "openshift"]:
 class MachineBuilder:
     def __init__(self, machine):
         self.machine = machine
+        self.iso = "boot" in self.machine.image
```
martinpitt, line 59: That heuristic is a bit uninutitive, but let's start with that, we can refine it later if we need to. But perhaps let's do `.endswith("-boot")`, just in case.

image-create
```diff
@@ -126,6 +128,9 @@ class MachineBuilder:
     def build(self):
         self.bootstrap_system()
 
+        if "boot" == self.machine.image:
+            return
```
martinpitt, line 132: This will never happen. It's also not necessary, as the code checks for the existence of a `.setup` script next, and as there (rightfully) isn't one, it returns.

martinpitt (changes requested):
Thanks! This is by and large fine, let's just clean it up a litte bit.

image-create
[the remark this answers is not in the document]
KKoukiou: Thanks for explaining! That was indeed slightly confusing naming, renamed.

image-create
KKoukiou: Done.

image-create
KKoukiou: Removed.

martinpitt:
Thanks for the updates! This LGTM, but still needs some massaging (git pre-push hook *cough* *cough*):

    image-create:171:121: E501 line too long (152 > 120 characters)

I added an image-refresh here, but I suppose it's not going to work yet as the image is not yet known. However, I ran `./image-create fedora-rawhide-boot` locally, and that works fine.

martinpitt (approved):
Thanks!

PR #2949, opened 2022-02-17

cockpituous:

### DIFF
--- a/images/centos-8-stream
+++ b/images/centos-8-stream
@@ -1,1 +1,1 @@
-centos-8-stream-f02cef3a6050cfe735d27d1b9932ca5a839d84c94b4f4af03ed775591403207d.qcow2
+centos-8-stream-36af53fea2b0cef8d517a120ca3bed129a77a50f6a6cd5df108966944fab5318.qcow2


martinpitt:
[vdo troubles](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-2949-20220217-023742-a005df9a-centos-8-stream-cockpit-project-cockpit/log.html#225-2)

marusak:
So it is not broken per se - but something fishy is surely happening. Sometimes `rpm` is stuck on `rpm -e --no-deps kmod-kvdo` - which we do as cleanup after `self.createPackage("kmod-kvdo", "999", "1")`.

I wonder, why we even create the package, as we have it in the VM:
```
kmod-kvdo-6.2.6.14-83.el8.x86_64
```
and when we run the `rpm -e kmod-kvdo` we remove the version from distro and not the version we created.

So either we don't need to create this package, or create it only when it is not present. I'll send PR to check if we need it first.

martinpitt:
I suppose we could restrict the test case to RHEL and CentOS, which actually have a vdo package. So far the test runs everywhere. Also note that the test covers on-demand installation, so you'll have to remove and build at least one fake package.

marusak:
> Also note that the test covers on-demand installation, so you'll have to remove and build at least one fake package.

Yes, it also does create `vdo` package: https://github.com/cockpit-project/cockpit/pull/17020/files#diff-963506329ee019eba7b52956adab34de2bddb814a265f291eeee24d518719db7L444

martinpitt:
Meh, [two failures](https://logs.cockpit-project.org/logs/pull-2949-20220217-155454-566aeafc-centos-8-stream-cockpit-project-cockpit/log.html#137-2), but they look familiar.

martinpitt:
Yep, [same failures as in another PR](https://logs.cockpit-project.org/logs/pull-17016-20220217-154950-74cea3f0-centos-8-stream/log.html), so unrelated.

marusak (approved):

PR #2952, opened 2022-02-17
testmap: Drop Fedora 34 from our projects

martinpitt:
We dropped that from our cockpituous control files, and now only release
to Fedora 35 and 36.


marusak (approved):
Thanks!

PR #2959, opened 2022-02-17
images: add fedora-36 image

jelly:
Fedora 36 is the current development release.

 * [x] image-refresh fedora-36


lib/testmap.py
```diff
@@ -48,6 +48,7 @@
         ],
         # These can be triggered manually with bots/tests-trigger
         '_manual': [
+            'fedora-36',
```
martinpitt, line 51: I recommend a separate PR which adds fedora-36 to *all* our projects (to `_manual`), so that we can enable them one by one.

martinpitt (commented):
For the final version, please also copy fedora-36.install, so that image-prepare works in cockpit. We will most likely also need a bunch of naughties copied from fedora-35, but we can do this after the first cockpit run. But let's start with starter-kit and podman, and work our way up :grin: 

This seems good enough to trigger an image build already, thanks!

cockpituous:
Log: https://logs.cockpit-project.org/logs/image-refresh-2959-20220217-162929/

cockpituous:
image-refresh fedora-36 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-36-20220217-164646

martinpitt:
@jelle: Note that you can't yet trigger f36 tests -- that's why it really needs a separate PR with the `_manual` entries first, so that our CI (bots main) knows about it. However, since this is a brand new image, it's fine. Please just add the actual image build commit here, then it's ready to land.

images/scripts/fedora-36.install
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
```
martinpitt, line 1: https://github.com/cockpit-project/cockpit/pull/16935 just landed, so we don't need the `*.install` file any more now. However, if you don't need to re-push, it's fine -- I'll send a PR to drop them wholesale anyway.

martinpitt (approved):
Thanks! Please trigger starter-kit, to make sure that at least that one works. Then good to land, test podman, and add starter-kit and podman to testmap for main branches.

images/scripts/fedora-36.install
jelly: Ok

Issue #2703, opened 2021-12-09
"tuned-adm recommend" reports non-existing "atomic-guest"


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2030580

We only see this in packit today (2021-12-09), as the Testing Farm infrastructure makes tuned believe it's an "atomic-guest". On our infra and locally it says "virtual-guest".

martinpitt:
bump, this only happens on packit, and the downstream bug is still open

martinpitt:
Reopening, see #2853

martinpitt:
Bump, this only happens on c9s, i.e. in packit. That can't update naughties.

martinpitt:
Bump, [still happens on c9s](https://artifacts.dev.testing-farm.io/2f01d8ca-7599-4d83-adf8-c8082ec31dca/work-allglKM1J/log.txt) and bz is still open.

martinpitt:
Bump, downstream bug is still open.

martinpitt:
Bump.

martinpitt:
Bump

martinpitt:
Bump

martinpitt:
Bump

PR #2966, opened 2022-02-18
naughty: Close 2703: "tuned-adm recommend" reports non-existing "atomic-guest"

cockpituous:
Known issue which has not occurred in 22 days

"tuned-adm recommend" reports non-existing "atomic-guest"

Fixes #2703


martinpitt:
Tthis only happens on c9s, i.e. in packit. That can't update naughties.

PR #2979, opened 2022-02-21
testmap: adjust image used for the anaconda tests

KKoukiou:


martinpitt:
This test context is not known yet (it's not in `_manual`), so I did some local trickery to queue it.

martinpitt:
Nice! That test took 101s, even though it downloaded the boot.iso. [A previous test](https://logs.cockpit-project.org/logs/pull-3868-20220215-104536-701539bb-fedora-35-rawhide/log) took 267s, the download alone took over two minutes.

martinpitt (approved):

PR #2981, opened 2022-02-21

cockpituous:


martinpitt:
@KKoukiou : You triggered this manually -- note that you can increase the refresh period in image-trigger; it currently uses the default 7 days, but e.g. fedora-testing is updated every 3 days.

PR #2982, opened 2022-02-21
image-customize: Add --build-options option

martinpitt:
This can pass arbitrary additional options to mock/pbuilder/arch
builder. We need it at least for cockpit's rhel-7.9 branch to pass an
additional RPM macro definition.

---

Tested/used by https://github.com/cockpit-project/cockpit/pull/17029 , otherwise triggering a few tests here to ensure this does not regress.

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -27,6 +27,7 @@ from machine import testvm
 
 opt_quick = False
 opt_verbose = False
+opt_build_options = None
 
 
 def prepare_install_image(base_image, install_image, resize, fresh):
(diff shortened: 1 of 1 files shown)


martinpitt:
Argh, this tickled the deployment test. But I don't want to force-push here, as so many integration tests are running. As repentence I fixed that in PR #2983.

martinpitt:
Eh, what? What do all these podman failures have to do with this change?

marusak (approved):

PR #2986, opened 2022-02-22
images: Drop *.install files

martinpitt:
[1] and [2] moved the cockpit specific VM preparation to cockpit.git,
similar to what starter-kit and derivatives have always done. These
files are cockpit specific and thus should not be in bots.

[1] https://github.com/cockpit-project/cockpit/pull/16935
[2] https://github.com/cockpit-project/cockpit/pull/17029


marusak (approved):
whooo \o/

Issue #2991, opened 2022-02-23
debian: ubuntu: virt-install fails to reinstall VM when automatic UEFI firmware is set


KKoukiou:
Downstream report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006324

(I just reported the bug on the debian tracker)

KKoukiou:
debian-testing
Ooops, it happened again<hr>
```
# testConfigureBeforeInstall (__main__.TestMachinesCreate)

DevTools listening on ws://127.0.0.1:9589/devtools/browser/9c220fb1-740b-4224-b9da-754164c0b6ba
[0223/143631.902473:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 6519: Permission denied (13)
[0223/143631.903391:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0223/143631.904783:ERROR:viz_main_impl.cc(161)] Exiting GPU process due to errors during initialization
[0223/143631.907199:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 6526: Permission denied (13)
[0223/143631.909684:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 6531: Permission denied (13)
[0223/143631.910382:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0223/143631.911698:ERROR:viz_main_impl.cc(161)] Exiting GPU process due to errors during initialization
[0223/143631.913744:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 6539: Permission denied (13)
[0223/143631.914294:ERROR:gpu_init.cc(457)] Passthrough is not supported, GL is disabled, ANGLE is 
[0223/143632.168253:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 6571: Permission denied (13)
Generating a RSA private key
................................................................................++++
...............................................................++++
writing new private key to 'key.pem'
-----
[...]

PR #2992, opened 2022-02-23
naughty: Add pattern for apparmor blocking VM installation when UEFI is set

KKoukiou:
Known issue: #2991
Downstream report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1006324

### DIFF
--- a/naughty/debian-stable/2991-apparmor-virt-install-uefi-firmware
+++ b/naughty/debian-stable/2991-apparmor-virt-install-uefi-firmware
@@ -1,0 +1,3 @@
+# testConfigureBeforeInstall (__main__.TestMachinesCreate)
+*
+*apparmor="DENIED" operation="open" profile="libvirt-*" name="/var/lib/libvirt/qemu/nvram/VmNotInstalled_VARS.fd"*
--- a/naughty/debian-testing/2991-apparmor-virt-install-uefi-firmware
+++ b/naughty/debian-testing/2991-apparmor-virt-install-uefi-firmware
@@ -1,0 +1,3 @@
+# testConfigureBeforeInstall (__main__.TestMachinesCreate)
+*
+*apparmor="DENIED" operation="open" profile="libvirt-*" name="/var/lib/libvirt/qemu/nvram/VmNotInstalled_VARS.fd"*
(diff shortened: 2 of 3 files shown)


KKoukiou:
Tests are running here https://github.com/cockpit-project/cockpit-machines/pull/583

martinpitt (dismissed):
Thanks! A bit sad to introduce a regression (from user's POV) that way, but let's hope it gets fixed soon.

KKoukiou:
@martinpitt oh indeed, good catch. As you imagined, ubuntu-2004 fails with:

```error: unrecognized arguments: --reinstall``` I guess I need to block this PR on newer virt-install on ubuntu.

martinpitt:
Thanks @KKoukiou : It's not *that* serious -- at least Debian does not have AppArmor installed by default (Ubuntu does), and I have some reasonably good connections to be able to pull some strings to fix that. AppArmor is also significantly easier to fix than SELinux, I am happy to give it a shot even.

However, `--reinstall` not supported in 20.04 LTS is more serious. In two months 22.04 LTS gets released, and then I'll stop backports to 20.04 LTS. As this is a really cool simplification, I'm also happy to stop backporting now already (it doesn't make much of a difference really).

KKoukiou:
> It's also worth filing that bug on https://bugs.launchpad.net/ubuntu/+source/apparmor -- if you have time to do that, I'd appreciate; otherwise I'll do it soon.

https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/1962035

martinpitt (approved):

martinpitt:
@KKoukiou : I sent a fix to https://gitlab.com/libvirt/libvirt/-/merge_requests/140 in case you want to try this (it's easy enough to patch in a running system)

KKoukiou:
> @KKoukiou : I sent a fix to https://gitlab.com/libvirt/libvirt/-/merge_requests/140 in case you want to try this (it's easy enough to patch in a running system)

Sure, I can test this gladly.

martinpitt:
Rebased to run against current images in https://github.com/cockpit-project/cockpit-machines/pull/583

martinpitt:
See https://github.com/cockpit-project/cockpit-machines/pull/583#issuecomment-1060235455 , the previous pattern does not match. Updated to match on the AppArmor message instead.

martinpitt:
Let's land it now -- we are now needing this for debian-testing in PR #3108 anyway, even without the `--reinstall` rewrite in https://github.com/cockpit-project/cockpit-machines/pull/583

martinpitt (approved):

PR #3003, opened 2022-02-25
images: move acl to test dependencies

jelly:


martinpitt (approved):
Thanks! I think we can get away with not rebuilding this.

PR #3020, opened 2022-03-02
testmap: Add cockpit rhel-9.0 branch

martinpitt:
We need this at least for updating translations, and smuggling in simple fixes like a backport of https://github.com/cockpit-project/cockpit/pull/17073

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -47,6 +47,9 @@ REPO_BRANCH_CONTEXT = {
             'rhel-7-9',
             'centos-7',
         ],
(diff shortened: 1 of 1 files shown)


martinpitt:
Skipping the route over `_manual` -- after this lands, I'll do an initial PR against rhel-9.0 to see if the tests work. If not, I'll fix them in that PR, as that's the first thing that should happen to this branch. This blocks updating translations and backporting "real" fixes.

marusak (approved):

PR #3028, opened 2022-03-04
testmap: test sub-man 1.29.26 with `rhel-9-0`

ptoscano:
The `subscription-manager-1.29.26` branch of subscription-manager targets
RHEL 9.0, so test it with the `rhel-9-0` image only.


martinpitt (approved):
I triggered the test, succeeded. Thanks!

PR #3027, opened 2022-03-05
Image refresh for rhel-9-0

cockpituous:
Image refresh for rhel-9-0
 * [x] image-refresh rhel-9-0


martinpitt:
@jkozol : I retried the cockpit-composer test three times, it keeps timing out in TestImage.testOSTreeCommit..

jkozol:
@martinpitt Thanks for the ping. Changes incoming.

martinpitt:
c-composer has had the same TestOSTree timeouts for a long time on rhel-8-6. Let's not block on this any further, we are falling behind in testing current RHEL 9.

martinpitt (approved):

PR #3034, opened 2022-03-07
machine: Add SSHConnection.reboot()

martinpitt:
Cockpit's tests repeat this code a lot, so let's add a proper API for
it.

---

See `git grep m.spawn.*reboot test/verify` in cockpit.git.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -146,6 +146,11 @@ class SSHConnection(object):
         else:
             raise exceptions.Failure("Timeout waiting for system to reboot properly")
 
+    def reboot(self, timeout_sec=180):
+        self.spawn("reboot", "reboot", check=False)
+        if timeout_sec:
+            self.wait_reboot(timeout_sec)
+
     def _start_ssh_master(self):
         self._kill_ssh_master()
 
@@ -253,7 +258,7 @@ class SSHConnection(object):
 
     def execute(self, command=None, script=None, input=None, environment={},
                 stdout=None, quiet=False, direct=False, timeout=120,
-                ssh_env=["env", "-u", "LANGUAGE", "LC_ALL=C"]):
+                ssh_env=["env", "-u", "LANGUAGE", "LC_ALL=C"], check=True):
         """Execute a shell command in the test machine and return its output.
 
         Either specify @command or @script
@@ -359,7 +364,7 @@ class SSHConnection(object):
                             proc.stdin.close()
             proc.wait()
 
-        if proc.returncode != 0:
+        if proc.returncode != 0 and check:
             raise subprocess.CalledProcessError(proc.returncode, command, output=output)
         return output
 
@@ -471,7 +476,7 @@ class SSHConnection(object):
             cmd += " && chmod '%s' '%s'" % (perm, dest)
         self.execute(command=cmd, input=content)
 
-    def spawn(self, shell_cmd, log_id):
+    def spawn(self, shell_cmd, log_id, check=True):
         """Spawn a process in the test machine.
 
         Arguments:
@@ -481,4 +486,8 @@ class SSHConnection(object):
         Returns:
             The pid of the /bin/sh process that executes the command.
         """
-        return int(self.execute("{ (%s) >/var/log/%s 2>&1 & }; echo $!" % (shell_cmd, log_id)))
+        res = self.execute(f"{{ ({shell_cmd}) >/var/log/{log_id} 2>&1 & }}; echo $!",
+                           check=check)
+        if not check:
+            return None
+        return int(res)


martinpitt:
Used/tested by https://github.com/cockpit-project/cockpit/pull/17089

allisonkarlitskaya (commented):
Why the `sync` and the `sleep`?

I understand that this is just copying existing usage, but if we're making an API, probably it makes sense to think about why this is necessary.  They both look hella suspicious.

marusak (approved):

martinpitt:
Right, I already reduced the `sync && sync && sync` to a single `sync`, but it really shouldn't be necessary -- rebooting already syncs. So I'll take it out.

The "sleep" is a race condition. It needs to postpone the reboot long enough for the m.execute() to return, so that it doesn't prematurely kill SSH.

allisonkarlitskaya:
> The "sleep" is a race condition. It needs to postpone the reboot long enough for the m.execute() to return, so that it doesn't prematurely kill SSH.

Can we find another way to do that?

martinpitt:
If you have a better idea, I'm all ears.

allisonkarlitskaya:
`check=False` parameter to `machine.execute()`?

martinpitt:
I don't like that. That would just hide errors, in case we break something in the VM so that rebooting would fail.

martinpitt:
I mean, we *can* try to drop the sleep entirely, but it would still be fundamentally the same race condition. But this never got in the way.

allisonkarlitskaya:
I actually don't understand why the sleep makes any difference at all here.  It just delays the `reboot` call.  It doesn't cause the ssh connection to terminate any quicker relative to the `reboot` call.  It still waits until after it's invoked.

allisonkarlitskaya:
Also: why do we pass both a `command` and `script` parameter?  It seems `script` gets ignored in the case that `command` is given...

allisonkarlitskaya:
> @allisonkarlitskaya : It does delay it -- this is m.spawn() (asynchronous), _not_ m.execute()

Right.  Thanks for the clarification.

allisonkarlitskaya:
What about something like `cat && reboot`?  ie: block until stdin is closed and then proceed.  It might even make sense to bake that into the shell stuff in the implementation of `spawn()` in that case...

martinpitt:
OK, moving to draft for later. This really isn't the bit that I wanted to do right now, and it's not at all important.

allisonkarlitskaya:
So using this in the shell:

```
ps -o ppid= $$
```

gets you the ppid, which is equal to the sshd instance that was forked off to handle the current connection.  We just need to wait for that process to be gone.




> I don't like that. That would just hide errors, in case we break something in the VM so that rebooting would fail.

The logic here is false: the entire point of this is that we don't call `reboot` until after the ssh session closes, which means that we won't catch errors.

martinpitt:
I experimented with `cat`, and that seems to work well enough -- we don't actually close stdin in execute anywhere, unless we give a stdin input string. Doing a run against that in https://github.com/cockpit-project/cockpit/pull/17089

allisonkarlitskaya:
I'm not super keen on the `cat` idea because it depends on the implementation of `sshd` and could just be trading one race for another.

I really start to think the correct solution here is to give a way to ignore `execute()`/`spawn()` failures...

allisonkarlitskaya:
But: I feel like we've gone around enough times and thought about it enough that I'm happy to stand out of the way and let you decide which approach you prefer.

martinpitt:
> the entire point of this is that we don't call reboot until after the ssh session closes, which means that we won't catch errors.

Right, good point! And we'll have the output in /var/log anyway. Let's give this a try

machine/machine_core/ssh_connection.py
```diff
@@ -146,6 +146,12 @@ def wait_reboot(self, timeout_sec=180):
         else:
             raise exceptions.Failure("Timeout waiting for system to reboot properly")
 
+    def reboot(self, timeout_sec=180):
+        # run cat to let spawned process wait for stdin to close (when m.execute() finishes)
```
allisonkarlitskaya, line 150: This comment is no longer accurate.

machine/machine_core/ssh_connection.py
```diff
@@ -253,7 +259,7 @@ def __execution_opts(self, direct=False):
 
     def execute(self, command=None, script=None, input=None, environment={},
                 stdout=None, quiet=False, direct=False, timeout=120,
-                ssh_env=["env", "-u", "LANGUAGE", "LC_ALL=C"]):
+                ssh_env=["env", "-u", "LANGUAGE", "LC_ALL=C"], ignore_exit_status=False):
```
allisonkarlitskaya, line 262: There is a very strong tradition for this parameter to be called `check`.

allisonkarlitskaya (changes requested):
(blocking because of the wrong comment.  the naming thing is a bikeshed.)

machine/machine_core/ssh_connection.py
martinpitt: Dropped.

martinpitt:
Tests in https://github.com/cockpit-project/cockpit/pull/17089 are happy, so I think this is what I want to go with.

PR #3045, opened 2022-03-08
naughty: Adjust #2412 pattern for m.reboot()

martinpitt:
https://github.com/cockpit-project/cockpit/pull/17089 will change the
`m.wait_reboot()` into `m.reboot()`. Adjust the pattern to match on
both APIs.


martinpitt:
Testing the new API in https://github.com/cockpit-project/cockpit/pull/17089 , testing the old API right here.

PR #3048, opened 2022-03-09
images: add libappstream-glib-devel in RHEL 8

ptoscano:
Pull it so it is possible to use gettext to merge translations in
AppStream XML files named .metainfo.xml; gettext in RHEL 8 does not
know about them, and libappstream-glib-devel provides the extra metadata
for gettext to handle them.

This will be needed for cockpit-project/starter-kit#534.

 - [ ] image-refresh centos-8-stream
 - [ ] image-refresh rhel-8-6

### DIFF
--- a/images/scripts/lib/build-deps.sh
+++ b/images/scripts/lib/build-deps.sh
@@ -37,4 +37,12 @@ case "$OS_VER" in
     *) EXTRA_DEPS="$EXTRA_DEPS nodejs" ;;
 esac
 
+# libappstream-glib-devel is needed for merging translations in AppStream XML files in starter-kit and derivatives
+# on RHEL 8 only: gettext in RHEL 8 does not know about .metainfo.xml files, and libappstream-glib-devel
+# provides /usr/share/gettext/its/appdata.{its,loc} for them
+case "$OS_VER" in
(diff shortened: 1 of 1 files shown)


images/scripts/lib/build-deps.sh
```diff
@@ -32,8 +32,11 @@ EXTRA_DEPS="libappstream-glib rpmlint"
 
 # nodejs for starter-kit and other projects which rebuild webpack during RPM build; but it can't be installed on RHEL 8
 # mocks due to missing modularity package indexes
+# libappstream-glib-devel is needed for merging translations in AppStream XML files in starter-kit and derivatives
+# on RHEL 8 only: gettext in RHEL 8 does not know about .metainfo.xml files, and libappstream-glib-devel
+# provides /usr/share/gettext/its/appdata.{its,loc} for them
 case "$OS_VER" in
-    rhel*8) ;;
+    rhel*8) EXTRA_DEPS="$EXTRA_DEPS libappstream-glib-devel" ;;
```
martinpitt, line 39: We need this on `centos*8` as well, but there we also need `nodejs`. So this should better become its own separate `case`/comment block.

martinpitt (changes requested):

martinpitt:
Sorry, I forgot that external contributors can't do image refresh PRs. @ptoscano , I copied this to #3049, I'll take it from here. Thanks!

PR #3059, opened 2022-03-10
log.html: show total time it took to run tests

jelly:
image


martinpitt (approved):
Thanks! I'm not familiar with the mustache syntax, but this looks plausible.

Issue #3068, opened 2022-03-14
avc: denied { create } for comm="timedatex" scontext=system_u:system_r:timedatex_t:s0 tcontext=system_u:system_r:timedatex_t:s0 tclass=unix_dgram_socket


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2063888

martinpitt:
Bump, this is rare -- let's keep this for another round.

PR #3069, opened 2022-03-14
naughty: Add pattern for timedatex D-Bus SELinux rejection

martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2063888
Known issue #3068

---

[seen here](https://logs.cockpit-project.org/logs/pull-17109-20220310-121234-6061e64b-centos-8-stream/log.html#70), and tested against that log.

### DIFF
--- a/naughty/rhel-8/3068-timedatex-naughty
+++ b/naughty/rhel-8/3068-timedatex-naughty
@@ -1,0 +1,3 @@
+audit: * avc:  denied  { create } for * comm="timedatex" scontext=system_u:system_r:timedatex_t:s0 tcontext=system_u:system_r:timedatex_t:s0 tclass=unix_dgram_socket*
+*
+testlib.Error: FAIL: Test completed, but found unexpected journal messages:


naughty/rhel-8/3068-timedatex-naughty
```diff
@@ -0,0 +1,3 @@
+audit: * avc:  denied  { create } for * comm="timedatex" scontext=system_u:system_r:timedatex_t:s0 tcontext=system_u:system_r:timedatex_t:s0 tclass=unix_dgram_socket*
+*
+testlib.Error: FAIL: Test completed, but found unexpected journal messages:
```
marusak, line 3: we rarely add also this line to avc, don't we?

martinpitt: This AVC does not actually make any particular test fail, it's completely random background activity. I don't want a real failure to be ignored because of that unrelated SELinux message. I think we should do the same for all naughties which are similar. We of course also have SELinux blockers which *do* break functionality that we test.

marusak (approved):
Good call!

PR #3073, opened 2022-03-15
testmap: test sub-man 1.28.29 with `rhel-8-6`

ptoscano:
The subscription-manager-1.28.29 branch of subscription-manager targets
RHEL 8.6, so test it with the `rhel-8-6` image only.


martinpitt:
Fudged it so that the test runs here.

martinpitt (approved):
Grazie!

PR #3077, opened 2022-03-16
Image refresh for rhel-8-4

cockpituous:
Image refresh for rhel-8-4
 * [x] image-refresh rhel-8-4


martinpitt (approved):

martinpitt:
@ptoscano : Do you still care about your -1.28 branch? RHEL 8.4 is really old by now, from an upstream's and RHEL packager's POV. This would be an easy image to retire.

Issue #2898, opened 2022-02-05
useradd causes avc: denied { execute } for pid=57884 comm="chpasswd|chage" name="sss_cache"


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2050952

cockpituous:
fedora-35
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testLibvirt (__main__.TestMachinesLifecycle)

DevTools listening on ws://127.0.0.1:9608/devtools/browser/38132afc-7a03-4901-9976-ad11eee0be58
[0205/091425.286810:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 16746: Permission denied (13)
[0205/091425.287844:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: cannot open shared object file: No such file or directory
[0205/091425.289526:ERROR:viz_main_impl.cc(161)] Exiting GPU process due to errors during initialization
[0205/091425.290790:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 16760: Permission denied (13)
[0205/091425.293595:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 16775: Permission denied (13)
[0205/091425.294272:ERROR:gpu_init.cc(453)] Passthrough is not supported, GL is disabled, ANGLE is 
[0205/091425.547747:ERROR:zygote_host_impl_linux.cc(263)] Failed to adjust OOM score of renderer with pid 16845: Permission denied (13)
error: Failed to define network from /etc/libvirt/qemu/networks/default.xml
error: operation failed: network 'default' already exists with uuid bd983666-3243-4b7a-af2e-a118e6a95aa5
> error: Modal: Specify at least one of: title, aria-label, aria-labelledby.
> error: You are currently using minified code outside of NODE_ENV === "production". This means that you are running a slower development build of Redux. You can use loose-envify (https://github.com/zertosh/loose-envify) for browserify or setting mode to production in webpack (https://webpack.js.org/concepts/mode/) to ensure you have the correct code for your production build.
> error: Scrollbar test exception: TypeError: Cannot read properties of null (reading 'appendChild')
[...]

PR #3097, opened 2022-03-16
naughty: Add alternative chage pattern for #2898

martinpitt:
This does not only affect sss_cache running `chpasswd`, but `chage` as
well. See https://bugzilla.redhat.com/show_bug.cgi?id=2050952#c3

---

See [this failure](https://logs.cockpit-project.org/logs/pull-17135-20220315-133402-92539075-fedora-35-firefox/log.html#307-2), which was previously hidden by a bug in check-users. That gets fixed in https://github.com/cockpit-project/cockpit/pull/17135

I validated this with `bots/test-failure-policy -o fedora-35 < /tmp/out` against that log from above.


marusak (approved):
Thanks!

PR #3026, opened 2022-03-04
WIP: 3-ci-srv-03: [no-test] Image refresh for rhel-8-6

cockpituous:
Image refresh for rhel-8-6
 * [ ] image-refresh rhel-8-6

PR #3103, opened 2022-03-17
Image refresh for rhel-8-6

cockpituous:
Image refresh for rhel-8-6
 * [x] image-refresh rhel-8-6


martinpitt:
cockpit-composer repeated failures, same as in the last refresh #3026 and before. Continuing to ignore (we have a meeting about that tomorrow)

martinpitt (approved):

PR #3102, opened 2022-03-17

cockpituous:


martinpitt:
These repeatedly failed, and I can't download the image locally either. Somehow it got lost, overzealous pruning? (storage space might get a little low?) Let's re-do this.

PR #3104, opened 2022-03-17
testmap: add my test of cockpit split from sub-man

ptoscano:
Add my local repository with the test of the split of the cockpit plugin
of subscription-manager out of subscription-manager.git; this way I can
test the cockpit CI on that testing repository before the official
split.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -178,6 +178,18 @@ REPO_BRANCH_CONTEXT = {
         '_manual': [
         ]
     },
+    'ptoscano/subscription-manager-cockpit': {
+        'main': [
+        ],
+        '_manual': [
+            'rhel-8-5',
+            'rhel-8-6',
+            'rhel-8-6/subscription-manager-1.28.29',
+            'rhel-9-0',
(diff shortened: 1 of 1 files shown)


lib/testmap.py
```diff
@@ -178,6 +178,18 @@
         '_manual': [
         ]
     },
+    'ptoscano/subscription-manager-cockpit': {
+        'main': [
+        ],
+        '_manual': [
+            'rhel-8-5',
+            'rhel-8-6',
+            'rhel-8-6/subscription-manager-1.28.29',
```
martinpitt, line 187: Are you sure these are scenarios, not branch names?

ptoscano: Yes, as my idea is to use the scenario feature to test the `main` branch of sub-man-cockpit with multiple branches of sub-man, to ensure it is compatible.

ptoscano: On top of that, a longer term idea of mine is to add a "system" scenario to use system packages rather than a git version of sub-man.

martinpitt (approved):
Ack, thanks!

PR #3106, opened 2022-03-17
images: build Arch Linux packages offline

jelly:
This changes the way we build Arch Linux packages for tests on our
images to allow for offline building. The standard way of building using
extra-x86_64-build first synchronises the Arch mirrors and fetches all
build dependencies when required. In this commit a build chroot is
pre-generated with the required build dependencies for cockpit
cockpit-podman and cockpit-machines. Using makechrootpkg to build the
package in the existing chroot should also speed up the build a bit as
this no longer runs checkpkg by default.

 * [x] image-refresh arch


jelly:
Test it here https://github.com/cockpit-project/cockpit-machines/pull/626

cockpituous:
image-refresh arch done: https://github.com/cockpit-project/bots/commits/image-refresh-arch-20220317-162828

cockpituous:
image-refresh in progress on 4-cockpit-7.
Log: https://logs.cockpit-project.org/logs/image-refresh-3106-20220317-162828/

cockpituous:
image-refresh arch done: https://github.com/cockpit-project/bots/commits/image-refresh-arch-20220317-164242

images/scripts/arch.setup
```diff
@@ -106,6 +106,33 @@ echo "%wheel ALL=(ALL) ALL" > /etc/sudoers.d/90-cockpit-wheel
...
+npm \
+pcp \
+perl-json \
+perl-locale-po \
+xmlto \
+"
+
+# Create build chroot for cockpit projects for offline builds, the base chroot
+# is created in /var/tmp/archbuild/root
+mkdir /var/tmp/archbuild
```
martinpitt, line 133: Nitpick, but /var/tmp/ is prone to getting cleaned up automatically. Better use /var/lib/ or /var/cache/.

images/scripts/arch.setup
```diff
@@ -106,6 +106,33 @@ echo "%wheel ALL=(ALL) ALL" > /etc/sudoers.d/90-cockpit-wheel
 useradd -c Builder builder
 echo "builder ALL = NOPASSWD: /usr/sbin/makechrootpkg, /usr/sbin/mkarchroot, /usr/bin/*-x86_64-build" > /etc/sudoers.d/90-archbuild
 
+BUILD_DEPS="\
```
martinpitt, line 109: Is there some way to tell it "install all build dependencies of cockpit"? Then we'd only have a small (possilbly empty) list of extra packages. By far not a blocker though, it doesn't change often.

martinpitt (changes requested):
Tests fail with `==> ERROR: No chroot dir defined, or invalid path '/var/tmp/archbuild'`. Nevertheless, that's probably easy to fix -- I really appreciate seeing this, thanks for figuring out!

images/scripts/arch.setup
jelly: We actually do record the makedepends now in the pacman database. The sync database (db pulled from the mirrors) is a simple tarball with packagename-$version/desc where `desc` is a simple file describing the package:

```
%FILENAME%
zstd-1.5.2-3-x86_64.pkg.tar.zst

%NAME%
zstd

%BASE%
zstd

%VERSION%
1.5.2-3

%DESC%
Zstandard - Fast real-time compression algorithm

%CSIZE%
671794

%ISIZE%
2487249

%MAKEDEPENDS%
gtest
```

cockpituous:
image-refresh in progress on centosci-tasks-jzqvf.
Log: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/image-refresh-3106-20220318-080202/

images/scripts/arch.setup
jelly: Raised questions for the pacman team, it would be nice to have something like apt build-deps has.

cockpituous:
Task failed: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/image-refresh-3106-20220318-080202/

jelly:
> Task failed: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/image-refresh-3106-20220318-080202/

http 400, requires a retry.

cockpituous:
image-refresh in progress on 1-ci-srv-04.
Log: https://logs.cockpit-project.org/logs/image-refresh-3106-20220318-084141/

cockpituous:
image-refresh arch done: https://github.com/cockpit-project/bots/commits/image-refresh-arch-20220318-085656

cockpituous:
image-refresh in progress on 4-cockpit-9.
Log: https://logs.cockpit-project.org/logs/image-refresh-3106-20220318-085656/

image-customize
```diff
@@ -189,10 +189,7 @@ class BuildAction(ActionBase):
         machine.execute(f"su builder {opt_build_options} /var/tmp/mkbuild.sh")
 
         # build binaries
-        machine.execute("cd /var/tmp/build; su builder -c extra-x86_64-build", timeout=1800)
-
-        # install packages
-        machine.execute("pacman -U --noconfirm /var/tmp/build/*.pkg.tar.zst")
+        machine.execute("cd /var/tmp/build; makechrootpkg -r /var/lib/archbuild/cockpit -U builder -- -i", timeout=1800)
```
jelly, line 192: -i tells makepkg to install the build packages, however this is run **in the chroot** not outside! *faceplam* :face_with_head_bandage:

jelly: As alternative we could call `makepkg --packagelist` and use that to install (it needs to be run from the PKGBUILD directory) so we don't hardcode the package extension here

images/scripts/arch.setup
jelly: So, I'll need to extend https://gitlab.archlinux.org/pacman/pacman/-/blob/master/src/pacman/util.c#L1145 so pacman can print the makedepends as well. Currently this is limited to a few formatting strings.

jelly:
Offline tests running here for c-machines https://logs.cockpit-project.org/logs/pull-626-20220318-130109-9655f4a4-arch-bots-3106/log.html

martinpitt:
I'm glad that the other failure is gone, and we are only left with this rather obvious [sudo message change](https://logs.cockpit-project.org/logs/pull-3106-20220318-142316-83898246-arch-cockpit-project-cockpit/log.html#233-2). Can you please send a cockpit PR which just relaxes the expected string? We don't care that much whether sudo reports failed logins :grin: 

xkcd

jelly:
PR made here https://github.com/cockpit-project/cockpit/pull/17151 waiting on https://logs.cockpit-project.org/logs/pull-17151-20220318-154509-ecffe463-arch-bots-3106/log.html

jelly:
After landing the sudo fix, the test now runs succesfully https://github.com/cockpit-project/cockpit-machines/pull/626

PR #3113, opened 2022-03-18
images: add the CRB repository for RHEL

ptoscano:
RHEL has a CodeReady Linux Builder (CRB in short) with packages shipped
to customers, usually needed to build other software, and not supported:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/package_manifest/codereadylinuxbuilder-repository

Since we may want to install packages from there to enable building more
bits on RHEL, then add it to the repositories of each RHEL 8+ image.

### DIFF
--- a/image-create
+++ b/image-create
@@ -93,6 +93,7 @@ class MachineBuilder:
                 "EXTRAS_REPO_URL": os.environ.get("EXTRAS_REPO_URL", ""),
                 "BASEOS_REPO_URL": os.environ.get("BASEOS_REPO_URL", ""),
                 "APPSTREAM_REPO_URL": os.environ.get("APPSTREAM_REPO_URL", ""),
+                "CRB_REPO_URL": os.environ.get("CRB_REPO_URL", ""),
             }
             self.machine.message("run setup script on guest")
 
--- a/image-create-rhel-compose
+++ b/image-create-rhel-compose
@@ -25,6 +25,7 @@ if [[ "$COMPOSE" =~ RHEL-7 ]]; then
 else
     export BASEOS_REPO_URL="http://download.devel.redhat.com/rhel-8/rel-eng/RHEL-8/$COMPOSE/compose/BaseOS/x86_64/os"
     export APPSTREAM_REPO_URL="http://download.devel.redhat.com/rhel-8/rel-eng/RHEL-8/$COMPOSE/compose/AppStream/x86_64/os"
+    export CRB_REPO_URL="http://download.devel.redhat.com/rhel-8/rel-eng/RHEL-8/$COMPOSE/compose/CRB/x86_64/os"
     install_url=$BASEOS_REPO_URL
 fi
 
--- a/images/scripts/rhel.setup
+++ b/images/scripts/rhel.setup
@@ -33,7 +33,7 @@ set -x
 if [ ! -f "$SKIP_REPO_FLAG" ]; then
     # Configure repositories.
 
-    if [ -n "$BASEOS_REPO_URL" -a -n "$APPSTREAM_REPO_URL" ]; then
+    if [ -n "$BASEOS_REPO_URL" -a -n "$APPSTREAM_REPO_URL" -a -n "$CRB_REPO_URL" ]; then
     # disable all default repos
     rm -f --verbose /etc/yum.repos.d/*.repo
 cat <<EOF > /etc/yum.repos.d/rel-eng.repo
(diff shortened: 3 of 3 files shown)


martinpitt:
Why does this have to be on the image? We try to stay as close to the cloud images as possible. If some project needs additional repos/software, that can just be installed in the per-project test/vm.install ?

ptoscano:
> Why does this have to be on the image?

Because it is not a 3rd party repo, but an official RHEL one; this means it must be in sync with BaseOS and AppStream. Adding it manually in a script for image-customize is possible, however cumbersome and potentially fragile. In the end CRB (and theoretically BUILDROOT) are available for packages built in RHEL, so it seems limiting to not be able to use them to _build_ things in the test VM.

martinpitt:
But the image doesn't install any packages from it anyway, so by the time you *use* it in your project (i.e. install packages), the  repo is out of sync with the packages on the image *anyway*. Hence my question what the purpose of this is?

Note that you should not build packages in the test VM itself. All our VMs have a pre-cached `mock` environment (which also has the RHEL buildroot repo). If you are missing some BuildRequires in that, then let's just rather add it there.

ptoscano:
> by the time you _use_ it in your project (i.e. install packages), the repo is out of sync with the packages on the image _anyway_.

Sorry, I did not explain correctly. For "in sync" I mean that if BaseOS & AppStream refer to a certain compose, then CRB must use the same one too. Ditto for the nightly compose references. I cannot blindly put the URL to a nightly repo if BaseOS is using a certain compose and viceversa.

> Hence my question what the purpose of this is?

Install CRB packages from the compose of BaseOS & AppStream.

> Note that you should not build packages in the test VM itself.

Well, starter-kit kind of does that:
```make
$(VM_IMAGE): $(TARFILE) $(NODE_CACHE) bots test/vm.install
	bots/image-customize --no-network --fresh \
		--upload $(NODE_CACHE):/var/tmp/ --build $(TARFILE) \
		--script $(CURDIR)/test/vm.install $(TEST_OS)
```
the `--build` parameter in particular.

> All our VMs have a pre-cached `mock` environment (which also has the RHEL buildroot repo). If you are missing some BuildRequires in that, then let's just rather add it there.

Yes, I know that, and while I can understand that, this approach is too limiting for projects. My actual case is with libappstream-glib-devel, however I think the situation is not just this specific package: if tomorrow I add a new BR in the spec file, then CI will fail and I need a whole image refresh for it. This means adding more and more packages for an image that is, in the end, shared by all the cockpit components, and maybe you wouldn't want to add all the extra unneeded bits for them. Again this applies only to `--build`, and installing packages manually outside works fine, albeit limited to BaseOS & AppStream, which is what this PR is for.

martinpitt:
> I cannot blindly put the URL to a nightly repo if BaseOS is using a certain compose and viceversa.

But our images don't pin down a compose, they use the /nightly symlink. So if you build an image today, install a package today and another one tomorrow, they will be (potentially) *different* packages.

> Well, starter-kit kind of does that:

No, it doesn't. `image-customize --build` builds rpms in mock (and deb/arch as well, in their respective build chroots). As I said, we don't build anything directly on the VM, as it would require tons of build dependencies which have no place in integration tests.

> My actual case is with libappstream-glib-devel

The RHEL/CentOS 8 images have that (a certain Pino sent a PR for translating metainfo files, so we refreshed our images to have it :grin: )

> If tomorrow I add a new BR in the spec file, then CI will fail and I need a whole image refresh for it.

Correct. However, (1) that happens rarely enough, and (2) we want to do that *anyway*. Our images should not install packages for each test run, but image preparation should ideally be completely offline. That makes tests reproducible and independent from OS regressions. We want the latter to happen in image refreshes only, not as part of testing a project PR.

martinpitt (changes requested):
Moving to draft and explicitly -1'ing for now. I am strongly convinced that this is the wrong solution -- it's invasive, not representative of how actual RHEL packages are built, and does not actually provide the "attach to a particular compose" functionality. @ptoscano , I'm happy to do a gmeet to talk this over, and figure out a better solution. Thanks!

martinpitt:
@ptoscano , can we close this, or do you still want to do anything with changing the mock chroot?

ptoscano:
OK, let's close this for now.

Adding packages is still "easy enough" (FSVO) currently, so we can add packages to the mock chroots in case we need them.

PR #3118, opened 2022-03-21
images: add gettext & desktop-file-utils in RHEL mock

ptoscano:
Pull them so it is possible to build gettext translations, and to
validate desktop files

gettext was already implicitly pulled by something else, so this makes
it simply explicit.


martinpitt:
That makes sense, but it will apply to the Fedora images as well. I suppose that was your intention? I'll transfer this to an origin PR so that we can refresh/test images, and adjust the commit message (s/RHEL//).

ptoscano:
> That makes sense, but it will apply to the Fedora images as well. I suppose that was your intention?

I thought about RHEL only, however Fedora is fine too.

> I'll transfer this to an origin PR so that we can refresh/test images, and adjust the commit message (s/RHEL//).

Thanks!

martinpitt:
Replaced with #3119 origin PR for technical reasons.

PR #3131, opened 2022-03-24
images: Add ubuntu-2204 image

martinpitt:
Now is a good time to start testing on the upcoming Ubuntu LTS -- it is
in feature freeze, let's spot and fix major bugs before the release.

Add manual test contexts for the projects which test on Ubuntu, so that
we can enable them step by step.

----

I locally tested the new image with cockpit-podman, and it generally works. It just needs one test adjustment: https://github.com/cockpit-project/cockpit-podman/pull/947


cockpituous:
image-refresh in progress on centosci-tasks-jzqvf.
Log: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/image-refresh-3131-20220324-052626/

cockpituous:
image-refresh ubuntu-2204 done: https://github.com/cockpit-project/bots/commits/image-refresh-ubuntu-2204-20220324-054343

lib/testmap.py
```diff
@@ -56,6 +56,7 @@
             'fedora-testing/dnf-copr',
             'rhel-8-6-distropkg',
             'rhel-9-0-distropkg',
+            'ubuntu-2204',
```
marusak, line 59: Would be great, if these could land in separate PR so we can trigger at least starter kit to see if the image at least boots and has basic functionality. Or we could just trigger the new context manually.

marusak (approved):
But I trust your local testing as well, thanks! :)

lib/testmap.py
martinpitt: I triggered the test in https://github.com/cockpit-project/cockpit-podman/pull/947 already. I initially typo'ed the PR number, but still, it's running.

martinpitt:
https://logs.cockpit-project.org/logs/pull-947-20220324-065944-06607665-ubuntu-2204-bots-3131/log.html is not super great, but by and large functional. I'll land this so that we can trigger tests without dirty tricks.

PR #3141, opened 2022-03-24
Image refresh for rhel-8-6

cockpituous:
Image refresh for rhel-8-6
 * [x] image-refresh rhel-8-6


martinpitt (approved):
Looks good, c-composer is known.

PR #3147, opened 2022-03-25
Naughties for cockpit ubuntu-2204

martinpitt:
See https://github.com/cockpit-project/cockpit/pull/17182 , tests are running there.


martinpitt:
@marusak : It's "just" one new one, and some from existing ubuntu-stable also disappeared. I intentionally started with zero naughties instead of copying them all over wholesale.

PR #3148, opened 2022-03-26
Image refresh for centos-7

cockpituous:
Image refresh for centos-7
 * [x] image-refresh centos-7


martinpitt:
third retry, and still [image not found](https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-3148-20220326-075128-82adc9fd-centos-7-cockpit-project-cockpit-rhel-7.9/log.html). Usually that's just some "eventually consistent" thing from Linode, but it looks like this one actually got lost. Let's try this again.

Issue #3163, opened 2022-03-30
targetcli delete fails with: [Errno 20] Not a directory: '/sys/kernel/config/target/iscsi/cpus_allowed_list'


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2069952

At the time of writing this only affects cockpit-machines Fedora Rawhide tests, so we won't get automatic naughty tracking.

martinpitt:
fedora-37
Ooops, it happened again<hr>
```
# testStoragePoolsCreate (__main__.TestMachinesStoragePools)
error: Failed to define network from /etc/libvirt/qemu/networks/default.xml
error: operation failed: network 'default' already exists with uuid 46f2104e-e91f-4816-a380-95073c534790
restorecon: SELinux: Could not get canonical path for /tmp/tmp472c9_qi/.mozilla/firefox/* restorecon: No such file or directory.
*** You are running in headless mode.
Warning no default label for /tmp/tmp472c9_qi/.mozilla/firefox/Crash Reports
Warning no default label for /tmp/tmp472c9_qi/.mozilla/firefox/Pending Pings
Warning no default label for /tmp/tmp472c9_qi/.mozilla/firefox/pkz4u03h.blank
Warning no default label for /tmp/tmp472c9_qi/.mozilla/firefox/profiles.ini
*** You are running in headless mode.
[GFX1-]: glxtest: Unable to open a connection to the X server
[GFX1-]: glxtest: libEGL initialize failed
[GFX1-]: RenderCompositorSWGL failed mapping default framebuffer, no dt
DevTools listening on ws://localhost:9321/devtools/browser/bfda420d-beed-4a45-8526-dbc79d32a54b
CDP: {"source":"other","level":"error","text":"The resource from “http://localhost:9090/cockpit/$d83e6d1755557697edc77f378dae91262674fda64be0f29faec940828aa0a452/shell/po.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).","timestamp":1648579157147,"url":"http://localhost:9090/machines","lineNumber":0}
[...]

PR #3164, opened 2022-03-30
naughty: Add pattern for Fedora Rawhide targetcli regression

martinpitt:
The first test will fail on `targetcli delete`. As that breaks the
cleanup, the next test that will try to set up an iSCSI device will fail
on creation as the previous device already exists. Add the "-followup"
pattern for that.

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2069952
Known issue #3163

----

See [recent c-machines packit failure on rawhide](https://artifacts.dev.testing-farm.io/8ee1bdbc-fdcf-4b03-a4e7-41f9b223b5c3//work-alldJiNZH/log.txt). I tested the patterns against that log.


marusak (approved):
Thank you!

PR #3173, opened 2022-03-31
testmap: add official subscription-manager-cockpit repo

ptoscano:
The cockpit plugin of subscription-manager was split in its own repository, so add it with some contexts (for `main` and `_manual`) of combinations of distros & subscription-manager branch we want to test.

Also remove my own repository that was used to test the split.


marusak (approved):
whoo :tada: 
It assumes that 4 contexts are already working. Fine from me of course, just that these will be already triggered on PRs.

PR #3178, opened 2022-04-01
Image refresh for rhel-9-0

cockpituous:
Image refresh for rhel-9-0
 * [x] image-refresh rhel-9-0


marusak (approved):
c-composer has been broken for a while now

PR #3179, opened 2022-04-01
npm-update: Pull prerelease versions on PF

marusak:
I ran this in cockpit and it produced:
```diff
--- package.json
+++ package.json
@@ -3,11 +3,11 @@
   "description": "Cockpit isn't a Node package, these are devel time deps, not needed to build tarball either",
   "private": true,
   "dependencies": {
-    "@patternfly/patternfly": "4.159.1",
-    "@patternfly/react-core": "4.175.4",
-    "@patternfly/react-icons": "4.26.4",
-    "@patternfly/react-styles": "4.25.4",
-    "@patternfly/react-table": "4.44.4",
+    "@patternfly/patternfly": "4.185.1",
+    "@patternfly/react-core": "4.202.14",
+    "@patternfly/react-icons": "4.53.14",
+    "@patternfly/react-styles": "4.52.14",
+    "@patternfly/react-table": "4.71.14",
```

### DIFF
--- a/npm-update
+++ b/npm-update
@@ -59,6 +59,8 @@ def write_package_json(package_json_path, data):
 def prefix(name, version):
     if not version[0].isdigit():
         return version
+    if name.startswith("@patternfly"):
+        return "prerelease"
     if name not in FRAGILE:
         return "^" + version
     return "~" + version


martinpitt:
No objection, but I think we should first try to manually update https://github.com/cockpit-project/cockpit/pull/17201 to a prerelease and see how that works for us.

npm-update
```diff
@@ -59,6 +59,8 @@ def write_package_json(package_json_path, data):
 def prefix(name, version):
     if not version[0].isdigit():
         return version
+    if name.startswith("@patternfly"):
+        return "prerelease"
```
martinpitt, line 63: Is that always guaranteed to exist *and* be newer than the latest release? I.e. could it happen that when they do a new major release, the prerelease is still the older one?

martinpitt:
https://github.com/cockpit-project/cockpit/pull/17208 landed, so now we know that this approach works in general. Unblocked, but there's still an outstanding question above.

npm-update
marusak: It is something that is manually done - you need to tag it with `prerelease`. But pretty sure that PF team does this with script and that we can count on that.

martinpitt (approved):
Ack, thanks! Let's see how this goes.

PR #3183, opened 2022-04-01
Add centos-9-stream

ptoscano:
Generalize a bit the EPEL enablement to make it work also for c9s, and add a centos-9-stream image.

 * [x] image-refresh centos-9-stream


martinpitt:
@ptoscano , I take it you want this for subscription-manager? As keeping and updating yet another image is not a small burden, some questions:

 - sub-man tests could presumably be all `@nondestructive` with a little care -- I'm happy to look into this. Not only would it make the tests run faster, and make them easier to debug/develop, you could then also (1) run them in packit (again, happy to help you set that up, it's rather simple) and (2) run them as RHEL/Fedora/CentOS gating tests. That provides c8s, c9s, and all Fedoras. You'd only need our images/infra for rhel-*.
 - sub-man is only really interesting on RHEL. Is testing on c9s really giving you some advantage, or wouldn't it just be an extra burden for you?

ptoscano:
>     * sub-man tests could presumably be all `@nondestructive` with a little care -- I'm happy to look into this. Not only would it make the tests run faster, and make them easier to debug/develop, you could then also (1) run them in packit (again, happy to help you set that up, it's rather simple) and (2) run them as RHEL/Fedora/CentOS gating tests. That provides c8s, c9s, and all Fedoras. You'd only need our images/infra for rhel-*.

I (briefly) noticed the "nondestructive tests" concept, and what I missed (most likely I didn't check hard enough) was a definition of what "nondestructive" means and/or implies. One important bit in our case, which _I think_ could be a problem preventing the usage of packit, is that sub-man needs a Candlepin server; currently this is provided by the "services" cockpit image.

>     * sub-man is only really interesting on RHEL. Is testing on c9s really giving you some advantage, or wouldn't it just be an extra burden for you?

Since c9s is "future RHELs", then IMHO checking in advance for possible issues with newer c9s packages should be a good thing. At the time of this writing there's `rhel-9-0` as only RHEL image for sub-man-cockpit, which is the only RHEL/CentOS 9 cockpit image; it is a good approximation, however RHEL 9.0 will not have the result of the split sub-man-cockpit.

martinpitt:
> sub-man needs a Candlepin server; currently this is provided by the "services" cockpit image.

Aww yes, that wouldn't work in packit, at least not as a separate image. It could be deployed as part of the test setup, but it takes ages and is brittle.

cockpituous:
image-refresh in progress on 2-ci-srv-06.
Log: https://logs.cockpit-project.org/logs/image-refresh-3183-20220404-151818/

cockpituous:
image-refresh centos-9-stream done: https://github.com/cockpit-project/bots/commits/image-refresh-centos-9-stream-20220404-153535

PR #3189, opened 2022-04-05
Image refresh for ubuntu-2204

cockpituous:
Image refresh for ubuntu-2204


marusak:
Failed twice the same way on machines: https://logs.cockpit-project.org/logs/pull-3189-20220405-054054-a3fb0a7d-ubuntu-2204-cockpit-project-cockpit-machines/log.html#6-2

```
Failed to destroy domain 'VmNotInstalled'
error: Requested operation is not valid: domain is not running
Could not run '/usr/bin/swtpm_setup'. exitstatus: 1; Check error log '/var/log/swtpm/libvirt/qemu/VmNotInstalled-swtpm.log' for details.
Domain installation does not appear to have been successful.
```

marusak:
`/var/log/swtpm/libvirt/qemu/VmNotInstalled-swtpm.log` contains:
```
Cannot write random bytes:
4067F5CBF87F0000:error:12000079:random number generator:RAND_write_file:Cannot open file:../crypto/rand/randfile.c:240:Filename=./.rnd
```

When I remove `swtpm` package it starts to work again. This package was only now included in our image. I think I'll report this and then exclude this package as workaround.

martinpitt:
@marusak : Note that we soon want/need swtpm for https://issues.redhat.com/browse/COCKPIT-834. So IMHO a better workaround would be to remove/disable it during the test, or just even better, do this as a naughty. After all, users will actually see this. Having to rebuild the image twice and hacking dependencies for such things feels both intrusive and also not testing reality.

marusak:
So it fails due to changing of the boot firmware: https://github.com/cockpit-project/cockpit-machines/blob/main/test/check-machines-create#L1544

martinpitt:
Most direct reproducer:
```
touch /var/lib/libvirt/novell.iso
virt-install --name t1 --os-variant fedora28 --memory 128 --wait -1 --noautoconsole --disk 'size=0.25,format=qcow2' --cdrom /var/lib/libvirt/novell.iso --boot uefi
```

This is from https://launchpad.net/ubuntu/+source/libvirt/8.0.0-1ubuntu6 which introduced the recommendation for `swtpm`.

I filed https://bugs.launchpad.net/ubuntu/+source/libvirt/+bug/1968131 for this, and a knownissue: https://github.com/cockpit-project/bots/issues/3200

jelly (approved):

Issue #3191, opened 2022-04-05
debian-stable: Error waiting for logical volume object for 'lvol0': Timed out waiting for object


mvollmer:
Upstream: https://github.com/storaged-project/udisks/issues/966

The bug exists in all versions of UDisks2, but seems to be triggered only(?) on debian-stable.

mvollmer:
debian-stable
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testLvm (__main__.TestStorageLvm2)

DevTools listening on ws://127.0.0.1:9701/devtools/browser/204bc6cc-25e9-450a-9425-1413f7d05df4
ATTENTION: default value of option allow_rgb10_configs overridden by environment.
[0405/151921.909455:WARNING:bluez_dbus_manager.cc(248)] Floss manager not present, cannot set Floss enable/disable.
[0405/151921.922774:ERROR:sandbox_linux.cc(377)] InitializeSandbox() called with multiple threads in process gpu-process.
> info: %cDownload the React DevTools for a better development experience: https://fb.me/react-devtools font-weight:bold
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (Not Found)","timestamp":1649161164845.403,"url":"http://127.0.0.2:9091/cockpit/$83c1f5cb8a2272826ffa7e1a57bf7b07d58b80632ab37938bb16ed19f71c5b5e/system/po.js","networkRequestId":"39995.52"}
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (Not Found)","timestamp":1649161164849.0378,"url":"http://127.0.0.2:9091/cockpit/$83c1f5cb8a2272826ffa7e1a57bf7b07d58b80632ab37938bb16ed19f71c5b5e/base1/po.js","networkRequestId":"39995.50"}
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (Not Found)","timestamp":1649161165267.478,"url":"http://127.0.0.2:9091/cockpit/$83c1f5cb8a2272826ffa7e1a57bf7b07d58b80632ab37938bb16ed19f71c5b5e/updates/po.js","networkRequestId":"39995.66"}
[...]

PR #3192, opened 2022-04-05
naughty: Add pattern for #3191

mvollmer:


martinpitt:
I'm feeling a little nervous about this one -- could you restrict it to a particular test, or at least a test class, and add the place where it fails? Or is that place too random?

I'm afraid this might hide totally unrelated things.

mvollmer:
> I'm afraid this might hide totally unrelated things.

I think it shouldn't, but fair enough.  The message is always a "this should never happen" internal error.  It means that the UDisks2 object that has just been created did not actually show up in the object tree.

martinpitt (approved):
Thanks so much for debugging/reporting this! Curious that this only hits d-stable, but we can copy the naughty when necessary. I triggered a cockpit test run.

PR #3202, opened 2022-04-07
image-customize: add --cpus & --memory-mb

ptoscano:
Add a way to tweak the number of vCPUs and RAM for the virtual machine
used for customizing the image; these can be used in case there are
tasks to run that may require more resources than the default ones
(i.e. 1 vCPU & 1152 MB).

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -249,6 +249,10 @@ def main():
                         help="Additional options for mock/pbuilder/arch builder")
     parser.add_argument('--resize', help="Resize the image. Size in bytes with using K, M, or G suffix.")
     parser.add_argument('-n', '--no-network', action='store_true', help='Do not connect the machine to the Internet')
+    parser.add_argument('--cpus', type=int, default=None,
+                        help="Number of CPUs for the virtual machine")
+    parser.add_argument('--memory-mb', type=int, default=None,
+                        help="RAM size for the virtual machine")
     parser.add_argument('-v', '--verbose', action='store_true',
                         help='Display verbose progress details')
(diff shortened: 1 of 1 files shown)


image-customize
```diff
@@ -272,10 +276,17 @@ def main():
...
+        'maintain': True,
+        'verbose': args.verbose,
+        'networking': network.host(restrict=args.no_network),
+        'image': prepare_install_image(args.base_image, args.image, args.resize, args.fresh),
+    }
+    if args.cpus:
+        machine_args['cpus'] = args.cpus
+    if args.memory_mb:
+        machine_args['memory_mb'] = args.memory_mb
+    machine = testvm.VirtMachine(machine_args)
```
martinpitt, line 289: This looks really weird. Instead of passing kwargs you now pass a single positional argument of a dictionary type. I would understand it if it would be `VirtMachine(**machine_args)`.

However, this indirection is completely unnecessary. Both the `args.*` values and the kwargs of VirtMachine default to `None`, so you can just pass them as they are. I.e.

    VirtMachine(maintain=True, ..., cpus=args.cpus, memory_mb=args.memory_mb)

martinpitt (changes requested):
Thanks!

image-customize
ptoscano: > I would understand it if it would be `VirtMachine(**machine_args)`.

Yes, that was it. That's what I get by applying wrong patches that late in the evening ...

martinpitt:
Thanks! LGTM now, triggering a few tests for good measure.

martinpitt (approved):

PR #3231, opened 2022-04-12
npm-update updates

marusak:
The idea behind this is, that we would replace our `every second day do npm-update` with two tasks - one to do every other day npm update of anything except patternfly and then once per week do only PF update.
The reasoning behind this change is, that now that we moved to PF prerelease, we get every day new update. Since npm-update takes the first it sees, it might happen that we will be just updating PF every time and never get to anything else. Also we don't need to update PF every other day, once per week is often enough - especially since it might need a bit of manual labour to update pixel tests etc.
@martinpitt WDYT?

### DIFF
--- a/npm-update
+++ b/npm-update
@@ -22,11 +22,6 @@
 #  * [ ] npm-update patternfly
 #
 
-# Dependencies where minor updates break, only bump microversion
-FRAGILE = [
-    "jquery",
-]
-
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Cheers! That will give us some more flexibility indeed -- we can continue to run the thing daily, but only on Mondays we run it for PF. Thanks!

martinpitt (approved):

PR #3232, opened 2022-04-12
npm-update: adjust to npm changes in cockpit repo

allisonkarlitskaya:
We no longer ask for `/npm install` on the PR in order to get a new
`node_modules` — we are now responsible for building it ourselves.

Also: replace the semi-gross `rmdir node_modules` trick to just check if
the `tools/node-modules` script is present: we need to call it in order
to do this stuff anyway.

See also cockpit-project/cockpit#17230

### DIFF
--- a/npm-update
+++ b/npm-update
@@ -88,9 +88,6 @@ def run(context, verbose=False, **kwargs):
                     if task.verbose:
                         sys.stderr.write("Ignoring '{0}' as there is pending PR #{1}\n".format(pkg, issue["number"]))
 
-    # The only time git will checkout an empty subdir is if it's a gitlink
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! I wonder if that also requires `permissions:` changes in the workflows, but these need to be handled in the project repos anyway.

allisonkarlitskaya:
> Thanks! I wonder if that also requires `permissions:` changes in the workflows, but these need to be handled in the project repos anyway.

It's actually a really good question.

At the very least, we're going to need to give the workflow access to the `node-cache` environment...  I'll make that tweak in the cockpit PR.

PR #3237, opened 2022-04-13
github-upload-secrets: allow multiple keys per target

allisonkarlitskaya:
We're going to want this so that the npm-upload workflow can push to both `node-cache` and `cockpit` repos.

- [x] actually test this first

### DIFF
--- a/github-upload-secrets
+++ b/github-upload-secrets
@@ -94,17 +94,22 @@ def upload_secret(api, path, name, value, pubkey, opts):
 
 
 def upload_deploy_key(api, repo, title, content, opts):
-    keys_path = f"/repos/{repo}/keys"
+    keys_path = f"/repos/{repo.lstrip('+')}/keys"
 
-    # delete all existing keys
-    keys = api.get(keys_path)
-    for key in keys:
-        key_id = key['id']
-        key_path = f'{keys_path}/{key_id}'
-        if opts.dry_run or opts.verbose:
-            print(opts.dry_run and "Would delete" or "Deleting", key_path)
-        if not opts.dry_run:
-            api.delete(key_path)
+    if repo.startswith('+'):
+        # add this key to the existing keys
+        pass
+
+    else:
+        # delete all existing keys
+        keys = api.get(keys_path)
+        for key in keys:
+            key_id = key['id']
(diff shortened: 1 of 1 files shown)


github-upload-secrets
```diff
@@ -94,17 +94,22 @@ def upload_secret(api, path, name, value, pubkey, opts):
...
-        key_path = f'{keys_path}/{key_id}'
-        if opts.dry_run or opts.verbose:
-            print(opts.dry_run and "Would delete" or "Deleting", key_path)
-        if not opts.dry_run:
-            api.delete(key_path)
+    keys_path = f"/repos/{repo.lstrip('+')}/keys"
+
+    if repo.startswith('+'):
+        # add this key to the existing keys
+        pass
```
martinpitt, line 101: Bit weird to use `pass` instead of negating the condition, but ok..

github-upload-secrets
```diff
@@ -132,13 +137,13 @@ def main():
...
                       help="Generate an SSH key, upload private, public to stdout")
-    parser.add_argument('--deploy-to', metavar="OWNER/REPO", action="append",
+    parser.add_argument('--deploy-to', metavar="[+]OWNER/REPO", action="append",
                         help="with --ssh-keygen, upload public key as deploy key")
 
     opts = parser.parse_args()
 
     NAME_RE = r'[a-z][-0-9a-z_.]*'
-    REPO_RE = f'{NAME_RE}/{NAME_RE}'
+    REPO_RE = fr'\+?{NAME_RE}/{NAME_RE}'
```
martinpitt, line 146: I.e. from now on, all names must be in French, right? :grin:

Cool, TIL!

martinpitt (approved):
Cheers! The PR title is of course totally wrong!!1! (This is an N:M relationship) :grin:

github-upload-secrets
allisonkarlitskaya: I agree, but I really wanted the comment here.

allisonkarlitskaya:
This is working now.  See cockpit-project/cockpit#17253.

These two PRs can land independently.  The script only runs when one of us does it manually, so there's no harm, either way.

Issue #3042, opened 2022-03-08
RHEL 8.6/9.0: "systemctl start insights-client" broken


mvollmer:
https://bugzilla.redhat.com/show_bug.cgi?id=2060834

cockpituous:
rhel-8-6
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testSubAndInAndFail (__main__.TestSubscriptions)
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.
Waiting 5 seconds for https://services.cockpit.lan:8443/candlepin
Waiting for https://services.cockpit.lan:8443/candlepin
Waiting for https://services.cockpit.lan:8443/candlepin
/usr/lib/python3.6/site-packages/urllib3/connection.py:376: SubjectAltNameWarning: Certificate for services.cockpit.lan has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/shazow/urllib3/issues/497 for details.)
  SubjectAltNameWarning
[0308/101052.869876:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory

DevTools listening on ws://127.0.0.1:10014/devtools/browser/6322e68a-f801-4278-be31-2f651d8736c8
[0308/101052.946132:ERROR:viz_main_impl.cc(161)] Exiting GPU process due to errors during initialization
[0308/101052.956160:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0308/101052.958791:ERROR:viz_main_impl.cc(161)] Exiting GPU process due to errors during initialization
[0308/101053.192326:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0308/101053.194890:ERROR:viz_main_impl.cc(161)] Exiting GPU process due to errors during initialization
[...]

PR #3247, opened 2022-04-17
Image refresh for rhel-8-6

cockpituous:
Image refresh for rhel-8-6
 * [x] image-refresh rhel-8-6


martinpitt:
insights-client (0:3.1.7-4.el8 -> 0:3.1.7-5.el8)

This is the same change as on RHEL 9.0 in #3248, and causes the same SELinux issues (see #3042). I'll adjust the naughty patterns accordingly.

marusak (approved):

PR #3270, opened 2022-04-21
Add a new containerised 'npm' wrapper, use it from npm-update

allisonkarlitskaya:
I went back and forth on this a couple of times, but it basically comes down to this: the code for the `npm` container needs to live somewhere, and preferably only once.  If we want to use this code for `npm outdated` (and I very much do) then it needs to be with the bots.

I'll follow this up with a commit in the `cockpit` repo to use `bots/npm download` as part of our `tools/node-modules install` command.

One cool immediate advantage: we can drop the `apt install npm` from the start of our `npm-update` workflows.

### DIFF
--- a/npm
+++ b/npm
@@ -1,0 +1,88 @@
+#!/bin/sh
+
+# This is a helper script which downloads the `node:alpine` container and runs
+# `npm` commands inside of it, as an alternative to installing and running
+# `npm` on the host.
+
+set -eu
+
+cmd_sh() {
+    # Lots of cockpit developers are using toolbox, which can't recursively run
+    # podman, but flatpak-spawn offers a nice workaround for that.
+    if [ -f /run/.toolboxenv ]; then
+        exec flatpak-spawn --host -- "$0" sh "$@"
+        exit 1
+    fi
+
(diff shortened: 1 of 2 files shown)


npm
```diff
@@ -0,0 +1,88 @@
...
+
+cmd_install() {
+    rm -rf node_modules
+    mkdir node_modules
+    cat 'package.json' | cmd_download | tar --directory node_modules --exclude '.git*' --extract
+    cp node_modules/.package-lock.json package-lock.json
+}
+
+cmd_outdated() {
+    cat 'package.json' | cmd_sh '
```
martinpitt, line 64: Oh good old UUOC :grin:

martinpitt (approved):
I tested this locally a bit on starter-kit, and it works fine. Thanks!

[accidental PR](https://github.com/cockpit-project/starter-kit/pull/566) from `npm-update`

npm
allisonkarlitskaya: This was originally `cat "$*"` to allow specifying a filename, multiple filenames, `-` or nothing.

Then I realized that everyone is only going to want to use it with package.json and using the arguments as a passthrough was more useful.

PR #3272, opened 2022-04-22
npm: replace cat with redirection

allisonkarlitskaya:
Replace two unnecessary uses of `cat` with redirection.

This has an additional benefit, that if we can't open `package.json`, we
won't spawn the container.


martinpitt (approved):
:smile_cat:

PR #3273, opened 2022-04-22
npm: add no-op "prune" command

allisonkarlitskaya:
This doesn't do anything, but is nice for using this script as a drop-in
replacement for npm in $PATH.  Some projects (like cockpit-machines)
call `npm prune` after `npm install`.

### DIFF
--- a/npm
+++ b/npm
@@ -68,6 +68,10 @@ cmd_outdated() {
     '
 }
 
+cmd_prune() {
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
I thought I had included this yesterday, but it seems I forgot to push it, and we merged without it.

martinpitt (approved):

Issue #3295, opened 2022-04-25
grubby on RHEL-9 does not change the default kernel arguments


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2078397

cockpituous:
rhel-9-0
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testCPUSecurityMitigationsEnable (__main__.TestSystemInfo)

DevTools listening on ws://127.0.0.1:10159/devtools/browser/6f219f6e-fd96-4477-ac3a-fc406faee749
[0425/093355.166809:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0425/093355.169222:ERROR:viz_main_impl.cc(188)] Exiting GPU process due to errors during initialization
[0425/093355.173758:ERROR:gpu_init.cc(440)] Passthrough is not supported, GL is disabled, ANGLE is 
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (Not Found)","timestamp":1650879239535.839,"url":"http://127.0.0.2:9291/cockpit/$19beccd35d417d7cc0369d402b2757bd012ba7dfe678c14bdc3b354ddce8e4bc/base1/po.js","networkRequestId":"87185.50"}
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (Not Found)","timestamp":1650879239543.6,"url":"http://127.0.0.2:9291/cockpit/$19beccd35d417d7cc0369d402b2757bd012ba7dfe678c14bdc3b354ddce8e4bc/system/po.js","networkRequestId":"87185.52"}
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (Not Found)","timestamp":1650879239619.093,"url":"http://127.0.0.2:9291/cockpit/$19beccd35d417d7cc0369d402b2757bd012ba7dfe678c14bdc3b354ddce8e4bc/updates/po.js","networkRequestId":"87185.64"}
[...]

PR #3296, opened 2022-04-25
naughty: Add pattern for grubby default args regression in RHEL 9

martinpitt:
Current cockpit integration test does not catch that as it is too
lenient. https://github.com/cockpit-project/cockpit/pull/17279 makes
this tighter and spots this bug.

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2078397

---

Running tests in https://github.com/cockpit-project/cockpit/pull/17279


martinpitt:
Test in https://github.com/cockpit-project/cockpit/pull/17279 [succeeded](https://logs.cockpit-project.org/logs/pull-17279-20220425-090549-b56d753a-rhel-9-0-bots-3296/log.html) and updated the naughty issue in #3295.

jelly (approved):

PR #3277, opened 2022-04-23
Image refresh for fedora-rawhide-boot

cockpituous:
Image refresh for fedora-rawhide-boot
 * [x] image-refresh fedora-rawhide-boot

PR #3301, opened 2022-04-27

cockpituous:

### DIFF
--- a/images/fedora-rawhide-boot
+++ b/images/fedora-rawhide-boot
@@ -1,1 +1,1 @@
-fedora-rawhide-boot-d176981b0b2f4a3c331db7fa8e227af97642f4343defc3be4d7112bdbad9fdf3.iso
+fedora-rawhide-boot-52f6b3f71008f09c4043d552e51fa15915903ddd8eb96792c7dee9b3ef8127f6.iso
--- a/images/fedora-rawhide-boot
+++ b/images/fedora-rawhide-boot
@@ -1,1 +1,1 @@
-fedora-rawhide-boot-3ec60a0309b4c174240cf8362fc4e3aea08e447313d73f0568c5991dfcd3d775.iso
+fedora-rawhide-boot-d176981b0b2f4a3c331db7fa8e227af97642f4343defc3be4d7112bdbad9fdf3.iso


martinpitt:
@VladimirSlavik, @jkonecny12: So #3277 was not just a random failure, rawhide really does not boot any more. Could this possibly be related to [dropping legacy BIOS support](https://fedoraproject.org/wiki/Changes/DeprecateLegacyBIOS)? I.e. the image now needs to be booted with EFI?

VladimirSlavik:
@M4rtinK you have infra duty, do you think you will have time to look at this?

M4rtinK:
> @VladimirSlavik, @jkonecny12: So #3277 was not just a random failure, rawhide really does not boot any more. Could this possibly be related to [dropping legacy BIOS support](https://fedoraproject.org/wiki/Changes/DeprecateLegacyBIOS)? I.e. the image now needs to be booted with EFI?

Looks like the change in it's current form was rejected:

`As discussed in [today's meeting](https://meetbot.fedoraproject.org/fedora-meeting/2022-04-26/fesco.2022-04-26-17.00.log.html#l-131), I'm processing this change as rejected. FESCo will ask BIOS sig to submit a new change with plans/responsibilities.`

From: https://pagure.io/fesco/issue/2780#comment-794311

So its not it, at least not directly - they would certainly not push the switch before the change has been approved by FESCO but I can imagine possibly some boot related fixed/improvement being pushed as a result of the big discussion and boot tooling investigation triggered by the change proposal.

I'll fetch the image locally and try if it boots on my machine in BIOS & EFI modes. :)

jelly:
Loaded the iso into gnome-boxes (bios boot) and it boots fine but there seems to be some issue with a systemd unit

image

Our boot timeout is 1 minute? So I assume that's why the test fails.

VladimirSlavik:
That might be it, I think - I saw this too but did not think it would be significant. It's somewhat annoying.

Tagging @poncovka who's the infra duty this week...

cockpituous:
Log: https://logs.cockpit-project.org/logs/image-refresh-3301-20220505-080404/

cockpituous:
image-refresh fedora-rawhide-boot done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-rawhide-boot-20220505-080707

allisonkarlitskaya (approved):
Oops.  Obviously this shouldn't be two commits, but it's also harmless, so let's go with it.

VladimirSlavik:
Thank you!

PR #3303, opened 2022-04-27
Move cockpit's deploy-keys.sh script to bots/

allisonkarlitskaya:
...and take another look at how we handle deploying to the same repo from multiple environments.

### DIFF
--- a/github-upload-secrets
+++ b/github-upload-secrets
@@ -122,7 +122,7 @@ def main():
     api = task.github.GitHub()
 
     parser = argparse.ArgumentParser(description='Upload encrypted action secrets to GitHub')
-    parser.add_argument('--repository', '-r', '--receiver', metavar="OWNER/REPO",
+    parser.add_argument('--repository', '-r', '--receiver', metavar="OWNER/REPO", required=True,
                         help="The repository which will receive the secrets; default: repo of current checkout")
     parser.add_argument('-e', '--env', metavar="ENVNAME", required=True,
                         help="Upload secrets to given project environment")
@@ -147,21 +147,11 @@ def main():
     if opts.deploy_to and not opts.ssh_keygen:
         parser.error('--deploy-to can only be used with --ssh-keygen')
 
-    if opts.repository and not re.fullmatch(REPO_RE, opts.repository, re.I):
+    if not re.fullmatch(REPO_RE, opts.repository, re.I):
         parser.error('--repository specifies an invalid org or repository name')
 
     if opts.deploy_to and not re.fullmatch(REPO_RE, opts.deploy_to, re.I):
         parser.error('--deploy-to specifies an invalid repository name')
-
-    if not opts.repository:
-        if not opts.env:
-            # get organization of the current repo, if available
-            try:
-                opts.repository = api.get('/repos/' + api.repo)["organization"]["login"]
-            except KeyError:
-                opts.repository = api.repo
-        else:
-            opts.repository = api.repo
 
     env_path = f"/repos/{opts.repository}/environments/{urllib.parse.quote(opts.env)}"
     secrets_path = f"{env_path}/secrets"
--- a/github-upload-secrets
+++ b/github-upload-secrets
@@ -135,7 +135,7 @@ def main():
     megr.add_argument('--directory',
                       help="Upload given directory with one file per secret")
     megr.add_argument('--ssh-keygen', metavar="SECRET_NAME",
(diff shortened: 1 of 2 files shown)


github-upload-secrets
```diff
@@ -122,7 +122,7 @@ def main():
     api = task.github.GitHub()
 
     parser = argparse.ArgumentParser(description='Upload encrypted action secrets to GitHub')
-    parser.add_argument('--repository', '-r', '--receiver', metavar="OWNER/REPO",
+    parser.add_argument('--repository', '-r', '--receiver', metavar="OWNER/REPO", required=True,
```
martinpitt: Can you please give a rationale in the commit message? This seemed quite useful to me, to avoid duplicating the project name in the deploy-keys.sh script.

github-upload-secrets
```diff
@@ -52,7 +51,7 @@ def encrypt(public_key: str, secret_value: str) -> str:
     return b64encode(encrypted).decode("utf-8")
 
 
-def generate_ssh_key(pubkey):
+def generate_ssh_key():
     """Generates a Ed25519 keypair and outputs public/private in OpenSSH format"""
     """We could also do this, using cryptography.hazmat.primitives:
```
martinpitt, line 55: BTW, it's technically valid, but not really semantically legit to have two subsequent docstrings. The second one should be a comment.

github-upload-secrets
```diff
@@ -77,20 +76,30 @@ def generate_ssh_key(pubkey):
...
 
-    return public, encrypt(pubkey["key"], private)
+    return public, private
 
 
-def read_and_encrypt(filename, pubkey):
-    with open(filename, 'rb') as f:
-        return encrypt(pubkey["key"], f.read().strip())
+def upload_secrets(api, repo, env, secrets, opts):
+    env_path = f"/repos/{repo}/environments/{env}"
```
martinpitt, line 82: I think it was you who talked me into using proper URL encoding, so that envs with funny names would work. No veto, but it's not so complicated to do it right..

setup-deploy-keys
```diff
@@ -49,6 +49,7 @@ deploy_to cockpit-project/cockpit-weblate \
 # cockpit-machines
 deploy_to cockpit-project/cockpit-machines \
     --deploy-from \
+        cockpit-project/cockpit-machines/npm-update/SELF_DEPLOY_KEY \
```
martinpitt, line 52: The commit message for "setup-deploy-keys: add cockpit-machines to node-cache" says "we rename the COCKPIT_DEPLOY_KEY to SELF_DEPLOY_KEY", but this commit didn't actually do that.

martinpitt (changes requested):
Cheers! Please also adjust the [header comment](https://github.com/cockpit-project/bots/blob/deploy-keys/github-upload-secrets#L7), as this is no longer true.

github-upload-secrets
allisonkarlitskaya: We don't use this feature, though.  Take a look at our existing `deploy-keys.sh` files:
 - https://github.com/cockpit-project/cockpit-machines/blob/main/.github/deploy-keys.sh
 - https://github.com/cockpit-project/cockpit/blob/main/.github/deploy-keys.sh

both of those are doing, in one way or another, something like: `--receiver "${ORG}/${THIS}"`

github-upload-secrets
allisonkarlitskaya: I always considered this to be a matter of taste, but after a bit of reading it seems that most sources agree that non-docstring block comments are better done with `#` on each line.

github-upload-secrets
allisonkarlitskaya: The reason it's a bit weird in this case is because we now inject the environment as part of a org/repo/env/secret quad, and the main thing that you'd escape from URLs is the `/` itself, which is no longer possible to inject.

Also: this quad gets subjected to a regexp specifically to check that it's not exotic...

For what it's worth, I intend to kill off `--repository` and `--env` fairly soon, at which point this is all going to be substantially simplified.

github-upload-secrets
martinpitt: Ack. I suppose that was still from the days when we called this manually.

github-upload-secrets
martinpitt: Ack, good enough for sure. Resolving.

martinpitt (approved):
Cheers!

PR #3324, opened 2022-05-03
images: add rawhide image for testing

jelly:
I had a script in my ~ to obtain the latest rawhide image and run it
with with cloud-init so I can easily ssh into it for testing. It feels
like it makes sense to allow anyone to easily spin up a rawhide image
with `./image-create fedora-rawhide`.

 * [x] image-refresh fedora-rawhide

PR #3333, opened 2022-05-05
images: add fedora-rawhide to testmap and image-trigger

allisonkarlitskaya:
See #3324.  We need to propose this from origin.

### DIFF
--- a/image-trigger
+++ b/image-trigger
@@ -30,6 +30,7 @@ REFRESH = {
     "fedora-36": {},
     "fedora-testing": {"refresh-days": 3},
     "fedora-coreos": {},
+    "fedora-rawhide": {},
     "fedora-rawhide-boot": {},
     "ubuntu-2204": {},
     "ubuntu-stable": {},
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
Urgh.  It looks like the cockpituous tests tried to run the VM inside of Github Actions.  Of course that didn't work...

Let's split this up.

allisonkarlitskaya:
> Urgh. It looks like the cockpituous tests tried to run the VM inside of Github Actions. Of course that didn't work...
> 
> Let's split this up.

The fix in #3334 would have prevented this from occurring, but we should probably also figure out a way to prevent the cockpituous test from ever considering new images...

jelly (approved):

allisonkarlitskaya:
For the record, `starter-kit` is working with the new image.  I tested it locally, but also double-checked it with a trivial PR here: https://github.com/cockpit-project/starter-kit/pull/571 and log here: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/pull-571-20220505-110452-3877acb3-fedora-rawhide/log.html

PR #3336, opened 2022-05-06
image-prune: two tweaks to relieve space pressure

allisonkarlitskaya:
I've noticed that some files on Linode are getting pruned a bit early.  That turns out to be due to space pressure.  Here are a pair of fixes that will help resolve that in two different ways.


allisonkarlitskaya:
I've also deleted 5 stale branches from GitHub.

jelly (approved):

jelly:
Nice find, thanks for the fix!

PR #3360, opened 2022-05-10
workflows: Run naughty-prune directly

martinpitt:
Drop naughty-trigger and the possibility to run naughty-prune through a
task on our own CI. The latter mechanism isn't very reliable any more,
and often runs twice now (as in #3337). It is equally or more convenient
to just manually run the workflow on the actions page instead of
creating an issue with a naughty-prune task.

Creating PRs requires similar privileges as creating issues, there is no
need to run this on our own infrastructure.

Explicitly specify the permissions to avoid unnecessary ones. Push the 
branch through SSH so that the unit tests will run on the newly created
PR.

Only run once per week, so that we get the same effect as the previous
`DAYS = 7` in naughty-trigger.

----

For testing I used an [additional debug commit](https://github.com/cockpit-project/bots/commit/7764261e3f2e9e) which renames the workflow back (so that I could trigger it manually from the branch) and slightly lower the "last modified at" time, so that I'd get one stale issue. The [workflow succeeded](https://github.com/cockpit-project/bots/runs/6372438043?check_suite_focus=true) and produced PR #3364


naughty-prune
```diff
@@ -53,10 +53,6 @@ def run(unused, verbose=False, **kwargs):
     # The head
     head = execute("git", "rev-parse", "HEAD").strip()
 
-    # Don't replace this issue with the pull request(s)
-    if "issue" in kwargs:
```
allisonkarlitskaya, line 57: I mean, really, we should start converting these "tasks" scripts into normal Python...

.github/workflows/naughty-prune.yml
```diff
@@ -1,12 +1,16 @@
...
-    - cron: '30 1 * * *'
+    - cron: '30 1 * * 0'
   # can be run manually on https://github.com/cockpit-project/bots/actions
   workflow_dispatch:
 jobs:
   maintenance:
     runs-on: ubuntu-latest
+    permissions:
+      issues: read
+      pull-requests: write
```
allisonkarlitskaya, line 12: I guess this is going to end up pushing a branch and opening a PR, but not triggering the test runs (which are required for merging)...

martinpitt: Ah right, only the LGTM ones. So I suppose we'll need to create a deploy key and a "self" environment for bots.

martinpitt: Reworked, I updated the description for the new test run and resulting PR.

.github/workflows/naughty-prune.yml
```diff
@@ -1,20 +1,31 @@
...
+      statuses: write
     steps:
       - name: Set up secrets
         run: echo '${{ secrets.GITHUB_TOKEN }}' > ~/.config/github-token
 
       - name: Clone repository
         uses: actions/checkout@v2
         with:
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
```
allisonkarlitskaya, line 22: That's not gonna be enough.  The problem is that the action that would trigger the bots to run in this case is not the pushing of the branch, but the opening of the PR.  That's still done with the GitHub API via the GitHub Actions token.

This is why we need to do a separate force-push step in the `npm-update` workflow.

This is all terribly inconvenient...

allisonkarlitskaya (changes requested):

.github/workflows/naughty-prune.yml
martinpitt: Well, it is demonstrably enough -- PR #3364 has all tests, and it gets force-pushed so that the bot can add its `Closes #` line.

martinpitt: BTW, this is exactly the same why https://github.com/cockpit-project/cockpit/blob/main/.github/workflows/weblate-sync-po.yml works. The point of the ssh git push is not for the initial branch creation, but the force-push for the "closes #" which both po-refresh and naughty-prune do (via common code in tasks/).

allisonkarlitskaya (approved):
duh.  Forgot that we do a force-push from that script.  Thanks!

PR #3362, opened 2022-05-10
naughty: copy selinux podman naughty to fedora-36

jelly:
Observed in https://logs.cockpit-project.org/logs/pull-17282-20220509-160521-075a2a9d-fedora-36/log.html#297-2


martinpitt (approved):
[here](https://github.com/cockpit-project/cockpit/pull/17282#issuecomment-1121100489) you said that this affected rhel-9-1 as well? If not, please land as is, thanks!

jelly:
> [here](https://github.com/cockpit-project/cockpit/pull/17282#issuecomment-1121100489) you said that this affected rhel-9-1 as well? If not, please land as is, thanks!

Ah hmm I guess it does not always occur? Strange!

martinpitt:
@jelly : Let's just send another PR if we see it happen on 9.1, shall we?

PR #3365, opened 2022-05-10
image-prune tweaks

allisonkarlitskaya:
- general `image-prune` cleanups
 - simplify the "keepers" algorithm of `image-prune`
 - prune the oldest images first and stop once we have enough free space
 - if we can't create enough free space, fail
 - call `image-prune` from `image-upload` as part of `image-create`: the correct place to prune images is when we want to make space for new ones.

 * [x] image-refresh cirros

### DIFF
--- a/image-prune
+++ b/image-prune
@@ -187,8 +187,8 @@ class S3ImageStore(ImageCache):
 
 def main():
     parser = argparse.ArgumentParser(description='Prune downloaded images')
-    parser.add_argument("--debug", action="store_true", help="Enable debugging output")
-    parser.add_argument("--force", action="store_true", help="Delete images even if they aren't old")
+    parser.add_argument("--debug", '-d', action="store_true", help="Enable debugging output")
+    parser.add_argument("--force", '-f', action="store_true", help="Delete images even if they aren't old")
     parser.add_argument("-n", "--dry-run", dest="dryrun", action="store_true",
                         help="Don't actually delete images and links")
     parser.add_argument("-c", "--checkout-only", dest="checkout_only", action="store_true",
--- a/image-prune
+++ b/image-prune
@@ -105,7 +105,7 @@ def get_keepers(offline=False, checkout_only=False):
 
 class ImageCache:
     def list_files(self) -> Iterable[Tuple[str, float]]:
-        """yields tuples of (basename, mtime)"""
+        """yields tuples of (mtime, basename)"""
         raise NotImplementedError
 
     def delete_file(self, filename: str):
@@ -119,7 +119,8 @@ class ImageCache:
     def prune(self, keepers, force=False, dryrun=False):
         expiry_threshold = time.time() - IMAGE_EXPIRE * 86400
 
-        for image, mtime in self.list_files():
+        # Sort by mtime, oldest first.
+        for mtime, image in sorted(self.list_files()):
             if not any(image.endswith(ext) for ext in ['.iso', '.qcow2', '.partial']):
                 logger.debug('Skipping file %s with unknown extension', image)
                 continue
@@ -146,7 +147,7 @@ class LocalImageDirectory(ImageCache):
     def list_files(self):
         for entry in os.scandir(self.directory):
             if entry.is_file(follow_symlinks=False):
-                yield entry.name, entry.stat().st_mtime
+                yield entry.stat().st_mtime, entry.name
 
     def delete_file(self, filename):
         os.unlink(os.path.join(self.directory, filename))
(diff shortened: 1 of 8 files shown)


image-prune
```diff
@@ -24,125 +24,88 @@ import argparse
...
+        patterns = {'HEAD:refs/heads/head', 'refs/heads/*:refs/heads/origin/*'}
+        patterns.update(f'refs/pull/{nr}/head:refs/heads/pr/{nr}' for nr in open_prs)
+        quiet = ['--quiet'] if not logger.isEnabledFor(logging.DEBUG) else []
+        git('fetch', *quiet, f'https://github.com/{api.repo}', *patterns, cwd=tmpdir)
+
+        yield from get_images_in_commit('head', cwd=tmpdir)
+        yield from get_images_in_branches('head', 'refs/heads/origin/*', 'refs/heads/pr/*', cwd=tmpdir)
+
+
+def get_local_images(branches=True):
```
martinpitt, line 94: Not used anywhere; possibly missing fallback for `--offline`? Or just a leftover?

image-prune
```diff
@@ -24,125 +24,88 @@ import argparse
...
+def get_remote_images():
+    """Returns all images currently used by PRs and origin branches on GitHub"""
+
+    api = github.GitHub()
+    logger.debug('Querying open PRs on %s', api.repo)
+    open_prs = [pr['number'] for pr in api.pulls()]
+    logger.debug('Open PRs: %s', open_prs)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        git('init', cwd=tmpdir)
```
martinpitt, line 82: This could use a comment about the "why?". I suspect you do this to avoid cluttering the local checkout with lots of extra fetches, and avoid piling up cruft in .git?

martinpitt, line 82: Please add `-b main` (any branch name really), to avoid the unsightly message:
```
❱❱❱ ./image-prune -n 
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>

```

martinpitt (changes requested):
Thanks! That's some nice cleanup.

martinpitt:
FTR, I don't get this [weird unit test error](https://github.com/cockpit-project/bots/runs/6373296608?check_suite_focus=true) locally. I retried it.

image-prune
allisonkarlitskaya: I have the config option set, so I didn't see this locally.  Added.

allisonkarlitskaya: I added a comment.

image-prune
allisonkarlitskaya: Thanks!  Just a leftover.  It proved to be trivial enough to just implement directly in `get_keepers()`.

image-upload
```diff
@@ -113,6 +114,13 @@ def main():
         success = False
         for store in stores:
             dest = urllib.parse.urljoin(store, basename)
+
+            if args.prune_s3 and store in S3_STORES:
+                try:
+                    subprocess.check_call([f'{BOTS_DIR}/image-prune', '--s3', store])
+                except subprocess.CalledProcessError:
+                    continue
```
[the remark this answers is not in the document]
allisonkarlitskaya: I actually agree.  Changed.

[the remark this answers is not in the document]
allisonkarlitskaya: The reason I did this in the first place is that there is a potential that the key is missing.  I moved the prune to inside the `upload()` function now, where we check the key.

allisonkarlitskaya:
> FTR, I don't get this [weird unit test error](https://github.com/cockpit-project/bots/runs/6373296608?check_suite_focus=true) locally. I retried it.
> 
> **Update**: This is persistent. Ubuntu 20.04's pyflakes somehow does not like this.

This is caused by the `:=` introduced in `image-prune`.  See PyCQA/pyflakes#564.

cockpituous:
image-refresh in progress on centosci-tasks-7zgx7.
Log: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/image-refresh-3365-20220511-113030/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20220511-113131

PR #3367, opened 2022-05-11
lib: switch to Fedora 36 as default image

allisonkarlitskaya:
...and make a minor adjustment to the testmap.


jelly (approved):

martinpitt:
That was a little too quick -- we have never ran fedora-36/devel or /firefox on our projects. That should have happened on this PR. Well, let's see what the fallout is..

PR #3387, opened 2022-05-17
S3 streamer

allisonkarlitskaya:


s3-streamer
```diff
@@ -0,0 +1,319 @@
...
+            print(response.status, filename)
+
+    def delete(self, filenames):
+        # to do: multi-object delete API
+        for filename in filenames:
+            with s3.urlopen(self.url(filename), method='DELETE') as response:
+                print(response)
+
+
+class Index(Destination):
```
martinpitt, line 94: Constantly maintaining an index.html, from dozens of parallel bots feels error prone and like unnecessary churn. We have both the S3 CLI and our test-results.db for querying available log URLs. I'd rather drop this.

s3-streamer
```diff
@@ -0,0 +1,319 @@
...
+        for entry in os.scandir(self.path):
+            if not self.destination.has(entry.name) and entry.is_file(follow_symlinks=False):
+                with open(entry.path, 'rb') as file:
+                    data = file.read()
+                self.destination.write(entry.name, data)
+
+
+class ChunkedUploader:
+    SIZE_LIMIT = 1000000  # 1MB
+    TIME_LIMIT = 10       # 10s
```
martinpitt, line 145: This was probably fine for testing, but for production one minute feels enough. Our logs are not that huge, so of these two it'll be the time limit which has the practical effect.

s3-streamer
```diff
@@ -0,0 +1,319 @@
...
+
+    def __init__(self, index, filename, encoding=None):
+        self.input_decoder = codecs.getincrementaldecoder(encoding or locale.getpreferredencoding())()
```
martinpitt, line 148: I'd rather not depend on the locale here. This has bitten us all too often in the past, with bots running under `C` locale and regurgitating UTF-8 input.

As I said, I'd rather drop that whole decode → encode dance completely and just handle byte streams, with breaking on the last `b\n` of a read block (as sending half a line isn't that useful anyway). This avoids potential locale crashes, hangs with internal buffers of the encoders, and saves CPU cycles on top of that. You seemed very eager to do the encode/decode, but let's at least do a fixed `utf-8` here.

martinpitt (commented):
Cheers! Some initial questions/comments.

s3-streamer
allisonkarlitskaya: Each bot generates a unique subdirectory name and writes to it.  We never try to maintain an index.html for the toplevel.

All the same, I may indeed drop this code...

s3-streamer
allisonkarlitskaya: Let's chat about this.

My priorities here are:
 - it needs to be impossible to send a partial multi-byte character
 - i also want to never ever send non-utf8 data to the client.  that means we need to detect invalid sequences and transform them, anyway.
 - we should see all of the available output in the event of a hang.

The locale is going to be utf-8 in any reasonable universe.  I use the user's preferred locale out of a desire to be strictly correct.  I can turn that into an assert, if you like.

s3-streamer
allisonkarlitskaya: Keep in mind that this doesn't mean that we send a block every 10 seconds.  It means that after we see some data, we wait for more data for 10 seconds before sending the accumulated data as a block.

I thought we could start low and increase it if we find that we're sending too much data...

s3-streamer
allisonkarlitskaya: I want to come up with a replacement before I drop this, though.  In particular, I want to send a static `index.html` which queries the REST API for the file list and generates an index on the client side.  That shouldn't be hard, but it's not on the critical path to landing this, so I'd rather do it later.

martinpitt (changes requested):
Thanks, this is exciting! Nothing that blocks first experiments in a manual container, but let's resolve the discussion threads before landing.

martinpitt (approved):

PR #3391, opened 2022-05-17
testmap: Move cockpit-machines to rhel-8-7

martinpitt:
Tests were fixed in https://github.com/cockpit-project/cockpit-machines/pull/708


allisonkarlitskaya (approved):
Thank you :+1:

PR #3399, opened 2022-05-18
s3-streamer improvements

allisonkarlitskaya:
As discussed yesterday, after the first round of testing:
 - set Content-Encoding for .log.gz files
 - write hostname and command into the start of the log
 - sort index.html
 - insist on utf-8 locale, but accept non-utf-8 input

### DIFF
--- a/s3-streamer
+++ b/s3-streamer
@@ -84,10 +84,14 @@ class S3Destination(Destination):
         raise NotImplementedError('use Index')
 
     def write(self, filename, data):
+        content_type, content_encoding = mimetypes.guess_type(filename)
         headers = {
-            'Content-Type': mimetypes.guess_type(filename)[0] or 'text/plain',
+            'Content-Type': content_type or 'text/plain',
             s3.ACL: s3.PUBLIC
         }
+        if content_encoding:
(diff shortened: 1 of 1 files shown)


s3-streamer
```diff
@@ -164,8 +166,9 @@ class ChunkedUploader:
         self.pending = b''
         self.send_at = 0  # Send the first write immediately
 
-    def init(self):
-        self.destination.write(f'{self.filename}.chunks', json.dumps([]).encode('ascii'))
+    def start(self, data):
+        # Send the initial data immediately, to get the chunks file written out.
+        self.append_block(data.encode('utf=-8'))
```
martinpitt, line 175: That looks like a typo. Does that actually work?

martinpitt (changes requested):
Thanks! Can you please also amend the `status.post('pending', github.TESTING)` to include the hostname it's running on?

s3-streamer
allisonkarlitskaya: That looks like a typo.

allisonkarlitskaya: For the record, I did test it.

martinpitt (approved):
Cheers! Please land and do another test run.

PR #3400, opened 2022-05-18
test-scan tweaks

allisonkarlitskaya:
Some improvements to `test-scan` to increase its usability as a local tool for queuing up test runs:

 - look for the secrets in the user's `XDG_RUNTIME_DIR` (aka "Martin mode")
 - make the `-c` flag more powerful: the given contexts are always considered, even if they're not present in the PR
 - drop the `-f`-bomb: ignore all checks and enqueue unconditionally.  Temper this potentially-explosive option by only allowing it to be used with a tightly-restricted scope.

### DIFF
--- a/task/distributed_queue.py
+++ b/task/distributed_queue.py
@@ -84,6 +84,13 @@ class DistributedQueue(object):
         if no_amqp:
             raise ImportError('pika is not available')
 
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Neat, thanks!

PR #3401, opened 2022-05-18
s3-streamer: add a prettier header

allisonkarlitskaya:
The first line in the log is turned into the heading of the page, so
make it as useful as possible.

While we're at it, add some extra info below.

### DIFF
--- a/s3-streamer
+++ b/s3-streamer
@@ -25,6 +25,7 @@ import logging
 import mimetypes
 import os
 import platform
+import re
 import shlex
 import subprocess
 import tempfile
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya:
erm.  That didn't turn out like I expected:

allisonkarlitskaya:
OK.  Turns out that dedent() doesn't remove the leading newline in a case like

```python
    dedent(f"""
        First line here
    """)
````

So the first line was blank.

martinpitt (approved):
Cheers!

Issue #2787, opened 2022-01-10
avc: denied { ioctl } for comm="iptables" on containers


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2031022
For rhel-9-1: https://bugzilla.redhat.com/show_bug.cgi?id=2089257
For C8S: https://bugzilla.redhat.com/show_bug.cgi?id=2106396

cockpituous:
fedora-coreos
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testBondMac (__main__.TestNetworkingMAC)
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.
NAME                UUID                                  TYPE      DEVICE 
Wired connection 1  ba9e34a6-f27d-3b79-9e51-e2616c5ea40c  ethernet  ens14  
mcast1              4bf68e6e-d2d4-345b-aa95-2259ef3bc196  ethernet  ens15  


DevTools listening on ws://127.0.0.1:9315/devtools/browser/79f44b72-394e-4628-b64a-12628361bb48
[0110/094851.301165:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: /usr/lib64/chromium-browser/swiftshader/libGLESv2.so: cannot open shared object file: No such file or directory
[0110/094851.309324:ERROR:viz_main_impl.cc(161)] Exiting GPU process due to errors during initialization
[0110/094851.322643:ERROR:gpu_init.cc(453)] Passthrough is not supported, GL is disabled, ANGLE is 
> error: Modal: Specify at least one of: title, aria-label, aria-labelledby.
> warning: grep: /sys/class/dmi/id/power/autosuspend_delay_ms: Input/output error

52:54:01:00:00:03 -> ens7
52:54:01:00:00:03 -> ens7
52:54:01:00:00:04 -> ens8
52:54:01:00:00:04 -> ens8
[...]

PR #3404, opened 2022-05-18
images: remove some dependencies from Fedora

allisonkarlitskaya:
At 228MB installed size, glibc-all-langpacks is the single largest thing
in our images.  Remove it.

qemu-char-spice has a large number of heavy dependencies, including mesa
drivers (>100MB) and llvm-libs (109MB).

ansible has a large number of standard and community-contributed scripts
in the default install (327MB), none of which we use.  Replace it with
ansible-core (11MB).

 * [x] image-refresh fedora-36


cockpituous:
Log: https://logs.cockpit-project.org/logs/image-refresh-3404-20220518-161515/

cockpituous:
image-refresh fedora-36 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-36-20220518-163737

allisonkarlitskaya:
2.36GB → 2.19GB.  Not a huge game changer, but progress.

mvollmer:
> The failure is presumably caused not by the change here, but by an unrelated change in the image.

Looks like #2787 has made its way into fedora-36.

martinpitt (approved):
Cheers! I validated that even on this Fedora 36 image I can use cockpit in all supported languages. Some strings which come from the OS won't be translated of course, but that doesn't affect `check-pages` -- the point of this test is to make sure that our own translations work and don't cause a crash.

Also, the "select all languages" test only runs on RHEL, so it's not affected by Fedora image changes.

PR #3413, opened 2022-05-20
setup-deploy-keys: move website to its own environment

allisonkarlitskaya:
The release workflow in Cockpit currently runs as a monolithic job under
the "release" environment.

cockpit-project/cockpit#17365 changes that, so that each part of the
process runs in its own environment and the "release" environment
no longer exists.

Install the deploy key for the website into the "website" environment
instead.

I didn't run this yet, and we shouldn't run it until we land cockpit-project/cockpit#17365.

### DIFF
--- a/setup-deploy-keys
+++ b/setup-deploy-keys
@@ -53,7 +53,7 @@ deploy_to cockpit-project/cockpit-weblate \
 
 deploy_to cockpit-project/cockpit-project.github.io \
     --deploy-from \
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Ack, thanks! This part is easy to approve, but as you said, let's land it together with the cockpit PR. Adding blocked tag for now.

allisonkarlitskaya:
...and I ran it.

PR #3417, opened 2022-05-23
Reorganize stores on their hosting/TLS cert type

martinpitt:
Currently, image-upload conflates the store type (image vs. s3) with the 
certificate type (official or our own ca.pem). Let's decouple this.

All our stores are S3 now, so we can drop the 's3' type in lib/stores,
and drop the handling of the cockpit/image store types. Repurpose types
"public" and "redhat" for how they are hosted: On a public cloud with an
official/trusted TLS certificate, or hosted ourselves inside the Red Hat 
VPN with our Cockpituous CA. Supply the cockpituous CA on all URL 
openers when talking to a redhat store.

This also moves our old cockpit/image image store to our brand new S3 store on cockpit-11. This is being added by https://github.com/cockpit-project/cockpituous/pull/490, but that needs this PR.

I am not yet 100% satisfied with this -- e.g. I  don't like the repeated `PUBLIC_STORES + REDHAT_STORES`, this could be made more elegant. However, it's not too bad IMHO, so this can be done in a follow-up.


cockpituous:
image-refresh in progress on 2-cockpit-8.
Log: https://logs.cockpit-project.org/logs/image-refresh-3417-20220523-063636/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20220523-063636

cockpituous:
image-refresh in progress on 4-cockpit-8.
Log: https://logs.cockpit-project.org/logs/image-refresh-3417-20220523-063636/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20220523-063737

cockpituous:
image-refresh in progress on centosci-tasks-q8lrz.
Log: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/image-refresh-3417-20220523-063737/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20220523-063838

cockpituous:
image-refresh in progress on centosci-tasks-wl5d6.
Log: https://logs-https-frontdoor.apps.ocp.ci.centos.org/logs/image-refresh-3417-20220523-063838/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20220523-063838

martinpitt:
Image refresh [successful](https://logs.cockpit-project.org/logs/image-refresh-3417-20220523-063636/log), the image is on all our stores. I delete the branch again, we don't actually need to land this.

image-upload
```diff
@@ -28,46 +27,42 @@ import urllib.parse
...
-    else:
-        # No credentials?  That's not going to work...
-        sys.stderr.write(f"image-upload: no credentials for {dest}")
-        return False
+    curl_arg = s3.sign_curl(url, method='PUT', headers=headers, checksum=hash_value)
 
     # Only prune after the above checks, to ensure that we have credentials.
     if prune:
-        subprocess.check_call([f'{BOTS_DIR}/image-prune', '--s3', f'https://{url.hostname}/'])
+        subprocess.check_call([f'{BOTS_DIR}/image-prune', '--s3', store])
```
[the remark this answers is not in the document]
martinpitt: It does, I tested image pruning explicitly and also through `image-upload --s3-prune`. See the change in lib/s3.py, that does the same "apply ca.pem for REDHAT_STORES" approach as image-upload/download.

martinpitt:
Rebased for fixing the conflict, and re-tested with `./image-upload --prune-s3 --store https://cockpit-11.apps.cnfdb2.e2e.bos.redhat.com/images/ cirros`.

lib/s3.py
```diff
@@ -107,13 +109,19 @@ def urlopen(url: urllib.parse.ParseResult, method='GET', headers={}, data=b'') -
     """Same as sign_request() but calls urlopen() on the result"""
     headers = sign_request(url, method, headers, hashlib.sha256(data).hexdigest())
     request = urllib.request.Request(url.geturl(), headers=headers, method=method, data=data)
+
+    # is this a redhat store?
+    if any([s for s in REDHAT_STORES if url.netloc in s]):
```
allisonkarlitskaya, line 114: 1. This doesn't need to be a list comprehension.  A generator is better here.
2. Just put the condition at the front.

Like this:

```
    if any(url.netloc in store for store in REDHAT_STORES):
        ....
```

image-upload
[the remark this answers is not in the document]
allisonkarlitskaya: I missed this, thanks.

In the name of DRY and keeping design decisions hidden within a single module, I think it would be better to also do one of:

1) add the `curl --cacert` logic to the curl helper in `lib.s3`, possibly moving over to always using the "new" signing for curl requests, and dropping the old code

2) port everything to use the `s3.urlopen()` path.  I mostly kept using `curl` because I thought we cared about the progress bar, but since you got rid of that, we don't really have a reason anymore.

allisonkarlitskaya (changes requested):

lib/s3.py
allisonkarlitskaya: This would also make it absolutely trivial to add to the `curl` helper, if you wanted to go that route (which might be useful for the CLI interface of `s3.py` still).

martinpitt: Cleaned this up.

image-upload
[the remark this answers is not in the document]
martinpitt: I separated and cleaned up the CA handling from the hosting type now.

cockpituous:
image-refresh in progress on 1-ci-srv-03.
Log: https://logs.cockpit-project.org/logs/image-refresh-3417-20220524-103333/

cockpituous:
image-refresh in progress on 1-ci-srv-03.
Log: https://logs.cockpit-project.org/logs/image-refresh-3417-20220524-103333/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20220524-103434

lib/network.py
```diff
@@ -27,6 +28,38 @@
...
+def get_host_ca(hostname):
+    '''Return custom CA that applies to the given host name.
+
+    Self-hosted infrastructure uses CA_PEM, while publicly hosted infrastructure ought to have
+    an officially trusted TLS certificate. Return None for these.
+    '''
+    # strip off port
+    hostname = hostname.split(':')[0]
+
+    if any(hostname.endswith(domain) for domain in CA_PEM_DOMAINS):
```
allisonkarlitskaya, line 49: This isn't exactly correct.  It would also match `blogs.cockpit-project.org`, for example.

I'm willing to let it slide, though :)

allisonkarlitskaya (approved):
This is wonderful.  The commit to add the helper to get the cacert or tls context is really great.  I wonder if it would make sense to have some kind of a similar helper for curl so you could say:

curl_cmd.extend(network.get_curl_ca_arg(...)) or so?

You can get rid of `lib/stores` and turn that back into `stores.py` again, as I mentioned, if you want.

One more thought for some extra factorizing: what if we had some `lib.network.is_accessible()` helper that knew that `*.e2e.bos.redhat.com` is behind the VPN?  That would let us drop the responsibility for carrying this logic in the stores list.  I'm actually not sure if I like this idea or not: just thought I'd throw it out there.

lib/network.py
martinpitt: This nags me too much to let it slip -- I had it correct the first time, and then oversimplified it. Fixed.

PR #3422, opened 2022-05-24
WIP: 4-cockpit-9: [no-test] Image refresh for rhel-8-7

cockpituous:
Image refresh for rhel-8-7
 * [ ] image-refresh rhel-8-7


martinpitt:
Ah, I was looking forward to this -- cockpit still has tons of distropkg special cases from the time when we had base-2 units. We can drop *all* of them now, this should make this PR green again.

mvollmer:
Clean up happening in https://github.com/cockpit-project/cockpit/pull/17379

mvollmer (approved):

PR #3429, opened 2022-05-25
image-download: Avoid walrus operator (quick alternative to #3430)

martinpitt:
image-download is called in cockpit-machines tests, which also run on
RHEL/CentOS 8. This has a too old Python to recongnize the `:=`
operator, so replace it.

---

See [this machines c8s failure](https://artifacts.dev.testing-farm.io/98ac3343-6c32-4374-b3e8-c48312f7f750/)

### DIFF
--- a/image-download
+++ b/image-download
@@ -86,7 +86,8 @@ def find(name, stores, latest, quiet):
         url = urllib.parse.urlparse(urllib.parse.urljoin(store, name))
 
         args = []
-        if ca := get_host_ca(url.netloc):
+        ca = get_host_ca(url.netloc)
+        if ca:
             args += ['--cacert', ca]
 
         # First, check if this is an S3 store for which we have a key


martinpitt:
Integration test is blocked by https://github.com/cockpit-project/cockpituous/pull/490

martinpitt:
I just sent #3430 which will also fix this (but conflict). Depending on how fast this goes in, we may want to quick-land this instead, and I'll rebase #3430 then.

mvollmer:
LGTM, but `test / cockpituous` seems to have trouble, and it would be good to run it as it would probably execute image-download, no?  I'll wait a bit to see whether the network weather clears up.

martinpitt:
@mvollmer : bots tests are currently broken until https://github.com/cockpit-project/cockpituous/pull/490 lands. I agree, while I did test this locally, we should let tests pass for infra changes -- that's why we have them.

mvollmer:
Blocked on https://github.com/cockpit-project/cockpituous/pull/490 to get `test/cockpituous` green.

allisonkarlitskaya:
We don't need to wait for bots to merge something like this...

martinpitt:
> We don't need to wait for bots to merge something like this...

Well, it does modify image-download, and tests cover that. In general I do.

PR #3430, opened 2022-05-25
store/CA followup fixes

### LINKED ISSUE #3424: store/CA follow-ups
This is wonderful.  The commit to add the helper to get the cacert or tls context is really great.  I wonder if it would make sense to have some kind of a similar helper for curl so you could say:

`curl_cmd.extend(network.get_curl_ca_arg(...))` or so?

You can get rid of `lib/stores` and turn that back into `stores.py` again, as I mentioned, if you want.

One more thought for some extra factorizing: what if we had some `lib.network.is_accessible()` helper that knew that `*.e2e.bos.redhat.com` is behind the VPN?  That would let us drop the responsibility for carrying this logic in the stores list.  I'm actually not sure if I like this idea or not: just thought I'd throw it out there.

_Originally posted by @allisonkarlitskaya in https://github.com/cockpit-project/bots/pull/3417#pullrequestreview-983101916_

martinpitt:
Some leftovers from https://github.com/cockpit-project/bots/pull/3417#pullrequestreview-983101916

This also drops the walrus operator which [fails on RHEL 8](https://artifacts.dev.testing-farm.io/98ac3343-6c32-4374-b3e8-c48312f7f750/), and thus this replaces #3429.

Fixes #3424

 - Tests blocked on https://github.com/cockpit-project/cockpituous/pull/490


image-upload
```diff
@@ -42,10 +42,7 @@ def upload(store, source, public, prune=False):
         return False
 
     # Start building the command
-    cmd = ["curl", "--no-progress-meter", "--fail", "--upload-file", source]
-
-    if ca := get_host_ca(url.netloc):
-        cmd += ['--cacert', ca]
+    cmd = ["curl", "--no-progress-meter", "--fail", "--upload-file", source] + get_curl_ca_arg(url.netloc)
```
allisonkarlitskaya, line 45: I like this syntax: `["curl", ..., *get_curl_ca_arg(url.netloc)]`

image-download
```diff
@@ -85,9 +85,7 @@ def find(name, stores, latest, quiet):
     for store in stores:
         url = urllib.parse.urlparse(urllib.parse.urljoin(store, name))
 
-        args = []
-        if ca := get_host_ca(url.netloc):
-            args += ['--cacert', ca]
+        args = get_curl_ca_arg(url.netloc)
```
allisonkarlitskaya, line 88: I wonder if we should make this part of `s3.sign_curl()` to better match the fact that `s3.urlopen()` takes care of this for itself.

This would have the unfortunate side-effect of meaning that we need to "do it ourselves" when we don't have an S3 key.  Which brings the next question: perhaps, instead of failing, calling `s3.urlopen()` or `s3.sign_curl()` should simply do unsigned requests.

That's a follow-up, and I'm happy to take care of it.

allisonkarlitskaya (dismissed):
This can go now, if you want.  Happy to do another round, as well.

As for the followup ideas:
 - move to hostname based network detection
 - move ssl certificate handling into `s3.sign_curl`

Either one of us could do those?

image-upload
martinpitt: Sure :-) Done

allisonkarlitskaya (approved):
Thanks :)

martinpitt:
Unconflicted, and updated the array syntax.

> move to hostname based network detection

Not sure what you mean by that -- changing how `redhat_network()` works?

> move ssl certificate handling into s3.sign_curl

Ack, let's do this as a follow-up. Want to give it a go? I feel like you already have this worked out in your head.

PR #3433, opened 2022-05-25
s3-streamer: don't append a slash to coverage report

jelly:
The coverage report link is a link to /Coverage/index.html so adding a
slash will make the link invalid.

### DIFF
--- a/lib/s3-html/log.html
+++ b/lib/s3-html/log.html
@@ -189,8 +189,8 @@ let link_patterns = [
     },
     {
         "label": "coverage",
-        "pattern": "Code coverage report in ([A-Za-z0-9\\-\\.]+)$",
-        "url": "$1/"
+        "pattern": "Code coverage report in ([A-Za-z0-9\\-\\.\/]+)$",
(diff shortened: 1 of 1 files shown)


lib/s3-html/log.html
```diff
@@ -190,7 +190,7 @@
     {
         "label": "coverage",
         "pattern": "Code coverage report in ([A-Za-z0-9\\-\\.]+)$",
```
allisonkarlitskaya, line 192: I guess this isn't going to work: that regexp doesn't match `/`.

allisonkarlitskaya (changes requested):
Since I look at it the second time, I'll comment:  the commit message is wrong that it says `/Coverage/` and not `Coverage/`

lib/s3-html/log.html
jelly: Yes, thanks for catching that!

allisonkarlitskaya (approved):
Might be nice to test it locally, but looks good to me, and if it breaks, it's not a big deal.

PR #3437, opened 2022-05-27

cockpituous:
 * https://github.com/cockpit-project/cockpit-podman/pull/996


martinpitt:
Now we are back to the status quo from before commit fb0437f5045cd03346d16b282235ce7bc7a233b4. @marusak , @jelly , I think my original reasoning was wrong. The intermediate image is not from the cockpit/ws rebuild, but from `make vm` re-pulling new container images, and the old ones then end up as untagged. It seems I just forgot to drop that container image pull after adding them to the image. I sent https://github.com/cockpit-project/cockpit-podman/pull/996

martinpitt (approved):

PR #3449, opened 2022-05-31
setup-deploy-keys: add cockpit-podman

jelly:
Add the deploy keys from cockpit-podman/.github/deploy-keys.sh.

### DIFF
--- a/setup-deploy-keys
+++ b/setup-deploy-keys
@@ -73,6 +73,16 @@ deploy_to cockpit-project/cockpit-machines-weblate \
     --deploy-from \
         cockpit-project/cockpit-machines/cockpit-machines-weblate/DEPLOY_KEY
 
+# cockpit-podman
+deploy_to cockpit-project/cockpit-podman \
+    --deploy-from \
+        cockpit-project/cockpit-podman/npm-update/SELF_DEPLOY_KEY \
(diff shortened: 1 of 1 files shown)


setup-deploy-keys
```diff
@@ -73,6 +73,20 @@ deploy_to cockpit-project/cockpit-machines-weblate \
...
 
+# cockpit-podman
+deploy_to cockpit-project/cockpit-podman \
+    --deploy-from \
+        cockpit-project/cockpit-podman/npm-update/SELF_DEPLOY_KEY \
+        cockpit-project/cockpit-podman/self/DEPLOY_KEY
+
+deploy_to cockpit-project/cockpit-podman-dist \
+    --deploy-from \
+        cockpit-project/cockpit-podman/cockpit-podman-dist/DEPLOY_KEY
```
martinpitt, line 84: c-podman does not use this yet. Are you going to? I feel it's totally worth adding node_modules/ git caching, but webpack for podman is so fast, I think it's not worth adding the dist caching overhead.

jelly: Ok, let's skip it for now then.

martinpitt (approved):
Thanks! I'll run the script after landing, so that the keys get deployed.

PR #3452, opened 2022-05-31
s3: Break out of both loops in log.html

mvollmer:
When a log chunk is missing, we should go straight to reading the full
file.

### DIFF
--- a/lib/s3-html/log.html
+++ b/lib/s3-html/log.html
@@ -375,6 +375,8 @@ function sleep(seconds) {
     return new Promise(resolve => setTimeout(resolve, 1000 * seconds));
 }
 
+class NotFoundError extends Error { };
+
 async function fetch_from(url, offset) {
     for (let attempts = 0; attempts < 10; attempts++) {
         const options = {};
@@ -393,9 +395,11 @@ async function fetch_from(url, offset) {
                     return text;
                 }
             } else {
-                return null;
+                throw new NotFoundError;
             }
         } catch (error) {
+            if (error instanceof NotFoundError)
+                throw new NotFoundError;
             const delay = 2 ** attempts;
             console.log(`Failed to fetch ${url}.  Waiting ${delay}.`);
             await sleep(delay);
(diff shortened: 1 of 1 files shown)


lib/s3-html/log.html
```diff
@@ -406,38 +406,45 @@
     return null;
 }
 
+class StopError extends Error { };
```
allisonkarlitskaya, line 409: I think if we get any hard error (basically the branch where we return `null` in `fetch_content()` then we should throw this exception from that point (possibly with a comment to the effect).

That would let us remove the conditions on calling `fetch_content()`.  Probably we'd then rewrite the `while()` loop as some infinite iteration or so.

allisonkarlitskaya (changes requested):
Thanks!

The toplevel `try`/`catch` structure is pretty much exactly right, but pushing the `throw` down a layer will let us clean things up some.

lib/s3-html/log.html
```diff
@@ -413,31 +417,31 @@
...
-            const chunk_end = chunk_start + chunk_size;
+            for (const chunk_size of chunks) {
+                const chunk_end = chunk_start + chunk_size;
 
-            if (bytes < chunk_end) {
-                const offset = bytes - chunk_start;
-                if ((chunk = await fetch_from(`${filename}.${chunk_start}-${chunk_end}`, offset))) {
+                if (bytes < chunk_end) {
+                    const offset = bytes - chunk_start;
+                    chunk = await fetch_from(`${filename}.${chunk_start}-${chunk_end}`, offset);
```
allisonkarlitskaya, line 430: `const chunk`?

lib/s3-html/log.html
```diff
@@ -413,31 +417,31 @@
     let content = '';
     let bytes = 0;
 
-    let chunks;
-    while ((chunks = JSON.parse(await fetch_from(`${filename}.chunks`)))) {
-        let chunk_start = 0;
+    try {
+        while (true) {
+            let chunks = JSON.parse(await fetch_from(`${filename}.chunks`));
```
allisonkarlitskaya, line 422: `const chunks`?

allisonkarlitskaya (changes requested):

allisonkarlitskaya (approved):

PR #3455, opened 2022-05-31
naughty: Add new pattern for podman not sending new events

marusak:
https://github.com/cockpit-project/cockpit-podman/pull/1001 introduces new test that also hits this issue and thus needs to be covered by naughty

### DIFF
--- a/naughty/debian-testing/2463-no-pod-events-2
+++ b/naughty/debian-testing/2463-no-pod-events-2
@@ -1,0 +1,3 @@
+testCreateContainerInPodSystem
+*
+wait_js_cond(ph_is_present("#table-system_pod .create-container-in-pod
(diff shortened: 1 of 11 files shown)


naughty/debian-testing/2463-no-pod-events-2
```diff
@@ -0,0 +1 @@
+b.click("#table-system_pod .create-container-in-pod")
```
martinpitt, line 1: This is too thin. Can you please add some context like the affected test name,`Traceback:` and the actual exception, so that this doesn't catch unrelated errors?

martinpitt (changes requested):

naughty/debian-testing/2463-no-pod-events-2
marusak: better?

Issue #3456, opened 2022-05-31
rhel-9-1: Latest kmod-kvdo does not work with the latest RHEL-9.1.0 kernel 


mvollmer:
https://bugzilla.redhat.com/show_bug.cgi?id=2060486

cockpituous:
rhel-9-1
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testVdoMissingPackages (__main__.TestStoragePackagesVDO)
Failed to kill unit packagekit.service: Unit packagekit.service not loaded.
Failed to reset failed state of unit packagekit.service: Unit packagekit.service not loaded.
rm: cannot remove '/etc/systemd/system/dnf-automatic*': No such file or directory

DevTools listening on ws://127.0.0.1:9747/devtools/browser/2dae3916-7072-4064-a3eb-bf623d6d6450
[0531/154837.384727:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0531/154837.386629:ERROR:viz_main_impl.cc(188)] Exiting GPU process due to errors during initialization
[0531/154837.393479:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0531/154837.395338:ERROR:viz_main_impl.cc(188)] Exiting GPU process due to errors during initialization
[0531/154837.397970:ERROR:gpu_init.cc(446)] Passthrough is not supported, GL is disabled, ANGLE is 
> warning: failed to poll tuned 
  Creating devices file /etc/lvm/devices/system.devices
> error: Text input: Text input requires either an id or aria-label to be specified
> warning: Error: vdo is not available from any repository.
> warning: Error creating volume: Process reported exit code 3: modprobe: FATAL: Module kvdo not found in directory /lib/modules/5.14.0-101.el9.x86_64
  /usr/sbin/modprobe failed: 1
  vdo: Required device-mapper target(s) not detected in your kernel.
  Run `lvcreate --help' for more information.

Traceback (most recent call last):
  File "/usr/lib64/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
[...]

PR #3488, opened 2022-06-08
Image refresh for rhel-9-1

cockpituous:
Image refresh for rhel-9-1
 * [x] image-refresh rhel-9-1


skobyda:
Tried to reproduce the failing test with `/test/verify/check-storage-vdo TestStorageVDO.testVdo`. It seems that as storage volume's size baloons up, cockpit should be subscribed to a signal listening to size and re-rendering the UI to show the actual size. 
However, it seems that re-rendering gets stuck around 0.6GB, instead of showing the expected 1.2GB for the storage volume.
However, if you reload the page, then it will show correct size of 1.2GB. That leads me to believe that maybe storaged stops firing signals which let us now that a size of storage volume changes? Or maybe React doesn't rerender the UI anymore.

marusak:
So the current image (and for some time) were skipped, see for example [this run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17459-20220616-062339-7902cff6-rhel-9-1/log.html)  due to #3456 . It seems to get fixed but we might break it/or it got broken while we were skipping these tests. Looking.

marusak (approved):

PR #3491, opened 2022-06-09
Image refresh for fedora-36

cockpituous:
Image refresh for fedora-36
 * [x] image-refresh fedora-36


martinpitt:
> error: unsupported configuration: storage type 'dir' requires use of storage format 'fat'

@skobyda: I suppose some further libvirt package split happened, and we need to now depend on an extra package? (Both in the image and as c-machines Requires:)

martinpitt:
Actually, I've seen this error in [other PRs](https://cockpit-logs.us-east-1.linodeobjects.com/pull-737-20220609-050531-0f6167b5-fedora-36/log.html) as well, so this is some kind of VM destruction. Needs to be fixed, but not necessarily block this PR.

martinpitt (approved):

PR #3513, opened 2022-06-15
testmap: update subscription-manager entries

ptoscano:
- replace rhel-8-6 with rhel-8-7 for subscription-manager/main, to make
  sure to check backports to RHEL 8 (which are going first to 8.7
- remove rhel-8-7 from _manual, no more needed there
- add rhel-8-7 for subscription-manager/subscription-manager-1.28, as
  that is the target of that branch
- remove rhel-8-6/subscription-manager-1.28 from
  subscription-manager-cockpit, as it will not get in that version
- switch the services trigger to rhel-8-7, as it's the development
  version

The new combinations were tested successfully:
- https://cockpit-logs.us-east-1.linodeobjects.com/pull-3074-20220615-094740-686edabc-rhel-8-7/log.html
- https://cockpit-logs.us-east-1.linodeobjects.com/pull-3072-20220615-094746-4532c6b7-rhel-8-7/log.html

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -139,7 +139,7 @@ REPO_BRANCH_CONTEXT = {
     },
     'candlepin/subscription-manager': {
         'main': [
-            'rhel-8-6',
+            'rhel-8-7',
             'rhel-9-0',
             'rhel-9-1',
             'fedora-36',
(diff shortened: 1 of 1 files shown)


marusak (approved):

lib/testmap.py
```diff
@@ -148,6 +148,7 @@
             'rhel-8-4',
```
marusak, line 148: @ptoscano Are you still testing 8-4 and 8-5?

PR #3531, opened 2022-06-21
Export more info to the test processes

mvollmer:
So that the test run itself might post review comments that contain
links to attached files.

### DIFF
--- a/s3-streamer
+++ b/s3-streamer
@@ -294,8 +294,9 @@ def main():
         index = Index(destination)
         attachments_directory = AttachmentsDirectory(index, tmpdir)
         log_uploader = ChunkedUploader(index, 'log')
+        env = dict(os.environ, TEST_ATTACHMENTS=tmpdir, TEST_ATTACHMENTS_URL=destination.directory)
 
-        with subprocess.Popen(args.cmd, env=dict(os.environ, TEST_ATTACHMENTS=tmpdir),
+        with subprocess.Popen(args.cmd, env=env,
                               stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
                               stdin=subprocess.DEVNULL, pipesize=max_pipe_size) as process:
             # We want non-blocking reads so that we can send attachments and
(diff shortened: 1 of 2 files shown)


s3-streamer
```diff
@@ -295,7 +295,9 @@ def main():
         attachments_directory = AttachmentsDirectory(index, tmpdir)
         log_uploader = ChunkedUploader(index, 'log')
 
-        with subprocess.Popen(args.cmd, env=dict(os.environ, TEST_ATTACHMENTS=tmpdir),
+        with subprocess.Popen(args.cmd, env=dict(os.environ,
+                                                 TEST_ATTACHMENTS=tmpdir,
+                                                 LOGDIR_URL=destination.directory),
```
allisonkarlitskaya, line 300: Let's call it `TEST_ATTACHMENTS_URL`, and please move this to a separate statement.

allisonkarlitskaya (changes requested):

s3-streamer
mvollmer: Done!

tests-invoke
```diff
@@ -77,6 +77,9 @@ def main():
             entry_point = alt_entry_point
         cmd = ["timeout", "120m", entry_point]
 
+        os.environ["TEST_REVISION"] = opts.revision
```
martinpitt, line 80: We already [do this](https://github.com/cockpit-project/cockpit/blob/main/test/common/run-tests#L528)

mvollmer: But that's wrong half of the time, no? `HEAD` is the result of the rebase. We could use `ORIG_HEAD`, but that then assumes that a rebase has just been done, which might not be true.

I think it is best to set the values here, with authority.

martinpitt: Ah, and you need the original HEAD from the proposed branch, without a rebase? AFAICS the only place where we use this variable so far is in bots/test-failure-policy, and there it doesn't matter much. OK then.

mvollmer: I'd say the only reason to have this in the environment at all is to reliably point to the pre-rebase HEAD.  If you want HEAD, you can just look at it with `git rev-parse HEAD`.

mvollmer:
> Thanks! This looks reasonable, but integration tests fail. Can you please re-submit his from origin? I'll then quickly ack/land it.

#3548

PR #3535, opened 2022-06-21
testmap: Trigger RHEL test for cockpit PO refreshes

martinpitt:
cockpit's check-pages "all languages" test only runs on RHEL, so make
sure we actually run this.

---

Spotted in https://github.com/cockpit-project/cockpit/pull/17477


marusak (approved):
Danke!

PR #3561, opened 2022-06-27
testmap: Move fedora-coreos build image to Fedora 36

martinpitt:


martinpitt:
Wait, current FCOS is still F35 based.

PR #3562, opened 2022-06-27

cockpituous:


martinpitt:
[interesting failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3562-20220627-145320-b0a7c29f-fedora-coreos-cockpit-project-cockpit/log.html#13-2). I saw this in https://github.com/cockpit-project/cockpit/pull/17506 already, but here I have it in isolation.

martinpitt:
Ah, the problem is that in our label-install code:
```
# For podman, generate a systemd unit for starting on boot
if [ "${container:-}" = oci ] && [ -n "$IMAGE" ] && [ ! -e /host/etc/systemd/system/cockpit.service ]; then
```
... the first condition is now false:
```
+ '[' podman = oci ']'
```
:angry: 

I sent https://github.com/cockpit-project/cockpit/pull/17509 for this. After that, rebuild the container and refresh the image.

martinpitt:
Container rebuilt, let's do this again.

PR #3591, opened 2022-07-07
images: add an openSUSE image

jelly:


images/scripts/opensuse-leap-15-4.bootstrap
```diff
@@ -0,0 +1,26 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301 USA.
+
+set -eux
+
+URL='https://download.opensuse.org/repositories/Cloud:/Images:/Leap_15.4/images/'
```
marusak, line 22: Aha, I was wondering if we want the leap or tumbleweed. I guess either is good starting point

martinpitt:
I don't know how OpenSUSE versioning works, but "15.4" seems a little too specific? Don't we want a "leap-15" image or so, and test on the latest .Y available? Or is "15.4" by itself a long-term release with sub-versions, and there's an independently evolving "15.5"?

imobachgs:
> I don't know how OpenSUSE versioning works, but "15.4" seems a little too specific? Don't we want a "leap-15" image or so, and test on the latest .Y available? Or is "15.4" by itself a long-term release with sub-versions, and there's an independently evolving "15.5"?

openSUSE Leap is based on SLE and 15.X are considered minor versions. openSUSE releases a new version once a year, although they are supported up to 18 months since the release date.

Having said that, note that openSUSE Leap does not include Cockpit. Instead, Tumbleweed does, although it is a rolling release.

jelly:
https://github.com/lnussel/cockpit/pull/1/files
https://build.opensuse.org/package/show/openSUSE%3AFactory/cockpit

martinpitt:
Superseded by #5954 and/or #6011

PR #3606, opened 2022-07-12

cockpituous:
 * [ ] Fix virt-install to get along with rawhide boot.iso: ~https://bugzilla.redhat.com/show_bug.cgi?id=2107440 (duplicate)~ https://bugzilla.redhat.com/show_bug.cgi?id=2103835 -- worked around in https://github.com/cockpit-project/cockpituous/pull/510


martinpitt:
I tried to get straight to the point:

    sudo dnf install -y qemu-device-usb-redirect qemu-char-spice
    virt-install --debug --name r --os-variant fedora-rawhide --memory 2048 --autoconsole text --extra-args 'inst.sshd' --network none --wait -1 --location `pwd`/images/fedora-rawhide-boot

This works (that's main, and the old image). When I run it against ~/.cache/cockpit-images/fedora-rawhide-boot-d021f287d273193e4c902207ece1f9b9e2443beb03e4b6e67801c1b673ffce2c.iso instead (this new image), I get the same error as in the test log.

The old working one:
```
Starting install...
[Fri, 15 Jul 2022 06:55:07 virt-install 270634] DEBUG (urlfetcher:141) hasFile(/isolinux/vmlinuz) returning True
[Fri, 15 Jul 2022 06:55:07 virt-install 270634] DEBUG (urlfetcher:141) hasFile(/isolinux/initrd.img) returning True
[Fri, 15 Jul 2022 06:55:07 virt-install 270634] DEBUG (urlfetcher:320) Running isoinfo: ['isoinfo', '-J', '-i', '/var/home/martin/.cache/cockpit-images/fedora-rawhide-boot-33a9f28780f40d85b37ece165a9297808a20152af55c5c89b544c0f5872f7608.iso', '-x', '/isolinux/vmlinuz']
[Fri, 15 Jul 2022 06:55:07 virt-install 270634] DEBUG (urlfetcher:72) Fetching URI: /isolinux/vmlinuz
Retrieving file vmlinuz...                                                                                       |  12 MB  00:00:00     
[Fri, 15 Jul 2022 06:55:07 virt-install 270634] DEBUG (urlfetcher:158) Saved file to /var/home/martin/.cache/virt-manager/boot/virtinst-ot9hdqm_-vmlinuz
```

The new failing one:
```
Starting install...
[Fri, 15 Jul 2022 06:57:11 virt-install 270771] DEBUG (urlfetcher:141) hasFile(/isolinux/vmlinuz) returning False
[Fri, 15 Jul 2022 06:57:11 virt-install 270771] ERROR (cli:256) Couldn't find kernel for install tree.
```

```
isoinfo -J -i fedora-rawhide-boot-33a9f28780f40d85b37ece165a9297808a20152af55c5c89b544c0f5872f7608.iso -l
# ...
# Directory listing of /isolinux/
# ----------   0    0    0        12967088 Jun 10 2022 [ 376187 00]  vmlinuz
```

Whereas the new image does not have an `/isolinux/` directory at all any more. There is only a vmlinuz and initrd.img in `/images/pxeboot`, that may be one that virt-install could use.

There is no existing bugzilla report (at least not any that I can find), so I reported that at https://bugzilla.redhat.com/show_bug.cgi?id=2107440 . Blocking on that.

CC @M4rtinK , @jkonecny12

martinpitt:
@jkonecny12, @M4rtinK : Where does boot.iso actually get defined and built? I.e. where can we check whether dropping /isolinux was deliberate or a bug?

VladimirSlavik:
@martinpitt I can't provide any links or references immediately, but I saw some discussions about this in bugzilla. The definitions are in lorax templates, owned by @bcl who is already aware of something related to this, as well as some Fedora folks and IIRC the virt people too. I don't know the full details, but the bug I saw was moved to some virt component.

bcl:
@martinpitt this was the result of moving to grub2 for BIOS booting - https://fedoraproject.org/wiki/Changes/BIOSBootISOWithGrub2 so yes, it's on purpose. Ends up virt was using the isolinux directory instead of the images/pxeboot/ directory. The kernel/initrd files in there are the same, the isolinux ones were hardlinks to the pxeboot ones which should remain stable for the foreseeable future :)

martinpitt:
Thanks @bcl! I'm happy to see that the [fix](https://gitlab.com/libosinfo/osinfo-db/-/merge_requests/481) landed a week ago. This matches my own observation, that the pxeboot/ files are identical and should suffice. Looking forward to seeing this land!

martinpitt:
To test the pustream fix, I ran this as root in one bot:

    sed -i 's_isolinux/_images/pxeboot/_' /usr/share/osinfo/os/fedoraproject.org/fedora-rawhide.xml

and then ran the command from the previously failed test run:

    ./make-checkout --verbose --repo=rhinstaller/anaconda master && cd make-checkout-workdir && TEST_OS=fedora-rawhide-boot BASE_BRANCH=master COCKPIT_BOTS_REF=325b50c85307842669ea626d33a64481dfe08dcb ../tests-invoke --pull-number 3606 --revision 325b50c85307842669ea626d33a64481dfe08dcb --repo cockpit-project/bots

which gets me a little further at least, although it still fails. But let's apply this workaround for now, and see how far we get.

martinpitt:
Hooray, green! So let's validate and land https://github.com/cockpit-project/cockpituous/pull/510 and then land this.

martinpitt (approved):

PR #3619, opened 2022-07-15
lib: Stop assuming non-TAP formatting in log.html

martinpitt:
We are going to clean up our test logs. Let's stop assuming that each
test starts with a `# ------` line, and instead just parse TAP:
everything in between two "(not) ok" lines is the test output.
    
Also drop the weird double-`$` terminator in the `tap_result` regexp.

-----

I tested this with a bunch of current logs (e.g. [this one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17436-20220607-110853-26dcc7ed-fedora-36-firefox/log)) as well as with the reduced logs from https://github.com/cockpit-project/cockpit/pull/17441 (e.g. [this one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17441-20220608-052614-9143fb37-fedora-36-firefox/log)). With main's log.html that is [a mess](https://cockpit-logs.us-east-1.linodeobjects.com/pull-17441-20220608-052614-9143fb37-fedora-36/log.html) , but it looks fine with this fix.

Tested as usual by downloading "log" into a directory, symlinking log.html, and running `python3 -m http.server`


marusak:
Hmm, can we test this locally? I knew how to test this with `task/log.html` but not sure how to do it with s3 version

martinpitt:
@marusak : There's no difference. In the end it's just HTTP-serving log and log.html, see description.

marusak (approved):
Thanks!

PR #3635, opened 2022-07-20
testmap: update cockpit-composer's test distro list

jkozol:
cockpit-composer should now test against f35, f36, RHEL 8.7, and 9.1

See https://github.com/osbuild/cockpit-composer/pull/1447 for test passes

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -124,16 +124,15 @@ REPO_BRANCH_CONTEXT = {
     },
     'osbuild/cockpit-composer': {
         'main': [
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! I'll trigger the firefox test (which was missing from https://github.com/osbuild/cockpit-composer/pull/1447), but I assume it'll be good.

martinpitt (approved):
Thanks @jkozol for fixing the tests! I rebased this to avoid landing red.

PR #3660, opened 2022-07-25
task: Switch over main AMQP server to AWS

martinpitt:
Our CentOS CI project will go away soon, so let's move to an EC2
instance as our primary webhook/queue. See
https://github.com/cockpit-project/cockpituous/pull/509

 - Land together with https://github.com/cockpit-project/cockpituous/pull/509 and updating all projects webhooks in their settings.


martinpitt:
@marusak : Please just review/approve, but don't land -- I'll coordinate that in the cockpituous PR. Dakujem!

marusak (approved):

PR #3664, opened 2022-07-26
images: Add rsync to Debian images

martinpitt:
This will allow us to replace the deprecated `scp` in our bots API. All
other images already have rsync installed.

----

 * [x] image-refresh debian-stable
 * [x] image-refresh debian-testing

PR #3665, opened 2022-07-26
machine_core: Use rsync for file downloading

martinpitt:
The scp protocol has been deprecated for a long time, so stop using it
and use rsync instead. After [1], all our images have rsync installed.

Commit 90a9f9ddf9c already moved SSHConnection.upload() to rsync, but
with a fallback to scp. We need to keep that fallback, as .upload() must
work on pristine cloud images, before running our setup scripts.

[1] https://github.com/cockpit-project/bots/pull/3664

----

The consumers of `.download()` are sub-man's check-subscription, cockpit's image-prepare, and cockpit's check-packagekit (only on rhel). I'll trigger these tests to validate.

 - [x] depends on PR #3664

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -433,13 +433,12 @@ class SSHConnection(object):
         dest = os.path.join(relative_dir, dest)
 
         cmd = [
-            "scp",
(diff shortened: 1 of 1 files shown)


marusak (approved):
Thank you!

PR #3706, opened 2022-08-08
machine_core: Run execute() commands with `set -e`

martinpitt:
In hindsight it was a major error to not run commands with `set -e`:

 - Tests silently ignore command failures, leading to test failures
   later on which are harder to debug

 - Tests accumulated a lot of `execute("a && b")` style calls, which
   silently succeed if `a` fails

 - Tests have to sprinkle `set -e` around a lot if they want to do the
   right thing.

Thus, do it by default, let's fix the fallout, and then clean up the
`&&` and `set -e` from tests afterwards.

Don't do this for the `execute(script=)` mode. It's obsolete and hardly
used (see https://github.com/osbuild/cockpit-composer/pull/1477).

----

See https://github.com/cockpit-project/cockpit-machines/pull/788 for the nasty effects that this did. 

We also have a lot of potential traps in cockpit:
```
❱❱❱ git grep '&&' test/verify  | wc -l
94
```
I'll clean these up once this PR lands.

Blocked on fixing tests for `set -e`:
 - https://github.com/cockpit-project/cockpit-machines/pull/788
 - https://github.com/cockpit-project/cockpit/pull/17624
 - https://github.com/candlepin/subscription-manager-cockpit/pull/36
 - https://github.com/cockpit-project/cockpit/pull/17630

```
     ./tests-trigger 3706 image:{fedora-36,debian-testing,ubuntu-stable,arch} image:rhel-{8-4,8-5,8-6,8-7,9-1}
```

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -302,7 +302,7 @@ class SSHConnection(object):
             self.ssh_address
         ]
         additional_ssh_params = []
-        cmd = []
+        cmd = ['set -e;']
 
         additional_ssh_params += self.__execution_opts(direct=direct)
 


martinpitt:
[this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3706-20220808-163616-549403cf-fedora-coreos-cockpit-project-cockpit/log.html#117-2) looks relevant.

martinpitt:
Yay, finally! I triggered a crap-ton of tests, and all green at last!

machine/machine_core/ssh_connection.py
```diff
@@ -307,15 +307,13 @@ def execute(self, command=None, script=None, input=None, environment={},
...
-                cmd += [command]
-                if not quiet:
-                    self.message("+", command)
-            else:
+            if isinstance(command, list):
                 # use shlex.join() once Python 3.8 is available everywhere
-                cmd.append(' '.join(shlex.quote(arg) for arg in command))
-                if not quiet:
-                    self.message("+", *command)
+                command = ' '.join(shlex.quote(arg) for arg in command)
```
allisonkarlitskaya, line 312: I like the `isinstance()` (although `list` is probably too specific — tuple might easily be used here).  I don't like replacing `command` with a new value of a different type.  I think this would be better with two separate `.append()` calls.

If you'd prefer to keep the DRY savings here, then please a new variable (but then you'll need an else branch again).

machine/machine_core/ssh_connection.py
```diff
@@ -311,7 +311,7 @@ def execute(self, command=None, script=None, input=None, environment={},
                 # use shlex.join() once Python 3.8 is available everywhere
                 command = ' '.join(shlex.quote(arg) for arg in command)
 
-            cmd += [command]
+            cmd += ['set -e; ' + command]
```
allisonkarlitskaya, line 314: SSH is so very evil...

allisonkarlitskaya (changes requested):

machine/machine_core/ssh_connection.py
martinpitt: It is, but do you have a better idea? Given how much crap and brokenness that I found recently with this, I am more convinced than ever that `set -e` by default is the right thing to do. Of course the "how" looks quite horrible..

machine/machine_core/ssh_connection.py
martinpitt: hmm, I'd like to rename the argument to `command_str_or_list`, but that's an API break. I wanted to avoid having "cmd", "command", and "command_string". In that case I'll just replace the getattr with an isinstance(str) (no basestr any more in py 3 \o/).

martinpitt: Hmm, no -- I really *really* want to get rid of that `else` branch, as it's just awkward to duplicate the `set -e` addition as well. I still don't understand why it's bad to transform one allowed data type variation into the canonical form..

martinpitt: Done without shadowing, and assigning to `command_str` now.

machine/machine_core/ssh_connection.py
allisonkarlitskaya: I was actually thinking to add `'set -x;`' as a separate commandline argument to ssh, with the knowledge that SSH is going to paste everything together again anyway (giving you the same result).

This would also allow you to move the `set -e` above the computation of the command itself (ie: the list-flattening stuff) and replace that with `.append()` or `.extend()` in order to avoid the extra variable assignment (and remove the comment about `shlex.join()`.

allisonkarlitskaya: Something like:

```python
cmd.append('set -e;')
if not isinstance(command, str):
    cmd.extend(shlex.quote(arg) for arg in command)
else:
    cmd.append(str)
````

martinpitt: But that's going to need the if/else case for the logging again. It really is the simplest way to just unify the two cases (list/string) very early IMHO. I can't find any iteration of that which is significantly simpler, so I'm just going to drop that commit. It's not necessary for what I want to achieve after all, and we can banter about this in a less intrusive PR (in terms of how many tests it needs).

machine/machine_core/ssh_connection.py
martinpitt: Dropped that commit.

machine/machine_core/ssh_connection.py
martinpitt: > I was actually thinking to add 'set -e;' as a separate commandline argument to ssh

That's a great idea! Done.

allisonkarlitskaya: I'd be happy enough to merge this already (since we bikeshed enough) but your changes after this message seem not to have been pushed yet, so I'll wait.

martinpitt: Argh! I pushed it to my fork instead of origin last night. There go 45 green tests..

martinpitt:
@allisonkarlitskaya : All green again -- PTAL?

allisonkarlitskaya (approved):

PR #3708, opened 2022-08-08
Fix LGTM alerts

martinpitt:
See https://lgtm.com/projects/g/cockpit-project/bots/alerts/?mode=list

Unfortunately I didn't find a way to tell lgtm to actually *fail* their test status on new alerts... :cry:

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -169,23 +169,21 @@ class VirtNetwork:
         except FileExistsError:
             pass
         for port in range(start, start + (100 * step), step):
-            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-            lockpath = os.path.join(resources, "network-{0}".format(port))
-            try:
-                lockf = os.open(lockpath, os.O_WRONLY | os.O_CREAT)
-                fcntl.flock(lockf, fcntl.LOCK_NB | fcntl.LOCK_EX)
-                sock.bind(("127.0.0.1", port))
-                self.locked.append(lockf)
-            except IOError:
-                if force:
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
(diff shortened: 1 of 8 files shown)


martinpitt:
Okay, this [fixes 8 alerts](https://lgtm.com/projects/g/cockpit-project/bots/rev/pr-a2646c734a06ee7140daaf88ec1afa10b46e309e), but there are [11 current ones](https://lgtm.com/projects/g/cockpit-project/bots/alerts/?mode=list). It seems it doesn't count the three alerts that I ignored? I might have messed up the [alert suppression](https://lgtm.com/help/lgtm/alert-suppression), possibly the missing "py/" prefix. Trying with that.

martinpitt:
Okay, :man_shrugging: -- maybe it just doesn't consider ignored alerts as "fixed" then? Let's at least land these 11? I trigger a few tests, just to make sure the changes don't break anything, and also to validate https://github.com/cockpit-project/cockpituous/pull/517.

machine/machine_core/machine_virtual.py
```diff
@@ -169,23 +169,21 @@ def _lock(self, start, step=1, force=False):
...
+                    fcntl.flock(lockf, fcntl.LOCK_NB | fcntl.LOCK_EX)
+                    sock.bind(("127.0.0.1", port))
+                    self.locked.append(lockf)
+                except IOError:
+                    if force:
+                        return port
+                    os.close(lockf)
+                    continue
+                else:
                     return port
```
allisonkarlitskaya, line 186: Unrelated, but another cleanup to this too-complicated code:

Can you drop the `else:` clause by pulling the `return port` into the `try:` or just unindenting it?  The `except:` clause never returns.

You could also save a line by inverting the condition in the `except:` clause and allowing the fall-through to the `return`.

allisonkarlitskaya (approved):
Feel free to commit as is, or bonus points for the suggested change.

In either case, thanks for the cleanups :)

machine/machine_core/machine_virtual.py
martinpitt: Good point! I'm not done with bots anyway, I have some more cleanups in the pipe once the current set of entangled PRs lands. I'll do that as a follow-up!

PR #3710, opened 2022-08-08
test: Move to flake8

martinpitt:
This easier to use, more flexible as pycodestyle errors can be
overridden on specific lines, and IDEs/vim-ale can read .flake8 and
stop complaining about the line length and imports.


jelly (approved):

jelly:
Thanks!

PR #3728, opened 2022-08-11
Add some optional typing

jelly:


martinpitt:
@jelly : Please re-propose from origin, so that the integration tests can run. See the "cockpituous" failed test.

machine/machine_core/constants.py
```diff
@@ -3,8 +3,8 @@
 from lib.constants import OSTREE_IMAGES, MACHINE_DIR, BOTS_DIR, BASE_DIR, TEST_DIR, GIT_DIR, \
     IMAGES_DIR, SCRIPTS_DIR, DEFAULT_IDENTITY_FILE, TEST_OS_DEFAULT, DEFAULT_IMAGE
 
-__all__ = (OSTREE_IMAGES, BOTS_DIR, MACHINE_DIR, BASE_DIR, TEST_DIR, GIT_DIR,
-           IMAGES_DIR, SCRIPTS_DIR, DEFAULT_IDENTITY_FILE, TEST_OS_DEFAULT, DEFAULT_IMAGE,)
+__all__ = ("OSTREE_IMAGES", "BOTS_DIR", "MACHINE_DIR", "BASE_DIR", "TEST_DIR", "GIT_DIR",
```
martinpitt, line 6: TIL! TBH I'm not sure what `__all__` is even good for. `import *` is just evil. Anyway, thanks for catching this!

machine/machine_core/machine_virtual.py
```diff
@@ -139,8 +149,8 @@ def stdchannel_redirected(stdchannel, dest_filename):
 
 
 class VirtNetwork:
-    def __init__(self, network=None, image="generic"):
-        self.locked = []
+    def __init__(self, network: Optional[int] = None, image: str = "generic"):
```
martinpitt, line 152: Can we also rename `network` to `id` or so?

martinpitt (changes requested):
This is a nice first step! As I said, please re-propose from origin, otherwise this is good to go.

machine/machine_core/constants.py
jelly: Yes, it's only for star imports and bad. Maybe we should drop it.

machine/machine_core/machine_virtual.py
jelly: Sure.

PR #3734, opened 2022-08-15
machine_core: Simplify MachineSSHConnection.execute()

allisonkarlitskaya:
Our logic about when to pass the `input` and the `stdin` parameters
seems to cause some issues with specific versions of Python (like that
found in CentOS 8 stream), where we get complaints about giving both
kwargs.

The only case where the logic has a gap in it (the input string is not
truthy, but also not `None`) should result in both kwargs being `None`.
I suspect that this happens as a result of the default value of `stdin`
in that version of Python not actually being `None`, but nothing else.

In any case, let's always provide an `input=` kwarg, possibly as an
empty string, and leave `stdin` alone.  An empty string should hand over
a pipe which immediately reads EOF, which is fairly equivalent to
reading from `/dev/null`.

This should help with cockpit-project/cockpit#17634 (@skazi0).

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -314,8 +314,7 @@ class SSHConnection(object):
 
         with timeoutlib.Timeout(seconds=timeout, error_message="Timed out on '%s'" % command, machine=self):
             res = subprocess.run(command_line,
-                                 stdin=None if input is not None else subprocess.DEVNULL,
-                                 input=input.encode("UTF-8") if input else None,
+                                 input=input.encode("UTF-8") if input else b'',
                                  stdout=stdout or subprocess.PIPE,
                                  check=check)
 


allisonkarlitskaya:
I'm not really sure how to test this against the thing which it claims to fix.  Let's do a test run to make sure it doesn't make everything explode and then we can land it and open a dummy PR on cockpit project.  Unless someone else has a better idea...

allisonkarlitskaya:
> open a dummy PR on cockpit project

→ cockpit-project/cockpit#17650

machine/machine_core/ssh_connection.py
```diff
@@ -314,8 +314,7 @@ def execute(self, command, input=None, environment={},
 
         with timeoutlib.Timeout(seconds=timeout, error_message="Timed out on '%s'" % command, machine=self):
             res = subprocess.run(command_line,
-                                 stdin=None if input is not None else subprocess.DEVNULL,
-                                 input=input.encode("UTF-8") if input else None,
+                                 input=input.encode("UTF-8") if input else b'',
```
jelly, line 317: Why change it to `b''` the default is `None`, isn't our main issue using `stdin` and `input` together. Input also defaults to `None` so that sounds good.

allisonkarlitskaya: Because I'm not sure if `None` means "pipe" "`/dev/null`" or "empty string".  I'm 100% sure what `b''` means.

allisonkarlitskaya: This fixes the problem, as shown by cockpit-project/cockpit#17650

jelly: Good question what `None` does the docs say:

```
The input argument is passed to [Popen.communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) and thus to the subprocess’s stdin. If used it must be a byte sequence, or a string if encoding or errors is specified or text is true. When used, the internal [Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) object is automatically created with stdin=PIPE, and the stdin argument may not be used as well.
```

Anyway this works so let's merge it.

jelly (approved):

PR #3735, opened 2022-08-15
Add optional typing

jelly:
Originally opened here https://github.com/cockpit-project/bots/pull/3728

 - [x] Blocked on https://github.com/cockpit-project/cockpit/pull/17733

### DIFF
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -18,7 +18,7 @@ import time
 import urllib.parse
 import urllib.request
 import xml.etree.ElementTree as ET
-from typing import IO, Dict, Iterable, List
+from typing import IO, Any, Dict, Generator, List, Union
 
 from .directories import xdg_config_home
 from .network import host_ssl_context
@@ -129,22 +129,23 @@ def urlopen(url: urllib.parse.ParseResult, method='GET', headers={}, data=b'') -
             raise
 
 
-def list_bucket(url: urllib.parse.ParseResult) -> ET:
+def list_bucket(url: urllib.parse.ParseResult) -> ET.Element:
     """Get the ListBucketResult as a xml.etree.ElementTree"""
     with urlopen(url) as response:
         return ET.fromstring(response.read())
 
 
-def parse_list(result: ET, *keys) -> Iterable[Iterable[str]]:
+def parse_list(result: ET.Element, *keys) -> Generator[Union[str, None, Any], None, None]:
     """For each item in the bucket, return the given keys"""
     # 'http' url is API: see https://doc.s3.amazonaws.com/2006-03-01/AmazonS3.wsdl
     xmlns = {'s3': 'http://s3.amazonaws.com/doc/2006-03-01/'}
     for child in result.findall('s3:Contents', xmlns):
-        yield (child.find(f's3:{key}', xmlns).text for key in keys)
+        yield (child.find(f's3:{key}', xmlns).text for key in keys)  # type: ignore
 
 
 def sign_url(url: urllib.parse.ParseResult, method='GET', headers=[], duration=12 * 60 * 60) -> str:
     """Returns a "pre-signed" url for the given method and headers"""
+    assert url.hostname, f'hostname is None for {url.path}'
     access, secret = get_key(url.hostname)
     bucket = url.hostname.split('.')[0]
 
--- a/lib/stores.py
+++ b/lib/stores.py
@@ -15,16 +15,18 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
+from typing import List
+
 from lib.directories import xdg_config_home
 
 # hosted on public internet
-PUBLIC_STORES = [
+PUBLIC_STORES: List[str] = [
     "https://cockpit-images.eu-central-1.linodeobjects.com/",
     "https://cockpit-images.us-east-1.linodeobjects.com/",
 ]
 
 # hosted behind the Red Hat VPN
-REDHAT_STORES = [
+REDHAT_STORES: List[str] = [
     # e2e down for maintenance
     # "https://cockpit-11.apps.cnfdb2.e2e.bos.redhat.com/images/",
 ]
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -25,6 +25,8 @@ import subprocess
 import sys
 import tempfile
 import time
+from collections import namedtuple
+from typing import IO, Any, Dict, List, Optional
 
 import libvirt
 import libvirt_qemu
@@ -39,10 +41,13 @@ sys.path.insert(1, BOTS_DIR)
 MEMORY_MB = 1152
 
 
+Disk = namedtuple('Disk', ['path', 'serial', 'filename', 'dev', 'index', 'type', 'extra'])
+
+
 # based on http://stackoverflow.com/a/17753573
 # we use this to quieten down calls
 @contextlib.contextmanager
-def stdchannel_redirected(stdchannel, dest_filename):
+def stdchannel_redirected(stdchannel: IO[str], dest_filename: str):
     """
     A context manager to temporarily redirect stdout or stderr
     e.g.:
@@ -142,15 +147,21 @@ TEST_USERNET_XML = """
 
 
 class VirtNetwork:
-    def __init__(self, network=None, image="generic"):
-        self.locked = []
+    hostnet: int
+    image: str
+    last: int
+    locked: List[int] = []
+    network: int
+    offset: int
+
+    def __init__(self, id: Optional[int] = None, image: str = "generic"):
         self.image = image
 
-        if network is None:
+        if id is None:
             offset = 0
             force = False
         else:
-            offset = network * 100
+            offset = id * 100
             force = True
 
         # This is a shared port used as the identifier for the socket mcast network
@@ -165,7 +176,7 @@ class VirtNetwork:
         # Unique hostnet identifiers
         self.hostnet = 8
 
-    def _lock(self, start, step=1, force=False):
+    def _lock(self, start: int, step: int = 1, force: bool = False) -> int:
         resources = os.path.join(tempfile.gettempdir(), ".cockpit-test-resources")
         os.makedirs(resources, 0o755, exist_ok=True)
         for port in range(start, start + (100 * step), step):
@@ -186,7 +197,7 @@ class VirtNetwork:
         raise Failure("Couldn't find unique network port number")
 
     # Create resources for an interface, returns address and XML
-    def interface(self, number=None):
+    def interface(self, number: Optional[int] = None):
         if number is None:
             number = self.last + 1
         if number > self.last:
@@ -203,7 +214,8 @@ class VirtNetwork:
         }
         return result
 
-    def host(self, number=None, restrict=False, isolate=False, forward={}):
+    def host(self, number: Optional[int] = None, restrict: bool = False, isolate: bool = False,
+             forward: Dict[str, int] = {}):
         '''Create resources for a host, returns address and XML
 
         isolate: True for no network at all, "user" for QEMU user network instead of bridging
@@ -240,12 +252,18 @@ class VirtNetwork:
 
 
 class VirtMachine(Machine):
-    network = None
-    memory_mb = None
-    cpus = None
+    cpus: int
+    graphics: bool
+    memory_mb: int
+    networking: VirtNetwork
+    run_dir: str
+    virt_connection: libvirt.virConnect
+    _disks: List[Disk] = []
+    _domain: Optional[libvirt.virDomain] = None
+    _transient_image: Any = None
 
-    def __init__(self, image, networking=None, maintain=False, memory_mb=None, cpus=None,
-                 graphics=False, overlay_dir=None, **args):
+    def __init__(self, image: str, networking=None, maintain=False, memory_mb=None, cpus=None,
+                 graphics: bool = False, overlay_dir=None, **args):
         self.maintain = maintain
 
         self.memory_mb = memory_mb or VirtMachine.memory_mb or MEMORY_MB
@@ -277,7 +295,6 @@ class VirtMachine(Machine):
 
         self.virt_connection = self._libvirt_connection(hypervisor="qemu:///session")
 
-        self._disks = []
         self._domain = None
         self._transient_image = None
 
@@ -442,13 +459,13 @@ class VirtMachine(Machine):
             # Normally only one pass
             break
 
-    def stop(self, timeout_sec=120):
+    def stop(self, timeout_sec: int = 120):
         if self.maintain:
             self.shutdown(timeout_sec=timeout_sec)
         else:
             self.kill()
 
-    def _cleanup(self, quick=False):
+    def _cleanup(self, quick: bool = False):
         self.disconnect()
         try:
             for disk in self._disks:
@@ -475,7 +492,7 @@ class VirtMachine(Machine):
                 sys.stderr.write(f"WARNING: Destroying machine failed: {e}\n")
         self._cleanup(quick=True)
 
-    def wait_poweroff(self, timeout_sec=120):
+    def wait_poweroff(self, timeout_sec: int = 120):
         # shutdown must have already been triggered
         if self._domain:
             start_time = time.time()
@@ -499,7 +516,7 @@ class VirtMachine(Machine):
                     raise
         self._cleanup(quick=True)
 
-    def shutdown(self, timeout_sec=120):
+    def shutdown(self, timeout_sec: int = 120):
         # shutdown the system gracefully
         # to stop it immediately, use kill()
         self.disconnect()
@@ -540,36 +557,36 @@ class VirtMachine(Machine):
         if self._domain.attachDeviceFlags(disk_desc, libvirt.VIR_DOMAIN_AFFECT_LIVE) != 0:
             raise Failure("Unable to add disk to vm")
 
-        disk = {
-            "path": image,
-            "serial": serial,
-            "filename": image,
-            "dev": dev,
-            "index": index,
-            "type": type,
-            "extra": extra,
-        }
+        disk = Disk(
+            path=image,
+            serial=serial,
+            filename=image,
+            dev=dev,
+            index=index,
+            type=type,
+            extra=extra,
+        )
 
         self._disks.append(disk)
         return disk
 
-    def rem_disk(self, disk, quick=False):
+    def rem_disk(self, disk: Disk, quick: bool = False):
         if not quick:
             disk_desc = TEST_DISK_XML % {
-                'file': disk["filename"],
-                'serial': disk["serial"],
-                'unit': disk["index"],
-                'dev': disk["dev"],
-                'type': disk["type"],
-                'extra': disk["extra"],
+                'file': disk.filename,
+                'serial': disk.serial,
+                'unit': disk.index,
+                'dev': disk.dev,
+                'type': disk.type,
+                'extra': disk.extra
             }
 
             if self._domain:
                 if self._domain.detachDeviceFlags(disk_desc, libvirt.VIR_DOMAIN_AFFECT_LIVE) != 0:
                     raise Failure("Unable to remove disk from vm")
-        os.unlink(disk['filename'])
+        os.unlink(disk.filename)
 
-    def _qemu_monitor(self, command):
+    def _qemu_monitor(self, command: str):
         self.message("& " + command)
         # you can run commands manually using virsh:
         # virsh -c qemu:///session qemu-monitor-command [domain name/id] --hmp [command]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,3 +11,19 @@ ignore = [
     "E731",  # Do not assign a `lambda` expression, use a `def`
 ]
 line-length = 118
+
+[[tool.mypy.overrides]]
+module = "pika.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "testmap.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "libvirt.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "libvirt_qemu.*"
+ignore_missing_imports = true
--- a/task/cache.py
+++ b/task/cache.py
@@ -27,6 +27,7 @@ import time
 import urllib.error
 import urllib.parse
 import urllib.request
+from typing import Any, Dict, Optional
 
 __all__ = (
     'Cache',
@@ -34,7 +35,12 @@ __all__ = (
 
 
 class Cache(object):
-    def __init__(self, directory, lag=None):
+    directory: str
+    lag: int
+    marked: float
+    pruned: bool = False
+
+    def __init__(self, directory: str, lag: Optional[int] = None):
         self.directory = directory
         self.pruned = False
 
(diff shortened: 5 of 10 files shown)


machine/machine_core/constants.py
```diff
@@ -3,8 +3,8 @@
 from lib.constants import OSTREE_IMAGES, MACHINE_DIR, BOTS_DIR, BASE_DIR, TEST_DIR, GIT_DIR, \
     IMAGES_DIR, SCRIPTS_DIR, DEFAULT_IDENTITY_FILE, TEST_OS_DEFAULT, DEFAULT_IMAGE
 
-__all__ = (OSTREE_IMAGES, BOTS_DIR, MACHINE_DIR, BASE_DIR, TEST_DIR, GIT_DIR,
-           IMAGES_DIR, SCRIPTS_DIR, DEFAULT_IDENTITY_FILE, TEST_OS_DEFAULT, DEFAULT_IMAGE,)
+__all__ = ("OSTREE_IMAGES", "BOTS_DIR", "MACHINE_DIR", "BASE_DIR", "TEST_DIR", "GIT_DIR",
```
allisonkarlitskaya, line 6: Did we decide to kill off all uses of `__all__` or not?

machine/machine_core/machine_virtual.py
```diff
@@ -37,10 +38,19 @@
 MEMORY_MB = 1152
 
 
+class Disk(TypedDict):
```
allisonkarlitskaya, line 41: The way we use this is actually pretty much exactly `namedtuple`.  Would you mind porting it over?

allisonkarlitskaya (changes requested):
Happy to see this getting started, but: precondition for adding type signatures was that we'd get them under some kind of CI.  I'd prefer to see the commit that introduces the testing in the same PR.

machine/machine_core/constants.py
jelly: Not yet, we can discuss if we want to do this everywhere.

machine/machine_core/machine_virtual.py
jelly: Sure!

machine/machine_core/constants.py
jelly: Found out yesterday, it's all over Python stdlib.

.mypy.ini
```diff
@@ -0,0 +1,7 @@
+[mypy]
```
martinpitt, line 1: Would it make sense to start using `pyproject.toml` right away, and add the mypy settings there? E.g. pylint can also store its configuration there.

jelly: Flake8 can't sadly https://github.com/PyCQA/flake8/issues/234

jelly:
Blocked on unit-tests container refresh.

martinpitt:
@jelly: This is unblocked at last! Most recent [unit-tests container build](https://github.com/cockpit-project/cockpit/actions/runs/3158211124/jobs/5139987364) now has mypy (and maybe yourpy as well :wink: )

martinpitt:
@jelly : We don't have a custom token, we use the default `secrets.GITHUB_TOKEN` for this. The [workflow permissions](https://github.com/cockpit-project/bots/blob/main/.github/workflows/test.yml#L20) are fairly strict, it may help to open them up a bit? But then again, fetching a list of pulls sounds very public already anyway, the [other permissions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) don't sound very plausible.

jelly:
> @jelly : We don't have a custom token, we use the default `secrets.GITHUB_TOKEN` for this. The [workflow permissions](https://github.com/cockpit-project/bots/blob/main/.github/workflows/test.yml#L20) are fairly strict, it may help to open them up a bit? But then again, fetching a list of pulls sounds very public already anyway, the [other permissions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) don't sound very plausible.

Ah, I amused the 401 is given because the token expired. But I wouldn't expect Github to block the pulls endpoint in a workflow? How did this even work before?!

jelly:
Ok, debugged this on my own fork, turns out.. I broke it somewhere in the PR https://github.com/jelly/bots/actions/runs/3182811947/jobs/5189241964

jelly:
Uhh, why is a failing unit-tests log https://localhost.localdomain:9000/logs/pull-3735-20221007-160949-a5e53271-unit-tests/log.html

martinpitt:
@jelly : This seems to be the problem, hmm:

    /work/bots/make-checkout-workdir/test/run: line 9: mypy: command not found

machine/machine_core/machine_virtual.py
```diff
@@ -27,6 +27,8 @@
 import tempfile
 import sys
 import time
+from collections import namedtuple
```
martinpitt, line 30: FTR: This should be available in RHEL 8/py 3.6, see https://docs.python.org/3/library/collections.html?highlight=namedtuple#collections.namedtuple

machine/machine_core/machine_virtual.py
```diff
@@ -543,34 +549,34 @@ def add_disk(self, size=None, serial=None, path=None, type='raw'):
...
 
-        disk = {
-            "path": image,
-            "serial": serial,
-            "filename": image,
-            "dev": dev,
-            "index": index,
-            "type": type,
-        }
+        disk = Disk(
```
martinpitt, line 552: This value is returned, thus this is an API break. Please run some cockpit and c-machines tests against this to make sure this works. Otherwise, let's maybe dial this back.

task/github.py
```diff
@@ -296,42 +299,44 @@ def delete(self, resource, accept=[]):
...
         while count == 100:
             data = self.get("commits/{0}/status?page={1}&per_page={2}".format(revision, page, count))
-            count = 0
-            page += 1
-            if "statuses" in data:
-                for status in data["statuses"]:
-                    if is_valid_context(status["context"], self.repo) and status["context"] not in result:
-                        result[status["context"]] = status
-                count = len(data["statuses"])
+            if data is not None:
```
martinpitt, line 316: Shouldn't this be more like an assertion? If we get a None data in the middle of the iteration, this is either a broken GitHhub API or some broken logic on our side. But we really should never get None here, at most an empty list?

task/github.py
```diff
@@ -296,42 +299,44 @@ def delete(self, resource, accept=[]):
...
         result = []
         page = 1
         count = 100
         while count == 100:
             data = self.get("statuses/{0}?page={1}&per_page={2}".format(revision, page, count))
-            count = 0
-            page += 1
-            result += data
-            count = len(data)
+            if data is not None:
```
martinpitt, line 332: dito

lib/s3.py
```diff
@@ -123,22 +123,24 @@ def urlopen(url: urllib.parse.ParseResult, method='GET', headers={}, data=b'') -
...
     xmlns = {'s3': 'http://s3.amazonaws.com/doc/2006-03-01/'}
     for child in result.findall('s3:Contents', xmlns):
-        yield (child.find(f's3:{key}', xmlns).text for key in keys)
+        yield (child.find(f's3:{key}', xmlns).text for key in keys)  # type: ignore
 
 
 def sign_url(url: urllib.parse.ParseResult, method='GET', headers=[], duration=12 * 60 * 60) -> str:
     """Returns a "pre-signed" url for the given method and headers"""
+    if url.hostname is None:
+        sys.exit(f'hostname is None for {url.path}')
```
martinpitt, line 143: This is a library -- I'd rather this was an assertion.

test/run
```diff
@@ -2,10 +2,12 @@
...
-# run static code checks like pyflakes and pep8
-PYEXEFILES="$(git grep -lI '^#!.*python') $(git ls-files "*.py")"
-flake8 $PYEXEFILES
+if [ -z "${SKIP_STATIC_CHECK:-}" ]; then
+	# run static code checks like pyflakes and pep8
+	PYEXEFILES="$(git grep -lI '^#!.*python') $(git ls-files "*.py")"
+	flake8 $PYEXEFILES
 
-mypy --no-error-summary --ignore-missing-imports --scripts-are-modules .
+	mypy --no-error-summary --ignore-missing-imports --scripts-are-modules .
```
martinpitt, line 10: Please squash this into the previous commit, or reorder them to do the "skip static checks" one first (which is cleaner).

test/run
```diff
@@ -2,10 +2,12 @@
 
 set -eu
 
-# run static code checks like pyflakes and pep8
-PYEXEFILES="$(git grep -lI '^#!.*python') $(git ls-files "*.py")"
-flake8 $PYEXEFILES
+if [ -z "${SKIP_STATIC_CHECK:-}" ]; then
+	# run static code checks like pyflakes and pep8
```
martinpitt, line 6: Please update the comment while you are at it, we don't run either of these two any more.

martinpitt (changes requested):
Thanks, and good job chasing this down!

lib/s3.py
jelly: Is it? It has a main and handles arguments. it doesn't really seem to be a library.

machine/machine_core/machine_virtual.py
jelly: https://github.com/cockpit-project/cockpit/pull/17801 Cockpit test

task/github.py
jelly: Seems I made a bogus change though with `if data and "statuses" in data:`

jelly: Ok, the handling if `None` is required if Github returns 404, if that's a fatal error, then we should assert yeah.

martinpitt:
Thanks! https://github.com/cockpit-project/cockpit/pull/17801 works, but note that you can run cockpit (or any other project) tests right in this PR; that's a bit easier.

jelly:
> Thanks! [cockpit-project/cockpit#17801](https://github.com/cockpit-project/cockpit/pull/17801) works, but note that you can run cockpit (or any other project) tests right in this PR; that's a bit easier.

Ah yeha, woops `image:fedora-36` is a lot of tests. Should have maybe picked fedora-37

martinpitt:
[These](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3735-20221011-080759-79c5b65c-fedora-coreos-cockpit-project-cockpit-ostree/log.html) [two](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3735-20221011-080755-79c5b65c-fedora-36-candlepin-subscription-manager/log.html) failures were due to npm cache breakage, I ran the Ansible script to reset them.

[This was](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3735-20221011-080747-79c5b65c-fedora-36-firefox-cockpit-project-cockpit-machines/log.html#3-3) $whatever, seemed unrelated.

But please re-trigger some tests here to make sure nothing gets broken.

lib/s3.py
martinpitt: No, that's just a `__main__` for interactively testing the library. It's normally being imported, e.g. [in image-upload](https://github.com/cockpit-project/bots/blob/main/image-upload#L32). This really isn't some "you specified the wrong arguments on the CLI" kind of situation, but an internal programming error when `sign_url` was called on invalid arguments. This should trigger a backtrace, and an assertion is exactly the right tool for function input validation.

test/run
```diff
@@ -6,6 +6,8 @@ if [ -z "${SKIP_STATIC_CHECK:-}" ]; then
 	# run static code checks like flake8 and mypy
 	PYEXEFILES="$(git grep -lI '^#!.*python') $(git ls-files "*.py")"
 	flake8 $PYEXEFILES
+
+	mypy --no-error-summary --ignore-missing-imports --scripts-are-modules .
```
martinpitt, line 10: Does this catch executables automatically? Most of these tools just look for modules. Otherwise this needs to be called on `$PYEXEFILES` as well. Or did you intentionally leave these out yet, as they need some further annotations/fixes?

martinpitt (changes requested):
Thanks! There's still a leftover thread from the previous round.

test/run
jelly: So `--scripts-are-modules` converts "executables" into modules so multiple programs don't conflict (multiple main()'s), but it seems this does not work at all. This is kinda frustrating.... I break image-prune by:

```
diff --git a/image-prune b/image-prune
index d4bde4f..87405e1 100755
--- a/image-prune
+++ b/image-prune
@@ -119,6 +119,9 @@ class ImageCache:

     def prune(self, keepers, force=False, dryrun=False):
         expiry_threshold = time.time() - IMAGE_EXPIRE * 86400
+        foo: bool = 1
+        foo = "str"
+        print(foo)
```

And mypy does not complain at all....

jelly: But changing to `PYEXEFILES` turns this into a another mess :|

```
[jelle@t14s][~/projects/cockpit-bots]%./test/run
pyproject.toml: Invalid value (at line 10, column 8)
test-failure-policy: error: Duplicate module named "test-failure-policy" (also at "task/test-failure-policy")
test-failure-policy: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
test-failure-policy: note: Common resolutions include: a) using `--exclude` to avoid checking one of them, b) adding `__init__.py` somewhere, c) using `--explicit-package-bases` or adjusting MYPYPATH
```

jelly: Ok, I just had to sort and unique `$PYEXEFILES` and now it works \o/

test/run
```diff
@@ -2,8 +2,10 @@
 
 set -eu
 
-# run static code checks like pyflakes and pep8
-PYEXEFILES="$(git grep -lI '^#!.*python') $(git ls-files "*.py")"
-flake8 $PYEXEFILES
+if [ -z "${SKIP_STATIC_CHECK:-}" ]; then
+	# run static code checks like flake8 and mypy
```
martinpitt, line 6: Over-indented.

martinpitt (commented):
The canonical renaming for task/test-failure-policy would be task/test-test-failure-policy. Or rename all of them to check-*.

martinpitt (approved):
Cheers!

martinpitt:
@allisonkarlitskaya : You still have a nack on this, can you please re-review?

machine/machine_core/machine_virtual.py
```diff
@@ -139,15 +144,15 @@ def stdchannel_redirected(stdchannel, dest_filename):
 
 
 class VirtNetwork:
-    def __init__(self, network=None, image="generic"):
-        self.locked = []
+    def __init__(self, id: Optional[int] = None, image: str = "generic"):
+        self.locked: List[int] = []
```
allisonkarlitskaya, line 148: Isn't it more usual to write it like this?

```python
class VirtNetwork:
    locked: List[int]
    image: str

    def __init__(self, ...
```

allisonkarlitskaya: I feel like annotations for `network`, `offset`, `last`, and `hostnet` might also be helpful here.

machine/machine_core/machine_virtual.py
```diff
@@ -281,7 +287,7 @@ def __init__(self, image, networking=None, maintain=False, memory_mb=None, cpus=
 
         self.virt_connection = self._libvirt_connection(hypervisor="qemu:///session")
 
-        self._disks = []
+        self._disks: List[Disk] = []
```
allisonkarlitskaya, line 290: *alarm bells*

task/cache.py
```diff
@@ -27,13 +27,15 @@
 import urllib.parse
 import urllib.error
 
+from typing import Any, Dict, Optional
```
allisonkarlitskaya, line 30: ditto comment about Python 3.9 and `dict`.

task/cache.py
```diff
@@ -48,10 +50,10 @@ def __init__(self, directory, lag=None):
         self.lag = lag
 
         # The mark tells us that stuff before this time is not "current"
-        self.marked = 0
+        self.marked: float = 0
```
allisonkarlitskaya, line 53: ditto comment about the variable annotation being on the class

task/cache.py
```diff
@@ -64,7 +66,7 @@ def prune(self):
             pass
 
     # Read a resource from the cache or return None
-    def read(self, resource):
+    def read(self, resource: str) -> Optional[Dict[Any, Any]]:
```
allisonkarlitskaya, line 69: Insofar as we can be sure that this is a dict at all (and not some other valid JSON object, like `None`, a number, or a list) then we can at least be pretty sure that the key type is `str`.

allisonkarlitskaya (changes requested):
I'd like to know if we have a good answer to the question about which Python version the bots targets.

There are also some small suggested changes below.  In general, I don't find annotations like `Optional[Any]` to be useful... if we're not able to be more specific, and we're only doing this ad-hoc at first, then I'd prefer to leave it out entirely...

martinpitt:
> I'd like to know if we have a good answer to the question about which Python version the bots targets.

They must work on RHEL/CentOS 8 for gating tests, so 3.6.

jelly:
> I'd like to know if we have a good answer to the question about which Python version the bots targets.
> 
> There are also some small suggested changes below. In general, I don't find annotations like `Optional[Any]` to be useful... if we're not able to be more specific, and we're only doing this ad-hoc at first, then I'd prefer to leave it out entirely...

We target Python 3.6 as per comment from @martinpitt https://github.com/cockpit-project/bots/pull/3735#issuecomment-1276237948

task/cache.py
jelly: Agreed, fixed.

pyproject.toml
```diff
@@ -11,3 +11,19 @@ ignore = [
     "E731",  # Do not assign a `lambda` expression, use a `def`
 ]
 line-length = 118
+
+[[tool.mypy.overrides]]
+module = "pika.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
```
allisonkarlitskaya, line 19: This one, at least, looks like a configuration problem.

pyproject.toml
```diff
@@ -11,3 +11,19 @@ ignore = [
...
+module = "testmap.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
```
allisonkarlitskaya, line 23: meh.  Is this about error messages in vim or about actual fails when linting as part of the static tests run?

machine/machine_core/machine_virtual.py
```diff
@@ -39,10 +41,13 @@
 MEMORY_MB = 1152
 
 
+Disk = namedtuple('Disk', ['path', 'serial', 'filename', 'dev', 'index', 'type', 'extra'])
```
allisonkarlitskaya, line 44: `typing.NamedTuple` is available in Python 3.6.

machine/machine_core/machine_virtual.py
```diff
@@ -142,15 +147,21 @@ def stdchannel_redirected(stdchannel, dest_filename):
 
 
 class VirtNetwork:
-    def __init__(self, network=None, image="generic"):
-        self.locked = []
+    hostnet: int
+    image: str
+    last: int
+    locked: List[int] = []
```
allisonkarlitskaya, line 153: please move the initializer back to `__init__`

machine/machine_core/machine_virtual.py
```diff
@@ -142,15 +147,21 @@ def stdchannel_redirected(stdchannel, dest_filename):
...
+    network: int
+    offset: int
+
+    def __init__(self, id: Optional[int] = None, image: str = "generic"):
```
allisonkarlitskaya, line 157: Your introduction of the `id` parameter here is going to clash with ruff `A` rules.

machine/machine_core/machine_virtual.py
```diff
@@ -203,7 +214,8 @@ def interface(self, number=None):
         }
         return result
 
-    def host(self, number=None, restrict=False, isolate=False, forward={}):
+    def host(self, number: Optional[int] = None, restrict: bool = False, isolate: bool = False,
```
allisonkarlitskaya, line 217: If it won't fit on one line, then please do one of these:

 - PEP8 hanging indent style like this:

```python
def host(
    self, number, restrict, ...
) -> None:
```

 - PEP8 hanging indent style like this:

```python
def host(
    self,
    number,
    restrict,
    ...
) -> None:
```

 - visual indent style like this:

```python
def host(self,
         number,
         restrict,
         ...) -> None:
```

machine/machine_core/machine_virtual.py
```diff
@@ -442,13 +459,13 @@ def start(self):
             # Normally only one pass
             break
 
-    def stop(self, timeout_sec=120):
+    def stop(self, timeout_sec: int = 120):
```
allisonkarlitskaya, line 462: Lots of missing return type annotations around...

task/cache.py
allisonkarlitskaya: Resolving.  We need py36 for now.

task/cache.py
allisonkarlitskaya: This isn't fixed.

task/cache.py
```diff
@@ -80,7 +86,7 @@ def read(self, resource):
             return None
 
     # Write a resource to the cache in an atomic way
-    def write(self, resource, contents):
+    def write(self, resource: str, contents: Dict[Any, Any]) -> None:
```
allisonkarlitskaya, line 89: I've acquired a rather powerful allergy to `Any` during the past year or so... practically the only place I see it as tolerable is on a `**kwargs` used as a passthrough to another function.

task/github.py
```diff
@@ -321,19 +324,20 @@ def statuses(self, revision):
                 count = len(data["statuses"])
         return result
 
-    def all_statuses(self, revision):
+    def all_statuses(self, revision: str) -> List[Any]:
```
allisonkarlitskaya, line 327: `List[Any]` isn't any better than just writing `list` as far as I know.  And both of them aren't really good enough...

test/run
```diff
@@ -13,8 +13,10 @@ if [ -z "${SKIP_STATIC_CHECK:-}" ]; then
     test_ruff
 
     # run static code checks like flake8 and mypy
-    PYEXEFILES="$(git grep -lI '^#!.*python') $(git ls-files "*.py")"
+    PYEXEFILES="$(echo $(git grep -lI '^#!.*python') $(git ls-files "*.py") | xargs -n 1 | sort -u)"
```
allisonkarlitskaya, line 16: Can you explain what you're doing here?

lib/stores.py
```diff
@@ -15,16 +15,18 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
+from typing import List
+
 from lib.directories import xdg_config_home
 
 # hosted on public internet
-PUBLIC_STORES = [
+PUBLIC_STORES: List[str] = [
```
allisonkarlitskaya, line 23: I feel like the type is superfluous here as easily inferable.

lib/s3.py
```diff
@@ -129,22 +129,23 @@ def urlopen(url: urllib.parse.ParseResult, method='GET', headers={}, data=b'') -
...
     # 'http' url is API: see https://doc.s3.amazonaws.com/2006-03-01/AmazonS3.wsdl
     xmlns = {'s3': 'http://s3.amazonaws.com/doc/2006-03-01/'}
     for child in result.findall('s3:Contents', xmlns):
-        yield (child.find(f's3:{key}', xmlns).text for key in keys)
+        yield (child.find(f's3:{key}', xmlns).text for key in keys)  # type: ignore
 
 
 def sign_url(url: urllib.parse.ParseResult, method='GET', headers=[], duration=12 * 60 * 60) -> str:
     """Returns a "pre-signed" url for the given method and headers"""
+    assert url.hostname, f'hostname is None for {url.path}'
```
allisonkarlitskaya, line 148: just write `is not None`

allisonkarlitskaya (changes requested):

allisonkarlitskaya (changes requested):
Commit messages need some looking at.  What's `mpypy`?  And `mypy.ini` isn't a thing anymore in this patch...

PR #3741, opened 2022-08-17
Image refresh for rhel-9-0

cockpituous:
Image refresh for rhel-9-0
 * [x] image-refresh rhel-9-0


allisonkarlitskaya:
Ya, so...

```
Removed:
  createrepo_c (0.17.7-4.el9_0)
  createrepo_c-libs (0.17.7-4.el9_0)
  distribution-gpg-keys (1.51-1.fc34)
  dwz (0.14-3.el9)
  ed (1.14.2-12.el9)
  efi-srpm-macros (6-2.el9_0)
  elfutils (0.186-1.el9)
  fonts-srpm-macros (1:2.0.5-7.el9.1)
  ghc-srpm-macros (1.5.0-6.el9)
  go-srpm-macros (3.0.9-9.el9)
  info (6.7-15.el9)
  kernel-srpm-macros (1.0-11.el9)
  libpkgconf (1.7.3-9.el9)
  lua-srpm-macros (1-6.el9)
  mock (2.9-2.fc34)
  mock-core-configs (34.2-1.fc34)
  mock-filesystem (2.9-2.fc34)
  ocaml-srpm-macros (6-6.el9)
  openblas-srpm-macros (2-11.el9)
  patch (2.7.6-16.el9)
  perl-srpm-macros (1-41.el9)
  pkgconf (1.7.3-9.el9)
  pkgconf-m4 (1.7.3-9.el9)
  pkgconf-pkg-config (1.7.3-9.el9)
  python-srpm-macros (3.9-52.el9)
  python3-pyroute2 (0.5.6-5.fc34)
  python3-templated-dictionary (1.1-2.fc34)
  qt5-srpm-macros (5.15.2-9.el9)
  redhat-rpm-config (194-1.el9)
  rpm-build (4.16.1.3-12.el9_0)
  rust-srpm-macros (17-4.el9)
  unzip (6.0-56.el9)
  zip (3.0-33.el9)
```

allisonkarlitskaya:
```
+ '[' '' -eq 1 ']'
/var/tmp/SETUP: line 239: [: : integer expression expected
```

which is from

```sh
# only install mock and build if DO_BUILD is 1
if [ "$DO_BUILD" -eq 1 ]; then
    if [ "${IMAGE#rhel-*}" != "$IMAGE" ]; then
        dnf install -y rpm-build
```

which is why we're not getting mock installed.  The weird thing is that `DO_BUILD` gets set like this:

```python
              env = {
                  "TEST_OS": self.machine.image,
                  "DO_BUILD": "0" if args.no_build else "1",
                  ...
```

so it should be either 0 or 1.

I think our environment handling got damaged in the execute simplifications.  This would explain a lot.

allisonkarlitskaya:
So ya.  Our addition of `set -e` to the front of our command to run mixes poorly with passing our environment variables by prefixing the command with (for example) `DO_BUILD=1 /var/tmp/SETUP`.

We end up with an invocation that looks like `DO_BUILD=1 set -e; /var/tmp/SETUP`.

Oof.

allisonkarlitskaya:
> So ya. Our addition of `set -e` to the front of our command to run mixes poorly with passing our environment variables by prefixing the command with (for example) `DO_BUILD=1 /var/tmp/SETUP`.
> 
> We end up with an invocation that looks like `DO_BUILD=1 set -e; /var/tmp/SETUP`.
> 
> Oof.

Fix in #3746.  Let's also land the image refreshes from there.

PR #3743, opened 2022-08-17

cockpituous:
 * [x] #3780 
 * [x] rhinstaller/anaconda#4287
 * [x] fix libsoup/webkit issue: rhinstaller/anacondal#4268

### DIFF
--- a/images/fedora-rawhide-boot
+++ b/images/fedora-rawhide-boot
@@ -1,1 +1,1 @@
-fedora-rawhide-boot-d021f287d273193e4c902207ece1f9b9e2443beb03e4b6e67801c1b673ffce2c.iso
+fedora-rawhide-boot-a321077ad99298ad489ef3d3e0b7f2a545b3457199837965bb1ad2114ff687bd.iso


allisonkarlitskaya:
Uh.  It's installing `cockpit-ws-259-1.fc35.x86_64.rpm`.  Interesting...

allisonkarlitskaya:
Aha.  rhinstaller/anaconda#4268 to the rescue?

M4rtinK:
> Aha. [rhinstaller/anaconda#4268](https://github.com/rhinstaller/anaconda/pull/4268) to the rescue?

Yep, that was hard coded before but should switch to fetching latest cockpit in [rhinstaller/anaconda#4268](https://github.com/rhinstaller/anaconda/pull/4268) - the thing is that even with the last known good image (20220712) currently in use + this Anaconda PR that has cockpit 274.1, it still does not work while it should as the new cockpit should be backwards compatible. 

So possibly something else going on as well.

allisonkarlitskaya:
I attached `virt-manager` to the running tests and got this:

image

Which means that this seems to be an issue in anaconda.

cockpituous:
image-refresh in progress on 4-cockpit-7.
Log: https://logs.cockpit-project.org/logs/image-refresh-3743-20220818-101414/

cockpituous:
image-refresh fedora-rawhide-boot done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-rawhide-boot-20220818-101630

allisonkarlitskaya:
This is the tasks container: https://github.com/cockpit-project/cockpituous/blob/main/tasks/Dockerfile

and indeed, it's not possible to choose which one you want.

Why not try building Anaconda in an image, instead?  Every other project does its build inside of the target image.  When the target image is a boot ISO, that's difficult of course, but I guess it would be possible to install mock?

Another option might be to do something similar to what we do for `fedora-coreos`.  See https://github.com/cockpit-project/bots/blob/main/lib/testmap.py#L194

M4rtinK:
> This is the tasks container: https://github.com/cockpit-project/cockpituous/blob/main/tasks/Dockerfile
> 
> and indeed, it's not possible to choose which one you want.
> 
> Why not try building Anaconda in an image, instead? Every other project does its build inside of the target image. When the target image is a boot ISO, that's difficult of course, but I guess it would be possible to install mock?

I'm afraid that's not going to work - the boot ISO needs to be very minimal (as its size drives, among other things, install time RAM requirements in important installation scenarios) and does not even have an on the image, so we can't really install to it. So I'm afraid this won't work, unless we can somehow use a different image (regular rawhide) for the build & then a different one for the test run (rawhide boot iso)

> 
> Another option might be to do something similar to what we do for `fedora-coreos`. See https://github.com/cockpit-project/bots/blob/main/lib/testmap.py#L194

I've noticed that as well, but seem kinda hardwired and special cased for coreos. If that was possible for us as well (eq. way to set what image or container to use for the anaconda build) I think it would fix the issue.

allisonkarlitskaya:
> I've noticed that as well, but seem kinda hardwired and special cased for coreos. If that was possible for us as well (eq. way to set what image or container to use for the anaconda build) I think it would fix the issue.

That is indeed exactly what I'm suggesting :)  I have no idea what it would look like, but it shouldn't be too hard.  I'd be happy to accept patches for it, in any case.

M4rtinK:
So, first wing we tried - using mock from inside the task container to build the Anaconda RPMs on Rawhide, so they get the correct Python version for the installation image:

https://github.com/rhinstaller/anaconda/pull/4280

Not yet working - would require mock to be added to the package list installed in the Cockpit task container and likely also some group (?) adjustments.

@allisonkarlitskaya What do you think ? :)

allisonkarlitskaya:
> @allisonkarlitskaya What do you think ? :)

I think that

 - the current problems are not caused by a version mismatch in a new image, because we haven't updated the image yet, and the tests are currently failing on anaconda master branch with the old image.  I think something changed in Anaconda.

 - i think that trying to build inside of a mock inside of the tasks container is maybe missing the point of what the tasks container is supposed to do.  Can we try using `image-customize` on the `fedora-36` (or newly-merged `fedora-37`) image to build in there?  They already have `mock` installed, and adding whatever dependencies we required would be easy.

M4rtinK:
> * the current problems are not caused by a version mismatch in a new image, because we haven't updated the image yet, and the tests are currently failing on anaconda master branch with the old image.  I think something changed in Anaconda.

That is correct in a sense, that newer Anaconda version uncovered that we actually don't use the Python code we inject from the F36 build RPMs as that ends up in Python 3.10 site-packages instead of the Python 3.11 ones the Rawhide image from July is already using.

> * i think that trying to build inside of a mock inside of the tasks container is maybe missing the point of what the tasks container is supposed to do.  Can we try using `image-customize` on the `fedora-36` (or newly-merged `fedora-37`) image to build in there?  They already have `mock` installed, and adding whatever dependencies we required would be easy.

As long as it can use a regular Fedora 37+ image, definitely. The image where the test run is far too cut down (including no RPM database) to build anything on it. 

But so far I have not found an easy way of doing that - ostree/coreos has some sort of support for that as far as I can tell, but it all seems very special cased and not readily usable without changes in both bots testmap and cockpit image-prepare as well.

But I might very well be missing an easy solution, as the tools do indeed some to be very powerful & there are many of them.

allisonkarlitskaya:
> > * the current problems are not caused by a version mismatch in a new image, because we haven't updated the image yet, and the tests are currently failing on anaconda master branch with the old image.  I think something changed in Anaconda.
> 
> That is correct in a sense, that newer Anaconda version uncovered that we actually don't use the Python code we inject from the F36 build RPMs as that ends up in Python 3.10 site-packages instead of the Python 3.11 ones the Rawhide image from July is already using.

Ah, right.  Makes sense :)

> But so far I have not found an easy way of doing that - ostree/coreos has some sort of support for that as far as I can tell, but it all seems very special cased and not readily usable without changes in both bots testmap and cockpit image-prepare as well.

These days, cockpit's `image-prepare` script is basically just a gloried wrapper around `image-customize`.  Do an `image-prepare` for `fedora-coreos` on cockpit project and watch what commands it invokes.  You could probably replace it with a reasonably small shell scripts, for your needs.

allisonkarlitskaya:
> These days, cockpit's `image-prepare` script is basically just a glorified wrapper around `image-customize`. Do an `image-prepare` for `fedora-coreos` on cockpit project and watch what commands it invokes. You could probably replace it with a reasonably small shell scripts, for your needs.

So I took a crack at it and that's in rhinstaller/anaconda#4287.  Tests there are green, and it wasn't too hard (although it wasn't possible to entirely drive the process from shell).

M4rtinK:
I've rebased the PR that makes sure latest cockpit is used on top of the package build changes that have just been merged:

https://github.com/rhinstaller/anaconda/pull/4268

That should hopefully fix the image refresh by pulling in >=274 cockpit that supports the 3.1 WebKit API present on current Rawhide images. :)

allisonkarlitskaya (approved):
:tada:

PR #3744, opened 2022-08-17
revert things

allisonkarlitskaya:
- [ ] image-refresh rhel-7-9


allisonkarlitskaya:
This is clearly not the cause of the anaconda issue, but it does fix RHEL.  The real fix for these problems is in #3746.

PR #3780, opened 2022-08-26
images: add Anaconda build deps to fedora-37

allisonkarlitskaya:
We're going to try using fedora-37 to build Anaconda RPMs (instead of
building them in the tasks container, as we do now).

 * [x] image-refresh fedora-37


cockpituous:
image-refresh in progress on rhos-01-2.
Log: https://logs.cockpit-project.org/logs/image-refresh-3780-20220826-092828/

cockpituous:
image-refresh fedora-37 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-37-20220826-094319

images/scripts/lib/build-deps.sh
```diff
@@ -25,6 +25,13 @@ esac
 
 echo "$spec" | rpmspec -D "$OS_VER" -D 'version 0' --buildrequires --query /dev/stdin | sed 's/.*/"&"/' | tr '\n' ' '
 
+# We build anaconda rpms on fedora-37
+if [ "$OS_VER" = 'fedora 37' ]; then
+    $GET https://raw.githubusercontent.com/rhinstaller/anaconda/master/anaconda.spec.in | \
```
M4rtinK, line 30: We do have a script that installs all necessary Anaconda dependencies:

https://raw.githubusercontent.com/rhinstaller/anaconda/master/scripts/testing/install_dependencies.sh

Maybe we can extend the script so that it can also just return all the dependencies as a string ? Then it could be fetched & executed to get the dependencies. That way it could be more robust in the future as we do use and maintain that script.

M4rtinK (approved):
Looks good to me, thanks a lot!

Just added a note about possibly using an anaconda deps enumeration script in the future, but nothing that should block this PR going in from my PoV. :)

images/scripts/lib/build-deps.sh
allisonkarlitskaya: I'd prefer not to execute stuff, if possible.  Could you have the dependencies in a flatfile, instead?

M4rtinK: Indeed, that's a good point - still it would be good if the flat file would be somehow automatically maintained so that we don't have to update the deps manually. I will think about something. :)

jelly (approved):

PR #3795, opened 2022-08-30
images: don't assume libvirt networking is disabled

allisonkarlitskaya:
The incoming ubuntu-stable image update contains a new version of
libvirt with the following change:

    - Autostart default bridged network (As upstream does, but not Debian).

This causes our unconditional attempt to enable the network to fail with
the following message:

    + virsh net-autostart default
    error: failed to mark network default as autostarted
    error: Failed to create symlink '/etc/libvirt/qemu/networks/autostart/default.xml' to '/etc/libvirt/qemu/networks/default.xml': File exists

Allow for both possibilities: check if the file exists before trying to
create it again.

### DIFF
--- a/images/scripts/debian.setup
+++ b/images/scripts/debian.setup
@@ -211,8 +211,8 @@ pbuilder --create --extrapackages "fakeroot $PBUILDER_EXTRA"
 /usr/lib/pbuilder/pbuilder-satisfydepends-classic --control /tmp/out/tools/debian/control --force-version --echo|grep apt-get | pbuilder --login --save-after-login
 rm -rf /tmp/out
 
-# Debian does not automatically start the default libvirt network
-virsh net-autostart default
+# (some) Debian does not automatically start the default libvirt network
+test -L /etc/libvirt/qemu/networks/autostart/default.xml || virsh net-autostart default
 
 # Disable services which we don't want running by default
 systemctl disable --now redis-server tuned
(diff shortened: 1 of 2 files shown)


cockpituous:
image-refresh in progress on 2-cockpit-7.
Log: https://logs.cockpit-project.org/logs/image-refresh-3795-20220830-120606/

cockpituous:
image-refresh in progress on rhos-01-9.
Log: https://logs.cockpit-project.org/logs/image-refresh-3795-20220830-120606/

cockpituous:
image-refresh ubuntu-stable done: https://github.com/cockpit-project/bots/commits/image-refresh-ubuntu-stable-20220830-121828

martinpitt:
This is a bug in Ubuntu's virsh. E.g. on Fedora 36, the "default" network is also auto-started by default, and that command should be idempotent:
```
# virsh net-list --all
 Name      State    Autostart   Persistent
--------------------------------------------
 default   active   yes         yes

# ls -l /etc/libvirt/qemu/networks/autostart/default.xml
lrwxrwxrwx. 1 root root 14 Aug 26 22:48 /etc/libvirt/qemu/networks/autostart/default.xml -> ../default.xml

# virsh net-autostart default
Network default marked as autostarted
```
I.e. we need to report this as a bug, and find a workaround.

mvollmer:
> and find a workaround.

This here is the workaround, no?  Let's call this out as a "HACK" or bug in the comment.

mvollmer:
I can't reproduce the virsh failure with the image in this PR....

```
root@ubuntu:~# virsh net-autostart default
Network default marked as autostarted

root@ubuntu:~# virsh net-autostart default
Network default marked as autostarted

root@ubuntu:~# virsh net-autostart default
Network default marked as autostarted

root@ubuntu:~# virsh net-autostart default
Network default marked as autostarted
```

mvollmer:
`virsh net-autostart` fails after the symlink is created manually while libvirtd is running and libvirtd is not restarted.  This might be happening to us when installing packages and then using them immediately without a reboot.

So a fix would be to restart libvirtd.service before running virsh.

martinpitt:
Closing in favor of #3842

PR #3797, opened 2022-08-31
machine: Ignore systemd-gpt-aut sys_admin SELinux violation on Fedora 37

martinpitt:
This happens during boot, thus affects every test which boots a fresh
VM. It's therefore too big for a naughty.

See https://bugzilla.redhat.com/show_bug.cgi?id=2083900

---

I tested this locally. This will unblock enabling f37 for our projects.


marusak (approved):
thanks!

PR #3805, opened 2022-08-31
machine: don't `set -e` for m.execute(check=False)

allisonkarlitskaya:
In case `check=False` was specified to Machine.execute(), skip the call
to `set -e`.

This pushes down the definition of `check=False` into the script itself:
in the case that multiple commands are given, we ignore failure of all
of them.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -287,7 +287,11 @@ class SSHConnection(object):
         ]
         additional_ssh_params = []
 
-        cmd = ['set -e;']
+        cmd = []
+
+        if check:
+            cmd.append('set -e;')
+
         cmd += [f'export {name}={shlex.quote(value)}; ' for name, value in environment.items()]
 
         additional_ssh_params += self.__execution_opts(direct=direct)


martinpitt (changes requested):
Oh, I wasn't even aware that .execute() has a `check=` kwarg. There is no consumer of that in any of our projects, including bots itself [1]. As I lamented in https://github.com/cockpit-project/cockpit/pull/17645#discussion_r944365747 , I find this a *terrible* API. It's not clear what "check" actually means (we don't document it), and tests (or shell scripts in general) really shouldn't be written in a way that "any command might fail". It is much clearer IMHO to mark individual commands with `|| true` -- this is both much easier to grep for, and also much more fine-grained than a blanket permission.

I would rather just drop this `check=` kwarg completely. 

[1] `cd ~/upstream; for d in *; do echo "==== $d ==="; git -C $d grep -A10 '\.execute' | grep check=; done`

allisonkarlitskaya:
The reason I like `check=` is that it's very very Pythonic.

martinpitt:
Right, but it's not Pythonic at all to run whole shell scripts. If we only use `check=False` for m.execute() with a single argv vector, then I'd agree -- but that's the case that already works today. However, for a lot of cases that would be too verbose or even too hard (as we often use shell features in setup snippets).

allisonkarlitskaya:
I feel like our solution here might be to push more in the direction of decomposing our little shell scripts into multiple `m.execute()` lines.

PR #3806, opened 2022-09-01
machine_core: Let SSHConnection.write() create the target directory

martinpitt:
Test suites are full of this pattern:

    m.execute("mkdir -p /target/dir")
    m.write("/target/dir/file.txt", "...")

Let's make our lifes a bit easier by always creating the target
directory of dest.

Also replace the wrong `&&` with separate execute() invocations, to
ensure that all the commands actually succeed (see commit b7d0edba5d),
and to avoid messing around with line breaks.

Replace % with f-strings.

----

Doing an image refresh just to make sure that this still works. I also trigger a few tests across projects to make sure this doesn't break anything.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -417,16 +417,18 @@ class SSHConnection(object):
             append: If True, append to existing file instead of replacing it
             owner: If set, call chown on the file with the given owner string
             perm: Optional file permission as chmod shell string (e.g. "0600")
+
+        The directory of dest is created automatically.
         """
         assert dest
         assert self.ssh_address
 
-        cmd = "cat %s '%s'" % (append and '>>' or '>', dest)
+        self.execute(["mkdir", "-p", os.path.dirname(dest)])
+        self.execute(f"cat {append and '>>' or '>'} {shlex.quote(dest)}", input=content)
         if owner:
-            cmd += " && chown '%s' '%s'" % (owner, dest)
+            self.execute(["chown", owner, dest])
         if perm:
-            cmd += " && chmod '%s' '%s'" % (perm, dest)
-        self.execute(cmd, input=content)
+            self.execute(["chmod", perm, dest])
 
     def spawn(self, shell_cmd, log_id, check=True):
         """Spawn a process in the test machine.


cockpituous:
image-refresh in progress on 2-ci-srv-04.
Log: https://logs.cockpit-project.org/logs/image-refresh-3806-20220901-061010/

cockpituous:
image-refresh in progress on rhos-01-10.
Log: https://logs.cockpit-project.org/logs/image-refresh-3806-20220901-061313/

cockpituous:
image-refresh debian-stable done: https://github.com/cockpit-project/bots/commits/image-refresh-debian-stable-20220901-063137

machine/machine_core/ssh_connection.py
```diff
@@ -417,15 +417,17 @@ def write(self, dest, content, append=False, owner=None, perm=None):
...
+        The directory of dest is created automatically.
         """
         assert dest
         assert self.ssh_address
 
-        cmd = "cat %s '%s'" % (append and '>>' or '>', dest)
+        cmd = f"mkdir -p '{os.path.dirname(dest)}'; cat {append and '>>' or '>'} '{dest}'"
         if owner:
-            cmd += " && chown '%s' '%s'" % (owner, dest)
+            cmd += f"; chown {owner} '{dest}'"
```
allisonkarlitskaya, line 428: Quoting, as appropriate.

allisonkarlitskaya (changes requested):
At the very least, this needs some `shlex.quote()` and I'd be happy to merge it with that, but see comments.

machine/machine_core/ssh_connection.py
```diff
@@ -417,15 +417,17 @@ def write(self, dest, content, append=False, owner=None, perm=None):
...
             owner: If set, call chown on the file with the given owner string
             perm: Optional file permission as chmod shell string (e.g. "0600")
+
+        The directory of dest is created automatically.
         """
         assert dest
         assert self.ssh_address
 
-        cmd = "cat %s '%s'" % (append and '>>' or '>', dest)
+        cmd = f"mkdir -p '{os.path.dirname(dest)}'; cat {append and '>>' or '>'} '{dest}'"
```
allisonkarlitskaya, line 426: Getting a little bit philosophical.  I start to dislike this `;` business.  This is a shells script we're writing, at the end of the day.  Why don't we do something like

```python
cmd = f"""
    mkdir -p {shlex.quote(os.path.dirname(dest))}
    cat {append and '>>' or '>'} {shlex.quote(dest)}
"""
```
(note: if you go the "one block of text" route then `shlex.quote()` in any case, please)

or even go crazy and invent a new mode for m.execute() and do it like

```python
cmd = [
    ['mkdir', '-p', os.path.dirname(dest)],
    ['tee', *['-a'] if append else [], dest]
]
```

allisonkarlitskaya: I have to admit: on first blush I found the nested lists thing to be ridiculous, but the more I think about it, the more I like it.  There's really just no other way to have multiple statements and also get quoting right without thinking about it all the time.

machine/machine_core/ssh_connection.py
martinpitt: Done.

machine/machine_core/ssh_connection.py
martinpitt: As discussed, let's not do this, at least not as part of this PR.

martinpitt:
@allisonkarlitskaya : I threw this around a few times (multi-line strings, adding `\n`, and this variant), and this both looks reasonable in `-v` output and reasonably consistent in code. WDYT?

machine/machine_core/ssh_connection.py
```diff
@@ -417,16 +417,23 @@ def write(self, dest, content, append=False, owner=None, perm=None):
...
             owner: If set, call chown on the file with the given owner string
             perm: Optional file permission as chmod shell string (e.g. "0600")
+
+        The directory of dest is created automatically.
         """
         assert dest
         assert self.ssh_address
 
-        cmd = "cat %s '%s'" % (append and '>>' or '>', dest)
+        qdest = shlex.quote(dest)
```
allisonkarlitskaya, line 426: wwg1wga

machine/machine_core/ssh_connection.py
```diff
@@ -417,16 +417,23 @@ def write(self, dest, content, append=False, owner=None, perm=None):
...
+            f"cat {append and '>>' or '>'} {qdest}",
+        ]
         if owner:
-            cmd += " && chown '%s' '%s'" % (owner, dest)
+            cmds.append(f"chown {owner} {qdest}")
         if perm:
-            cmd += " && chmod '%s' '%s'" % (perm, dest)
-        self.execute(cmd, input=content)
+            cmds.append(f"chmod {perm} {qdest}")
+        self.execute('\n'.join(cmds), input=content)
```
allisonkarlitskaya, line 436: You couldn't resist, huh? :)

I'd have preferred to see like

```python
cmd = """
  foo
  bar
"""

cmd+="baz\n"
cmd+="quux\n"
```

but this is fine too :)

allisonkarlitskaya (approved):

machine/machine_core/ssh_connection.py
allisonkarlitskaya: It's almost like it would be useful to have a variant of `m.execute()` that accepted multiple commands... ;)

allisonkarlitskaya: OR: would it really be so evil to just use separate `m.execute()` invocations?  Keep in mind that we use the SSH control channel mechanism.  It's probably lower overhead than we think...

martinpitt: I had that as well, it just looks more ugly in `-tv` output. Also this mixes multi-line with single-line and `\n`, so worst of both worlds. I have no strong opinion either way though, if you prefer that I can do that.

martinpitt: Right, separate execute() are fine. Reworked accordingly, PTAL?

machine/machine_core/ssh_connection.py
martinpitt: Not sure what that means, but if it's a complaint about my abbrev, I renamed it to `dest_quoted`.

machine/machine_core/ssh_connection.py
```diff
@@ -417,16 +417,20 @@ def write(self, dest, content, append=False, owner=None, perm=None):
...
         assert self.ssh_address
 
-        cmd = "cat %s '%s'" % (append and '>>' or '>', dest)
+        dest_quoted = shlex.quote(dest)
+
+        self.execute(f"mkdir -p {shlex.quote(os.path.dirname(dest))}")
+        self.execute(f"cat {append and '>>' or '>'} {dest_quoted}", input=content)
         if owner:
-            cmd += " && chown '%s' '%s'" % (owner, dest)
+            self.execute(f"chown {owner} {dest_quoted}")
```
allisonkarlitskaya, line 431: Sorry to do this, but why not `['chown', owner, dest]`?

allisonkarlitskaya: shell is the enemy :shrug:

martinpitt: done

allisonkarlitskaya (approved):

PR #3832, opened 2022-09-06
testmap: Test Anaconda also on fedora-rawhide

mvollmer:


mvollmer:
@M4rtinK, does this make sense?

mvollmer:
As discussed on IRC, this doesn't make sense, but #3833 does.

PR #3847, opened 2022-09-09
image-create: Don't swallow stdout of the setup scripts

mvollmer:


image-create
```diff
@@ -97,7 +97,7 @@ class MachineBuilder:
 
             try:
                 self.machine.execute(f"/var/tmp/SETUP {self.machine.image}",
-                                     environment=env, quiet=not self.machine.verbose, timeout=7200)
+                                     environment=env, stdout=sys.stdout, quiet=not self.machine.verbose, timeout=7200)
```
martinpitt, line 100: This feels like a hack. Why not fix .execute() to default to `stdout=subprocess.PIPE`, so that callers can override it to `stdout=None`? Feels cleaner to me.

mvollmer: That would change the meaning of `execute(stdout=None)` and we would need to make sure that nobody calls `execute` like that.  Is that worth it?  What about inventing `stdout=True` to mean "no-capturing"?

mvollmer: Or even `execute(capture=False)` to make it super obvious.  I'll see if I can find any users of `execute(stdout=...)`.

mvollmer: https://github.com/cockpit-project/bots/blob/7482e7d96b8cce70851fc2a3e1b6b8c1745f4e6e/vm-run#L92

Haha, so I propose we keep this hackish. :-)

martinpitt: that seems fine? `subprocess.PIPE` is -1, and all the flags are negative precisely so that they don't collide with fd numbers.

I grepped all our projects, and did find a few stdout=1.

But I don't want to dwell on this, it's not too bad.

martinpitt (approved):
Looks fine. Please run an image build to make sure that this doesn't break anything, or flood the log with too much garbage (hello, insights-client!)

image-create
mvollmer: > that seems fine?

Yes, but using a magic "1" seems even more hackish than sys.stdout, no? :)

cockpituous:
Log: https://logs.cockpit-project.org/logs/image-refresh-3847-20220909-124343/

cockpituous:
image-refresh fedora-37 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-37-20220909-130508

mvollmer:
> Looks fine. Please run an image build to make sure that this doesn't break anything, or flood the log with too much garbage (hello, insights-client!)

Done.

insights-client vomits on stderr, apparently, as can be seen in this old log that supresses stdout: https://logs.cockpit-project.org/logs/image-refresh-3830-20220906-095656/log

PR #3849, opened 2022-09-10
Image refresh for debian-stable

cockpituous:
Image refresh for debian-stable

### DIFF
--- a/images/debian-stable
+++ b/images/debian-stable
@@ -1,1 +1,1 @@
-debian-stable-ad384bc3c0ac9c7c71d8e76aac7895393423cde80522d672a4564dc3a618e2d2.qcow2
+debian-stable-49d98249feacfe89abd3ae30cbb2513c66dcef97dad25d7a6d6fd1815f7e5ba9.qcow2


martinpitt:
[storage failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3849-20220910-224701-d4f2b351-debian-stable-cockpit-project-cockpit/log.html#214-2) looks like a flake, retrying to compare.

martinpitt:
No, this is a real regression.

mvollmer:
I was only able to get this to fail once locally.  My best guess is that during the teardown phase of deletion (unmounting, removing entry from fstab), systemd has for some reason mount the partition back, and then deleting it failed.  This condition somehow persistet for the next retries.

Systemd mounting things unexpectedly is happening in other places, as well, and we try to combat it by reloading the daemon after changing fstab, which is something that should be done anyway, I guess.

So I don't think this is a regression, we just got very unlucky.

I'll retry once more here, and also make a PR to reload systemd after all tear-down actions.

mvollmer (approved):

martinpitt:
Ah, third time is the charm then!

PR #3858, opened 2022-09-13
log.html: Show also link to html dump

marusak:
It used to be there some years ago and I think it is often useful, yet only very few people know where to look for it.

Screenshot from 2022-09-13 14-48-48


martinpitt (approved):
Thanks! I trust you tested this locally.

We can ignore the cockpituous test.

Issue #3590, opened 2022-07-07
SELinux policy issues with insights-client


jelly:
Downstream reports

RHEL 8.7: https://bugzilla.redhat.com/show_bug.cgi?id=2124550
RHEL 9.1: https://bugzilla.redhat.com/show_bug.cgi?id=2124549

cockpituous:
rhel-8-6
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testSubAndInAndFail (__main__.TestSubscriptions)
Error: unknown connection 'eth1'.
Error: cannot delete unknown connection(s): 'eth1'.
Waiting 5 seconds for https://10.111.112.100:8443/candlepin
Waiting for https://10.111.112.100:8443/candlepin
Waiting for https://10.111.112.100:8443/candlepin
Waiting for https://10.111.112.100:8443/candlepin
Waiting for https://10.111.112.100:8443/candlepin
Waiting for https://10.111.112.100:8443/candlepin
Waiting for https://10.111.112.100:8443/candlepin
Waiting for https://10.111.112.100:8443/candlepin
/usr/lib/python3.6/site-packages/urllib3/connectionpool.py:847: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning)
/usr/lib/python3.6/site-packages/urllib3/connectionpool.py:847: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning)

DevTools listening on ws://127.0.0.1:9484/devtools/browser/98c60357-3cf8-4a7c-959d-abb4c3f4ac2a
[0713/101017.376378:ERROR:egl_util.cc(74)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0713/101017.379352:ERROR:viz_main_impl.cc(188)] Exiting GPU process due to errors during initialization
[...]

PR #3867, opened 2022-09-15
naughty: #3590 also happens on rhel-8-7

mvollmer:
Happens here, for example:

https://cockpit-logs.us-east-1.linodeobjects.com/pull-3128-20220915-072004-2ef4d87d-rhel-8-7/log.html


skobyda (approved):
Thanks

PR #3869, opened 2022-09-15

cockpituous:


skobyda:
I have a feeling that this might be already existing naughty, just selinux message changed a bit in a never version of NetworkManager:
https://bugzilla.redhat.com/show_bug.cgi?id=2112782

martinpitt:
@skobyda: Even easier -- this is https://bugzilla.redhat.com/show_bug.cgi?id=2122888 , and machine/machine_core/machine.py already has a pattern for it, it just needs to be applied to fedora-testing as well.

skobyda (approved):

PR #3881, opened 2022-09-20
images: only pull the pause container when required

jelly:
Podman 4.0 ships the pause image so only pull it when required.


skobyda:
We some cockpit-podman tests and pixel screenshots because docker://k8s.gcr.io/pause:3.5 was present.
Perhaps we should run cockpit-podman tests for this PR to see it doesn't affect them?

jelly:
> We changed some cockpit-podman tests and pixel screenshots because docker://k8s.gcr.io/pause:3.5 was present. Perhaps we should run cockpit-podman tests for this PR to see it doesn't affect them?

Pixel tests run on Fedora-35, which should still have the k8s image.

skobyda (approved):

PR #3883, opened 2022-09-20

cockpituous:


martinpitt:
@M4rtinK : ERR_CERT_AUTHORITY_INVALID still sounds like the old problem?

martinpitt:
Ah, it pulled yesterday's image, it needs the one from Sep 20. Let's try again tomorrow.

PR #3890, opened 2022-09-21
naughty: consolidate & generalize denial patterns for Insights

ptoscano:
Sadly, it looks like fixing the SELinux policy for insights-client is
not going to happen soon, and new issues (related to new tools used by
insights-client for data collection) happen every now and then.
Hence, consolidate all the RHEL 8 & RHEL 9 naughties of insights-client
into a single one that basically accepts any failure in a
`insights_client_t` context.


mvollmer (approved):
I like this!

martinpitt (approved):
Thanks! Just retrying unrelated flakes.

PR #3876, opened 2022-09-23
Image refresh for fedora-37

cockpituous:
Image refresh for fedora-37
 * [x] image-refresh fedora-37

### DIFF
--- a/images/fedora-37
+++ b/images/fedora-37
@@ -1,1 +1,1 @@
-fedora-37-af9203aea2b321bb23acc76caf55b7c90caf651ed469be41334d8f1b73fbb3bb.qcow2
+fedora-37-28dcf3f7c9c31e58c66df10da5c6c3e9db2bb07c16c02ec3ccf662435c1e736b.qcow2
--- a/naughty/fedora-37/3899-invalid-time-value
+++ b/naughty/fedora-37/3899-invalid-time-value
@@ -1,0 +1,1 @@
+RuntimeError: RangeError: Invalid time value


martinpitt:
RuntimeError: RangeError: Invalid time value

in `TestAccounts.testExpire`, this needs investigation

skobyda:
I tried debugging fedora-37 check-system-realms failures, and they all fail with ssh error "no-host" when trying to add "x0.cockpit.lan" to hosts. 
Weirdly, if I try to open terminal during the VM, and try `ssh x0.cockpit.lan`, it works, so maybe issue might be with our cockpit's ssh code.
Logs also show this error `Sep 23 09:25:30 x0.cockpit.lan cockpit-ssh[5772]: cockpit-ssh x0.cockpit.lan: -1 couldn't connect: Socket error: Connection reset by peer 'x0.cockpit.lan' '22'`

martinpitt:
@skobyda : The big change here is the upgrade of libssh from 0.9. to 0.10.4. That represents some 1.5 years of development. Indeed that had some [other issue](https://gitlab.com/libssh/libssh-mirror/-/issues/145) already, but I managed to prevent that one from landing. debian-testing got the new version already and passed, but we don't run the realms tests there as that lacks FreeIPA. I.e. this is *very* likely an issue with libssh/cockpit-ssh.

martinpitt:
I retry the tests which failed due to the broken npm cache. See https://github.com/cockpit-project/cockpituous/pull/521

marusak:
Let's redo this, the naughty has been fixed in the meantime as well

Issue #3899, opened 2022-09-23
User expiration fails with 'Invalid time value'


skobyda:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2129336
Downstream report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1021697

Upstream report: https://github.com/shadow-maint/shadow/commit/e503fd574b7dbf6b21b1168e20938f0922807916

cockpituous:
fedora-37
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testExpire (__main__.TestAccounts)
[0929/100258.922215:ERROR:bus.cc(398)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
[0929/100258.922391:ERROR:bus.cc(398)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory

DevTools listening on ws://127.0.0.1:9372/devtools/browser/5667055b-c48e-44df-b39a-55e5f4908fb6
[0929/100258.925783:WARNING:bluez_dbus_manager.cc(247)] Floss manager not present, cannot set Floss enable/disable.
[0929/100258.925760:ERROR:egl_util.cc(55)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0929/100258.927634:ERROR:viz_main_impl.cc(186)] Exiting GPU process due to errors during initialization
[0929/100258.932891:ERROR:egl_util.cc(55)] Failed to load GLES library: /usr/lib64/chromium-browser/libGLESv2.so: /usr/lib64/chromium-browser/libGLESv2.so: cannot open shared object file: No such file or directory
[0929/100258.934545:ERROR:viz_main_impl.cc(186)] Exiting GPU process due to errors during initialization
[0929/100258.937679:ERROR:gpu_init.cc(486)] Passthrough is not supported, GL is disabled, ANGLE is 
[0929/100259.618313:ERROR:devtools_http_handler.cc(636)] Using unsafe HTTP verb GET to invoke /json/new. This action will stop supporting GET and POST verbs in future versions.
[...]

PR #3900, opened 2022-09-23
Add naughty for 'Invalid time value'

skobyda:
Relevant to one of the failures at https://github.com/cockpit-project/bots/pull/3876
Issue: #3899


martinpitt (changes requested):
Ouch, what a sad regression in shadow-utils -- one would think that'd be caught by some tests? Thanks for filing!

However, please add that naughty to #3876 so that it can test itself.

PR #3929, opened 2022-10-02
Image refresh for fedora-35

cockpituous:
Image refresh for fedora-35
 * [x] image-refresh fedora-35


marusak:
Why are we even still testing 35, we have 37 now, I guess we can start dropping it.

marusak:
@jkozol tests here try to pick `Fedora IoT Commit (.tar)` but the list does not contain it as seen for example [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-3929-20221002-231125-238a82d7-fedora-35-osbuild-cockpit-composer/TestWizard-testOSTreeStep-fedora-35-127.0.0.2-2401-FAIL.png)

Any idea why is this missing? Were there any changes?

jkozol:
@marusak. Hmmm, strange. I will look into that but in general we no longer need to support f35 for cockpit-composer. [This PR](https://github.com/osbuild/cockpit-composer/pull/1568) will bump our release script and as you can see in the tests I am preparing to bump the testmap

PR #3935, opened 2022-10-05
Drop Fedora 35

marusak:


marusak:
I will then do cleanups in projects if they have 35 special cases

jelly:
When do we stop testing Fedora 35? It's not EOL yet according to these docs https://docs.fedoraproject.org/en-US/releases/eol/

jelly (approved):

PR #3944, opened 2022-10-07
Investigate/fix tests

martinpitt:
@jelly runs into testing trouble in #3735  . Let's confirm that they actually broke on main, and then investigate separately.


martinpitt:
Hmm, so main is actually fine. Bisect o'clock in #3735 then.

PR #3950, opened 2022-10-10
task: update legacy url

jelly:
The commit statuses endpoint still used the legacy route which makes it hard to find it in the Github docs and surely Github will drop this at some point.

See https://docs.github.com/en/rest/commits/statuses#list-commit-statuses-for-a-reference for the mention of the legacy url. This is used in store-tests, so should I trigger a test job somewhere?

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -324,7 +324,7 @@ class GitHub(object):
         page = 1
         count = 100
         while count == 100:
-            data = self.get("statuses/{0}?page={1}&per_page={2}".format(revision, page, count))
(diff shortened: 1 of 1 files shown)


marusak:
this won;t be self testing...  I guess we can just risk it and land it :)

marusak:
Also it is not tested if I am not wrong, but would not hurt to push from upstream so we can actually run tests.

jelly:
Ok, opened https://github.com/cockpit-project/bots/pull/3956

PR #3956, opened 2022-10-11
task: update legacy url

jelly:
The commit statuses endpoint still used the legacy route which makes it hard to find it in the Github docs and surely Github will drop this at some point.


marusak (approved):
Thank you!

PR #3957, opened 2022-10-11
testmap: update cockpit composer testing

jkozol:
https://github.com/osbuild/cockpit-composer/pull/1568

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -126,19 +126,16 @@ REPO_BRANCH_CONTEXT = {
     'osbuild/cockpit-composer': {
         'main': [
             'fedora-36',
-            'fedora-36/firefox',
-            'rhel-8-7',
-            'rhel-9-1',
-        ],
-        # These can be triggered manually with bots/tests-trigger
-        '_manual': [
             'fedora-37',
             'fedora-37/firefox',
-            'fedora-rawhide',
             'rhel-8-8',
             'centos-8-stream',
             'rhel-9-2',
             'centos-9-stream',
+        ],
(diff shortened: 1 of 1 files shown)


lib/testmap.py
```diff
@@ -128,15 +128,16 @@
     },
     'osbuild/cockpit-composer': {
         'main': [
-            'fedora-35',
             'fedora-36',
-            'fedora-36/firefox',
-            'rhel-8-7',
-            'rhel-9-1',
+            'fedora-37',
```
marusak, line 132: Neither of these new tests ran in https://github.com/osbuild/cockpit-composer/pull/1568
Wouldn;t it be better to add them to manual, trigger them in the given PR and then enable them in main?

jkozol: Do you have any idea why the rhel tests aren't running but fedora-37 is? I just trigger them using my local checkout of bots with the updated testmap. That seems to work fine. Or is that why the tests are stuck pending?

marusak:
I see these context triggered in https://github.com/osbuild/cockpit-composer/pull/1568 and they are red. Marking this as blocked.

jkozol:
@marusak Most tests are passing now but there seems to be some issue with the fedora-37 image. Whatever repo is being used is only pulling osbuld-composer 61 even though 67 is in stable: https://bodhi.fedoraproject.org/updates/FEDORA-2022-d62111bac7

marusak:
We refreshed 37 5 days ago. The update you linked has been pushed to stable 5 days ago as well.

jkozol:
Can you click merge for me?

PR #3966, opened 2022-10-14

cockpituous:


M4rtinK:
> @M4rtinK any idea what is going wrong here?

Thanks for the heads-up, will take a look. :)

M4rtinK:
So looks like this is a wider Rawhide regression, with SSH access to even regular images does not work if inst.sshd is passed to Anaconda. 

I've tested an Oct 5 image and that still works. Given that the image refresh boot iso is from Oct 13, it was broken already back then. So there is quite a thin time-frame in which this stopped working.

Will try to bisect further. :)

M4rtinK:
Aaand its a rawhide bug, where its simply no longer possible to SSH into the image after passing inst.sshd for some reason:
https://bugzilla.redhat.com/show_bug.cgi?id=2136916

M4rtinK:
This is caused by a change in openssh, the bug has been reassigned to openssh and the openssh team is aware of the issue & is looking into it, time permitting. Will continue to monitor this. :)

martinpitt:
This is becoming really stale, and thus Anaconda testing incrementally useless. Is it possible to add a workaround to anaconda or the image itself?

M4rtinK:
> This is becoming really stale, and thus Anaconda testing incrementally useless. Is it possible to add a workaround to anaconda or the image itself?

Yeah, the timing is not good on this. :P I'll try to ping the openssh maintainers again, but that still realistically means even if finally fixed a ~week until it percolates to compose & then to mirrors - not great. :package: 

As for a hotfix - we need to be careful not to break openssh by whatever we do, as Anaconda ends up installed also on peoples systems for various reasons. But I think we could inject something via the image-reparation machinery in a safe manner - eq. for example with this:

https://github.com/rhinstaller/anaconda/blob/master/ui/webui/test/prepare-updates-img

Inject some hotfix RPM fetched from COPR or our fedorapeople web space which would override the offending openssh service files could do the job.

M4rtinK:
So I'll try to put something together. :P

M4rtinK:
It has been decided to go for a fix/hotfix in Lorax: 

https://github.com/weldr/lorax/pull/1283

That should hopefully get us rid of this issue in the next few days - once it hits the Lorax package & when next Rawhide compose is built with this Lorax version.

This is because [the bug](https://bugzilla.redhat.com/show_bug.cgi?id=2136916) that openssh managed to trigger is apparently deeper than we though & has been reasigned to systemd, as its still unclear how the given change in openssh actually activates the sshd socket.

martinpitt:
lorax with the workaround is in rawhide now: https://bodhi.fedoraproject.org/updates/FEDORA-2022-35f9f24baa

So let's retry the refresh.

cockpituous:
image-refresh in progress on 3-fedora.
Log: https://logs.cockpit-project.org/logs/image-refresh-3966-20221111-052626/

cockpituous:
image-refresh fedora-rawhide-boot done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-rawhide-boot-20221111-053148

martinpitt (approved):
Look at that, green! :four_leaf_clover: Thanks @M4rtinK for investigating and fixing!

PR #3974, opened 2022-10-19
Image refresh for fedora-37

cockpituous:
Image refresh for fedora-37
 * [x] image-refresh fedora-37
 * [x] https://github.com/cockpit-project/cockpit/pull/17833
 * [x] https://github.com/cockpit-project/cockpit/pull/17834

PR #3982, opened 2022-10-19
lib: handle empty files

jelly:
When the `image-stores` file is empty, we still add an entry to the store as in Python `''.split('\n')` outputs `['']`. This breaks image-download which does not check if the parsed url is valid and has a hostname which is not None.

Tested here: https://github.com/cockpit-project/bots/pull/3974


martinpitt (approved):
Ouch, that was my fault! Thanks!

PR #3913, opened 2022-09-28
arch: disable systemd-time-wait-sync

jelly:
This causes a 30 second delay as enabling systemd-time-wait-sync seems to cause systemd-timesyncd to hit a timeout of 30 seconds. As we run our images offline for tests, we want to disable this anyway.

 * [x] image-refresh arch

PR #3987, opened 2022-10-20
Fix Arch Linux Refresh

jelly:
Detailed description [here](https://github.com/cockpit-project/bots/pull/3913#issuecomment-1285403269), but the TL;DR is our test environment blocks NTP which makes the first boot hang on Arch Linux as sshd.service is never started as systemd-timesyncd hangs on reaching a NTP server.

Tests succeeded here, but as the PR needs to be made against `origin` I opened this one https://github.com/cockpit-project/bots/pull/3913


martinpitt (approved):
Niice, thank you!

PR #3992, opened 2022-10-21
s3: retry 503 errors on S3

### LINKED ISSUE #3991: s3-streamer crashes with HTTP Error 503: Slow Down
We are currently getting tons of tasks container crashes due to this:
```
   File "/work/bots/./s3-streamer", line 225, in write
     self.append_block(self.pending)
   File "/work/bots/./s3-streamer", line 205, in append_block
     self.destination.write(f'{self.filename}.chunks', json.dumps(chunk_sizes).encode('ascii'))
   File "/work/bots/./s3-streamer", line 97, in write
     with s3.urlopen(self.url(filename), data=data, method='PUT', headers=headers) as _:
   File "/work/bots/lib/s3.py", line 117, in urlopen
     response = urllib.request.urlopen(request, context=host_ssl_context(url.netloc))
   File "/usr/lib64/python3.10/urllib/request.py", line 216, in urlopen
     return opener.open(url, data, timeout)
   File "/usr/lib64/python3.10/urllib/request.py", line 525, in open
     response = meth(req, response)
   File "/usr/lib64/python3.10/urllib/request.py", line 634, in http_response
     response = self.parent.error(
   File "/usr/lib64/python3.10/urllib/request.py", line 563, in error
     return self._call_chain(*args)
   File "/usr/lib64/python3.10/urllib/request.py", line 496, in _call_chain
     result = func(*args)
   File "/usr/lib64/python3.10/urllib/request.py", line 643, in http_error_default
     raise HTTPError(req.full_url, code, msg, hdrs, fp)
 urllib.error.HTTPError: HTTP Error 503: Slow Down
ERROR:root:./s3-streamer --repo cockpit-project/cockpit-machines --test-name pull-835-20221021-145814 --github-context centos-8-stream --revision 6bd666705c12a02c8189af45fd4349f89100877d -- /bin/sh -c "PRIORITY=0005 ./make-checkout --verbose --repo=cockpit-project/cockpit-machines --rebase=main 6bd666705c12a02c8189af45fd4349f89100877d && cd make-checkout-workdir && TEST_OS=centos-8-stream BASE_BRANCH=main COCKPIT_BOTS_REF=main ../tests-invoke --pull-number 835 --revision 6bd666705c12a02c8189af45fd4349f89100877d --repo cockpit-project/cockpit-machines" failed with exit code 1
```

@allisonkarlitskaya does this ring a bell? Are we hitting some quota?

allisonkarlitskaya:
According to the S3 documentation[1] we should retry if we get 503 errors.

The default configuration for various other S3 tools is 3 tries, so let's do the same.

Fixes #3991

[1] https://aws.amazon.com/premiumsupport/knowledge-center/s3-resolve-503-slowdown-throttling/

### DIFF
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -111,16 +111,22 @@ def sign_curl(url: urllib.parse.ParseResult, method='GET', headers={}, checksum=
 
 def urlopen(url: urllib.parse.ParseResult, method='GET', headers={}, data=b'') -> IO:
     """Same as sign_request() but calls urlopen() on the result"""
-    headers = sign_request(url, method, headers, hashlib.sha256(data).hexdigest())
-    request = urllib.request.Request(url.geturl(), headers=headers, method=method, data=data)
-    try:
-        response = urllib.request.urlopen(request, context=host_ssl_context(url.netloc))
-    except urllib.error.HTTPError as exc:
-        logger.debug('%s %s %s → %s:', method, url.geturl(), headers, exc.status)
-        logger.debug('  %s', exc.read())
-        raise
-
-    return response
+    tries = 3
+    while True:
+        headers = sign_request(url, method, headers, hashlib.sha256(data).hexdigest())
+        request = urllib.request.Request(url.geturl(), headers=headers, method=method, data=data)
+        try:
+            return urllib.request.urlopen(request, context=host_ssl_context(url.netloc))
+        except urllib.error.HTTPError as exc:
+            logger.debug('%s %s %s → %s:', method, url.geturl(), headers, exc.status)
+            logger.debug('  %s', exc.read())
+            if exc.status == 503:
+                tries -= 1
+                if tries != 0:
+                    # First wait 1s, then wait 10s
+                    time.sleep(10 ** (2 - tries))
+                    continue
+            raise
 
 
 def list_bucket(url: urllib.parse.ParseResult) -> ET:


martinpitt:
Hmm, [test fails](https://github.com/cockpit-project/bots/actions/runs/3298876106/jobs/5441559720) in S3, that's suspicious.

martinpitt:
This reproduces locally with

    COCKPIT_BOTS_BRANCH=idiot-slowdown tasks/run-local.sh

in cockpituous (and works fine with main).

lib/s3.py
```diff
@@ -111,14 +111,20 @@ def sign_curl(url: urllib.parse.ParseResult, method='GET', headers={}, checksum=
...
-    except urllib.error.HTTPError as exc:
-        logger.debug('%s %s %s → %s:', method, url.geturl(), headers, exc.status)
-        logger.debug('  %s', exc.read())
-        raise
+    tries = 3
+    while True:
+        headers = sign_request(url, method, headers, hashlib.sha256(data).hexdigest())
+        request = urllib.request.Request(url.geturl(), headers=headers, method=method, data=data)
+        try:
+            response = urllib.request.urlopen(request, context=host_ssl_context(url.netloc))
```
martinpitt, line 119: This needs to be returned. Otherwise, it will indefinitely re-do the loop on success, eventually leading to 403.

lib/s3.py
```diff
@@ -111,14 +111,20 @@ def sign_curl(url: urllib.parse.ParseResult, method='GET', headers={}, checksum=
...
+        except urllib.error.HTTPError as exc:
+            logger.debug('%s %s %s → %s:', method, url.geturl(), headers, exc.status)
+            logger.debug('  %s', exc.read())
+            if exc.status == 503:
+                tries -= 1
+                if tries != 0:
+                    continue
```
martinpitt, line 126: Should we really retry *immediately* here? Usually there's some exponential back-off, like `time.sleep(2**2**i)` (with `i = 3 - tries`, or you write the loop the other way around, etc.)

martinpitt (changes requested):
I validated that with the correct `return` this works fine. However, I'd like to at least discuss the sleep.

lib/s3.py
allisonkarlitskaya: I contemplated this.  I didn't do it, since this function isn't documented as "might sleep", but on the other hand, it's not exactly expected to be fast, either.

lib/s3.py
allisonkarlitskaya: I thought I had that in there.  I wonder how it got dropped...

allisonkarlitskaya: Fixed.

lib/s3.py
allisonkarlitskaya: I added a sleep of `10 ** (2 - tries)` which works out to 1s on the first iteration and 10s on the second.

martinpitt (approved):
Thanks!

PR #4006, opened 2022-10-27
Image refresh for fedora-36

cockpituous:
Image refresh for fedora-36
 * [x] image-refresh fedora-36


jelly:
```
ls: cannot access '*.src.rpm': No such file or directory
```

Source rpm can't be build for cockpit-ostree, while it works locally for me.

martinpitt:
c-ostree fix landed, retried tests.

martinpitt (approved):

PR #4012, opened 2022-10-31

cockpituous:

### DIFF
--- a/images/fedora-testing
+++ b/images/fedora-testing
@@ -1,1 +1,1 @@
-fedora-testing-162e66d8a6cbed96cb9c36d4a9be311a188fbff17d547432f05d384ace77764c.qcow2
+fedora-testing-69d9305f5312642404faebd1d3ba1c24b2a4543461546c0436fbe0f35aebdb39.qcow2


jelly:
Both tests failed with:

```Traceback (most recent call last):
  File "/work/bots/make-checkout-workdir/test/verify/check-system-info", line 821, in testOverview
    wait(lambda: progressValue(1) < 20)
  File "/work/bots/make-checkout-workdir/test/common/testlib.py", line 2339, in wait
    raise Error(msg or "Condition did not become true.")
testlib.Error: Condition did not become true.
```

So that doesn't seem like a flake.

jelly:
This isn't a flake `pmie.service` uses a ton of CPU

image

So maybe we mask it and unmask?

martinpitt:
The [damage is done now](https://bodhi.fedoraproject.org/updates/FEDORA-2022-18bbe14f70), pcp went into stable and can't be -1'ed any more. So either this was a temporary glitch (unlikely, though), or it will now break the next fedora-37 image refresh.

jelly:
> The [damage is done now](https://bodhi.fedoraproject.org/updates/FEDORA-2022-18bbe14f70), pcp went into stable and can't be -1'ed any more. So either this was a temporary glitch (unlikely, though), or it will now break the next fedora-37 image refresh.

Yes, sorry for not following up quickly enough. I'll report a bugzilla

naughty/fedora-36/4031-pmie-check-cpu-usage
```diff
@@ -0,0 +1,6 @@
+Traceback (most recent call last):
+  File "/work/bots/make-checkout-workdir/test/verify/check-system-info", line *, in testOverview
+    wait(lambda: progressValue(1) < 20)
```
marusak, line 3: This is bit too generic as there might be a lot of reasons why the load does not come under 20.  I would prefer if we would just disable `pmie.service`. Or is there some reason why we cannot do that?

marusak: right, I was thinking about editing of the test. Good point about disabling pcp service in metrics test but this naughty does not say anything about what consumes the processor. If there is another service in the meantime we will still be blocking this one naughty and the test might never run to completion.

jelly: So instead of checking the overall CPU, check the table of processes?

jelly: Ok, created the workaround here https://github.com/cockpit-project/cockpit/pull/17892

jelly (approved):

PR #4026, opened 2022-11-04
Image refresh for arch

cockpituous:
Image refresh for arch
 * [x] image-refresh arch


jelly:
cockpit does not seem to listen on 172.27.0.15 now, which is kinda strange?
```
tcp       LISTEN     0          10                    127.0.0.1:9000              0.0.0.0:*         users:(("cockpit-ws",pid=533,fd=3))
```

jelly:
Curl output changed and I didn't notice. This is now handled in https://github.com/cockpit-project/cockpit/pull/17894

After this is merged we should enable the Arch stratis tests.

jelly (approved):

PR #4030, opened 2022-11-07
machine: Support adding a new boot disk

mvollmer:
This is done by using per-disk <boot> elements instead of a global one, and allowing a new disk to be ordered in front of the original boot disk.


martinpitt (approved):
Makes sense, thank you!

PR #4040, opened 2022-11-09
naughty: also match bond issues

jelly:
In TestBondingVirt this violation also randomly occurs but as name="22_clhm_tbond.issue".


martinpitt (approved):
Danke!

PR #4048, opened 2022-11-11
Fix CodeQL scan HTML/JavaScript alerts

martinpitt:
This takes care of all relevant HTML/JS alerts in PR #4039. Let's land this first and then re-run, to avoid losing the alerts the next time (due to comparison with the previous scan).

I tested this locally with downloading the raw log from [this test](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4044-20221110-231108-a4e66841-fedora-37-cockpit-project-cockpit/log.html#71-2), as well as downloading https://logs-https-frontdoor.apps.ocp.ci.centos.org/test-results.db and putting that next to tests.html; and then `python3 -m http.server` and checking http://localhost:8000/lib/s3-html/log.html and localhost:8000/tests.html


marusak (approved):
Thank you!

PR #4049, opened 2022-11-11
Fix CodeQL scan Python alerts

martinpitt:
This addresses all Python alerts in PR #4039.

 * [x] image-refresh cirros

:arrow_double_up:  not actually going to use/commit that image refresh, just proving that `image-refresh` still works

### DIFF
--- a/image-download
+++ b/image-download
@@ -224,11 +224,12 @@ def committed_target(image):
     # The image file in the images directory, may be same as dest
     image_file = os.path.join(IMAGES_DIR, os.readlink(link))
 
-    # Double check that symlink in place but never make a cycle.
+    # Double check that symlink in place
     if os.path.abspath(dest) != os.path.abspath(image_file):
         try:
             os.symlink(os.path.abspath(dest), image_file)
         except FileExistsError:
+            # .. but never make a cycle
             pass
 
     return dest
--- a/image-download
+++ b/image-download
@@ -49,7 +49,6 @@ from lib.stores import PUBLIC_STORES, REDHAT_STORES
 from lib.testmap import get_test_image
 from lib import s3
 
-DEVNULL = open("/dev/null", "r+")
 EPOCH = "Thu, 1 Jan 1970 00:00:00 GMT"
 
 if os.isatty(1):
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -68,7 +68,10 @@ class Machine(ssh_connection.SSHConnection):
         if not browser:
             browser = address
 
-        super(Machine, self).__init__(user, ssh_address, ssh_port, identity_file, verbose=verbose)
+        if not label and image != "unknown":
+            label = "{}-{}-{}".format(image, ssh_address, ssh_port)
+
+        super(Machine, self).__init__(user, ssh_address, ssh_port, identity_file, verbose=verbose, label=label)
 
         self.arch = arch
         self.image = image
@@ -78,12 +81,6 @@ class Machine(ssh_connection.SSHConnection):
         else:
             self.web_address = browser
             self.web_port = web_port
-        if label:
-            self.label = label
-        elif self.image != "unknown":
-            self.label = "{}-{}-{}".format(self.image, self.ssh_address, self.ssh_port)
-        else:
-            self.label = "{}@{}:{}".format(self.ssh_user, self.ssh_address, self.ssh_port)
 
         # The Linux kernel boot_id
         self.boot_id = None
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -104,7 +104,7 @@ class Machine(ssh_connection.SSHConnection):
         """Overridden by machine classes to stop the machine"""
         self.message("Not shutting down already running machine")
 
-    def wait_poweroff(self):
+    def wait_poweroff(self, timeout_sec=120):
         """Overridden by machine classes to wait for a machine to stop"""
         assert False, "Cannot wait for a machine we didn't start"
 
@@ -112,7 +112,7 @@ class Machine(ssh_connection.SSHConnection):
         """Overridden by machine classes to unconditionally kill the running machine"""
         assert False, "Cannot kill a machine we didn't start"
 
-    def shutdown(self):
+    def shutdown(self, timeout_sec=120):
         """Overridden by machine classes to gracefully shutdown the running machine"""
         assert False, "Cannot shutdown a machine we didn't start"
 
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -46,7 +46,7 @@ class SSHConnection(object):
     ssh_default_opts = ["-o", "StrictHostKeyChecking=no", "-o", "UserKnownHostsFile=/dev/null",
                         "-o", "IdentitiesOnly=yes", "-o", "BatchMode=yes", "-o", "PKCS11Provider=none"]
 
-    def __init__(self, user, address, ssh_port, identity_file, verbose=False):
+    def __init__(self, user, address, ssh_port, identity_file, verbose=False, label=None):
         self.verbose = verbose
 
         # Currently all images are x86_64. When that changes we will have
@@ -58,7 +58,7 @@ class SSHConnection(object):
         self.ssh_master = None
         self.ssh_process = None
         self.ssh_reachable = False
-        self.label = "{}@{}:{}".format(self.ssh_user, self.ssh_address, self.ssh_port)
+        self.label = label if label else "{}@{}:{}".format(self.ssh_user, self.ssh_address, self.ssh_port)
 
     def disconnect(self):
         self.ssh_reachable = False
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -89,8 +89,7 @@ class SSHConnection(object):
                         self.ssh_reachable = True
                         return True
                 except IOError:
-                    pass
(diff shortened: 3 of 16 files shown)


cockpituous:
image-refresh in progress on rhos-01-16.
Log: https://logs.cockpit-project.org/logs/image-refresh-4049-20221111-090909/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20221111-091012

cockpituous:
image-refresh in progress on 3-fedora.
Log: https://logs.cockpit-project.org/logs/image-refresh-4049-20221111-095555/

martinpitt:
OK, here it is, all "empty excepts" fixed as well. *phew*

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20221111-095551

task/test-github
```diff
@@ -60,7 +60,11 @@ def mockServer():
             self.wfile.write(value.encode('utf-8'))
             self.wfile.flush()
 
-        def replyJson(self, value, headers={}, status=200):
+        def replyJson(self, value, headers=None, status=200):
+            if headers:
```
allisonkarlitskaya, line 63: I'm not a huge fan of the shadowing here.

In any case I feel like this is candidate for a one-liner

```
headers = headers.copy() if headers else {}
```

I also sort of prefer this pattern: `new=dict(old, key=val)` but because "Content-type" is not a valid kwarg name, we can't do it...

Maybe something like

```
our_headers = { "Content-type": "application/json" }

... self.replyData(... headers=dict(headers, **our_headers))
```
?

Or just inline the entire thing...

image-download
```diff
@@ -236,11 +236,10 @@ def committed_target(image):
 def wait_lock(target):
     lockfile = os.path.join(os.path.dirname(target), ".lock." + os.path.basename(target))
 
-    # we need to keep the lock fd open throughout the entire runtime, so remember it in a global-scoped variable
-    wait_lock.f = open(lockfile, "w")
+    lock_fd = open(lockfile, "w")
```
allisonkarlitskaya, line 240: Please don't call this `fd` — it's not.  It's makes the `fd.close()` line below very confusing.

image-download
```diff
@@ -279,12 +278,13 @@ def download_images(image_list, force, quiet, state, store):
...
 
             # don't download the same thing multiple times in parallel
-            lockfile = wait_lock(target)
+            lockfile, lock_fd = wait_lock(target)
 
             if force or state or not os.path.exists(target):
                 download(target, force, state, quiet, store)
 
             # clean up lock file; parallel runs may remove it already
+            lock_fd.close()
```
allisonkarlitskaya, line 288: Strictly speaking, that should live in a `finally:` block.

How about using the `@contextmanager` decorator with a coroutine and a `with:` statement?  I feel like that would be all-around safer, more foolproof, and also easier to read.

task/test-github
```diff
@@ -179,7 +179,7 @@ class TestGitHub(unittest.TestCase):
 
     def testAllowlist(self):
         allowlist = self.api.allowlist()
-        self.assertTrue(len(allowlist) > 0)
```
allisonkarlitskaya, line 178: I wonder about these.  It seems to me that `pytest` has recently gained some magical ability to dissect normal `assert a > b` statements and get the values out...

But in any case, `assertGreater(a, b)` is definitely more correct than `assertTrue(a < b)`

machine/machine_core/ssh_connection.py
```diff
@@ -145,7 +145,7 @@ def wait_reboot(self, timeout_sec=180):
                 if self.boot_id != boot_id:
                     break
             except exceptions.Failure:
-                pass
+                pass  # try again
```
marusak, line 147: any comment makes codeQL happy? :)

marusak (approved):
Danke!

task/test-github
```diff
@@ -111,19 +111,19 @@ def mockServer():
...
-    # prctl(PR_SET_PDEATHSIG, SIGTERM)
-    try:
-        libc = ctypes.CDLL('libc.so.6')
-        libc.prctl(1, 15)
-    except OSError:
-        pass
-
-    httpd.serve_forever()
-    os._exit(1)
+    if child == 0:
```
allisonkarlitskaya, line 113: This code is horrifying, considering why codeql thinks that we might return from this function in the child (and I can consider another case: an exception).

Can you change this to use a thread or `multiprocessing` module or something?

machine/machine_core/ssh_connection.py
martinpitt: That's at least what the [detailed view claims](https://github.com/cockpit-project/bots/security/code-scanning/29). Once this lands, I'll rebase #4039, then we'll see whether it really worked. :grin:

store-tests
```diff
@@ -72,7 +72,7 @@ def main():
     statuses = api.all_statuses(opts.revision)
     if not statuses:
         logging.warning("Revision %s has no statuses", opts.revision)
-        return 0
+        return
```
allisonkarlitskaya, line 75: This case is probably better written as `return f"Revision {opts.revision} has no statuses"`.  In case `sys.exit()` gets a string, it'll write it to stderr and exit non-zero.

allisonkarlitskaya (changes requested):
Thanks for a huge round of cleanups.

Most of these are good.  Comments below with some suggested improvements.

store-tests
martinpitt: But then we again have the case that some exit paths return a value and some don't. Also, this is a potential breakage in our bots (if the command fails). I can either take out the commit, and block #4039 further, or we do this in a follow-up?

task/test-github
martinpitt: Reworked as per IRC.

image-download
martinpitt: Dropped that commit as you wanted to rework this.

image-download
martinpitt: Dropped that commit as you wanted to rework this.

store-tests
allisonkarlitskaya: Conceptually is this an error, or a valid case where we merely want to issue a warning?

If it's the latter, then probably we should leave the code as it is..

In any case, if our `main()` function is always returning `None` then it makes absolutely no sense to pass its return value to `sys.exit()`.  So drop the `sys.exit()` wrapping of `main()` at the bottom of the file.  And if you want to deal with this case as an error then you can just use `sys.exit()` directly here...

martinpitt: I can't say off-hand whether this should be fatal and crash the bot. My gut feeling is "fatal", but that'd be a behaviour change. But I'm willing to try. Reworked.

task/cache.py
```diff
@@ -55,13 +56,13 @@ def prune(self):
         if not os.path.exists(self.directory):
             return
         now = time.time()
-        try:
-            for filename in os.listdir(self.directory):
-                path = os.path.join(self.directory, filename)
-                if os.path.isfile(path) and os.stat(path).st_mtime < now - 7 * 86400:
+        for filename in os.listdir(self.directory):
```
allisonkarlitskaya, line 59: Missed this before.

This is screaming out for porting to `scandir`, which would let you avoid the pathname computations and also the stat call for `os.path.isfile()`.  Note also the `.stat()` method on the returned object (but no `.unlink()`).

task/test-github
martinpitt: Resolved in #4050

cockpituous:
image-refresh in progress on 1-fedora.
Log: https://logs.cockpit-project.org/logs/image-refresh-4049-20221111-114242/

cockpituous:
image-refresh cirros done: https://github.com/cockpit-project/bots/commits/image-refresh-cirros-20221111-114250

store-tests
```diff
@@ -72,7 +72,7 @@ def main():
     statuses = api.all_statuses(opts.revision)
     if not statuses:
         logging.warning("Revision %s has no statuses", opts.revision)
-        return 0
+        sys.exit(1)
```
allisonkarlitskaya, line 75: I mean to write here `sys.exit(f"Revision {opts.revision} has no statuses")`

store-tests
```diff
@@ -256,4 +256,4 @@ def test_parser(logfile):
 
 
 if __name__ == '__main__':
-    sys.exit(main())
+    main()
```
allisonkarlitskaya, line 259: :+1:

allisonkarlitskaya (changes requested):

store-tests
martinpitt: Hmm, I really like use `logging` consistently. We get time stamps in the logs and all that. How strongly do you feel about this?

martinpitt: Done

task/cache.py
martinpitt: Done in #4051, I'll drop this commit.

machine/machine_core/machine.py
```diff
@@ -68,7 +68,10 @@ def __init__(self, address="127.0.0.1", image="unknown", verbose=False, label=No
         if not browser:
             browser = address
 
-        super(Machine, self).__init__(user, ssh_address, ssh_port, identity_file, verbose=verbose)
+        if not label and image != "unknown":
+            label = "{}-{}-{}".format(image, ssh_address, ssh_port)
```
allisonkarlitskaya, line 72: I mean, once more around if you feel like turning that into an f-string, but otherwise I wouldn't worry...

allisonkarlitskaya (approved):
Thanks for the iteration!

PR #4051, opened 2022-11-11
task: clean up and robustify cache pruning

allisonkarlitskaya:
Make some improvements to the cache pruning code:

 - avoid a very pointless check for the existence of the cache directory: when the prune() function is called from write() we'll have had just created the directory.  Meanwhile, if our concern is that the directory may have been deleted out from under us, then checking it in this way is still a TOCTOU: rather just try to open it and handle the exception.  That's more Pythonic, anyway.

 - use os.scandir() instead of os.listdir(), allowing us to avoid an extra stat() syscall per item, as well as removing the need to do manual path calculations

 - improve handling of failing os.remove() calls.  Previously, a failing call would interrupt the entire process, and print no error of any kind.  Now we catch the exception per-file, and display a warning, before proceeding on to the next file.

More split-outs from #4049

### DIFF
--- a/task/cache.py
+++ b/task/cache.py
@@ -21,6 +21,7 @@
 import json
 import os
 import stat
+import sys
 import tempfile
 import time
 import urllib.request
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Nice!

PR #4057, opened 2022-11-12
More CodeQL fixes, fix cache crash regression

martinpitt:
[Next round of CodeQL scans](https://github.com/cockpit-project/bots/pull/4039/checks?check_run_id=9437227290) in #4039, only 5 left! This fixes them, including spotting a regression from #4051 (which I didn't spot on review either).

I also pushed the changes to #4039 to get a run with the fixes (... which got green). Of course I'll rebase them away again once this PR lands.

I re-tested the log.html changes with an existing log and a local webserver.

### DIFF
--- a/lib/s3-html/log.html
+++ b/lib/s3-html/log.html
@@ -234,7 +234,7 @@ function extract(text) {
     const entries = [];
     const testingElem = document.querySelector('#testing');
     const testingProgressElem = document.querySelector('#testing-progress');
-    const tap_range_match = tap_range.exec(text)
+    const tap_range_match = tap_range.exec(text);
     if (tap_range_match) {
         first = parseInt(tap_range_match[1], 10);
         last = parseInt(tap_range_match[2], 10);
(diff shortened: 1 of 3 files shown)


martinpitt:
Yup, this works, we got a [clean bill of health](https://github.com/cockpit-project/bots/pull/4039/checks?check_run_id=9447521939) now.

lib/s3-html/log.html
```diff
@@ -285,7 +285,8 @@
                           interesting: false,
                           links: [],
                           text: segment};
-            if (m = tap_result.exec(segment)) {
+            const m = tap_result.exec(segment);
```
allisonkarlitskaya, line 288: Maybe a nice chance for a more descriptive name for this variable, as well...

allisonkarlitskaya (approved):
Looks good.  Thanks for the cache fix.

My excuse: I originally wrote that like this:

```
try:
    for entry in os.scandir():
        ....
except FileNotFoundError:
    pass
```

but decided this was too complicated.  That's when I split the `scandir()` call out from the loop and pulled the `except:` clause up, but forgot to change it to `return`.

PR #4068, opened 2022-11-16
image-customize: add support for installing wheels

allisonkarlitskaya:
Here's a proposal to make the installation of Python binary packages (as per cockpit-project/cockpit#17922) more elegant.

It's possible to implement the functionality in the other PR without the addition of the option to `image-customize`, but:
 - we should install `pip` into at least `fedora-36` (and soon `fedora-37`) to avoid downloading it each time (and potentially failing due to infra issues, etc).
 - the abspath fix here makes sense in any case
 - after those two changes, the patch itself is rather small

Note: the reordering of the package-manager check vs. the file upload here made the logic a bit easier because it allowed an early exit for the python case.  Otherwise, we'd have ended up with either something like this:

```
   if whl:
      command = pip
   else:
      out = execute('which...')
      if out == 'dnf':
          ....
```
(ie: over-indented block)

or

```
    out = execute('which...')
    if whl:
        command = pip
    elif out == 'dnf':
        ...
```
(ie: unnecessary `which` invocations for the wheel case).

I'm open to implementing this as either of those options.  The re-odering was just my personal preference.

 * [x] image-refresh fedora-36

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -85,6 +85,11 @@ class InstallAction(ActionBase):
             machine_instance.upload([os.path.abspath(package)], dest)
             package = dest
 
+        # requesting install of Python wheel?
+        if package.endswith('.whl'):
+            machine_instance.execute(f"pip install --no-index --prefix=/usr {package}", timeout=120)
+            return
+
         # this will fail if neither is available -- exception is clear enough, this is a developer tool
         out = machine_instance.execute("which dnf || which yum || which apt-get")
         if 'dnf' in out:
--- a/image-customize
+++ b/image-customize
@@ -77,6 +77,14 @@ class InstallAction(ActionBase):
     '''Install local rpm or distro package'''
     @staticmethod
     def execute(machine_instance, package):
+        # If we have a '/' in the package name, or if a file with that name
+        # exists in the current directory, then assume that this is a package
+        # we're uploading from the host.
+        if '/' in package or os.path.isfile(package):
+            dest = "/var/tmp/" + os.path.basename(package)
+            machine_instance.upload([os.path.abspath(package)], dest)
+            package = dest
+
         # this will fail if neither is available -- exception is clear enough, this is a developer tool
         out = machine_instance.execute("which dnf || which yum || which apt-get")
         if 'dnf' in out:
@@ -85,14 +93,6 @@ class InstallAction(ActionBase):
             install_command = "yum --setopt=skip_missing_names_on_install=False -y install"
         else:
             install_command = "apt-get install -y"
-
-        if os.path.isfile(package):
-            pname = os.path.basename(package)
-            dest = "/var/tmp/" + pname
-            machine_instance.upload([os.path.abspath(package)], dest)
-            package = dest
-        elif '/' in package:
-            sys.stderr.write("Bad package name or path: %s\n" % package)
 
         machine_instance.execute(f"{install_command} {package}", timeout=1800)
 
--- a/image-customize
+++ b/image-customize
@@ -89,7 +89,7 @@ class InstallAction(ActionBase):
         if os.path.isfile(package):
             pname = os.path.basename(package)
             dest = "/var/tmp/" + pname
-            machine_instance.upload([package], dest)
+            machine_instance.upload([os.path.abspath(package)], dest)
             package = dest
         elif '/' in package:
             sys.stderr.write("Bad package name or path: %s\n" % package)
--- a/images/scripts/fedora.setup
+++ b/images/scripts/fedora.setup
@@ -40,6 +40,7 @@ openssl \
 PackageKit \
 pcp \
 pcp-libs \
+python3-pip \
 python3-tracer \
 qemu-block-curl \
 qemu-char-spice \


cockpituous:
Task failed: https://logs.cockpit-project.org/logs/image-refresh-4068-20221116-123636/

images/scripts/fedora.setup
```diff
@@ -40,6 +40,7 @@ openssl \
 PackageKit \
 pcp \
 pcp-libs \
+python3-pip \
```
martinpitt, line 43: I'm not a fan, but let's discuss in https://github.com/cockpit-project/cockpit/pull/17922/files#r1024140524 . This encourages users to do the wrong thing -- we strive for projects to prepare VM and run tests entirely offline, so that it becomes reproducible.

image-customize
```diff
@@ -89,7 +89,7 @@ class InstallAction(ActionBase):
         if os.path.isfile(package):
             pname = os.path.basename(package)
             dest = "/var/tmp/" + pname
-            machine_instance.upload([package], dest)
+            machine_instance.upload([os.path.abspath(package)], dest)
```
martinpitt: OK, but this needs checking our existing projects that they don't rely on the old behaviour. I did a quick grep, and it should be okay.

image-customize
```diff
@@ -77,6 +77,14 @@ class InstallAction(ActionBase):
     '''Install local rpm or distro package'''
     @staticmethod
     def execute(machine_instance, package):
+        if os.path.isfile(package):
+            pname = os.path.basename(package)
+            dest = "/var/tmp/" + pname
+            machine_instance.upload([os.path.abspath(package)], dest)
+            package = dest
```
martinpitt, line 84: This should `return`, otherwise you still do the package manager detection in vain.

images/scripts/fedora.setup
allisonkarlitskaya: `pip` is only ever being used to install a wheel that was generated locally.  This use is really more like `rpm` than it is `dnf`.

image-customize
allisonkarlitskaya: True enough.  But why isn't that warning an error?

martinpitt: It should be IMHO -- better than silently ignoring a broken argument.

image-customize
allisonkarlitskaya: I'm reasonably certain that this would absolutely always work.  We only ever enter this block if `os.path.isfile()` found the named file, relative to the current working directory, and in that case `os.path.abspath()` is going to definitely give the same result.

The only time this might break down is if we have a file called `a`, and also a file called `test/a` and we specify to upload `a`, which is then found in the current directory, but we actually wanted `test/a`.  I can't imagine that anyone is making use of [this sort of thing](https://xkcd.com/1172/)...

martinpitt: You are right, thanks!

images/scripts/fedora.setup
martinpitt: Maybe, but I'd still find it a bit more elegant/robust (and way easier) to do this with a mere directory copy, instead of (1) rebuilding *all* our images first, and (2) suggesting that it was okay to run `pip` in general on our images. But as I said, let's discuss in the cockpit PR.

allisonkarlitskaya:
So some research has revealed that the standard way of handling modern python projects packaging in RPM is via `%pyproject_wheel` and `%pyproject_install`.  Both of those macros are ultimately invoking `pip`, and doing so in a very similar way to the approach I take in https://github.com/cockpit-project/cockpit/pull/17922.

I feel like what we're doing here is fundamentally sound if we're trying to align with a future where we build cockpit-bridge RPMs via setuptools rather than autotools.  (And this is my intention).

martinpitt:
@allisonkarlitskaya : Ack, I resolved the comment above; so let's just fix the broken error exit paths, then we can refresh/land (note that this is blocked by the pixel update, but this should clear up in the next hour or two)

allisonkarlitskaya:
note: I don't intend to land the "fed36" commit, but am using it for testing :)

allisonkarlitskaya:
> note: I don't intend to land the "fed36" commit, but am using it for testing :)

Actually, scratch that idea.  We can't test the new scenario with the bots anyway, until the new tasks container lands, and that's blocked by pixels for now.  Pushing now so that we can get this one reviewed and landed in parallel.

images/scripts/fedora.setup
martinpitt, line 43: Note: This is temporary. Eventually this needs to go into the mock environment, not the VM host.

cockpituous:
image-refresh in progress on 3-cockpit-7.
Log: https://logs.cockpit-project.org/logs/image-refresh-4068-20221117-101818/

martinpitt:
Thanks! LGTM. Triggering two tests and an image refresh, to confirm that this doesn't beak stuff (at least too badly :grin: )

martinpitt (approved):

cockpituous:
image-refresh fedora-36 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-36-20221117-104147

martinpitt:
OK, image build works, that's all I wanted to know.

allisonkarlitskaya:
I actually need the image changes :sweat_smile:

images/scripts/fedora.setup
mvollmer: Maybe both?

PR #4069, opened 2022-11-17
Image refresh for arch

cockpituous:
Image refresh for arch
 * [x] image-refresh arch


jelly:
Arch now also suffers from the expire issue in passwd.

martinpitt (approved):

PR #4082, opened 2022-11-18
Image refresh for rhel-8-7

cockpituous:
Image refresh for rhel-8-7
 * [x] image-refresh rhel-8-7


jkozol:
cockpit-composer main no longer supports 8.7 or 9.1. Branches https://github.com/osbuild/cockpit-composer/tree/rhel-8.7.0 and https://github.com/osbuild/cockpit-composer/tree/rhel-9.1.0 should still pass tests on 87 and 91 machines.

martinpitt:
@jkozol : ah, interesting -- indeed it's not in the testmap any more, I wonder how it even ended up here as a triggered test. I'll have a look.

martinpitt:
Ah, commit 2f53070af984eedc49596903e934fc449b77c93e landed yesterday only, I see. So let's just rebase.

martinpitt (approved):

PR #4129, opened 2022-12-06
store-tests: Temporarily ignore pybridge scenario

martinpitt:
This manual scenario currently fails hundreds of tests. But we do want to run it often to check the progress. Don't count these failures into our statistics, as it completely ruins them.

---

Tested with the run  from https://github.com/cockpit-project/cockpit/pull/17988, by running
```
./store-tests --repo cockpit-project/cockpit --verbose 9dc7c742def34f79c40b6741e6cdfc3afabc5afd
./prometheus-stats --verbose
```
before/after.  You see that it previously considers lots and lots of tests flaky, but that's wrong.


jelly (dismissed):
Makes sense, thanks!

martinpitt:
Integration test repeatedly fails. I'm investigating in PR #4130 (as it also happens on main) and https://github.com/cockpit-project/cockpituous/pull/538 . Our integration tests don't currently cover store-tests, so this failure is unrelated. @jelly , can you please re-review?

jelly (approved):

jelly:
> Integration test repeatedly fails. I'm investigating in PR #4130 (as it also happens on main) and [cockpit-project/cockpituous#538](https://github.com/cockpit-project/cockpituous/pull/538) . Our integration tests don't currently cover store-tests, so this failure is unrelated. @jelly , can you please re-review?

Ok! Saw the debugging PR so approving and merging!

allisonkarlitskaya (approved):

PR #4130, opened 2022-12-06
XXX debug test

martinpitt:
test/cockpituous from https://github.com/cockpit-project/bots/pull/4129 repeatedly fails, and it's not obvious why. Using this as a guinea pig and for possible fixes.


martinpitt:
I [hacked](https://github.com/cockpit-project/bots/pull/4130/commits/60443ec7cdd979be33623dac334b8a629404421c) bots to run against this branch in https://github.com/cockpit-project/bots/pull/4130  and tests pass now.

PR #4174, opened 2022-12-16
naughty: relax regexp for rawhide nfs regression

KKoukiou:
As it affects more than one tests.


martinpitt:
Can you please copy these to fedora-38/ as well? I'm going to drop that "fedora-38" == "fedora-36" hack in cockpit.

jelly (approved):

martinpitt (approved):
Cheers! I'll send a cockpit PR now which makes use of that, then we can clean up the naughty from fedora-36.

PR #4175, opened 2022-12-16
vm-run: support adding multiple disks

jelly:
Allow -S to be repeated so one can easily test RAID and other multi disk features in Cockpit.

### DIFF
--- a/vm-run
+++ b/vm-run
@@ -33,7 +33,8 @@ parser.add_argument('-G', '--graphics', action='store_true', help='Display a gra
 parser.add_argument('-q', '--quiet', action='store_true', help="Don't connect text console")
 parser.add_argument('-c', '--client', action='store_true', help="Open Cockpit Client (implies -q)")
 parser.add_argument('-C', '--cpus', default=None, type=int, help='Number of cpus in the target machine')
-parser.add_argument('-S', '--storage', default=None, type=str, help='Add a second qcow2 disk at this path')
+parser.add_argument('-S', '--storage', default=[], action='append',
+                    help='Add a qcow2 disk(s) at this path')
 parser.add_argument('-e', '--execute', metavar='COMMAND', action='append',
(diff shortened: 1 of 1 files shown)


vm-run
```diff
@@ -33,7 +33,8 @@ parser.add_argument('-G', '--graphics', action='store_true', help='Display a gra
 parser.add_argument('-q', '--quiet', action='store_true', help="Don't connect text console")
 parser.add_argument('-c', '--client', action='store_true', help="Open Cockpit Client (implies -q)")
 parser.add_argument('-C', '--cpus', default=None, type=int, help='Number of cpus in the target machine')
-parser.add_argument('-S', '--storage', default=None, type=str, help='Add a second qcow2 disk at this path')
+parser.add_argument('-S', '--storage', type=str, action='append',
+                    help='Add a qcow2 disk(s) at this path')
```
martinpitt, line 37: Drop the `type=str` (it's the default anyway), and instead add `default=[]`. Then...

vm-run
```diff
@@ -78,7 +79,8 @@ try:
     machine.start()
 
     if args.storage:
-        machine.add_disk(path=args.storage, type='qcow2')
+        for disk in args.storage:
```
martinpitt, line 82: ... you can drop the `if` here, and just leave the `for`.

martinpitt (changes requested):
Thanks! Let's clean this up a bit, but LGTM otherwise.

vm-run
martinpitt: While you are at it, `default=None` is pointless, you can also drop it from all other options. Perhaps do a first commit of dropping type=str and default=None?

jelly: Good idea!

vm-run
jelly: That is a lot nicer, thanks!

martinpitt (approved):
Cheers!

PR #4180, opened 2022-12-19
lib: Treat TODO TAP entries as skips in log.html

martinpitt:
The TAP specification defines "todo" tests [1]. Format failed TODO tests the same way as skipped tests, so that they stop appearing as failures, but still stand out the same way as skips are.

[1] https://testanything.org/tap-specification.html#todo-tests

----

I tested this on [this run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18045-20221215-133023-56f9e5e3-fedora-37/log.html) from https://github.com/cockpit-project/cockpit/pull/18045 , which currently looks like this:

image
[...]
image

To test this PR, download the raw `log` next to log.html, start `python3 -m http.server`, and http://localhost:8000/log.html now looks like this:

image
[...]
image

We can certainly debate the visual design, but IMHO that's at least some improvement.

### DIFF
--- a/lib/s3-html/log.html
+++ b/lib/s3-html/log.html
@@ -137,8 +137,9 @@
         <script>
 
 const tap_range = /^([0-9]+)\.\.([0-9]+)$/m;
-const tap_result = /^(ok|not ok) ([0-9]+) (.*)(?: # duration: ([0-9]+s))?(?: # SKIP .*)?$/gm;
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (commented):
Don't we want some kind of visual indication of TODO tests?  Like orange in the bar and a mention in the status line?

allisonkarlitskaya (approved):
good 'nuf

martinpitt:
Maybe -- TBH I didn't think deeply about this, I just find that `# TODO` is semantically very close to what we do with "skip", so it's a "good enough" hack until it annoys us :grin:

PR #4185, opened 2022-12-20
testmap: Drop rhel-8-7 for cockpit-podman

martinpitt:
These are released RHELs, and history for us. 8.8/9.2 is the current
focus. Unfortunately podman on 9.2 is completely broken, so keep 9.1 for 
now, but add rhel-9-2 as a manual context.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -78,7 +78,6 @@ REPO_BRANCH_CONTEXT = {
     'cockpit-project/cockpit-podman': {
         'main': [
             'arch',
-            'rhel-8-7',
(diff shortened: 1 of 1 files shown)


martinpitt:
Argh, podman on 9.2 is an [unmitigated disaster](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4185-20221220-101015-232ad874-rhel-9-2-cockpit-project-cockpit-podman/log.html#5), this needs *waaaay* more work first, I'm afraid. So let's at least drop 8.7.

jelly (approved):

PR #4186, opened 2022-12-20
testmap: Drop rhel-8-7 and rhel-9-1 for cockpit

martinpitt:
These are released RHELs, and history for us. 8.8/9.2 is the current focus.

---

This introduces rhel-8-8-distropkg, which needs to be tested first.


jelly (approved):

jelly:
333 tests, 78 passed, 0 failed, 255 skipped, 0 to go (0 retries of failures, 0 retries of successes), took 24 minutes to run.

All green \o/

PR #4188, opened 2022-12-20
testmap: Move cockpit-machines from rhel-9-1 to -2

martinpitt:
Tests got fixed in
https://github.com/cockpit-project/cockpit-machines/pull/897


jelly (approved):
Danke!

PR #4203, opened 2022-12-25
machine_core: Avoid colons in static NM connection names

martinpitt:
These trigger a bug somewhere between kdump and NM: https://bugzilla.redhat.com/show_bug.cgi?id=2151504


martinpitt:
TestKDumpNFS [succeeds on rhel-9-2](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4203-20221225-182049-261ef991-rhel-9-2-cockpit-project-cockpit/log.html) with this. This is clearly a workaround and "proof of bug". Not sure if we really want to land it, but file names with colons are rather unusual.

KKoukiou (approved):

PR #4230, opened 2023-01-02
naughty: add alternative expression to match failures for selinux gssproxy regression

KKoukiou:
This will mark as xfail noumerous tests on cockpit-machines rawhide test runs.

### DIFF
--- a/naughty/fedora-38/4160-selinux-gssproxy-alternative-0
+++ b/naughty/fedora-38/4160-selinux-gssproxy-alternative-0
@@ -1,0 +1,1 @@
+error: Failed to start pool nfs-pool*
--- a/naughty/fedora-38/4160-selinux-gssproxy-alternative-1
+++ b/naughty/fedora-38/4160-selinux-gssproxy-alternative-1
@@ -1,0 +1,1 @@
+RuntimeError: Timed out on 'virsh pool-start nfs-pool*'


martinpitt (approved):
Thanks!

KKoukiou:
@martinpitt ./test-failure-policy fedora-38 < /tmp/c matches the naughty, however seems like this here is not. Any idea why?

martinpitt:
@KKoukiou : It doesn't work because on our CI we don't translate "fedora-rawhide" to `TEST_OS=fedora-38`. In tmt we do.

KKoukiou:
> @KKoukiou : It doesn't work because on our CI we don't translate "fedora-rawhide" to `TEST_OS=fedora-38`. In tmt we do.

So I guess we should override TEST_OS=fedora-38 when it's rawhide in the test/run file?

martinpitt:
> So I guess we should override TEST_OS=fedora-38 when it's rawhide in the test/run file?

Yes, that would work. It's another thing that goes stale, but it's easy enough to update.

martinpitt:
Done in https://github.com/cockpit-project/cockpit-machines/pull/902

martinpitt:
That doesn't work. Let's just land this and see how far it gets us. There is no risk here.

Issue #4231, opened 2023-01-02
Starting iscsi pool falis on login to iscsi server when selinux is enabled   


KKoukiou:
Downstream report: Failed https://bugzilla.redhat.com/show_bug.cgi?id=2157732

cockpituous:
fedora-rawhide
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testCreateThenInstall (__main__.TestMachinesCreate.testCreateThenInstall)
error: Failed to define network from /etc/libvirt/qemu/networks/default.xml
error: operation failed: network 'default' already exists with uuid ce64eac5-0a71-43c8-8541-67ff75edc10c
...+......+...+.....+......+.......+........+.+...+..+...+......+.+.........+......+.....+.+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++*..+..+...+......+.......+........+.......+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++*.....+...+.+..+.......+...+..+...+.............+.....+......+....+..+...+....+..+.........+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
[...]

PR #4232, opened 2023-01-02
Add pattern for selinux prohibiting iscsi pool to start

KKoukiou:
Known issue #4231
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2157732

### DIFF
--- a/naughty/fedora-38/4231-iscsi-selinux
+++ b/naughty/fedora-38/4231-iscsi-selinux
@@ -1,0 +1,3 @@
+# testAddDiskSCSI (__main__.TestMachinesDisks.testAddDiskSCSI)
+*
+RuntimeError: Timed out on 'virsh pool-start iscsi-pool'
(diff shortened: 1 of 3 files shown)


naughty/fedora-38/4231-iscsi-selinux
```diff
@@ -0,0 +1 @@
+error: Failed to start pool iscsi-pool
```
martinpitt, line 1: There's not enough context here IMHO. Could this grow a test case name/traceback context, if we don't get an actual error message in the test output? E.g. this could easily happen if we deliberately test adding an iSCSI  pool with invalid parameters.

martinpitt (changes requested):

naughty/fedora-38/4231-iscsi-selinux
KKoukiou: Indeed. ok let me repost

KKoukiou: Verified with logs from https://artifacts.dev.testing-farm.io/d66356c1-cdc2-4b3b-9b1a-d482967799f7/

martinpitt (approved):
Epharisto!

PR #4235, opened 2023-01-03
naughty: add fedora-rawhide symlink to fedora-38

KKoukiou:
When testing rawhide in upstream CI TEST_OS is fedora-rawhide but for tmt test runs we override it to fedora-38. Let's symlink the folders so the naughties cover both scenarios.


martinpitt (approved):
Thanks! Should this be sufficient for a `fedora-rawhide@cockpit-project/cockpit-machines` test run?

KKoukiou:
> Thanks! Should this be sufficient for a `fedora-rawhide@cockpit-project/cockpit-machines` test run?

Yes let's try before merging.

martinpitt:
Much better, down to two failures.

    error: internal error: process exited while connecting to monitor: audio: Unknown audio driver `spice'

That sounds unrelated. The `TestMachinesStoragePools.testStoragePoolsCreate` failure sounds like an incomplete naughty match? But either way, this run properly skipped e.g. the NFS tests, so this works. I think with that we can even drop the `$TEST_OS` mangling in c-machines' tmt test, right?

KKoukiou:
@martinpitt indeed the storage test failure from c-machines looks related, but missing naughty. 
I will merge this and follow up with another naughty if I reproduce locally.

PR #4241, opened 2023-01-05
lib: bump TEST_DEFAULT_OS to Fedora-37

jelly:
Fedora-37 was released last year!


jelly:
needs to be made from origin.

PR #4261, opened 2023-01-11
naughty: add kenrel port conflict issue to fedora-testing

jelly:
See https://cockpit-logs.us-east-1.linodeobjects.com/pull-1171-20230111-112515-186a35bc-fedora-testing/log.html#11-2

Fedora-testing is still F36 btw, shouldn't it be F37?


martinpitt (dismissed):
Ooops, yes, it should. images/scripts/fedora-testing* and naughty/fedora-testing symlinks still point to fedora-36*. In the meantime, let's land this to unblock the refresh. Please fix the typo with squash&merge.

But why did you send that here, instead of in #4256? That would be a lot simpler and safer?

jelly:
> Ooops, yes, it should. images/scripts/fedora-testing* and naughty/fedora-testing symlinks still point to fedora-36*. In the meantime, let's land this to unblock the refresh. Please fix the typo with squash&merge.

Yes, I can fix that later
 
> But why did you send that here, instead of in #4256? That would be a lot simpler and safer?

I did not notice there was  PR open, but it's unrelated we don't run podman tests yet in fedora-testing image refreshes.

martinpitt (approved):
(Huh, I apparently forgot to select "Approve"..)

PR #4225, opened 2023-01-02
Image refresh for fedora-rawhide-boot

cockpituous:
Image refresh for fedora-rawhide-boot
 * [x] image-refresh fedora-rawhide-boot
 * #4224

PR #4287, opened 2023-01-18

martinpitt:
Retry of PR #4225 with a current image.

 - [x] https://github.com/rhinstaller/anaconda/pull/4515


martinpitt:
The anaconda PR landed, so this is finally unblocked.

jelly (approved):

M4rtinK:
For the record - we do run with SELinux disabled on purpose, as we don't really expect the Fedora SELinux policy to handle all the special cases of an installer putting a system together from zero.

PR #4317, opened 2023-01-26
rhel4edge: Preparation

marusak:


machine/machine_core/machine.py
```diff
@@ -183,7 +183,7 @@ def audit_messages(self, type_pref, cursor=None):
 
     def allowed_messages(self):
         allowed = []
-        if self.image.startswith('debian') or self.image == 'fedora-coreos':
+        if self.image.startswith('debian') or self.image == 'fedora-coreos' or self.image == 'rhel4edge':
```
martinpitt, line 186: Let's please do that differently. Please add rhel4edge to OSTREE_IMAGES in constants.py (which should happen anyway, so that tests are correct), and then change this to `self.ostree_image`.

martinpitt (changes requested):
A small step for a developer.. over the edge! :wink:

machine/machine_core/machine.py
marusak: > add rhel4edge to OSTREE_IMAGES in constants.py

Right, I have that in separate commit, I can move it here.

> and then change this to `self.ostree_image`

Much better! Will update

martinpitt (approved):
Cheers!

PR #4339, opened 2023-01-30
image-customize: Call pip as module

martinpitt:
`pip` does not exist on RHEL/CentOS 8, but it can be called as module name.

---

With that, `test/image-prepare -q --python centos-8-stream` succeeds. Now it's failing left and right on walruses and other py3.6-isms, but with this we can at least start tests.


allisonkarlitskaya (approved):
yes plz!!

PR #4343, opened 2023-01-31
testmap: Run rhel4edge by default on c-ostree

marusak:
Also testing podman


martinpitt:
for fedora-coreos we added a hack to Makefile to just copy dist/, let's do the same for rhel4edge?

martinpitt (approved):

PR #4346, opened 2023-02-01
testmap: Define rhel4edge base image

marusak:


lib/testmap.py
```diff
@@ -218,6 +218,7 @@
 # their non-Atomic siblings.
 OSTREE_BUILD_IMAGE = {
     "fedora-coreos": "fedora-36",
```
martinpitt, line 220: Oops --- that's out of date.

martinpitt (approved):
Thanks!

PR #4371, opened 2023-02-06
testmap: /pybridge updates

martinpitt:
composer was enabled in https://github.com/osbuild/cockpit-composer/pull/1765 , machines got fixed through various actual pybridge fixes, see https://github.com/cockpit-project/cockpit-machines/pull/935


jelly (approved):
If you feel like it, also add candlepin/subscriptionmanager as `_manual` target?

martinpitt:
@jelly: These are already present. The only missing one was ostree.

PR #4375, opened 2023-02-07
images: Install python3 into fedora-coreos

martinpitt:
Our future Python bridge will require this. Once we switch officially, the cockpit-bridge package will grow a `python3` package dependency; until that happens, install it explicitly so that we can run a /pytest scenario.


cockpituous:
image-refresh fedora-coreos done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-coreos-20230207-082425

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-coreos-20230207-082424.log

jelly (approved):

martinpitt:
This isn't sufficient yet -- there's no `pip`, nor does it do the right thing -- we *never* should install stuff into `/usr`, just `/usr/local`. This requires some more work on image-customize, but indeed we can already land this image refresh -- it helps a lot.

PR #4381, opened 2023-02-09
testmap: trigger cockpit-podman on fedora-testing refreshes

jelly:
We found that fedora-testing worked fine with cockpit-podman and we should really test it when we update fedora-testing. (podman usually lands in updates testing as now with 4.4.0)

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -225,6 +225,7 @@ IMAGE_REFRESH_TRIGGERS = {
     "fedora-testing": [
         "fedora-testing@cockpit-project/cockpit",
         "fedora-testing/dnf-copr@cockpit-project/cockpit",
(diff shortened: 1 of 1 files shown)


jelly:
Same goes for machines, but that's for later :)

martinpitt (approved):
:pilot: :heavy_check_mark: 

Does that succeed right now? Or do you want to sort out possible failures in the next image refresh?

jelly:
> 



> pilot heavy_check_mark
> 
> Does that succeed right now? Or do you want to sort out possible failures in the next image refresh?

Last time I ran the tests was some time ago, but I'm willing to fix them sure! (4.4.0 seems to go fine on rawhide already tested through packit).

PR #4394, opened 2023-02-12

cockpituous:

### DIFF
--- a/images/fedora-rawhide-boot
+++ b/images/fedora-rawhide-boot
@@ -1,1 +1,1 @@
-fedora-rawhide-boot-f48e2062b65fe7e3a199dd3a3f362319a8394c5103b061168b2a1d594e576ff3.iso
+fedora-rawhide-boot-d336ac417adccbbfd5d6548f9f4afc7c7a90860eb27b8a23bb80fc63936381ea.iso


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-rawhide-boot-20230212-224040.log

jelly:
@M4rtinK  We updated our tasks container so we have to update all the pixel tests now, but one new pixel test will be wrong as the iso seems to miss a Japanese and Chinese font? So I am unsure if we should update, then again that is probably better then leaving it broken? If I understand it correctly these fonts need to be on the iso which we have no control over.

https://cockpit-logs.us-east-1.linodeobjects.com/pull-4394-20230212-224131-6cf64f53-fedora-rawhide-boot-rhinstaller-anaconda/pixeldiff.html#TestLanguage-testLanguageSwitching-language-step-basic

https://cockpit-logs.us-east-1.linodeobjects.com/pull-4394-20230212-224131-6cf64f53-fedora-rawhide-boot-rhinstaller-anaconda/TestLanguage-testLanguageSwitching-language-step-basic-pixels.png

VladimirSlavik:
Disclaimer: I am no @M4rtinK, just dumping my 2c.

If this still results in the same - missing fonts breaking pixel tests - then it might make sense to just go forward with it and update images here and tests on our side. A month ago it would not be not worth it, but as it is now, rawhide will start moving faster, so it could be better to keep up with the system underneath rather than wait for two pictures.

The fonts are nothing specially set up for these tests, so I don't see the fix happening here anyway, this side is just in for the ride along with what Fedora bakes. The fonts are also set up in lorax, so coordinating a potential fix across 3 parts is likely more work than updating the bits 2*2 times.

@M4rtinK feel free to disagree, you remain the one making the decision.

mvollmer (approved):
All pixels green!

PR #4397, opened 2023-02-13
image-customize: Install wheels into /usr/local/

martinpitt:
Installing into /usr does not work on OSTree. And even on writable /usr it is unfriendly, as it clobbers distro files. With that we can much more easily compare the py and C bridge behaviour on the same image.

----

This is going to fail for cockpit, as some of the tests have /usr/bin/cockpit-bridge hardcoded. I'll fix that, but want to get a first test run.

 - https://github.com/cockpit-project/cockpit/pull/18333

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -87,7 +87,7 @@ class InstallAction(ActionBase):
 
         # requesting install of Python wheel?
         if package.endswith('.whl'):
-            machine_instance.execute(f"python3 -m pip install --no-index --prefix=/usr {package}", timeout=120)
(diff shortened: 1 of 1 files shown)


martinpitt:
Cockpit tests were updated, so this works now. @allisonkarlitskaya are you d'accord with this?

allisonkarlitskaya (approved):
Not just OK with it — I actually really like it.

But what happened to not having pip installed on fedora-coreos?

martinpitt:
@allisonkarlitskaya : That's still the case. But at least it's one less thing to worry about.

PR #4403, opened 2023-02-14
testmap: add Fedora-38 to manual contexts

jelly:
Allow testing of Fedora 38 for Cockpit, podman, machines and starter-kit.


martinpitt (changes requested):
Let's please add it for the other projects as well (sub-man, composer, certificates, etc.).

martinpitt (approved):
Thanks!

PR #4413, opened 2023-02-15
testmap: Enable fedora-38 for cockpit-certificates

martinpitt:
Also drop the obsolete manual fedora-36 context, and move the RHELs to the current versions.


jelly:
Did you miss a trigger for rhel-8-8, rhel-9-2?

martinpitt:
@jelly These are just in _manual, we've never actually run the tests on RHEL.

jelly (approved):

PR #4432, opened 2023-02-20
testmap: drop rhel-8 from subscription-manager/main

ptoscano:
We recently decided to not keep support for Python 3.6 anymore in the 'main' branch of subscription-manager, as it is targeting RHEL 9.x only.


allisonkarlitskaya (approved):
Ok.

PR #4436, opened 2023-02-21
Add anaconda f38 to testmap

VladimirSlavik:
SSIA. I don't know what I 'm doing, and didn't find a readme, so advice appreciated. I assume there's some semi-manual testing step to do for the image? Or for both? Do I need to split this into two PRs, so that the image is first taken in and then added to tests?

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -210,6 +210,9 @@ REPO_BRANCH_CONTEXT = {
         'master': [
             'fedora-rawhide-boot',
         ],
+        'fedora-38': [
+            'fedora-38-boot',
+        ],
         '_manual': [
             'fedora-rawhide-boot/devel',
         ]


mvollmer:
> Do I need to split this into two PRs, so that the image is first taken in and then added to tests?

Yes, I think so.  Otherwise the tests might try to run without the fedora-38-boot being available for download.

I am also not sure what permissions you would need... let me just make a PR for the fedora-38-boot image, and then we can use this PR to change the testmap.

mvollmer:
#4437 has the image creation.

mvollmer:
I think I messed up the branching stuff... now the test runs on your "master", but it should run on the "fedora-38" branch, right?

mvollmer:
Ok, fedora-38-boot@rhinstaller/anaconda/fedora-38 is the right context.

VladimirSlavik:
Thanks! Let's see now...

VladimirSlavik:
Uhm, the test seems to be doing this:
```
=> https://cockpit-images.us-east-1.linodeobjects.com/fedora-37-aea45a597a2748296209259f2420d9dca6ddb84d1647f02eb18531caa8366c03.qcow2
```

Does this need a rebase now too?

mvollmer:
I think you hard-code TEST_OS=fedora-rawhide-boot, no? See https://github.com/rhinstaller/anaconda/blob/fedora-38/ui/webui/Makefile.am#L29

mvollmer:
> Uhm, the test seems to be doing this:
> 
> ```
> => https://cockpit-images.us-east-1.linodeobjects.com/fedora-37-aea45a597a2748296209259f2420d9dca6ddb84d1647f02eb18531caa8366c03.qcow2
> ```
> 
> 37 != 38, I think :) Does this need a rebase now too?

That is test/build-rpms.  It defaults to fedora-37 and test/prepare-updates-img doesn't override it:

https://github.com/rhinstaller/anaconda/blob/fedora-38/ui/webui/test/build-rpms#L53

mvollmer:
> Thanks - missed these completely, sorry. I will fix these and come back to this.

I think we can add the fedora-38-boot context for your fedora-38 branch to testmap already now, since it passes just fine.  It is testing the wrong thing, but you need to fix that in the fedora-38 branch itself, and having testmap already setup can only help with that.

mvollmer (approved):

PR #4449, opened 2023-02-23
testmap: Enable centos-8-stream/pybridge for Cockpit

martinpitt:
This was fixed in https://github.com/cockpit-project/cockpit/pull/18318 Let's make sure it stays that way!


martinpitt:
Meh, [some flakes](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4449-20230223-085755-3cd7e23a-centos-8-stream-pybridge-cockpit-project-cockpit/log.html#286-2) on the Services page. Unfortunately the Services page flakes quite often especially on the rhos-01 runners (like this run). This is next on my list of things to address, that page is just unbelievably expensive. We don't count /pybridge for statistics yet, but it shouldn't block PRs too much. Retrying once.

mvollmer (approved):

PR #4456, opened 2023-02-25
Image refresh for fedora-36

cockpituous:
Image refresh for fedora-36
 * [x] image-refresh fedora-36

PR #4474, opened 2023-03-01
image-customize: Use 2 GiB RAM by default

mvollmer:
Running dnf might need this now.

### DIFF
--- a/image-customize
+++ b/image-customize
@@ -266,7 +266,7 @@ def main():
     parser.add_argument('-n', '--no-network', action='store_true', help='Do not connect the machine to the Internet')
     parser.add_argument('--cpus', type=int, default=None,
                         help="Number of CPUs for the virtual machine")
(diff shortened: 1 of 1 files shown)


mvollmer:
This is needed for https://github.com/cockpit-project/bots/pull/4456. subman and cockpit-composer canÄt run their "dnf" customizations otherwise: https://cockpit-logs.us-east-1.linodeobjects.com/pull-4456-20230301-120317-9937de87-fedora-36-candlepin-subscription-manager/log.html

mvollmer:
subman already has "--memory-mb 2048" in one of its image-customize invocations, but not in the other, and I think this is a general problem that should be fixed in image-customize.

mvollmer:
Hmm, let's put this right into the image refresh PR...

martinpitt:
Ooh, thanks for tracking this down! 1 GiB of memory!?!? Ugh. FTR, +1 for the change.

PR #4475, opened 2023-03-01
testmap: add subscription-manager branches for rhel 8.8 & 9.2

ptoscano:


martinpitt (commented):
Did you test them? I.e. do you already want that landed, or add to _manual first? Once we land, this will block image refreshes and your PRs, so they should be green.

martinpitt (approved):
Ack, thanks!

PR #4519, opened 2023-03-13
npm: always pull latest node image

jelly:
To avoid pull requests which use an old npm version always pull the latest container. Github CI always has the latest container as it has to pull it.


martinpitt (approved):
Thanks! We can ignore the "cockpituous" test.

martinpitt:
Triggering a podman test, just to be sure.

PR #4530, opened 2023-03-17
refresh fedora-rawhide-boot

cockpituous:

### DIFF
--- a/images/fedora-rawhide-boot
+++ b/images/fedora-rawhide-boot
@@ -1,1 +1,1 @@
-fedora-rawhide-boot-885bac4a0f3ba9b9e7f7df65acfb27edb09720451dc55b0a47f92cc9cb0a366b.iso
+fedora-rawhide-boot-d4420db02c0e4d8dd6ee5c9996ed5c4332a2a015b0908c70a753edb6bc6cf122.iso


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-rawhide-boot-20230317-155555.log

jkonecny12:
The test seems to be failing because of:
```
ERROR    The requested volume capacity will exceed the available pool space when the volume is fully allocated. (15360 M requested capacity > 7849 M available) (Use --check disk_size=off or --check all=off to override)
```

Please extend libvirt pool capacity to store disk for Anaconda tests.

martinpitt:
How do we do that? There is no explicit libvirt pool as such. Perhaps this is running out of RAM or disk space in the tasks container?

KKoukiou:
@jkonecny12  It's the default pool available space in the tasks container, that is less than the requested disk size. Of course qcow s are COW so that's not  a problem if you don't use the full disk. So, if you increased it for interactive testing outside of the CI but the tests don't need 15GB, you can ignore disk size checks, with `--check disk_size=off`  when creating the VM in Use --check disk_size=off or --check all=off to override

jkonecny12:
Hi, @KKoukiou that is true that we don't probably need the full capacity. However, I can imagine that end-to-end testing could take more than 7849 M because of the payload. Especially if we start testing OSTree variants and the space would go even smaller.

I'm a bit afraid of switching of the disk_size check because in case the space won't be enough we could get into situation of I/O errors during the test which could be hard to discover.

jkonecny12:
Nice example of the missing space in pool would be if you build new images for Rawhide in parallel to F38 (we need to update both of them). In that case you have to get two such images into the same pool.

KKoukiou (approved):

PR #4544, opened 2023-03-20
prometheus-stats: add code coverage

tomasmatus:
Add new prometheus stat for code coverage by project.


tomasmatus:
opening again as #4543 was pushed from wrong origin

jelly:
> opening again as #4543 was pushed from wrong origin

CI succeeds so I think we can close this one.

PR #4554, opened 2023-03-23
README.md: document github.py's git config option

jelly:
When a user for example has their own fork as origin, tests-trigger won't work without specifying --repo. Our code however has a special git config option which is tried before trying to read the git remote origin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ want to change that, you can set the `cockpit.bots.images-data-dir` variable
 with `git config` to a directory where to store the pristine virtual machine
 images.  For example:
 
-    $ git config cockpit.bots.images-data-dir /srv/cockpit/images
+    git config cockpit.bots.images-data-dir /srv/cockpit/images
 
 ## Tests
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -170,6 +170,14 @@ run tests-trigger with `--allow`:
 Of course, you should make sure that the pull request is proper and
 doesn't execute evil code during tests.
 
+### tests-trigger with a different origin
+
+If you need to specify --repo in tests-trigger as your remote is different from
+cockpit-project/cockpit you can set a git configuration option from which
```
martinpitt, line 176: Run-away sentence. Please at least add a comma before the "you", but would be better to split the sentence.

README.md
```diff
@@ -170,6 +170,14 @@ run tests-trigger with `--allow`:
...
+tests-trigger reads the repo. This has to be set per cockpit project.
+
+    $ git config cockpit.bots.github-repo cockpit-project/cockpit
```
martinpitt, line 179: I find it really evil to have these `$` everywhere, as they break cut & paste. See https://github.com/cockpit-project/cockpit/commit/fe906e6bc92468ebf130d348d7fd1a7edc008c94 . While you are at it, would you mind dropping them everywhere in that file?

martinpitt (changes requested):
This is a nice trick indeed, thanks!

martinpitt (approved):
Cheers!

PR #4542, opened 2023-03-19
WIP: 4-ci-srv-06: [no-test] Image refresh for services

cockpituous:
Image refresh for services
 * [ ] image-refresh services

PR #4587, opened 2023-03-30
services: switch to my unofficial Candlepin container image

ptoscano:
The current way to test registration is to deploy Candlepin from sources in a container; while this initially worked fine, it started to grow a number of drawbacks:
- building Candlepin from sources (including importing the test data for it) takes a lot of time, at least 5-6 minutes even on fast systems
- the Candlepin project changes the supported OSes over time, and they recently stopped any effort to support Fedora
- the currently used Ansible role is no more used/supported by the Candlepin project, and thus it is rotting

In the end, we really don't need to build Candlepin from sources, but rather only use an already deployed Candlepin.

The Candlepin people are working on an official Candlepin container image usable also outside of their CI system; since this process will take some time, I built an unofficial container image of Candlepin:
  https://github.com/ptoscano/candlepin-container-unofficial
It contains everything (DB + Candlepin + test data) in a single image. While this is unofficial, it is not worse than the previously used way of deploying Candlepin, and it has the advantage of being stable (i.e. existing tags should still work fine regardless of upstream changes).

Hence, switch to this image for now:
- pull this image directly for the 'candlepin' container
- wait for Candlepin to start, so we check it works
- create the /root/run-candlepin script to run the container from scratch so it removes itself once stopped, and use it to start it for the extraction of the files
- regenerate the Candlepin certificates, so the hostname embedded in them is the actual hostname of the system (and `subjectAltName` will match on SSL connection)

The rest of the bits related to the 'candlepin' container still applies, i.e. the certificates are in the same locations.

 * [x] image-refresh services


martinpitt:
Oooh, this is nice! Is there any kind of automation and smoke-test around this? i.e. how reliable will this be in the future? Is this being used for CI in any other project (like sub-man)?

Note that this is currently blocked on PR #4542, which in turn is blocked on updating cockpit's tests against the new grafana.

ptoscano:
Also note that unfortunately so far the image is quite big (~2.1GB), and thus the generated services image is bigger:
- currently: 2384351232 bytes (2.4 GB, 2.2 GiB) copied, 12.7044 s, 188 MB/s
- with this change: 3788910592 bytes (3.8 GB, 3.5 GiB) copied, 8.6451 s, 438 MB/s

martinpitt:
Thanks for the details! If it's all automated, then it can easily be transplanted to some other org, and there's no secret sauce. :grin: 

> generated services image is bigger

Uh -- is that because your container image could get some basic cleanup (likd `dnf clean all`, or even building distro-less), or did newer candlepin versions just grew so enormously?

ptoscano:
Hmm I was able to cleanup few things in the image, and it is now ~1.3GB (which seems OK-ish to me). Yet when I rebuild the services image I now get:
- 4660838400 bytes (4.7 GB, 4.3 GiB) copied, 10.4388 s, 446 MB/s

OTOH the resulting qcow2 is ~2.3G, compared to a ~3.1G I see from a freshly-downloaded image in my cache (?) -- seems OK, I guess?

ptoscano:
> If it's all automated, then it can easily be transplanted to some other org, and there's no secret sauce.

The creation and push of the image is automated, you can check the details in the aforementioned repository: there are some files there, and it's all driven by a GitHub action/workflow that I run manually specifying the version (tag) of Candlepin, whether push it, and whether tag it as `:latest`.

The testing, as I mentioned earlier, is done either manually or using the currently-WIP PR for the `rhc` system role.

cockpituous:
image-refresh services done: https://github.com/cockpit-project/bots/commits/image-refresh-services-20230405-060155

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/services-20230405-060101.log

martinpitt:
Thanks! The current services VM is 3.1 GB, the candlepin container on it is 227 MB (I think that's uncompressed). Your container is 1.3 GB. Why is that so much bigger? Is that just because it derives from c9s instead of fedora? Or does it have extra features?

cockpituous:
image-refresh services done: https://github.com/cockpit-project/bots/commits/image-refresh-services-20230405-072353

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/services-20230405-072323.log

martinpitt:
The "rhel-8-8@candlepin/subscription-manager" is bogus, this doesn't exist any more in the regular testmap. I'll send a PR to fix that, and triggered rhel-8-8@candlepin/subscription-manager/subscription-manager-1.28 instead.

martinpitt:
"Waiting for https://services.cockpit.lan:8443/candlepin" times out.. still no joy, @ptoscano :cry:

ptoscano:
> "Waiting for https://services.cockpit.lan:8443/candlepin" times out..

Hmm the image definitely works, and so the podman invocation (otherwise the file extraction during build wouldn't work); I'll need to debug this further, then.

cockpituous:
image-refresh services done: https://github.com/cockpit-project/bots/commits/image-refresh-services-20230406-113521

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/services-20230406-113535.log

martinpitt:
I ran `./tests-trigger 4587 image:services` again to test this. (@ptoscano , you should be able to do that, too)

ptoscano:
> I ran `./tests-trigger 4587 image:services` again to test this. (@ptoscano , you should be able to do that, too)

Tests will still fail, however now I know what's the issue (and I'm working on it).

BTW: managed to shrink my image a bit more: 1.3G -> 1.19G

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/services-20230406-132525.log

ptoscano:
OK, it looks like it should work fine now.

The reason for the issue were the certificates: during the Candlepin deployment, a self-signed CA is generated using the hostname of the system, which is a random string as it is a build container. Hence, when trying to validate the SSL connection, the `subjectAltName` of the Candlepin certificate did not match the hostname (i.e. `services.cockpit.lan`).

The approach I chose here is to run the candlepin container, and regenerate the Candlepin certificates so they have the expected hostname.

ptoscano:
Yes! It looks like a success! PTAL, thanks :)

martinpitt (approved):
Thanks! It would have been easier to fix this by setting a static host name in the container (--hostname), but this is fine as well. Triggering the missing test, and ignoring the failed invalid sub-man test.

ptoscano:
Thanks!

Also, I managed to squeeze few bits, so now it's ~1.04 GB -- wait for the next periodic refresh ;)

PR #4588, opened 2023-03-30
image-diff: Use default format for ubuntu and debian

marusak:
Default format is `${binary:Package}\t${version}`. Difference between binary and source is that source shows name of source package while binary shows name of the binary package. This view shows actual package names which user can upgrade, downgrade, remove, install... The source package often does not align with actual package names.

@allisonkarlitskaya and @martinpitt what was the reason behind using source package name?


martinpitt (approved):
I don't know any more why I picked the source package name. Probably because it's easier to follow on https://tracker.debian.org? But binary packages make more sense indeed here.  Thanks!

PR #4618, opened 2023-04-07
services: less aggressive wait for grafana startup

ptoscano:
According to what I noticed in refreshes of the services image either done automatically by cockpituous or manually, it seems grafana requires at least 10-15 seconds to start answering after the startup of its contained.

Hence, poll for its finished setup less aggressively than once per seconds, doing it only once every 5 seconds. In the worst case, it will take up to 4 seconds more to wait for grafana, with the advantage of spamming less with requests.

### DIFF
--- a/images/scripts/services.setup
+++ b/images/scripts/services.setup
@@ -191,7 +191,7 @@ podman run -d --rm --name grafana -p 3000:3000 \
     docker.io/bitnami/grafana
 
 # wait until set up completed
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! I think that's harmless enough to do without a full image refresh/test cycle. But also, a curl request ought to be rather cheap, so it shouldn't make much difference either way. It's probably most useful to clean up the logs a bit.

PR #4634, opened 2023-04-11
testmap: Rename "dnf-copr" scenario to "daily"

mvollmer:
This follows the same change in Cockpit, where this scenario now also installs daily udisks2 rpms.

https://github.com/cockpit-project/cockpit/pull/18616


martinpitt (approved):
Thanks! These should land in lockstep, so not landing just yet.

PR #4641, opened 2023-04-12
Image refresh for rhel-8-6

cockpituous:
Image refresh for rhel-8-6
 * [x] image-refresh rhel-8-6


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/rhel-8-6-20230412-230606.log

martinpitt:
@ptoscano : sub-man fails on rhel-8-6 because 8.6's cockpit version [has an incompatibility with newer browsers](https://cockpit-project.org/blog/login-issues.html). Our CI container got updated yesterday, so this starts happening now. I'll apply the `sed` hack on this image, but I'd like to use the occasion to clean up old images: rhel-8-4, -8-6, -8-7, and rhel-9-0, -9-1. AFAIUI, at least the odd-numbered ones should be able to go, and you can drop the subscription-manager-1.28.32 and subscription-manager-1.29.30 branches from the testmap?

cockpituous:
image-refresh rhel-8-6 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-8-6-20230413-080727

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/rhel-8-6-20230413-080707.log

marusak (approved):
Oh no, we are still affected by this :/
Havn't we shipped z-stream fixes? Also why 8.4 and 8.6 but not 8.5?

martinpitt:
That's a good point -- https://bugzilla.redhat.com/show_bug.cgi?id=2129055 was the 8.6.z update. Apparently we don't install z-stream updates into our images? I'll land this to unbreak sub-man (as that affects the current image already, this was fallout from the tasks container update), and then will do another refresh to revert the hack and install z-stream updates. Same issue with 9.0.z, which we fixed in https://bugzilla.redhat.com/show_bug.cgi?id=2129056 -- so I'll figure that out in #4645.

> Also why 8.4 and 8.6 but not 8.5?

The even-numbered minors are EUS releases, while the odd ones aren't, so sub-man doesn't test on those.

PR #4661, opened 2023-04-17
testmap: a couple of subscription-manager updates

ptoscano:
- test `rhel-8-9` with `subscription-manager/subscription-manager-1.28`
- drop `rhel-9-2` from `subscription-manager-cockpit`


martinpitt:
Thanks, LGTM! I triggered the new test to make sure.

PR #4662, opened 2023-04-17
machine_core: Reduce journal sync sleep to 1s

martinpitt:
This wait time is quite expensive, as it adds 3s to each single test run time. This happens after the actual test is done, so the machine ought to be quiet. 1s should be enough time for services to catch up on log writing.

----

See https://issues.redhat.com/browse/COCKPIT-979 for the baseline numbers. I'll run the tests several times here to get some reliable data, as our test runtimes vary a lot: Chromium ø 5634s, σ 381s; Firefox: ø 5929s, σ 525s


jelly (approved):
Nice, as someone who often works on c-podman it seems it affected @nd tests mostly is not an issue at all :grin:

PR #4703, opened 2023-04-27
images/scripts: hard-code some extra build-deps

allisonkarlitskaya:
The cockpit repo is introducing dynamic BuildRequires[^1] in its spec   
file, which is not supported by our build-deps.sh script, and would not    
be easy to add.    
    
We avoid this issue by simply hard-coding the necessary dependencies    
directly here.    
    
It could be argued that it would be better to add those extra    
dependencies to the .spec file itself, but I chose this approach for two    
reasons:    
    
- having the redundant information in the .spec file is inelegant
    
- whenever we need to add extra dependencies, it's a bit of a
complicated lock-step dance between adding the new dependencies to
the spec file, landing it on main, then updating the images before
actually being able to use those dependencies.
    
Hard-coding the dependencies directly here, although also inelegant,
lets us skip the first step and remove the catch-22: we just update
the images via a PR to bots/ and then can proceed to hack on the
cockpit repo.
    
[^1]: https://fedoraproject.org/wiki/Changes/DynamicBuildRequires

<hr>

Note: this PR can't land in its current state.  It has a hacked up temporary commit to pull the `cockpit.spec` file from cockpit-project/cockpit#18730 (which in turn, can't land until we get new images).

 * [x] image-refresh fedora-37: succeeded as fedora-37-31d79577954d1e3c0e2d7d26db24e75588d29102814d0e904825dcc3839bfe17.qcow2 , but failed during image-prune
 * [x] image-refresh fedora-38

### DIFF
--- a/images/scripts/lib/build-deps.sh
+++ b/images/scripts/lib/build-deps.sh
@@ -19,7 +19,7 @@ case "$OS_VER" in
         ;;
 
     *)
-        spec=$($GET "$COCKPIT_GIT/main/tools/cockpit.spec")
+        spec=$($GET "$COCKPIT_GIT/procps-ng-buildrequire/tools/cockpit.spec")
         ;;
 esac
 


cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-37-20230427-201313.log

cockpituous:
image-refresh fedora-38 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-38-20230427-203854

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-38-20230427-203838.log

martinpitt:
We did this several times before, like in commit 4bd317227ec6fa63926cf4954a44f1e73ff1f139. So I'm not sure why you need the HACKING commit here? I'd propose the following:

 * Have the first commit here with the new deps. Build a Fedora and a rhel-8-9 image to prove that all these build deps are available in RHEL 8 as well (I have a feeling they are not)
 * Land a PR with these two image refreshes (as refreshing them all in a single PR is too much noise)
 * Create three more issues with refreshing the remaining Fedora,  RHEL, and CentOS images
 * Use the new deps in cockpit

Fedora 37 actually succeeded:
```
Uploading to https://cockpit-images.eu-central-1.linodeobjects.com/fedora-37-31d79577954d1e3c0e2d7d26db24e75588d29102814d0e904825dcc3839bfe17.qcow2
Uploading to https://cockpit-images.us-east-1.linodeobjects.com/fedora-37-31d79577954d1e3c0e2d7d26db24e75588d29102814d0e904825dcc3839bfe17.qcow2
```
It just failed on a temporary network glitch during pruning. So I suggest you just do that symlink update commit manually and avoid another refresh.

allisonkarlitskaya:
>     * Have the first commit here with the new deps. Build a Fedora and a rhel-8-9 image to prove that all these build deps are available in RHEL 8 as well (I have a feeling they are not)

Indeed.  This is turning into a major nightmare.

RHEL8 is missing `%generate_buildrequires` entirely along with the entire `pyproject-rpm-macros` stuff, so we're sort of stuck there.  That's going to need to be handled conditionally.

Additionally, all of the extra pytest-* packages are only in Fedora.  Even RHEL9 is lacking those.

I'd like to run pytest on Fedora systems, at least.... so I think the two image updates here *are* valuable...

But in general, this is going to need a bit of thinking through before we proceed....

allisonkarlitskaya:
I start to wonder if we could build a new way of avoiding the dependency catch-22 situation without `#HACK` lines in the script or `HACK:` commits to fetch from a different branch: we could have our invocation of `rpmspec` define a macro that's only defined for this script.  That would let us introduce dependencies to the spec that are only intended to be picked up by this script and won't otherwise cause build failures on main before the image refreshes happen....

On the other hand... yick.

allisonkarlitskaya:
→ #4707

PR #4720, opened 2023-05-02
setup-deploy-keys: Add "self" environment for cockpit-podman

martinpitt:
We need that to properly run weblate-sync-po.yml.

----

This is the prerequisite for fixing [podman PO refreshes](https://github.com/cockpit-project/cockpit-podman/pull/1264) to actually be able to run "protection checks". That only works for SSH pushes, not for HTTP pushes with the github default tokoen (as these can't trigger workflows).

I ran this, and https://github.com/cockpit-project/cockpit-podman/settings/environments now has a "self" environment. It's also a good opportunity to rotate the keys, we last did that 4 months ago.


allisonkarlitskaya (approved):
It is good.

PR #4726, opened 2023-05-03
images: Drop fedora-38-boot [was: Image refresh for fedora-38-boot]

cockpituous:
Fedora 38 is released, so Anaconda's fedora-38 branch will not get changes any more.

### DIFF
--- a/image-trigger
+++ b/image-trigger
@@ -29,7 +29,6 @@ REFRESH = {
     "fedora-36": {},
     "fedora-37": {},
     "fedora-38": {},
-    "fedora-38-boot": {},
     "fedora-testing": {"refresh-days": 3},
     "fedora-coreos": {},
     "fedora-rawhide": {},
--- a/images/fedora-38-boot
+++ b/images/fedora-38-boot
@@ -1,1 +1,0 @@
-fedora-38-boot-8268db8d64ecaedda312b63202cefe2461b1bcbc1917fcac9a4e72253f852147.iso
--- a/images/scripts/fedora-38-boot.bootstrap
+++ b/images/scripts/fedora-38-boot.bootstrap
@@ -1,8 +1,0 @@
-#!/bin/sh
-set -eux
-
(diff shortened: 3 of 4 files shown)


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-38-boot-20230503-223838.log

martinpitt:
```
Serving HTTP on 0.0.0.0 port 8000 (http://0.0.0.0:8000/) ...
ERROR    Error validating install location: Could not find an installable distribution at URL '/work/bots/make-checkout-workdir/ui/webui/bots/images/fedora-38-boot'

The location must be the root directory of an install tree.
```

@jkonecny12 @jelly can you please have a look?

VladimirSlavik:
F38 is released, there is no further work for us, the install media is done. Thus the tests are guaranteed to be unused... and practically irrelevant. Maybe the right step would be disabling the updates?

martinpitt:
@VladimirSlavik Good point! And you'll wait with a fedora-39-boot until that gets branched off [in August](https://fedorapeople.org/groups/schedule/f-39/f-39-key-tasks.html)?

VladimirSlavik:
Yes, I believe it will be exactly like that. 99% of anaconda work must happen before Fedora ga/final, so I expect we could generally stop updating images after that.

This needs confirmation from the team members working on webui - once again I'm the first responder but not the person to make the decision. However, the logic seems solid enough to me, and I wanted to save you from useless work :-)

As an aside, the `fedora-38-boot` image was originally meant as a "learn branching with webui" attempt. I don't think it ever served an actual useful purpose.

martinpitt:
Okay, I changed this PR to drop the image instead. @jkonecny12 ,  @jelly can you please confirm?

KKoukiou (approved):
indeed the did the release note commit 5 days ago https://github.com/rhinstaller/anaconda/pull/4718 for this branch

PR #4746, opened 2023-05-09
Image refresh for fedora-rawhide

cockpituous:
Image refresh for fedora-rawhide
 * [x] https://github.com/cockpit-project/starter-kit/pull/648


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-rawhide-20230509-230202.log

martinpitt:
This is most curious. This worked last week in https://github.com/cockpit-project/starter-kit/pull/646 (last change), and now it's clearly broken despite pinning down the dependencies exactly. Is it possible to re-publish the same version on npm with changes?

This is *not* some issue with the internal NPM miror, it happens locally too. I take a look.

martinpitt:
> ✘ [ERROR] Could not resolve "@patternfly/react-styles/css/components/AlertGroup/alert-group"

The file moved to node_modules/@patternfly/react-styles/css/components/Alert/alert-group/

We also don't import this directly. So the problem is that we *don't* pin down the react-styles dependency, it's an indirect one which just got pulled in as a newer version. We do pin it down in other projects, let's just fix that.

I sent https://github.com/cockpit-project/starter-kit/pull/648

martinpitt (approved):

PR #4768, opened 2023-05-16
testmap: Enable pybridge for subscription-manager-cockpit

martinpitt:
I set this up over two months ago: https://github.com/candlepin/subscription-manager-cockpit/pull/47 -- But then forgot to enable it in the testmap :sweat:


allisonkarlitskaya (approved):
thx :+1:

PR #4799, opened 2023-05-24
Ruffification

jelly:


martinpitt:
@jelly We only install it into cockpit's unit-tests container, so you'll need to install it explicitly into the GitHub runner.

test/run
```diff
@@ -2,8 +2,11 @@
 
 set -eu
 
-# run static code checks like pyflakes and pep8
-PYEXEFILES="$(git grep -lI '^#!.*python') $(git ls-files "*.py")"
-flake8 $PYEXEFILES
+test_ruff() {
+    test -n "$(ruff check --show-files . 2>/dev/null)" || return 0  # no python changes
```
martinpitt, line 6: This will silently ignore a failing or non-existing `ruff`, which may be your intention. But at least showing some "skip: ruff not installed" would be nice?

allisonkarlitskaya (commented):
I'm pretty happy with this.  You could probably select a bit more — "A", "B", "G", "D", etc. are mostly worthwhile, but they can also come in a follow-up.

I'm afraid we can't expel flake8 just yet, unfortunately.  ruff is still missing a lot of pretty basic whitespace checking rules.  See https://github.com/charliermarsh/ruff/issues/2402.

martinpitt (changes requested):
tests.yml runs in cockpit's unit-tests container, so ruff actually should be available.

Let's please keep flake8 for now, and show a "skip" message if ruff is not installed. Thanks!

jelly:
> tests.yml runs in cockpit's unit-tests container, so ruff actually should be available.
> 
> Let's please keep flake8 for now, and show a "skip" message if ruff is not installed. Thanks!

I was under the same impression, so re-added flake8 and made it skip ruff if not available.

jelly:
> I'm pretty happy with this. You could probably select a bit more — "A", "B", "G", "D", etc. are mostly worthwhile, but they can also come in a follow-up.
> 
> I'm afraid we can't expel flake8 just yet, unfortunately. ruff is still missing a lot of pretty basic whitespace checking rules. See [charliermarsh/ruff#2402](https://github.com/charliermarsh/ruff/issues/2402).

Let's do that in a follow up with mypy :)

jelly:
And ruff is :100: installed https://github.com/cockpit-project/bots/actions/runs/5077170161/jobs/9120121035?pr=4799 \o/

jelly:
Hmm I think the test fails and rebase / craft a new git commit id to make the integration tests succeed? Or can we skip that?

test/run
```diff
@@ -2,6 +2,14 @@
 
 set -eu
 
+test_ruff() {
+    command -v ruff >/dev/null || echo 'no ruff installed'
```
martinpitt, line 6: It needs to `return` if ruff is not installed.

test/run
```diff
@@ -2,6 +2,14 @@
...
+    test -n "$(ruff check --show-files . 2>/dev/null)" || return 0  # no python changes
```
martinpitt, line 7: I don't like suppressing stderr here. Is that just for the case that ruff isn't installed? (see above)

machine/machine_core/ssh_connection.py
```diff
@@ -32,7 +32,7 @@
 # HACK: some projects directly import ssh_connection before adjusting sys.path; add bots root dir
 sys.path.insert(1, os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__)))))
 
-from lib.constants import TEST_DIR
+from lib.constants import TEST_DIR  # noqa: E402
```
martinpitt, line 35: Perhaps we should just disable that rule globally for now? (Not a strong opinion)

martinpitt (changes requested):
Thanks! The integration test ought to go green, but we don't run it very often -- perhaps something in GitHub changed in between.

machine/machine_core/ssh_connection.py
martinpitt: Ack.

test/run
```diff
@@ -2,6 +2,14 @@
 
 set -eu
 
+test_ruff() {
+    command -v ruff >/dev/null || echo 'no ruff installed' && return 0
```
martinpitt, line 6: That now returns if ruff is available. You need `|| { echo 'no ruff installed'; return 0; }`

martinpitt (changes requested):

martinpitt (approved):
Good eruff! :grin:  Danjkewel!

martinpitt:
Triggering at least one quick test for paranoia's sake.

PR #4809, opened 2023-05-26
Enable more ruff rules

jelly:
Also sneaked in a unit test improvement as waiting for 6 seconds got annoying ;-)

### DIFF
--- a/machine/machine_core/timeout.py
+++ b/machine/machine_core/timeout.py
@@ -45,7 +45,7 @@ class Timeout:
             signal.signal(signal.SIGALRM, self.handle_timeout)
             signal.alarm(self.seconds)
 
-    def __exit__(self, type, value, traceback):
+    def __exit__(self, _type, value, traceback):
         if self.seconds:
             signal.alarm(0)
             signal.signal(signal.SIGALRM, signal.SIG_DFL)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,7 @@ select = [
     "E",       # pycodestyle
     "F",       # pyflakes
     "I",       # isort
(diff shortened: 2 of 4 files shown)


task/test-cache
```diff
@@ -55,26 +55,26 @@ class TestCache(unittest.TestCase):
         self.assertEqual(result, other)
 
     def testCurrent(self):
-        c = cache.Cache(self.directory, lag=3)
+        c = cache.Cache(self.directory, lag=0.5)
```
allisonkarlitskaya, line 58: I don't like this test, and I particularly don't like your changes to it... this timing-dependent stuff is asking for trouble, and making the window smaller makes it more dangerous.

Why not introduce a better separation between linting and unit tests?

martinpitt (approved):
LGTM, but please finish the discussion with Lis first. Dankjewel!

task/test-cache
jelly: I ran the test ~ 10 times in a row and that seemed fine, this is mostly a pre-push hook annoyance as that takes 6 seconds. Hmm can we have a different approach of testing this.

jelly: Dropped it, as I don't feel like bikeshedding in this PR, we can discuss it separate from this PR :)

martinpitt (approved):

allisonkarlitskaya (approved):
Fine by me too, but the aliasing fix looks like you were hoping to `select = [ "A" ]` as well?

jelly:
> Fine by me too, but the aliasing fix looks like you were hoping to `select = [ "A" ]` as well?

Yup, but that requires more possibly API breaking changes, as some public API uses builtin's

PR #4818, opened 2023-05-29
Image refresh for arch

cockpituous:
Image refresh for arch
 * [x] image-refresh arch


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/arch-20230529-225454.log

martinpitt:
@jelly: [This](https://cockpit-logs.us-east-1.linodeobjects.com/pull-4818-20230529-225518-e3ebe3fb-arch-cockpit-project-cockpit/log.html#266-2) "error: failed to synchronize all databases (unable to lock database)" happens randomly. There are several google hits for that, including the [arch wiki](https://wiki.archlinux.org/title/Pacman#.22Failed_to_init_transaction_.28unable_to_lock_database.29.22_error), but they all just recommend to remove the lock file. But I suppose that happens because there is actually some other pacman running. The journal does not have any trace of it, though. Do you happen to know what that could be?

martinpitt:
I don't see any "pacman" related cronjob, just archlinux-keyring-wkd-sync.service. Starting it does not involve pacman, though. It will eventually fail because the VM is offline, but it seems unrelated. So perhaps it actually is a leftover from a previous test, but there are no previous failures. It certainly shouldn't be a previous `pacman -Sy`, as that's hopefully synchronous, and would fail on interruption then?

We could try to run `pkill pacman` on test cleanup, but that's not solving the root cause.

jelly:
I was unable to reproduce it locally, extended packagelib to clean up a possible leftover lockfile https://github.com/cockpit-project/cockpit/pull/18869

PR #4830, opened 2023-05-31
naughty: add podman hang issue for rhel4edge

jelly:


jelly:
Example https://cockpit-logs.us-east-1.linodeobjects.com/pull-1235-20230526-162637-03f5f1da-rhel4edge/log.html#26

naughty/rhel4edge/4829-podman-hang
```diff
@@ -0,0 +1,3 @@
+  File "*", line *, in testPruneUnusedContainersSystem
```
martinpitt, line 1: You could have unified the two patterns with `testPruneUnusedContainers*`, but this is fine of course.

martinpitt (approved):
Thanks! I triggered a rhel4edge run.

naughty/rhel4edge/4829-podman-hang
jelly: Yup, I chose to make it explicit as that's a bit easier.

PR #4874, opened 2023-06-14
testmap: Accept arbitrary scenarios in test triggers

martinpitt:
We want to use scenarios a lot more soon for parallelizing tests. Having to always pre-declare them as `_manual` context is annoying for development and experimentation.

Change `is_valid_context` to cut off and ignore the scenario when checking a context against the testmap.

----

Plus a few small cleanups.

I did a `bots/tests-trigger 18778 fedora-38/ohlook` in cockpit with this, and https://github.com/cockpit-project/cockpit/pull/18778 now has the /ohlook scenario pending. Of course that won't actually be considered by our CI until this PR lands. After that, it should be run and fail immediately with "unknown scenario", as cockpit checks that.

Other projects like c-machines don't check for unknown scenarios, and will just run the default one. We can/should fix that, but IMHO this is acceptable for now.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -290,24 +290,25 @@ def split_context(context):
 
 def is_valid_context(context, repo):
     image_scenario, _bots_pr, context_repo, branch = split_context(context)
+    image = image_scenario.split('/')[0]
     # if the context specifies a repo, use that one instead
     branch_contexts = tests_for_project(context_repo or repo)
     if context_repo:
         # if the context specifies a repo, only look at that particular branch
         try:
-            repo_contexts = branch_contexts[branch or get_default_branch(context_repo)].copy()
+            repo_images = [c.split('/')[0] for c in branch_contexts[branch or get_default_branch(context_repo)]]
         except KeyError:
             # unknown project
             return False
         # also allow _manual tests
-        repo_contexts.extend(branch_contexts.get('_manual', []))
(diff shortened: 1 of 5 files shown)


lib/testmap.py
```diff
@@ -285,29 +273,30 @@ def split_context(context):
...
+            repo_images = [c.split('/')[0] for c in branch_contexts[branch or get_default_branch(context_repo)]]
         except KeyError:
             # unknown project
             return False
         # also allow _manual tests
-        repo_contexts.extend(branch_contexts.get('_manual', []))
+        repo_images.extend([c.split('/')[0] for c in branch_contexts.get('_manual', [])])
     else:
         # FIXME: if context is just a simple OS/scenario, we don't know which branch
         # is meant by the caller; accept known contexts from all branches for now
```
mvollmer, line 295: We could change the comment to also say "accept known images"

lib/testmap.py
```diff
@@ -285,29 +273,30 @@ def split_context(context):
...
-        repo_contexts = set(itertools.chain(*branch_contexts.values()))
+        repo_images = set([c.split('/')[0] for c in itertools.chain(*branch_contexts.values())])
 
     # Valid contexts are the ones that exist in the given/current repo
-    return os_scenario in repo_contexts
```
mvollmer, line 310: "Valid images"?

mvollmer (approved):

PR #4911, opened 2023-06-22
testmap: Don't run cockpit rhel-8-9-distropkg/networking

### LINKED ISSUE #4910: skip some testing scenarios with test splitting
We tried to produce evenly-split buckets for the sub-scenario splitting, but didn't really think too much about skips.  For example, rhel-8 skips every single test in `/networking`.

https://cockpit-logs.us-east-1.linodeobjects.com/pull-18992-20230622-135331-5770d0e3-rhel-8-9-distropkg-networking/log.html

At the very least we shouldn't run the scenarios in which no tests will be run.

martinpitt:
All tests are skipped anyway.

Fixes #4910

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -49,7 +49,8 @@ REPO_BRANCH_CONTEXT = {
             # no udisks on CoreOS → skip storage
             *contexts('fedora-coreos', COCKPIT_SCENARIOS - {'storage'}),
             *contexts('rhel-8-9', COCKPIT_SCENARIOS),
-            *contexts('rhel-8-9-distropkg', COCKPIT_SCENARIOS),
+            # all skipped
+            *contexts('rhel-8-9-distropkg', COCKPIT_SCENARIOS - {'networking'}),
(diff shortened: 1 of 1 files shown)


lib/testmap.py
```diff
@@ -49,7 +49,8 @@ def contexts(image, *scenarios: List[str]):
             # no udisks on CoreOS → skip storage
             *contexts('fedora-coreos', COCKPIT_SCENARIOS - {'storage'}),
             *contexts('rhel-8-9', COCKPIT_SCENARIOS),
-            *contexts('rhel-8-9-distropkg', COCKPIT_SCENARIOS),
+            # all skipped
+            *contexts('rhel-8-9-distropkg', COCKPIT_SCENARIOS - {'networking'}),
```
allisonkarlitskaya, line 53: I'm not a huge fan of how brittle this is.  We skip this because we know that there are no tests there, but maybe one day we do have tests there and falsely conclude that they're passing...

I understand that this is a temporary measure, but ya... let's try to figure out something better asap.

allisonkarlitskaya (approved):

PR #4935, opened 2023-06-27
lib: testmap: update fedora-rawhide to trigger anaconda tests

KKoukiou:
Anaconda test preparation builds the RPMs in the fedora-rawhide image which needs to be gated.


martinpitt (approved):
Ah, does it now? Nice!

PR #4949, opened 2023-06-30
testmap: Unsplit Cockpit's /devel scenario

martinpitt:
This produces coverage reports, which are much less useful when they get split into four parts.

### DIFF
--- a/lib/test-testmap.py
+++ b/lib/test-testmap.py
@@ -92,8 +92,9 @@ class TestTestMap(unittest.TestCase):
         self.assertIn("arch/networking", main_tests)
         self.assertIn("debian-testing/other", main_tests)
         # scenario options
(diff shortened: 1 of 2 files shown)


mvollmer (approved):
Thanks!

allisonkarlitskaya:
@martinpitt I guess you don't feel like re-proposing from origin...

mvollmer:
> get meaningful coverage results for the JS of a given page

The main problem right now are the comments posted by the bots: "This added line is not executed by any test".  These comments are now wrong and plentiful.  For example, if a PR changes something pkg/storaged, the fedora-38/devel-networking scenario will find that this modification is not covered by any test, and complain loudly.

We could also tune the coverage complain machinery to know that devel-networking should never complain about stuff in pkg/storaged. That also requires a source file -> tests mapping.

PR #5002, opened 2023-07-14
naughty: Add yet another pattern for podman pod run hang

martinpitt:
Pattern -4 only applied to `testPruneUnusedContainers`, but the same can happen to `testCreatePod`. As -5 was a different pattern, move -5 to -6, and make the new -5 a copy of 4 that applies to the latter test.

---

Spotted here: https://cockpit-logs.us-east-1.linodeobjects.com/pull-1324-20230714-100146-0d0a164d-ubuntu-2204/log.html#19-1

The gift that keeps on giving! :gift_heart:


skobyda (approved):
Thanks

Issue #5016, opened 2023-07-18
sos immediately crashes: Could not initialize 'report': 'ConfigParser' object has no attribute 'readfp'


martinpitt:
Downstream issue: https://bugzilla.redhat.com/show_bug.cgi?id=2223526

cockpituous:
debian-testing
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testVerbose (__main__.TestSOS.testVerbose)
Failed to try-restart sssd.service: Unit sssd.service not found.
[0716/135055.410803:WARNING:sandbox_linux.cc(430)] InitializeSandbox() called with multiple threads in process gpu-process.

DevTools listening on ws://127.0.0.1:9619/devtools/browser/d3c8a37a-16d8-428f-97d2-913b9c982153
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (ERROR)","timestamp":1721137857704.081,"url":"http://127.0.0.2:9791/cockpit/@localhost/*/po.manifest.js","networkRequestId":"13269.19"}
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (ERROR)","timestamp":1721137857708.527,"url":"http://127.0.0.2:9791/cockpit/@localhost/*/po.js","networkRequestId":"13269.20"}
CDP: {"source":"security","level":"error","text":"Refused to execute script from 'http://127.0.0.2:9791/cockpit/@localhost/*/po.manifest.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.","timestamp":1721137857783.01,"url":"http://127.0.0.2:9791/sosreport"}
CDP: {"source":"security","level":"error","text":"Refused to execute script from 'http://127.0.0.2:9791/cockpit/@localhost/*/po.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.","timestamp":1721137857783.139,"url":"http://127.0.0.2:9791/sosreport"}
> warning: failed to poll tuned {"problem":"disconnected","name":"null","message":"Server has closed the connection.","toString":""}
> warning: failed to poll tuned {"problem":"disconnected","name":"null","message":"Server has closed the connection.","toString":""}
> error: Failed to call sos report: {"problem":null,"exit_status":1,"exit_signal":null,"message":""}
Traceback (most recent call last):
  File "/work/make-checkout-workdir/test/verify/check-sosreport", line 144, in testVerbose
[...]

PR #5017, opened 2023-07-18
naughty: Add pattern for Fedora rawhide sos crash

martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2223526 Known issue #5016

----

Fedora Rawhide works again in TF in principle since yesterday evening. It now fails "properly" on this sos regression, see https://artifacts.dev.testing-farm.io/382d0430-4269-4e71-8fca-2f874a491f91/

### DIFF
--- a/naughty/fedora-39/5016-sosreport-readfp
+++ b/naughty/fedora-39/5016-sosreport-readfp
@@ -1,0 +1,4 @@
+  File "tests/test/verify/check-sosreport", line *
+    b.wait_not_present("#sos-dialog")
+*
+testlib.Error: timeout


marusak (approved):
The pattern is bit too generic (dialog does not get closed) but I guess it is what we got. There is already activity on the downstream report so this should get fixed quickly. Thanks!

martinpitt:
@marusak : Indeed, and I improve that in https://github.com/cockpit-project/cockpit/pull/19104 . Should be ready soonish :crossed_fingers:

martinpitt:
Also, the fix is already being landed upstream, they promised a new release end of July. So we can lose this again soon.

Issue #2947, opened 2022-02-16
Image refresh for fedora-rawhide-boot


KKoukiou:
Image refresh for fedora-rawhide-boot
 * [ ] FAIL: image-refresh fedora-rawhide-boot

cockpituous:
image-refresh in progress on 2-cockpit-8.
Log: https://logs.cockpit-project.org/logs/image-refresh-2947-20220216-125151/

martinpitt:
I fixed the cockpituous token permisisons; unfortunately, for image builds we still need `public_repo`.

cockpituous:
Task failed: https://logs.cockpit-project.org/logs/image-refresh-2947-20220216-125151/

cockpituous:
image-refresh in progress on 1-cockpit-9.
Log: https://logs.cockpit-project.org/logs/image-refresh-2947-20220216-125252/

martinpitt:
Closing this, this needs fixing image-upload. Please iterate in a PR.

PR #5057, opened 2023-07-28
Store anaconda payload in the cockpit image server

KKoukiou:
Image refresh for fedora-rawhide-anaconda-payload
 * [x] image-refresh fedora-rawhide-anaconda-payload

-------

This is the payload used for the anaconda tests for completing the installation.

Right now it is not gated and updating the images on the fedorapeople server is limited to specific people due to lack of permissions.

Additionally since there is not any caching happening, the tests that need the payload download it each time, making the whole concept very inefficient.

Lastly this will help to fix check-progress TestInstallationProgress.testBasic flake, which fails very often on timeout, because of the payload download.

Intructions on how to update it exist here: https://spaces.redhat.com/display/FRONTDOOR/Payload+creation

However, this does not need to change often, we can hardcode this for now.

### DIFF
--- a/image-create
+++ b/image-create
@@ -56,7 +56,13 @@ class MachineBuilder:
     def __init__(self, machine):
         self.machine = machine
         self.iso = self.machine.image.endswith("boot")
-        self.suffix = ".iso" if self.iso else ".qcow2"
+        self.payload = self.machine.image.endswith("payload")
+        if self.iso:
+            self.suffix = ".iso"
+        elif self.payload:
+            self.suffix = ".tar.gz"
+        else:
+            self.suffix = ".qcow2"
 
         # Use /var/tmp/ as this is going to be a huge file; /tmp/ is commonly tmpfs
         self.target_file = self.machine.image_file
@@ -163,7 +169,7 @@ class MachineBuilder:
         if not os.path.exists(self.machine.image_file):
             raise testvm.Failure("Nothing to save.")
 
-        if not self.iso:
+        if not self.iso and not self.payload:
             rebuild = os.path.join(data_dir, self.machine.image + ".rebuild")
 
             # Copy image via convert, to make it sparse again
@@ -171,16 +177,16 @@ class MachineBuilder:
 
         # Hash the image here
         (sha, x1, x2) = subprocess.check_output(
-            ["sha256sum", self.machine.image_file if self.iso else rebuild], universal_newlines=True
+            ["sha256sum", self.machine.image_file if (self.iso or self.payload) else rebuild], universal_newlines=True
         ).partition(" ")
         if not sha:
             raise testvm.Failure("sha256sum returned invalid output")
 
         name = self.machine.image + "-" + sha + self.suffix
         data_file = os.path.join(data_dir, name)
-        shutil.move(self.machine.image_file if self.iso else rebuild, data_file)
+        shutil.move(self.machine.image_file if (self.iso or self.payload) else rebuild, data_file)
 
-        if not self.iso:
+        if not self.iso and not self.payload:
             # Remove temp image file
             os.unlink(self.machine.image_file)
 
--- a/image-trigger
+++ b/image-trigger
@@ -44,6 +44,7 @@ REFRESH = {
     "fedora-coreos": {},
     "fedora-rawhide": {},
     "fedora-rawhide-boot": {},
+    "fedora-rawhide-anaconda-payload": {"refresh-days": 30},
     "ubuntu-2204": {},
     "ubuntu-stable": {},
     "rhel-7-9": {},
--- a/images/scripts/fedora-rawhide-anaconda-payload.bootstrap
+++ b/images/scripts/fedora-rawhide-anaconda-payload.bootstrap
@@ -1,0 +1,6 @@
+#!/bin/sh
+set -eux
+
+OUTPUT="$1"
+
+python3 create-anaconda-payload --output=$OUTPUT
--- a/create-anaconda-payload
+++ b/create-anaconda-payload
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+# This file is part of Cockpit.
+#
+# Copyright (C) 2023 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+# create-anaconda-payload  -- Create a payload to be used by anaconda installer tests.
+
+import argparse
+import os
+import subprocess
+
+from lib.constants import BOTS_DIR
+from machine import testvm
+
+KICKSTART = """\
+cmdline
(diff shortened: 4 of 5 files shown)


KKoukiou (commented):
@martinpitt Are you fine with storing this here?

martinpitt:
Not sure what the "payload" is, is that like the "rpm bundle" that the installer actually puts on disk? In that regard it's similar to boot.iso, and of course it's fine to gate/store it, if it helps you.

Please complete this by actually building that "image", and add it to image-trigger for regularly refreshing it. Also I wonder, how is that not release specific? Shouldn't there be some "38" or "rawhide" in both the source URL and the bots "image" name?

KKoukiou:
> Not sure what the "payload" is, is that like the "rpm bundle" that the installer actually puts on disk? In that regard it's similar to boot.iso, and of course it's fine to gate/store it, if it helps you.
> 
> Please complete this by actually building that "image", and add it to image-trigger for regularly refreshing it. Also I wonder, how is that not release specific? Shouldn't there be some "38" or "rawhide" in both the source URL and the bots "image" name?

It's some RPM bundle indeed.  From my understanding it should be Fedora release specific, you 're right. 

Just a note: the anaconda fedora people account is not something open (I personally don't have access there so I can't update the stored payload). So when I have time I will add the payload build script and run it here (followup) so that I can eliminate that anaconda fedorapeople as intermediate storage..

martinpitt:
FTR, my prefered approach would be to replace the "curl" with the actual thing (lorax? osbuild-composer?) that *builds* that payload. This fixes multiple major problems: That payload stops having to be built and scp'ed by a human, it gets automatically gated, and it doesn't need special credentials.

martinpitt:
That said, feel free to do the curl approach as an intermediate stepping stone, but it doesn't help to lessen confusion: there are even more moving parts involved afterwards, and the timing between manual builds and updates to S3 aren't synced up at all.

KKoukiou:
> That said, feel free to do the curl approach as an intermediate stepping stone, but it doesn't help to lessen confusion: there are even more moving parts involved afterwards, and the timing between manual builds and updates to S3 aren't synced up at all.

The only reason i wanted this change in fast was the one test (see grafana testProgress) fails sometimes with timeout, as it pulls from fedorapeople always. But I will try today to see how far I get with the proper solution.

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-rawhide-anaconda-payload-20230731-082929.log

cockpituous:
image-refresh fedora-rawhide-anaconda-payload done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-rawhide-anaconda-payload-20230731-084128

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-rawhide-anaconda-payload-20230731-084141.log

cockpituous:
image-refresh fedora-rawhide-anaconda-payload done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-rawhide-anaconda-payload-20230731-092908

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-rawhide-anaconda-payload-20230731-092929.log

cockpituous:
image-refresh fedora-rawhide-anaconda-payload done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-rawhide-anaconda-payload-20230731-094354

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-rawhide-anaconda-payload-20230731-094343.log

VladimirSlavik:
Just wanted to +1 the "let's make it automatic" part.

KKoukiou:
See run at https://github.com/rhinstaller/anaconda/pull/4986

KKoukiou:
@martinpitt this is ready for re-review. Tests are passing https://github.com/rhinstaller/anaconda/pull/4986

images/scripts/fedora-rawhide-anaconda-payload.bootstrap
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -eux
+
+OUTPUT="$1"
+
+python3 create-anaconda-payload --output=$OUTPUT
```
martinpitt, line 6: Hmm -- Would you mind renaming create-anaconda-payload to *this* bootstrap script, drop the `--output` argument, and avoid this extra shell wrapper?

martinpitt (approved):
This is most excellent, thanks for figuring this out!

Please add it to `image-trigger`, so that it gets automatically refreshed regularly. Depending on how urgent this is for you, I'm also okay with these two points to happen in an immediate follow-up, so approving with two closed eyes :grin:

KKoukiou:
> This is most excellent, thanks for figuring this out!
> 
> Please add it to `image-trigger`, so that it gets automatically refreshed regularly. Depending on how urgent this is for you, I'm also okay with these two points to happen in an immediate follow-up, so approving with two closed eyes grin

Not urgent I will repost. However, I don't really need this to be refreshed once per week like all images, it's just unneeded noise. Files creates once per fedora major release is fine - so I think this does not need cron refresher.

images/scripts/fedora-rawhide-anaconda-payload.bootstrap
KKoukiou: I see two small problems with the move. 

First is that imports like BOTS_DIR and testvm will not be really possible without some hack to change the python path.

Second is that the seperate script will allow us in the future to reuse this for creating the payload for other fedora versions. 

What do you think?

martinpitt:
> Files creates once per fedora major release is fine

Then how about making this similar to the services image, which is in a similar position? Refreshing every 30 days seems like a good balance between avoiding noise, and making breakages debuggable without having to go back several months.

images/scripts/fedora-rawhide-anaconda-payload.bootstrap
martinpitt: Sure. Not really important.

image-create
```diff
@@ -56,7 +56,13 @@ class MachineBuilder:
...
         self.machine = machine
         self.iso = self.machine.image.endswith("boot")
-        self.suffix = ".iso" if self.iso else ".qcow2"
+        self.payload = self.machine.image.endswith("payload")
+        if self.iso:
+            self.suffix = ".iso"
+        elif self.payload:
+            self.suffix = ".tar.gz"
+        else:
+            self.suffix = ".qcow2"
```
VladimirSlavik, line 65: Is this OK in terms of cockpit infra? I find deciding based on name a bit hacky, but don't know how correct this needs to be.

image-trigger
```diff
@@ -44,6 +44,7 @@ REFRESH = {
     "fedora-coreos": {},
     "fedora-rawhide": {},
     "fedora-rawhide-boot": {},
+    "fedora-rawhide-anaconda-payload": {"refresh-days": 30},
```
VladimirSlavik, line 47: I assume this can be triggered manually at from this, right?

VladimirSlavik (approved):
Thank you, LGTM.

I don't know what would be needed to do this from the anaconda side, but if manual triggering is possible, we could eventually do that on our side so that there's something telling us we own this process....

image-trigger
KKoukiou: Similarly to triggering image refreshes- this can be triggered by creating an issue like this.
https://github.com/cockpit-project/bots/issues/2947

image-create
KKoukiou: it's fine for now.

image-trigger
martinpitt: Yes, or on the CLI with `./image-trigger fedora-rawhide-anaconda-payload` (which creates such a GitHub issue via the API)

martinpitt (approved):

PR #5065, opened 2023-07-31
images: add wireguard-tools to rhel-9

subhoghoshX:
* [x] image-refresh rhel-9-3
 * [x] image-refresh rhel-8-9


images/scripts/rhel.setup
```diff
@@ -208,6 +208,7 @@ tlog \
 valgrind \
 vdo \
 virt-install \
+wireguard-tools \
```
martinpitt, line 211: This doesn't exist on rhel 8, so will fail. Keep it here, but remove it for centos-8-stream and rhel-8* in a rule similar to line 219.

subhoghoshX: Done.

Question: The backslash `\` method of removing something out of a variable isn't part of bash, right? I couldn't reproduce it in a normal script.

martinpitt: I'm not familiar with `\` being used to remove something from a variable. The forward-slash substitution is bash syntax. The backslash here is just to split a single logical line into multiple physical ones, to make them easier to patch and read.

images/scripts/rhel.setup
```diff
@@ -220,6 +221,11 @@ if [ "$IMAGE" = "centos-8-stream" ]; then
     TEST_PACKAGES="${TEST_PACKAGES/kpatch-dnf /}"
 fi
 
+# wireguard-tools not available on rhel-8* and centos-8-stream
+if [ "$IMAGE" = "rhel-8*" ] || [ "$IMAGE" = "centos-8-stream" ]; then
```
martinpitt, line 225: This "globbing" doesn't work. The way to write that in shell is `[ "${IMAGE#rhel-8*}" != "$IMAGE" ]`, see e.g. line 53.

images/scripts/rhel.setup
subhoghoshX: Sorry I meant forward slash :man_facepalming:. So it is a builtin bash feature(?) Nice!

images/scripts/rhel.setup
subhoghoshX: Fixed it, thanks!

cockpituous:
image-refresh rhel-9-3 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-9-3-20230801-073918

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/rhel-9-3-20230801-073939.log

cockpituous:
image-refresh rhel-8-9 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-8-9-20230801-080056

martinpitt (approved):
Thanks!

PR #5070, opened 2023-08-01
Remove ubuntu-2004 leftover scripts

skobyda:
We no longer test anything on ubuntu-2004 since it was dropped at https://github.com/cockpit-project/bots/pull/3294, so remove its leftover setup scripts.


martinpitt (approved):
:broom: Dakujem!

PR #5072, opened 2023-08-01
testmap: Drop fedora-38/pybridge

martinpitt:
The most recent fedora-38 image refresh picked up cockpit 297, which ships the Python bridge. Hence, the /pybridge scenarios for TEST_OS_DEFAULT aka fedora-38 are entirely redundant now.

This only leaves cockpit's centos-8-stream/pybridge scenario, which we want to keep for a fair while, to test the "connect to RHEL 8 system with beibooted bridge" case.


jelly (approved):
:snake:

PR #5073, opened 2023-08-01

cockpituous:


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/centos-8-stream-20230801-230000.log

martinpitt:
Meh, the dreaded [LUKS OOM](https://cockpit-logs.us-east-1.linodeobjects.com/pull-5073-20230801-230145-b3981e21-centos-8-stream-pybridge-storage-cockpit-project-cockpit/log.html#28-2). This could do any of:

 - Retry and close our yes
 - Preferably, configure LUKS' PBKDF to use less memory, then we can convert a lot of destructive tests to nondestructive
 - If the above is too hard, convert TestStorageMounting.testNeverAuto to a destructive test with provisioning more RAM
 - Provide the nondestructive VMs with more RAM in general. The TF ones have enough.

skobyda:
> 

We already have tests for LUKS where we provision more memory: https://github.com/cockpit-project/cockpit/blob/297/test/verify/check-storage-mounting#L405

The test which fails here uses LUKS, but is not part of TestStorageMountingLUKS class. Maybe it should be moved under `TestStorageMountingLUKS` as other tests

martinpitt:
> Maybe it should be moved under TestStorageMountingLUKS as other tests

Yes, that would be the third option from my list above.

PR #5081, opened 2023-08-04
test-failure-policy: add option to test other naughties

jelly:
When a test fails, compare it to all known naughties and show this in the failed test.

This requires Cockpit changes as well made [here](https://github.com/cockpit-project/cockpit/pull/19173) and a hack to remove a known naughty which is applied to two other distros.

See https://cockpit-logs.us-east-1.linodeobjects.com/pull-19173-20230804-085354-1b0ea8fb-ubuntu-stable-other/log.html#27

### DIFF
--- a/test-failure-policy
+++ b/test-failure-policy
@@ -38,6 +38,8 @@ def main():
     parser = argparse.ArgumentParser(description='Check a traceback for a known issue')
     parser.add_argument('-o', "--offline", action='store_true',
                         help="Work offline, don't fetch new data or contact servers")
+    parser.add_argument('-a', '--all', action='store_true',
+                        help='Check for matching naughties in all other images')
     parser.add_argument('image', help="The image to check against")
     opts = parser.parse_args()
 
@@ -61,6 +63,11 @@ def main():
         if number:
             print("Known issue #{0}".format(number))
             return 77
+        elif opts.all:
+            number, img = check_all_known_issues(api, output, opts.image)
+            if number:
+                print(f"Known issue #{number} in {img}")
+                return 78
 
         if checkRetry(output):
             print("due to failure of test harness or framework")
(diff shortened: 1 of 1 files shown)


test-failure-policy
```diff
@@ -66,6 +68,12 @@ def main():
             print("due to failure of test harness or framework")
             return 1
 
+        if not number and opts.all:
```
martinpitt, line 71: this would be better as an "else" branch after line 63, to keep the "known issue" stuff together. It's confusing and also inefficient to have the "checkRetry" thing in between.

test-failure-policy
```diff
@@ -66,6 +68,12 @@ def main():
...
+            number, img = check_all_known_issues(api, output, opts.image)
+            if number:
+                print(f"Known issue #{number} in {img}")
+                return 78
```
martinpitt, line 75: This isn't a magic number, but I don't mind much.

test-failure-policy
```diff
@@ -109,6 +117,22 @@ def checkRetry(trace):
 # -----------------------------------------------------------------------------
 # Known Issue Matching and Filing
 
+
+def check_all_known_issues(api, trace, image):
+    img = None
+    number = None
+    naughty_dir = os.path.join(BOTS_DIR, "naughty")
+
+    for img in os.listdir(naughty_dir):
```
martinpitt, line 126: This will also iterate over symlinks, and half of the entries are links. Please move to https://docs.python.org/3/library/os.html#os.scandir and filter out links. This happens a lot in tests and check_known_issue() is expensive, so not being overly inefficient here matters IMHO.

martinpitt (commented):
Thanks! Please do another test run with the second "break" commit, to ensure everything still works.

martinpitt (changes requested):
(see above, wrong button)

test-failure-policy
martinpitt: Nevermind, it becomes magic in https://github.com/cockpit-project/cockpit/pull/19173

test-failure-policy
jelly: Good point, having fast CI is worth the effort.

jelly:
Triggered another run here: https://cockpit-logs.us-east-1.linodeobjects.com/pull-19173-20230824-092328-1204a7a1-ubuntu-stable-other/log.html

Waiting on the result and then we can drop the commit, and move it to `ready for review`.

martinpitt (approved):
Cheers!

PR #5138, opened 2023-08-23
testmap: add Fedora-39 to manual contexts

tomasmatus:
Allows testing on Fedora 39.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -63,6 +63,7 @@ REPO_BRANCH_CONTEXT = {
         '_manual': [
             'fedora-testing',
             'fedora-rawhide',
+            'fedora-39',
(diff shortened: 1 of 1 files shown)


lib/testmap.py
```diff
@@ -127,6 +130,7 @@ def contexts(image, *scenarios: Iterable[str]):
         # These can be triggered manually with bots/tests-trigger
         '_manual': [
             'rhel-8-6/osbuild-composer',
+            'fedora-39',
```
martinpitt, line 133: Lorax isn't a fedora thing, so that's overzealous. But it doesn't really hurt, of course.

martinpitt (approved):
Thanks!

PR #5140, opened 2023-08-23
images: add Fedora-39 image

tomasmatus:
* [x] image-refresh fedora-39

PR #5141, opened 2023-08-23
testmap: Test starter-kit on fedora-39

martinpitt:
Already tested in #5140, but will trigger again.


tomasmatus (approved):
So far so good!

PR #5142, opened 2023-08-23
lib/testmap: update cockpit-composer testing for rhel 89 and 93

jkozol:
Tests passing here: https://github.com/osbuild/cockpit-composer/pull/1904


martinpitt (approved):
Thanks! Note that we have fedora-39 since yesterday, you may want to enable that as well?

PR #5145, opened 2023-08-24
testmap: Test c-podman on fedora-39

tomasmatus:


martinpitt (approved):
Smooth!

As we dropped F37 from releasing in packit, I think it's also time to stop running tests on it, yes. You can either update this PR, or drop f37 wholesale after f39 got enabled everywhere, your choice. Thanks, and good luck with c-machines and cockpit!

tomasmatus:
Ok! I will drop f37 once f39 is enabled everywhere.

PR #5146, opened 2023-08-24
testmap: test c-machines on fedora-39

tomasmatus:
* [x] image-refresh fedora-39
 * [x] image-refresh fedora-37

PR #5148, opened 2023-08-24
images: Add qemu-audio-spice as dependency

tomasmatus:
image refreshes from #5146


martinpitt (approved):
Thanks! The test results from #5146 are good enough.

PR #5149, opened 2023-08-24
Fedora nightly scenario

jelly:
In general the idea is:

* Create a new Github workflow which runs nightly
* Github workflow invokes `./bots/tests-trigger --repo cockpit-project/cockpit-podman "-" $DEFAULT_OS/nightly`
* Commit status is posted, webhook runs and the we end up running: `/tests-scan -s $sha --context $DEFAULT_OS/nightly --repo cockpit-project/cockpit-podman`
* Issue is created if fails, otherwise :heavy_check_mark: should be shown on the commit.

Open discussion points:

* [ ] It seems we can't get the direct logs url from s3-streamer, we can either pol statuses or set an `env` variable so `s3-streamer`'s children know where to look @martinpitt what is your opinion here?
* [ ] Another discussion point, do we drop `--pull-number 0` and add a new argument to `tests-invoke`.

Testing ideas:

We could test the new `tests-invoke` option if we:
* [ ] Provide a mock Github issues endpoint where we can create an issue
* [ ] mock `BASE_DIR` so we can control the executed `test/run` and simply `exit 1` or `exit 0`
* [ ] mock / override `time.sleep(60)` or change the polling to be more aggressive as waiting 1 minute in a test is bonkers.

### DIFF
--- a/task/test-tests-scan
+++ b/task/test-tests-scan
@@ -25,6 +25,7 @@ import tempfile
 import unittest
 
 from lib.constants import BOTS_DIR
+from task import github
 from task.test_mock_server import MockHandler, MockServer
 
 ADDRESS = ("127.0.0.7", 9898)
@@ -69,11 +70,20 @@ class Handler(MockHandler):
             self.replyJson([])
         elif self.path.startswith('/repos/project/repo/pulls'):
             self.replyJson([self.server.data['/repos/project/repo/pulls/1']])
+        elif self.path.endswith("/issues"):
+            issues = self.server.data['issues']
+            self.replyJson(issues)
         else:
             self.send_error(404, 'Mock Not Found: ' + self.path)
 
     def do_POST(self):
-        self.send_error(405, 'Method not allowed: ' + self.path)
+        if self.path.startswith("/repos/cockpit-project/cockpit/issues"):
+            content_len = int(self.headers.get('content-length'))
+            data = json.loads(self.rfile.read(content_len).decode('utf-8'))
+            self.server.data['issues'] = [data]
+            self.replyJson(data)
+        else:
+            self.send_error(405, 'Method not allowed: ' + self.path)
 
 
 class TestTestsScan(unittest.TestCase):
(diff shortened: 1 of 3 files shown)


jelly:
`tests-invoke` is sadly not easy to test without waiting ages on `time.sleep(60)`. Or'd we have to hacky invoke `main()`, but I'll skip that for now.

tests-invoke
```diff
@@ -94,6 +93,8 @@ def main():
                 time.sleep(60)
         return_code = p.returncode
         sys.stderr.write("Test run finished, return code: {0}\n".format(return_code))
+        if not opts.pull_number:  # Create an issue
+            ...  # TODO implement
```
martinpitt: Yeah, that still needs to happen :)

tests-scan
```diff
@@ -287,6 +287,21 @@ def cockpit_tasks(api, update, contexts, repo, pull_data, pull_number, sha, amqp
...
+            "title": f"{sha}",
+            "number": 0,
+            "head": {
+                "sha": sha,
+                "user": {
+                    "login": "cockpit-project"
+                }
+            },
+            "base": {
+                "ref": "main"  # TODO: anaconda/cockpit-certificates
```
martinpitt, line 301: perhaps change the comment to "projects which have a "master" branch, to clarify what it means? But this is actually easy to do, testmap.py has a `get_default_branch(repo)` function.

tests-invoke
```diff
@@ -93,8 +93,15 @@ def main():
                 time.sleep(60)
         return_code = p.returncode
         sys.stderr.write("Test run finished, return code: {0}\n".format(return_code))
-        if not opts.pull_number:  # Create an issue
-            ...  # TODO implement
+        # Nightly tests use --pull-number 0
```
martinpitt, line 96: Ah, please squash this.

tests-invoke
```diff
@@ -93,8 +93,15 @@ def main():
...
+        if opts.pull_number == "0" and return_code != 0:
```
martinpitt, line 97: Any chance this could still be `if not opts.pull_number`, so that it could be `None`? That would be cleaner than `"0"`.

martinpitt (changes requested):
Cheers! Could you please add a unit test for the `tests-scan` part of this, using your shiny new test-tests-scan?

jelly:
Pushed with a test commit which tests the issue creation, but this leads to 60 second delay in tests due to https://github.com/cockpit-project/bots/blob/main/tests-invoke#L94

task/test-tests-scan
```diff
@@ -206,6 +216,26 @@ class TestTestsScan(unittest.TestCase):
...
+                fp.write("#!/bin/bash\nexit 1")
+            os.system(f"chmod +x {testdir}/run")
+            proc = subprocess.Popen([script, *args], stdout=subprocess.PIPE, universal_newlines=True,
+                                    env={"GITHUB_API": f"http://{ADDRESS[0]}:{ADDRESS[1]}"}, cwd=tempdir)
+            output, stderr = proc.communicate()
+            api = github.GitHub(f"http://{ADDRESS[0]}:{ADDRESS[1]}/")
+
+            self.assertEqual(output, "")
+            self.assertEqual(proc.returncode, 1)
+            self.assertEqual(len(api.get("issues")), 1)
```
martinpitt, line 236: Could this check the content of the issue? that the expected SHA made it into the body? Or doesn't our mock server go that far?

martinpitt (dismissed):
This looks really cool, and I don't want to block this for any longer. Thanks!

I suggest to change the sleep to something like `int(os.getenv("TEST_INVOKE_SLEEP", "60"))` and set the env variable to "1" in the unit test. I leave it up to you if you want to do this here or in a follow-up.

tests-scan
```diff
@@ -287,6 +291,21 @@ def cockpit_tasks(api, update, contexts, repo, pull_data, pull_number, sha, amqp
...
+            "head": {
+                "sha": sha,
+                "user": {
+                    "login": "cockpit-project"
+                }
+            },
+            "base": {
+                "ref": testmap.get_default_branch(repo)
+            },
+            "labels": [{"name": "main"}],
```
martinpitt, line 307: Ah, does this also need to change to get_default_branch()? If that is a github label, I'm not sure why it would even be called "main", instead of the "nightly" one you defined in tests-invoke?

jelly:
> This looks really cool, and I don't want to block this for any longer. Thanks!
> 
> I suggest to change the sleep to something like `int(os.getenv("TEST_INVOKE_SLEEP", "60"))` and set the env variable to "1" in the unit test. I leave it up to you if you want to do this here or in a follow-up.

I'll do it here, I do actually **really** dislike such hacks but there isn't a better way :(

PR #5155, opened 2023-08-25
naughty: Make patterns for #5090 less specific

mvollmer:
The message is not triggered by the function in the backtrace, but sometime earlier.

The tests in https://github.com/cockpit-project/cockpit/pull/19079 have changed and now we need to update these naughties.


martinpitt (approved):
I suppose the "> warning" message is specific enough. Thanks!

PR #5160, opened 2023-08-27

cockpituous:
 * [x] https://github.com/cockpit-project/cockpit/pull/19265

### DIFF
--- a/images/centos-8-stream
+++ b/images/centos-8-stream
@@ -1,1 +1,1 @@
-centos-8-stream-ff085b5a037ae2b76055b460b3c4b1b8be30b0717ea7720cee743292412f8e9d.qcow2
+centos-8-stream-895a8ab606dee80aab943ada5ee0bbc2b1b4f20abffa06384b273bb9db7a4536.qcow2


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/centos-8-stream-20230827-225858.log

jelly:
Retrying to see if this is a SELinux flake.

martinpitt:
It also [happens on the current image](https://cockpit-logs.us-east-1.linodeobjects.com/pull-19069-20230831-104720-4f88c3d0-centos-8-stream-pybridge-other/log.html), so it's not a regression -- but one of our top flakes right now. I'll look at it ASAP. It may just take longer, but first I'll see whether I can reproduce it.

martinpitt:
I think the reason is that the string got wider, and the original reference marks only the original area as "ignore me"; that happens via transparency or some other magic value in the ref png. So what happens is that the wider string now flows outside that marked area, and thus becomes diff-y; if we accept that new reference, that will just make the ignored area wider, and thus should work with both the old and new image, and woud also conceptually be correct; does that make sense @mvollmer , or am I misunderstanding how ignoring works?

martinpitt:
Sent fix to https://github.com/cockpit-project/cockpit/pull/19265 , blocking on that.

martinpitt (approved):

PR #5185, opened 2023-08-29
Capture VM's console if VM hangs on boot or poweroff

skobyda:
* [ ] FAIL: image-refresh centos-9-stream

### DIFF
--- a/image-create
+++ b/image-create
@@ -38,6 +38,7 @@ parser.add_argument('-v', '--verbose', action='store_true', help='Display verbos
 parser.add_argument('-s', '--sit', action='store_true', help='Sit and wait if setup script fails')
 parser.add_argument('-n', '--no-save', action='store_true', help='Don\'t save the new image')
 parser.add_argument('-u', '--upload', action='store_true', help='Upload the image after creation')
+parser.add_argument('--capture-console', action='store_true', help='Capture VM console to stderr')
 parser.add_argument('--no-build', action='store_true', dest='no_build',
                     help='Don''t build packages and create the vm without build capabilities')
 parser.add_argument("--store", default=None, help="Where to send images")
@@ -207,7 +208,12 @@ try:
         memory_mb = 3072
     else:
         memory_mb = 2048
-    machine = testvm.VirtMachine(verbose=args.verbose, image=args.image, memory_mb=memory_mb, maintain=True)
+
+    machine = testvm.VirtMachine(verbose=args.verbose,
+                                 capture_console=args.capture_console,
+                                 image=args.image,
+                                 memory_mb=memory_mb,
+                                 maintain=True)
     builder = MachineBuilder(machine)
     builder.build()
     if not args.no_save:
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -71,7 +71,12 @@ class Machine(ssh_connection.SSHConnection):
         if not label and image != "unknown":
             label = "{}-{}-{}".format(image, ssh_address, ssh_port)
 
-        super(Machine, self).__init__(user, ssh_address, ssh_port, identity_file, verbose=verbose, label=label)
+        super(Machine, self).__init__(user,
+                                      ssh_address,
+                                      ssh_port,
+                                      identity_file,
+                                      verbose=verbose,
+                                      label=label)
 
         self.arch = arch
         self.image = image
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -75,7 +75,7 @@ TEST_DOMAIN_XML = """
     <acpi/>
   </features>
   <devices>
-    <disk type='file' snapshot='external'>
+    <disk type='file'>
       <driver name='qemu' type='qcow2' cache='unsafe'/>
       <source file='{drive}'/>
       <target dev='vda' bus='virtio'/>
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -86,8 +86,9 @@ TEST_DOMAIN_XML = """
     <graphics type='vnc' autoport='yes' listen='127.0.0.1'>
       <listen type='address' address='127.0.0.1'/>
     </graphics>
-    <console type='pty'>
+    <console type='{console_type}'>
       <target type='serial' port='0'/>
+      {console_source}
     </console>
     <disk type='file' device='cdrom'>
       <source file='{iso}'/>
@@ -245,12 +246,14 @@ class VirtMachine(Machine):
     cpus = None
 
     def __init__(self, image, networking=None, maintain=False, memory_mb=None, cpus=None,
-                 graphics=False, **args):
+                 capture_console=False, graphics=False, **args):
         self.maintain = maintain
 
         self.memory_mb = memory_mb or VirtMachine.memory_mb or MEMORY_MB
         self.cpus = cpus or VirtMachine.cpus or 1
         self.graphics = graphics
+        if capture_console:
+            self.console_file = tempfile.NamedTemporaryFile(suffix='.log', prefix='console-')
 
         # Set up some temporary networking info if necessary
         if networking is None:
@@ -331,6 +334,8 @@ class VirtMachine(Machine):
             "memory_in_mib": self.memory_mb,
             "drive": image_to_use,
             "iso": os.path.join(BOTS_DIR, "machine", "cloud-init.iso"),
+            "console_type": "file" if self.console_file else "pty",
+            "console_source": f"<source path='{self.console_file.name}'/>" if self.console_file else "",
         }
 
         if os.path.exists("/dev/kvm"):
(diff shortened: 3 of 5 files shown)


skobyda:
Tried to test it locally with currently failing "rhel-9-3", it managed to capture the console log successfully:
https://gist.githubusercontent.com/skobyda/1b85557c87ad1d91be954e1e00b9f42a/raw/47c9359805184bf3c2a5283184d413213e79eda3/gistfile1.txt

machine/machine_core/ssh_connection.py
```diff
@@ -69,6 +69,22 @@ def message(self, *args):
         if self.verbose:
             sys.stderr.write(" ".join(args) + '\n')
 
+    def print_console_log(self):
+        """Prints VM's console to stderr"""
+        console_file = "/tmp/console.log"
+        try:
+            f = open(console_file)
+            console_log = f.read().strip()
```
martinpitt: Please reduce the scope of the try/except to the open/read pair, and catch all `OSError`s -- this is already in an exception handler, so this better not crash.  Print the exception when it happens for debugging.

Please also use a `with` context (see codeql)

machine/machine_core/ssh_connection.py
```diff
@@ -69,6 +69,22 @@ def message(self, *args):
...
+
+            if len(console_log) == 0:
+                sys.stderr.write(f"VM's console log file '{console_file}' is empty")
+                return
+
+            sys.stderr.write(f"---- Console log start's here ----\n{console_log}\n---- Console log end's here ----")
```
martinpitt: "starts" and "ends" (no apostrophes)

Also, write() does not automatically append a `\n`. Either do that, or use `print(..., file=sys.stderr)`

machine/machine_core/ssh_connection.py
```diff
@@ -69,6 +69,22 @@ def message(self, *args):
         if self.verbose:
             sys.stderr.write(" ".join(args) + '\n')
 
+    def print_console_log(self):
+        """Prints VM's console to stderr"""
+        console_file = "/tmp/console.log"
```
martinpitt: This looks wrong. Is this a test leftover? If this path is set anywhere, it won't work on our bots, as they run many VMs in parallel. Also, fixed paths in /tmp are a no-go.

martinpitt (changes requested):
Thanks for figuring this out!

machine/machine_core/ssh_connection.py
skobyda: Fixed

machine/machine_core/ssh_connection.py
skobyda: Fixed

machine/machine_core/ssh_connection.py
skobyda: Fixed

machine/machine_core/machine.py
```diff
@@ -53,8 +53,8 @@
 
 
 class Machine(ssh_connection.SSHConnection):
-    def __init__(self, address="127.0.0.1", image="unknown", verbose=False, label=None, browser=None,
-                 user="root", identity_file=None, arch="x86_64", ssh_port=22, web_port=9090):
+    def __init__(self, address="127.0.0.1", image="unknown", verbose=False, console_file=None, label=None,
+                 browser=None, user="root", identity_file=None, arch="x86_64", ssh_port=22, web_port=9090):
```
martinpitt, line 58: This is the wrong place. The *only* class where we can do this is VirtMachine. We can't control already running/external/Testing Farm etc. machine types. Please move all of this into VirtMachine and handle the argument there. Machine and SshConnection should not be touched by this.

machine/machine_core/ssh_connection.py
```diff
@@ -69,6 +70,24 @@ def message(self, *args):
         if self.verbose:
             sys.stderr.write(" ".join(args) + '\n')
 
+    def print_console_log(self):
+        """Prints VM's console to stderr"""
+        if not self.console_file:
+            return
+
+        try:
+            f = open(self.console_file)
```
martinpitt, line 79: This needs to go inside the `with` already. I. e. `with open(...) as f: log = read()`. The processing of `console_log` shouldn't be in the context.

machine/machine_core/ssh_connection.py
```diff
@@ -69,6 +70,24 @@ def message(self, *args):
...
+            console_log = f.read().strip()
+        except OSError as ex:
+            sys.stderr.write(f"Failed to open '{self.console_file}': {0}".format(ex.strerror))
```
martinpitt, line 82: Missing `\n`. Also, please put the exception into the f-string. Combining f-string and .format() is weird.

machine/machine_core/ssh_connection.py
```diff
@@ -69,6 +70,24 @@ def message(self, *args):
...
+
+        with f:
+            if len(console_log) == 0:
+                sys.stderr.write(f"VM's console log file '{self.console_file}' is empty")
```
martinpitt, line 86: Missing `\n`

machine/machine_core/machine_virtual.py
```diff
@@ -330,6 +331,8 @@ def _start_qemu(self):
             "memory_in_mib": self.memory_mb,
             "drive": image_to_use,
             "iso": os.path.join(BOTS_DIR, "machine", "cloud-init.iso"),
+            "console_type": "file" if self.console_file else "pty",
+            "console_source": f"<source path='{self.console_file}'/>" if self.console_file else "",
```
martinpitt, line 335: This shouldn't be a separate commit, it belongs into the previous one.

image-create
```diff
@@ -207,7 +207,14 @@ try:
         memory_mb = 3072
     else:
         memory_mb = 2048
-    machine = testvm.VirtMachine(verbose=args.verbose, image=args.image, memory_mb=memory_mb, maintain=True)
+
+    console_file = tempfile.NamedTemporaryFile(suffix='.log', prefix='console-').name
```
martinpitt: Not keeping the `NamedTemporaryFile` object around will immediately close it, and thus also remove the file again. So the only effect of this is that you keep a random name. It's not super critical here, but please don't learn this pattern -- it's actively dangerous in "production" applications. Drop the `.name` here, and pass `console_file.name` to VirtMachine instead, so that the TemporaryFile object survives throughout image-create.

martinpitt (changes requested):
Thanks!

image-create
```diff
@@ -207,7 +207,14 @@ try:
         memory_mb = 3072
     else:
         memory_mb = 2048
-    machine = testvm.VirtMachine(verbose=args.verbose, image=args.image, memory_mb=memory_mb, maintain=True)
+
+    console_file = tempfile.NamedTemporaryFile(suffix='.log', prefix='console-')
```
martinpitt: This gets changed again in the third commit. Please move all the changes there.

martinpitt (changes requested):
Thanks! one more cleanup round, then it should be good!

image-create
skobyda: But console_file is necessary in second commit. Without it it doesn't work. Or do you mean to squash second and third commit?

martinpitt: If they are so interdependent, then just squash them, yes.

image-create
```diff
@@ -207,7 +208,16 @@ try:
...
     else:
         memory_mb = 2048
-    machine = testvm.VirtMachine(verbose=args.verbose, image=args.image, memory_mb=memory_mb, maintain=True)
+
+    console_file = None
+    if args.capture_console:
+        console_file = tempfile.NamedTemporaryFile(suffix='.log', prefix='console-')
+
+    machine = testvm.VirtMachine(verbose=args.verbose,
+                                 console_file=console_file.name,
```
martinpitt, line 217: Sorry that I didn't see this earlier, as it was too distributed : This is a somewhat weird API. You stuff in a path, but the caller never actually does anything with it. Instead it's handled all internally. I think it may be cleaner to make this a `capture_console=True` flag, and generate the file name inside of VirtMachine. That also fixes the temp file scope issue that you stumbled over before.

martinpitt (changes requested):
Thanks! This is a lot cleaner. I didn't see the API issue before, so one more round?

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/centos-9-stream-20230907-152222.log

martinpitt (approved):
Thanks! Image refresh runs now. Assuming that fails with a proper log.

martinpitt (changes requested):
Oh, good that we tested that, see https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/centos-9-stream-20230907-152222.log

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/centos-9-stream-20230907-154848.log

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/centos-9-stream-20230907-161515.log

martinpitt (approved):
Nice, now it fails in the expected manner!

PR #5186, opened 2023-08-30
testmap: Drop fedora-37

tomasmatus:


lib/testmap.py
```diff
@@ -247,7 +242,7 @@ def contexts(image, *scenarios: Iterable[str]):
         "rhel-8-8@candlepin/subscription-manager/subscription-manager-1.28",
         "rhel-9-3@candlepin/subscription-manager-cockpit",
     ],
-    # Anaconda builds in fedora-37 and runs tests in fedora-rawhide-boot
+    # Anaconda builds in fedora-39 and runs tests in fedora-rawhide-boot
```
tomasmatus, line 245: I believe this was outdated comment.

martinpitt:
Dropping the image itself is premature -- It first needs to go away from all projects testmaps. I don't know for how long sub-man and composer still want to support F37, that needs sign-off from @jkozol and @ptoscano. What I meant is to drop fedora-37 from the testmap for our team's projects, for those where we already dropped f37 from packit.

Thanks!

tomasmatus:
right

lib/testmap.py
martinpitt: Indeed it was. It should actually be `fedora-rawhide`, as the line directly below says. As this PR is cheap to update, would you mind fixing it?

ptoscano:
New versions of subscription-manager are generally uploaded to any supported versions of Fedora; the last version was uploaded to Fedora 37 to 40 (both included) last week.

I don't mind dropping `fedora-37`, however isn't it a bit premature?

lib/testmap.py
martinpitt: I pushed this myself.

martinpitt:
@ptoscano Yes, that's what I thought. It's of course fine to keep the image around as long as any of our users is still releasing/supporting there.

martinpitt (approved):
Thanks!

PR #5189, opened 2023-08-30
Image refresh for fedora-38

cockpituous:
Image refresh for fedora-38
 * [x] image-refresh fedora-38
 * [x] https://github.com/cockpit-project/cockpit/pull/19267


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-38-20230830-230202.log

cockpituous:
image-refresh fedora-38 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-38-20230830-230707

cockpituous:
Completed. Log: Error accessing /repos/cockpit-project/bots/pulls
  Status: 422
  Reason: Unprocessable Entity
  Response: {"message":"The specified issue is already attached to a pull request.","errors":[{"value":5189,"resource":"PullRequest","field":"issue","code":"invalid"}],"documentation_url":"https://docs.github.com/rest/pulls/pulls#create-a-pull-request"}

martinpitt:
[This may require a pixel update](https://cockpit-logs.us-east-1.linodeobjects.com/pull-5189-20230830-230425-9f66c7a0-fedora-38-networking-cockpit-project-cockpit/log.html)

martinpitt:
I think the reason is that the string got wider, and the original reference marks only the original area as "ignore me"; that happens via transparency or some other magic value in the ref png. So what happens is that the wider string now flows outside that marked area, and thus becomes diff-y; if we accept that new reference, that will just make the ignored area wider, and thus should work with both the old and new image, and woud also conceptually be correct; does that make sense @mvollmer , or am I misunderstanding how ignoring works?

martinpitt (approved):

PR #5200, opened 2023-09-01
Detect nightly triggers without an open PR better

jelly:
For direct triggered tests using `./tests-trigger` issues where being created for failing tests. Before 14b13cb3bc0cdf7df9740 an status event would trigger tests-scan and look up the pull request for the given sha, given the introduction of the `elif sha` condition it would now omit the `--pull-number` creating an issue when the test failed.

As we only want this functionality for the nightly scenario triggered on main, introduce a check if an open pull request exists.

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -376,6 +376,12 @@ class GitHub(object):
             result = result + issues
         return result
 
+    def has_open_prs(self, sha):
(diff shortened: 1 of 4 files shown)


martinpitt (approved):
:cowboy_hat_face:

PR #5212, opened 2023-09-05
Misc fixes

jelly:
Split off from my local typing additions.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -291,7 +291,7 @@ class SSHConnection(object):
 
         additional_ssh_params += self.__execution_opts(direct=direct)
 
-        if getattr(command, "strip", None):  # Is this a string?
+        if isinstance(command, str):
             cmd += [command]
             if not quiet:
                 self.message("+", command)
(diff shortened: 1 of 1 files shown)


machine/machine_core/ssh_connection.py
```diff
@@ -249,7 +249,7 @@ def _ensure_ssh_master(self):
             self._start_ssh_master()
 
     def __ssh_direct_opt_var(self, direct=False):
-        return os.getenv("TEST_SSH_DIRECT", direct)
+        return direct or os.getenv("TEST_SSH_DIRECT") is not None
```
martinpitt, line 252: That changes the semantics of `TEST_SSH_DIRECT= program` though, i. e. empty string now counts as True. I'm not sure how important that is, but usually an empty value in shell counts as false-y. Would you mind doing `bool(getenv(..))` instead?

martinpitt (changes requested):

machine/machine_core/ssh_connection.py
jelly: Alright, simplified it!

martinpitt (approved):
Cheers!

PR #5228, opened 2023-09-08
tests-invoke: add the scenario to the title

jelly:
We create an issue per test scenario, adding the scenario to the title makes it easier to spot the difference.


martinpitt (changes requested):
I'd actually prefer to have the full context here, e. g. "fedora-38/updates-testing". For a long time I have wished that we could run updates-testing (or COPRs) on multiple distros. E.g. fedora-39/updates-testing would also be highly interesting now. That shold be in TEST_OS, so not hard to get?

martinpitt (approved):
Looks good, thanks!

PR #5231, opened 2023-09-08
Python typing follow up

jelly:


machine/machine_core/timeout.py
```diff
@@ -16,13 +16,18 @@
...
+import machine_core.ssh_connection as ssh_connection
 
 
 class Timeout:
     """ Add a timeout to an operation
         Specify machine to ensure that a machine's ssh operations are canceled when the timer expires.
     """
-    def __init__(self, seconds=1, error_message='Timeout', machine=None):
+    def __init__(self, seconds: int = 1, error_message: str = 'Timeout',
+                 machine: Optional[ssh_connection.SSHConnection] = None) -> None:
```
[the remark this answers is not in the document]
jelly: This is old.

machine/machine_core/ssh_connection.py
```diff
@@ -205,13 +209,14 @@ def _start_ssh_master(self):
             else:
                 raise exceptions.Failure("SSH master process exited with code: {0}".format(proc.returncode))
 
+        assert proc, "SSH master process not created succesfully"
```
allisonkarlitskaya, line 212: This assert is bad and a rather direct result of the retry loop above being a bit of a disaster.  How about lifting that into a separate function which returns the subprocess unconditionally?

machine/machine_core/ssh_connection.py
```diff
@@ -221,14 +226,16 @@ def _kill_ssh_master(self):
             self.ssh_master = None
         if self.ssh_process:
             self.message("killing ssh master process", str(self.ssh_process.pid))
+            assert self.ssh_process.stdin, "no stdin for SSH master process"
```
allisonkarlitskaya, line 229: I'm not a fan of the commentary on `assert`s, and please use `is not None` here, if that's what you mean.

machine/machine_core/ssh_connection.py
```diff
@@ -248,26 +255,28 @@ def _check_ssh_master(self):
...
 
-    def __execution_opts(self, direct=False):
+    def __execution_opts(self, direct: bool = False) -> List[str]:
         direct = self.__ssh_direct_opt_var(direct=direct)
         if direct:
             return ["-i", self.identity_file]
-        else:
+        elif self.ssh_master:
             return ["-o", "ControlPath=" + self.ssh_master]
+        return []
```
allisonkarlitskaya, line 271: Tricky.  If you're going to take this approach to avoid the warning about `self.ssh_master` possibly being `None` then I'd prefer to see some kind of an error raised here instead of returning empty.

machine/machine_core/ssh_connection.py
```diff
@@ -248,26 +255,28 @@ def _check_ssh_master(self):
...
-                stdout=subprocess.PIPE, quiet=False, direct=False, timeout=120,
-                ssh_env=["env", "-u", "LANGUAGE", "LC_ALL=C"], check=True):
+    def execute(self, command: Union[List[str], str], input: Optional[str] = None, environment: Dict[str, str] = {},
+                stdout: int = subprocess.PIPE, quiet: bool = False, direct: bool = False,
+                timeout: int = 120, ssh_env: List[str] = ["env", "-u", "LANGUAGE", "LC_ALL=C"],
+                check: bool = True) -> Optional[str]:
         """Execute a shell command in the test machine and return its output.
 
-            command: The string or argument list to execute by /bin/sh (still with shell interpretation)
+            command: The string or argument list to execute by /bin/sh (still with shell interpretation) FIXME?!
```
allisonkarlitskaya, line 279: FIXME?!

machine/machine_core/timeout.py
```diff
@@ -16,13 +16,19 @@
 # along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
 
 import signal
+from types import FrameType, TracebackType
+from typing import TYPE_CHECKING, Optional, Type
+
+if TYPE_CHECKING:
+    import machine_core.ssh_connection as ssh_connection
```
allisonkarlitskaya, line 23: This is better spelled as `from machine_core import ssh_connection`.  I'm not a fan of the `TYPE_CHECKING` block, though.  This it the circular dependency thing, right?  I feel like we need to sort that out in a better way, possibly through the use of an interface...

machine/machine_core/timeout.py
```diff
@@ -32,20 +38,21 @@ def __init__(self, seconds=1, error_message='Timeout', machine=None):
...
         raise RuntimeError(self.error_message)
 
-    def __enter__(self):
+    def __enter__(self) -> None:
         if self.seconds:
             signal.signal(signal.SIGALRM, self.handle_timeout)
             signal.alarm(self.seconds)
 
-    def __exit__(self, _type, value, traceback):
+    def __exit__(self, _type: Optional[Type[BaseException]], _value: Optional[BaseException],
```
allisonkarlitskaya, line 54: This is pretty frustrating...

If we don't care about these, I've seen it done like:

```python
def __exit__(self, *_args: object) -> None:
    ...
```

machine/machine_core/machine.py
```diff
@@ -85,12 +89,9 @@ def __init__(self, address="127.0.0.1", image="unknown", verbose=False, label=No
             (self.web_address, unused, self.web_port) = browser.rpartition(":")
         else:
             self.web_address = browser
-            self.web_port = web_port
+            self.web_port = str(web_port)  # TODO: switch argument to type str
```
allisonkarlitskaya, line 92: ...why not switch the attribute type to int?

machine/machine_core/machine.py
```diff
@@ -152,7 +153,7 @@ def journal_messages(self, matches, log_level, cursor=None):
         self.execute("journalctl --sync 2>/dev/null || true; sleep 1; journalctl --sync 2>/dev/null || true")
 
         # Prepend "SYSLOG_IDENTIFIER=" as a default field, for backwards compatibility
-        matches = (m if re.match("[a-zA-Z0-9_]+=", m) else "SYSLOG_IDENTIFIER=" + m for m in matches)
+        all_matches = (m if re.match("[a-zA-Z0-9_]+=", m) else "SYSLOG_IDENTIFIER=" + m for m in matches)
```
allisonkarlitskaya, line 156: Nag about existing code: I don't like capturing iterators into variables like this — we've been bit by things like this before.  Either turn this into a list comprehension or pull the `" + " .join()` up here.

Doing it as a list comprehension would actually allow you to avoid the variable rename, since it's the same type...

machine/machine_core/machine.py
```diff
@@ -164,29 +165,33 @@ def journal_messages(self, matches, log_level, cursor=None):
         else:
             cursor_arg = ""
 
-        cmd = "journalctl 2>&1 %s -o cat -p %d %s || true" % (cursor_arg, log_level, " + ".join(matches))
-        messages = self.execute(cmd).splitlines()
+        cmd = "journalctl 2>&1 %s -o cat -p %d %s || true" % (cursor_arg, log_level, " + ".join(all_matches))
+        output = self.execute(cmd)
+        assert output, "journalctl did not return any output"
```
allisonkarlitskaya, line 170: Isn't the correct behaviour here rather to treat this as empty?  So like `execute(cmd) or ''`?  I'm not a big fan of `.execute()` returning `None` instead of the empty string.  Maybe we should actually fix that.

machine/machine_core/machine.py
```diff
@@ -192,6 +193,7 @@ def audit_messages(self, type_pref: str, cursor: Optional[str] = None) -> List[s
         return messages
 
     def allowed_messages(self) -> List[str]:
+        """Allowed journal messages"""
```
allisonkarlitskaya, line 196: Where's the verb? :)

allisonkarlitskaya (changes requested):
This looks good in general.  See comments.

I feel like using the addition of types as an opportunity to do a bit more code cleanups is not necessarily a bad thing...

I feel like we might want to have some more variants of `execute()` here, based on the higher-level interface of the command in question.  I can imagine:

 - should return no data and if it returns data, it's an error
 - should return a non-empty string on a single line, followed by a newline, which is returned to the user stripped.  No output or more than one line → it's an error
 - return zero or more lines as split lines, with `\n` stripped

machine/machine_core/ssh_connection.py
jelly: I think I didn't like the string or argument list and wanted to change that, but it might be a bit too late for that :)

machine/machine_core/timeout.py
jelly: This is a circular dependency yes, I don't feel inclined to fix that right now. Could be done later sure, but honestly just having some basic typing in testlib for machines is the goal for me. Like:

image

Which hmm seems wrong!

jelly: Gotta love Python :snake:

machine/machine_core/machine.py
jelly: I glanced over this comment, I'm all for changing m.execute to not return `None` this would be a quite big API breakage. From a quick scan in Cockpit I don't see anything relying on `m.execute` being able to return `None`.

jelly:
```
  File "/home/jelle/projects/cockpit-navigator/test/common/testlib.py", line 1591, in tearDown
    self.check_journal_messages()
  File "/home/jelle/projects/cockpit-navigator/test/common/testlib.py", line 1802, in check_journal_messages
    messages = machine.journal_messages(matches, 6, cursor=cursor)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jelle/projects/cockpit-bots/machine/machine_core/machine.py", line 171, in journal_messages
    assert output, "journalctl did not return any output"
AssertionError: journalctl did not return any output
```

Locally I get this.. which is fun!

Issue #5275, opened 2023-09-19
selinux policy violation in sssd


jelly:
```
audit: type=1405 audit(1695078863.815:488): bool=sssd_use_usb val=1 old_val=0 auid=4294967295 ses=4294967295 exactly the same
```

martinpitt:
[example 1](https://cockpit-logs.us-east-1.linodeobjects.com/pull-19350-20230919-120520-0b58fb67-fedora-39-networking/log.html#38), [example 2](https://cockpit-logs.us-east-1.linodeobjects.com/pull-19349-20230919-120150-b7598bb0-fedora-39-networking/log.html#38)

jelly:
```
* Thu Aug 31 2023 Zdenek Pytela <zpytela@redhat.com> - 38.27-1
- Allow fedora-third-party get generic filesystem attributes
- Allow sssd use usb devices conditionally
```
Interesting, this was the previous selinux-policy release note

jelly:
This is not a rejection but an event to indicate that "changes to boolean".

PR #5276, opened 2023-09-19
naughty: add sssd selinux policy violation

jelly:


martinpitt:
Thanks! Please report to bugzilla and add a ref to the commit message and https://github.com/cockpit-project/bots/issues/5275

PR #5286, opened 2023-09-22
README: Document trigger permissions

martinpitt:
There is no way in GitHub to grant someone permissions to POST /statuses without giving full "Write" permissions to the repository. Document that, so that we know what to do for future contributors.

---

@subhoghoshX can now run `bots/tests-trigger` on cockpit, I added him to https://github.com/cockpit-project/cockpit/settings/access . There are still the branch protection rules, so it's not *that* bad.


martinpitt:
Note, cockpituous test now failed not because of the README change, but the change to .github/workflows/. I think that's safe to ignore.

jelly (approved):

PR #5304, opened 2023-09-25
lib: Enable cockpit-navigator testing

KKoukiou:
Needs PR to enable CI on the cockpit-navigator repo.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -123,6 +123,16 @@ REPO_BRANCH_CONTEXT = {
             'fedora-rawhide',
         ],
     },
+    'cockpit-project/cockpit-navigator': {
+        '_manual': [
+            TEST_OS_DEFAULT,
+            'arch',
+            'rhel-9-4',
+            'fedora-39',
+            'debian-testing',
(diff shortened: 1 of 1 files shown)


martinpitt:
What is it blocked on? Do you really want to enable all these OSes right away, or perhaps start with `_manual`?

KKoukiou:
> What is it blocked on? Do you really want to enable all these OSes right away, or perhaps start with `_manual`?

I though I need to add the .cockpit-ci/ directory to the repo, but looks like not. Let me try first once all the OSes and see how the tests do.

martinpitt:
@KKoukiou the entry point is https://github.com/cockpit-project/bots/blob/main/test/run . `.cockpit-ci/run` is an alternate entry point indeed, but test/run is more common, especially if you have a test/ directory already anyway

martinpitt (approved):
Thanks! I'd add `TEST_OS_DEFAULT` (aka fedora-38) and at least ubuntu-stable as well, though. You can do that as a follow-up, or here right away.

martinpitt (approved):

PR #5320, opened 2023-09-28
testmap: auto-trigger CI against fedora test VMs for cockpit-navigator

KKoukiou:


KKoukiou:
Fedora is fine. https://github.com/cockpit-project/cockpit-navigator/pull/122

lib/testmap.py
```diff
@@ -127,13 +127,15 @@ def contexts(image, *scenarios: Iterable[str]):
         ],
     },
     'cockpit-project/cockpit-navigator': {
+        'main': [
+            'fedora-38',
+            'fedora-39',
+            'fedora-rawhide',
```
martinpitt, line 133: Missing `]`, which explains the unit test failure. You didn't actually run fedora-38 on our CI, but as it runs on packit, it should be fine.

martinpitt (approved):
Epharisto!

PR #5343, opened 2023-10-04
images: test anaconda payload updates

KKoukiou:


allisonkarlitskaya:
If someone can furnish me with a copy of `fedora-rawhide-anaconda-payload-b9cab41ceb8b52f530d090fa62aa34e8e00f093c078f388244eb9bd9cec4d371.tar.gz` then I'd be able to upload it to the server again....

martinpitt:
So that still looks like a boot failure with the previous image/payload?

KKoukiou:
> So that still looks like a boot failure with the previous image/payload?

Indeed - so the payload update was not the reason  for our broken CI. Debugging this now.

martinpitt:
@KKoukiou what's the current status of that? I had the impression that the thing which broke the test was the recent firefox update, which you pinned back. Should this work now?

KKoukiou:
> @KKoukiou what's the current status of that? I had the impression that the thing which broke the test was the recent firefox update, which you pinned back. Should this work now?


Yes -  but currently the anaconda test suite is extremely flaky (since end last week). I am waiting for some reviews on test improvements and I will re-run the tests here later today once the anaconda test fix PRs are in.

martinpitt (approved):
Cheers!

PR #5357, opened 2023-10-09
Stop testing webui on Fedora-39 branch since it's not delivered there

KKoukiou:
Currently only rawhide is shipping the new Anaconda UI.


martinpitt (approved):
Heh, ok, thanks!

martinpitt (changes requested):
Wait, I take it you want to remove the whole `fedora-39-boot` image? can you  please `git rm` it from images/ then?

martinpitt (approved):
:broom: epharisto!

PR #5368, opened 2023-10-10
tests-invoke: restrict issue creation condition

### LINKED ISSUE #5365: Nightly tests did not succeed on fedora-39/other
Tests failed on 677467cf02db6c5472404f691d9cf3d9aa55f9a1

jelly:
We've had cases where PR collisions triggered an issue to be created.

Closes #5365

---

Alternatively we raise an exception (as this is all in a try/catch block) when a collission is detected.

### DIFF
--- a/tests-invoke
+++ b/tests-invoke
@@ -81,12 +81,14 @@ def main():
         os.environ["TEST_REVISION"] = opts.revision
         os.environ["TEST_PULL"] = str(opts.pull_number)
 
+        collision_detected = False
         p = subprocess.Popen(cmd)
         while p.poll() is None:
             # Cancel test if PR gets updated while running
             ret = detect_collisions(opts)
             if ret:
                 sys.stderr.write(f'Collision detected, cancelling this test run: {ret}\n')
+                collision_detected = True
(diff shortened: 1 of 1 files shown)


tests-invoke
```diff
@@ -81,20 +81,22 @@ def main():
...
+                collision_detected = True
                 ret = None
                 p.terminate()
                 p.wait(timeout=60)
             else:
                 time.sleep(int(os.getenv("TEST_INVOKE_SLEEP", "60")))
         return_code = p.returncode
         sys.stderr.write(f"Test run finished, return code: {return_code}\n")
-        if not api.has_open_prs(opts.revision) and return_code != 0:
+        if not api.has_open_prs(opts.revision) and not collision_detected and return_code != 0:
```
martinpitt, line 99: Ah, we should probably do something like `os.W_IFEXITED(p.returncode)`, to ignore cases of `W_IFSIGNALED` etc. But oh well, good enough.

martinpitt (approved):
Thanks! Feels a bit ad-hoc, but good enough, let's not waste time on this.

PR #5376, opened 2023-10-12
images: add linux-system-roles for kdump automation

jelly:
The upcoming Ansible integration into the Kdump page will use the existing linux-system-roles (rhel-system-roles on RHEL) for configuring kdump using Ansible.

 * [x] image-refresh fedora-38

Just refreshing TEST_OS_DEFAULT now as that's required for kdump testing.

### DIFF
--- a/images/fedora-38
+++ b/images/fedora-38
@@ -1,1 +1,1 @@
-fedora-38-e6d3defaeb4e04beb148725ce9edc4b570a3485e6d8e1784fd7d2df4082c9f6e.qcow2
+fedora-38-4061eaf7f1654018a42de300d5ab5db24d49d9d7ea4fb20f814fafc59f992eac.qcow2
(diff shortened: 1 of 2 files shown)


cockpituous:
image-refresh fedora-38 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-38-20231012-100707

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-38-20231012-100707.log

jelly:
> OK for testing of course, but there needs to be some fallback if l-s-r is not installed/available. What will the page do then?

This feature is basically only for RHEL/Fedora. As is the kdump page incidently.

martinpitt:
Sure, but what I mean is: cockpit-system does not, and should not, depend on l-s-r, and this package isn't part of a default install. So this needs to be handled via some ansible structure to install it first, and then pull in the role from it (which you can do, it's just not entirely trivial).

Issue #5395, opened 2023-10-16
teardown fails on locked LUKS block dev: Call failed: Error synchronizing after initial wipe: Timed out waiting for object


martinpitt:
This only happens with the daily udisks COPR.

Upstream report: https://github.com/storaged-project/udisks/issues/1204

cockpituous:
fedora-38
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testOverMounting (__main__.TestStorageMountingLUKS.testOverMounting)

DevTools listening on ws://127.0.0.1:9424/devtools/browser/34afa6f2-2a1e-4641-a3a0-fdfd535b00c8
[1016/044109.067072:WARNING:sandbox_linux.cc(393)] InitializeSandbox() called with multiple threads in process gpu-process.
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (ERROR)","timestamp":1697431272087.634,"url":"http://127.0.0.2:9191/cockpit/@localhost/*/po.js","networkRequestId":"45954.20"}
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (ERROR)","timestamp":1697431272088.998,"url":"http://127.0.0.2:9191/cockpit/@localhost/*/po.manifest.js","networkRequestId":"45954.19"}
CDP: {"source":"security","level":"error","text":"Refused to execute script from 'http://127.0.0.2:9191/cockpit/@localhost/*/po.manifest.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.","timestamp":1697431272155.3071,"url":"http://127.0.0.2:9191/storage"}
CDP: {"source":"security","level":"error","text":"Refused to execute script from 'http://127.0.0.2:9191/cockpit/@localhost/*/po.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.","timestamp":1697431272155.609,"url":"http://127.0.0.2:9191/storage"}
[...]

PR #5396, opened 2023-10-16
naughty: Add udisks locked LUKS teardown regression

martinpitt:
Upstream report: https://github.com/storaged-project/udisks/issues/1204 Known issue #5395

Fixes https://github.com/cockpit-project/cockpit/issues/19445

### DIFF
--- a/naughty/fedora-38/5395-udisks-locked-luks-teardown
+++ b/naughty/fedora-38/5395-udisks-locked-luks-teardown
@@ -1,0 +1,7 @@
+*Error synchronizing after initial wipe: Timed out waiting for object
+*
+Traceback (most recent call last):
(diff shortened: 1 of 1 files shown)


mvollmer (approved):

mvollmer:
As discussed on "IRC", this regression should theoretically also affect `check-storage-hidden`... hmm.

martinpitt:
@mvollmer I don't see how -- that test doesn't try to reformat the drive with a new partition table and teardown after creating a luks device.

mvollmer:
> @mvollmer I don't see how -- that test doesn't try to reformat the drive with a new partition table and teardown after creating a luks device.

Yeah, but it was meant to test this feature, but it does it only incompletely.

PR #5400, opened 2023-10-17
naughty: #5395 affects RHEL 9 as well

martinpitt:
While that error happens reliably with the udisks daily COPR version, it provably happens rarely on RHEL 9.3/9.4 already. It almost surely affects Fedora as well then, so copy the naughty there as well -- our tracker will tell us over time.

---

See [rhel-9-4](https://cockpit-logs.us-east-1.linodeobjects.com/pull-19487-20231016-080950-f7ea2a06-rhel-9-4-storage/log.html#32) and [rhel-9-3](https://cockpit-logs.us-east-1.linodeobjects.com/pull-19489-20231016-144148-9757b8f4-rhel-9-3-storage/log.html#32) failures, I updated https://github.com/storaged-project/udisks/issues/1204

### DIFF
--- a/naughty/fedora-39/5395-udisks-locked-luks-teardown
+++ b/naughty/fedora-39/5395-udisks-locked-luks-teardown
@@ -1,0 +1,7 @@
+*Error synchronizing after initial wipe: Timed out waiting for object
+*
+Traceback (most recent call last):
+  File "test/verify/check-storage-mounting", line *, in testOverMounting
+    self.content_row_wait_in_col(1, 0, "Free space")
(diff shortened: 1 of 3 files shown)


mvollmer:
Hmm, ok. I think there is a real chance that the failures in rhel-9 and fedora-40 are honest time outs caused by slow test VMs. But we will know when https://github.com/storaged-project/udisks/issues/1204 has been debugged, and we might just as well ignore those flakes (with a record).

mvollmer (approved):

PR #5433, opened 2023-10-23
images: Add targetcli to arch image

mvollmer:
So that we can run the new LVM2 RAID tests.

 * [ ] FAIL: image-refresh arch


cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/arch-20231023-080606.log

mvollmer:
Right: `error: target not found: targetcli`. We can disable that test on arch, I guess.

PR #5439, opened 2023-10-24
tests-invoke: show a bodhi link for the updates-testing scenario

jelly:
When the updates-testing scenario include a link to the bodhi updates page for convenience.

* [ ] Unbreak cockpituous CI `/bin/sh: line 2: update-ca-trust: command not found`

### DIFF
--- a/tests-invoke
+++ b/tests-invoke
@@ -107,6 +107,14 @@ def main():
                 body += f", [logs]({logs_url}/log.html)"
             if test_scenario != "":
                 title += f"/{test_scenario}"
+
+            if 'updates-testing' in test_scenario:
+                assert test_os is not None
+                if test_os.startswith("fedora-"):
+                    rel = test_os.split("-")[-1]
+                    base_url = "https://bodhi.fedoraproject.org"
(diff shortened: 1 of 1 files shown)


tests-invoke
```diff
@@ -107,6 +107,12 @@ def main():
                 body += f", [logs]({logs_url}/log.html)"
             if test_scenario != "":
                 title += f"/{test_scenario}"
+
+            assert test_os is not None
+            rel = test_os.split('-')[-1]
```
martinpitt, line 112: This should be a bit stronger and check if it starts with `fedora-`. Also, let's check for `updates-testing` first, so that none of this runs for other scenarios.

jelly: Correct, adjusted it all.

jelly:
`update-ca-trust` is no longer a part of the minio container which makes our CI not run. Nor does it contain microdnf or dnf so we can't really update any packages.

Can we just append the certificate and try that?

jelly:
Or maybe?

```
${HOME}/.minio/CAs
  myCA.crt
```

martinpitt (approved):
Let's sort out the cockpituous test separately, you already have a PR for it. Thanks!

PR #5445, opened 2023-10-25
testmap: Drop rhel-9-3 from cockpit{,-machines,-podman}

martinpitt:
9.3 is a done deal, 9.4 is the current development release.


allisonkarlitskaya (approved):
OK

PR #5480, opened 2023-11-01
Add new anaconda-webui project in the setup-deploy-keys script

KKoukiou:


martinpitt:
As discussed in the chat: We need to split this script into three, one for each organization. The problem with merging it is that no person has admin permissions in all three orgs, so nobody is able to run the script like this. Also, the person who runs it instantly "owns" all commits that are done through the deploy keys.

So I propose to split this up into `setup-deploy-keys-{cockpit-project,rhinstaller,osbuild}`, i.e. per org.

martinpitt (approved):
Cheers! As per chat, this works now.

PR #5490, opened 2023-11-02
lib: add new anaconda-webui repository in the testmap

KKoukiou:
This will take the subdirectory of the webui out of the rhinstaller/anaconda project.
When the move finalizes we need to replace the image updates to run also against the new repository.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -222,6 +222,11 @@ REPO_BRANCH_CONTEXT = {
             'fedora-rawhide-boot',
         ],
     },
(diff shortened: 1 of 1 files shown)


KKoukiou:
@martinpitt I can't really test it for some reason, I triggered it - see checks here in the PR - but it's not running.

martinpitt:
Yeah, you can't trigger it before it's in the testmap. That's why we normally start with `_manual`, but it's ok as I wrote above.

PR #5496, opened 2023-11-03
weather: Make top 10 failures context specific

martinpitt:
Many (or even most) flakes and failures are very specific to their
OS/context. These previously got drowned by averaging over all contexts:
e.g. a test that failed on a single OS 70% of the time ended up with a
global failure rate of < 7%, and thus would formally be within our SLO
even though it would still break most test runs, and never survive a "3x
affected" run.

Change the query and table to track the top ten failures separately for
each context, which is a lot more useful to identify current problems.

In the other direction, results were inflated by test runs which break
due to regressions in Fedora updates-testing -- these tend to get very
high failure rates quickly. We may decide to just hide them, but at
least it is now very clear where they actually fail.

----

[current main weather report](https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit-machines&days=7) looks like all would be hunky-dory for machines:

image

But when you e.g. drill down to [testAddDiskSCSI](https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit-machines&days=7&test=%2Ftest%2Fcheck-machines-disks+TestMachinesDisks.testAddDiskSCSI), it's clear that on RHEL 8 it is on fire:

image

(Note: this very bug was fixed yesterday, so it'll go down from now on)

I deployed this PR to my server, and in [that overview](https://piware.de/tmp/tests.html?repo=cockpit-project%2Fcockpit-machines&days=7) it is now much more useful and realistic:

image

I realize that this duplicates the test detail view somewhat, but due to the "drowning in average", many hot problems don't even appear on this table.

### DIFF
--- a/tests.html
+++ b/tests.html
@@ -275,8 +275,8 @@
                     createTable("t2", "Unexpected messages", ["Message", "Count", "Logs"], data);
 
                 // Get all failures sorted by descending fail rate
-                const top10 = [];
-                const rows = db.exec(`
+                const top_failures = [];
+                const result = db.exec(`
                         SELECT allruns.testname, project, failruns.context, (failruns.failed * 100.0) / COUNT(run) AS percent
                         FROM Tests AS allruns
                         JOIN TestRuns ON allruns.run = TestRuns.id
@@ -289,11 +289,21 @@
                         ) AS failruns ON allruns.testname = failruns.testname AND TestRuns.context = failruns.context
                         WHERE TestRuns.time > ${since} AND failruns.failed > 1 AND project LIKE '${repo}'
                         GROUP BY allruns.testname, failruns.context
-                        ORDER BY percent DESC
-                        LIMIT 10`);
+                        ORDER BY percent DESC`);
 
-                if (rows.length)
-                    rows[0].values.forEach(row => {
+                if (result.length) {
+                    const rows = result[0].values;
+                    // If we have more than 10 results, truncate below 10% failure rate
+                    if (rows.length > 10) {
+                        for (let i = 0; i < rows.length; i++) {
+                            if (rows[i][3] < 10) {
+                                rows.splice(i);
+                                break;
+                            }
+                        }
+                    }
+
+                    rows.forEach(row => {
                         const urls = db.exec(`
                             SELECT DISTINCT TestRuns.url
                             FROM TestRuns JOIN Tests ON TestRuns.id = Tests.run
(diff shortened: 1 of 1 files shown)


martinpitt:
The numbers seem rather high to me, also for [cockpit](https://piware.de/tmp/tests.html?repo=cockpit-project%2Fcockpit&days=7), so I still have some disbelief that they are really that bad. OTOH they actually match my current gut feeling -- I've been staring at nothing but falkes and failures for over two weeks now. E.g. kdump NFS on RHEL 8 really is that bad, it also fails locally in about half of the runs. Same for TestSuperuser.testWrongPasswd (which I fixed this morning in https://github.com/cockpit-project/cockpit/pull/19562 as it was breaking pretty much every PR) and  TestReauthorize.testSudo (see #19328). So in that sense I think they are fairly real.. :cry:

martinpitt:
TODO:
 - [ ] `TestStorageLvm2.testLvm` is broken: just one log, implausibly high failure rate, and no details view
 - [ ] show all above 10% and if less than 10 items then add up to top 10?

martinpitt:
Looking at `TestStorageLvm2.testLvm`:
 - This happens because the "fedora-38" context is kind of wrong, and in the last week there really only was one run with it. Normally this test runs in the `/expensive` scenario. I sent PR #5497 to fix this. But with that in mind, the number is correct.
 - Doing the grouping with ignoring scenarios would have to happen on the JS side, SQL can't `GROUP BY` patterns AFAIK. But that's probably okay, after the above fix. We can ignore this, it'll fall off the radar next week anyway.
 - Later I will check the broken details view.

martinpitt:
@marusak: Implemented as discussed. The [cockpit overview](https://piware.de/tmp/tests.html?repo=cockpit-project%2Fcockpit&days=7) looks correct now, that TestStorageLvm2.testLvm is still visible [with 14 days](https://piware.de/tmp/tests.html?repo=cockpit-project%2Fcockpit&days=14), but it works (detail view) and is correct with that "wrong context" caveat, see above.

A [short list](https://piware.de/tmp/tests.html?days=21&repo=rhinstaller%2Fanaconda) shows tests which fail < 10%, a [long list](https://piware.de/tmp/tests.html?days=21&repo=cockpit-project%2Fcockpit) hides tests < 10%.

marusak (approved):
Nice, thanks!

PR #5509, opened 2023-11-06
po-refresh: make coverage threadhold configurable

KKoukiou:
Anaconda project does not have a threshold.

Anaconda is special in the sense, that whatever language the user chooses from the UI is used for translating it and also for setting the language in the target system. Because the target system is allowed to have any language, when presenting the list of languages, we don't filter only the high translated ones.

### DIFF
--- a/po-refresh
+++ b/po-refresh
@@ -98,7 +98,8 @@ def run(context, verbose=False, **kwargs):
         with open("po/LINGUAS", "r", encoding='utf-8') as lngs:
             current_linguas = lngs.read().strip().split()
 
-    # Remove all files that have less than 50% coverage
+    # Remove all files that have less than X% coverage
+    # By default 50% is used but can be overridden by COVERAGE_THRESHOLD env variable
     for po in glob.glob("po/*.po"):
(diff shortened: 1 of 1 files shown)


po-refresh
```diff
@@ -98,15 +98,17 @@ def run(context, verbose=False, **kwargs):
...
+    # By default 50% is used but can be overridden by COVERAGE_THRESHOLD env variable
     for po in glob.glob("po/*.po"):
         all_types = output_with_stderr("msgfmt", "--statistics", po).split(", ")
         translated = int(all_types[0].split(" ")[0])
         untranslated = 0
         for u in all_types[1:]:
             untranslated += int(u.split(" ")[0])
         coverage = translated / (translated + untranslated)
-        if coverage < 0.5:
+        coverage_limit = os.getenv("COVERAGE_THRESHOLD", 0.5)
```
martinpitt, line 110: I suppose mypy would have something to say about mixing `str` and `float` here. It'd be cleaner to do this with `"0.5"`. But as we don't yet run mypy on bots, and it's trivial, it's fine.

martinpitt (approved):

PR #5516, opened 2023-11-07
naughty: Add pattern for Fedora podman-next image push regression

martinpitt:
Upstream issue: https://github.com/containers/podman/issues/20611 Known issue #5515


martinpitt:
This will stop the avalanche of failed podman PR tests, and we can concentrate on fixing the bug on the podman side, or lifting the "same image" assumption on our side. Let's discuss that with the podman devs first, though.

jelly (approved):

PR #5522, opened 2023-11-08
naughty: add new machines snapshot revert variant

jelly:
Example log: https://artifacts.dev.testing-farm.io/0a63986c-a4fa-4872-9e67-d1f8abdd496d/work-storagep7trdk0u/plans/all/storage/execute/data/guest/default-0/test/browser/storage-1/output.txt


martinpitt (approved):
Hmm, I would just have replaced the "0" with a "*", but ok..

jelly:
> Hmm, I would just have replaced the "0" with a "*", but ok..

The whole is_visible/is_present line is different so that wasn't as easy.

PR #5524, opened 2023-11-08
testmap: reflect on the testmap the anaconda repository split

KKoukiou:
Also drop anaconda conditional for previous fedora versions in the build-deps.sh script. The anaconda-webui will be maintaining one branch for all active fedora versions.


martinpitt (approved):
Nice! Whenever you are ready :rocket:

martinpitt:
I land this now as it started to break image refreshes, see PR #5560

PR #5542, opened 2023-11-13
image-prune: Fix prune URL

martinpitt:
Don't assume that the S3 objects live in the root path, but keep the path given by the store S3 URL. The path is `/images/` for our minio store on rhos-01-1, and trying to delete an image in / will cause a HTTP error 400.

Fixes https://github.com/cockpit-project/cockpituous/issues/563

---

 - [x] I have a test for this: https://github.com/cockpit-project/cockpituous/pull/564
 - [x] Test this with our actual Linode S3 store

### DIFF
--- a/image-prune
+++ b/image-prune
@@ -177,7 +177,7 @@ class S3ImageStore(ImageCache):
 
     def delete_file(self, filename):
         try:
-            with s3.urlopen(self.url._replace(path='/' + filename), method='DELETE'):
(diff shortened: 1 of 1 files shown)


martinpitt:
I went into rhos-01-1, and ran
```
cd bots
git checkout origin/fix-prune
./image-prune --force -c --s3 https://cockpit-images.eu-central-1.linodeobjects.com/
```
This works, so the fix doesn't break our public stores.

jelly (approved):

jelly:
Thanks for fixing this!

PR #5580, opened 2023-11-22
images: Go back to Fmstrat/samba-domain container

martinpitt:
Commit 392d6b21a15 moved to quay.io/samba.org/samba-ad-server, but this has a serious and difficult bug [1] with connecting to the Global Directory.

Go back to the previous https://github.com/Fmstrat/samba-domain container. The official dockerhub image actually works very well now, but we still have to build ourselves due to the docker.io pull rate limits.

Also don't re-add the external volumes -- we are not interested in permanently keeping any Samba data.

[1] https://github.com/samba-in-kubernetes/samba-container/issues/160

----

I built this locally, and it looks promising. Let's give it a good workout!

 * [x] image-refresh services


cockpituous:
image-refresh services done: https://github.com/cockpit-project/bots/commits/image-refresh-services-20231122-054654

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/services-20231122-054646.log

images/scripts/services.setup
```diff
@@ -66,37 +66,37 @@ poweroff
...
-    }
-  }
-}
-EOF
-
-# See https://github.com/samba-in-kubernetes/samba-container#ad-dc
-podman pull quay.io/samba.org/samba-ad-server
+# See https://github.com/Fmstrat/samba-domain ; docker.io/nowsci/samba-domain works fine
+# but we have to build ourselves due to docker pull rate limits
+git clone --depth=1 https://github.com/Fmstrat/samba-domain /var/tmp/samba-domain
```
jelly, line 71: Should we pin this at least?

martinpitt: We didn't pin it in the past, and I'd actually like to get updates. We can pin it if/when it breaks, but I wouldn't like to do that in advance.

jelly (approved):

images/scripts/services.setup
jelly: Alright, approved! We pin it indirectly with a refresh so we are quite flexible in that regard.

PR #5605, opened 2023-11-28
images: Update rhel-9-4 image

martinpitt:


martinpitt:
Nice, kdump is fixed :tada:

jelly (approved):

Issue #5444, opened 2023-10-25
systemd cannot clean up kdump.service's temp files: avc: denied { unlink remove_name rmdir } scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:kdumpctl_tmp_t:s0


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2246046

cockpituous:
fedora-39
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testBasic (__main__.TestKdump.testBasic)

DevTools listening on ws://127.0.0.1:10129/devtools/browser/2b5b9a68-dbe2-4922-93e3-0cf636319b5a
[1025/090234.651030:WARNING:sandbox_linux.cc(393)] InitializeSandbox() called with multiple threads in process gpu-process.
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (ERROR)","timestamp":1698224557606.3982,"url":"http://127.0.0.2:9191/cockpit/@localhost/*/po.manifest.js","networkRequestId":"130767.19"}
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (ERROR)","timestamp":1698224557610.1438,"url":"http://127.0.0.2:9191/cockpit/@localhost/*/po.js","networkRequestId":"130767.20"}
CDP: {"source":"security","level":"error","text":"Refused to execute script from 'http://127.0.0.2:9191/cockpit/@localhost/*/po.manifest.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.","timestamp":1698224557665.756,"url":"http://127.0.0.2:9191/kdump"}
CDP: {"source":"security","level":"error","text":"Refused to execute script from 'http://127.0.0.2:9191/cockpit/@localhost/*/po.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.","timestamp":1698224557666.149,"url":"http://127.0.0.2:9191/kdump"}
Created symlink /etc/systemd/system/multi-user.target.wants/kdump.service → /usr/lib/systemd/system/kdump.service.
[...]

PR #5616, opened 2023-11-30
naughty: Add another pattern for #5444

martinpitt:
With latest systemd this can now also happen straight from pid 1 instead of systemd-tmpfile.

---

This started happening recently, e.g. [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18614-20231130-141040-88c30181-fedora-39-expensive/log.html) or [here](https://cockpit-logs.us-east-1.linodeobjects.com/pull-18614-20231130-174158-a87c3dbc-fedora-39-devel/log.html).


allisonkarlitskaya (approved):
`fedora-40`, eh?

martinpitt:
> `fedora-40`, eh?

Yes, also known as "rawhide" :grin:

allisonkarlitskaya:
> > `fedora-40`, eh?
> 
> Yes, also known as "rawhide" 😁

fascinating.  I didn't know we did this mapping for naughties.  TIL.

martinpitt:
Well, our own bots call the image "fedora-rawhide", but in Testing Farm it's called "fedora-40" as that gets constructed from /etc/os-release. That's why we have that symlink in naughty/.

PR #5623, opened 2023-12-01
images: Drop openshift

martinpitt:
This image has been unbuildable for many years, and thus is useless for validating Cockpit's rhel-7.9 branch openshift integration by now. The latter got dropped now [1], so we can finally put this zombie to rest.

[1] https://github.com/cockpit-project/cockpit/pull/19674


jelly (approved):
First time I see it, looks good to drop!

PR #5642, opened 2023-12-06
testmap: auto-trigger fedora-eln-boot scenario for anaconda-webui tests

KKoukiou:


lib/testmap.py
```diff
@@ -248,10 +246,12 @@ def contexts(image, *scenarios: Iterable[str], repo: Optional[str] = None):
     # Anaconda builds in fedora-rawhide and runs tests in fedora-rawhide-boot
     "fedora-rawhide": [
         "fedora-rawhide-boot@rhinstaller/anaconda-webui"
+        "fedora-eln-boot@rhinstaller/anaconda-webui"
     ],
     # Anaconda payload updates can affect tests
     "fedora-rawhide-anaconda-payload": [
         "fedora-rawhide-boot@rhinstaller/anaconda-webui"
+        "fedora-eln-boot@rhinstaller/anaconda-webui"
```
martinpitt, line 254: Yeah, and that is a bug that still needs to be fixed in the anaconda-webui test.

martinpitt (approved):
Thanks!

PR #5740, opened 2023-12-31

cockpituous:

### DIFF
--- a/images/fedora-coreos
+++ b/images/fedora-coreos
@@ -1,1 +1,1 @@
-fedora-coreos-02b8ae709ab251617e270e24e389396a73808fd3f26d0c822f6815dcd1cf2ae1.qcow2
+fedora-coreos-1fbba117e42bd7a609a0b88e1e5710337eec228d9b2a03d5fbd549fe4dfc9bf1.qcow2


cockpituous:
image-refresh fedora-coreos done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-coreos-20231231-224841

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/fedora-coreos-20231231-224848.log

martinpitt:
Odd, it's the fixed libssh, but still failed the IPv6 address check..

martinpitt:
Ah! The problem is that the cockpit/ws *container* on this image has libssh 0.10.6-1.fc39. Which is a bit weird, as https://quay.io/repository/cockpit/tasks?tab=tags says that the last refresh was on Nov 20, while that libssh version only happened on Dec 22. But that's obviously bogus, as the real tag says "8 days ago". :man_shrugging: 

The [latest refresh run](https://github.com/cockpit-project/cockpit/actions/runs/7374259478/job/20064434110) already pulled in -2. So let's just try to refresh the image again.

PR #5752, opened 2024-01-04
testmap: Run firefox tests for validating services image

martinpitt:
As we have seen in https://github.com/cockpit-project/cockpit/pull/19794 there can be Firefox specific failures for a services image refresh. Thus run the Fedora tests with Firefox to cover these.


mvollmer (changes requested):
Hmm, what about adding the firefox scenario for the default OS, without removing the chromium one? I think we can afford one more scenario for image refreshes.

martinpitt:
I thought in the spirit of pairwise testing this would be enough. But sure, we don't refresh the services image that much. Updated.

mvollmer (approved):
Thanks!

PR #5771, opened 2024-01-09
testmap: Retire cockpit rhel-9.3 branch

martinpitt:
We have not received any further regression reports in RHEL 9.3, and can now stop maintaining this branch.


mvollmer (approved):
Nice!

PR #5781, opened 2024-01-11
testmap: enable debian-testing for navigator

jelly:
Test on debian-testing in pull requests now that we have packaging for it.


martinpitt (approved):
Yippie!

PR #5810, opened 2024-01-19
add customization script for fleet.ks in rhel4edge iso file

jscotka:
Image refresh for rhel4edge


cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/rhel4edge-20240119-220000.log

jscotka:
@allisonkarlitskaya Hi Lis. Is it possible to install and run `podman`, to be able to run container with tooling for KS customization? I'm not sure if it is allowed or there should change something? Or bots run it under user with admin rights so that I can do whatever I want there?

martinpitt:
@jscotka that doesn't work, sorry. Our bots run themselves in (unpriv) podman containers. This needs to be done some other way, perhaps fetching fleet.ks from its originating git?

jscotka:
Hi @martinpitt , I expected that, So that I have to use original tooling without container, Is It possible to ask to install some package inside this container? Or I should try to use copletely another solution? IT Is not just about extracting fleet.ks, but also give them back to ISO a tweak ISO file little bit, I have to look at all necessary changes.

jscotka:
Last commit with rebase depends on changes in https://github.com/cockpit-project/cockpituous/pull/571

images/scripts/rhel4edge.bootstrap
```diff
@@ -55,8 +55,33 @@ if [ "$STATUS" != "SUCCESS" ]; then
...
+
+# ==========================================
+# customize fleet.ks file inside ISO, to be able to do installation without user interaction
+DIR_NAME=$(mktemp -d -p $BASE)
+FLEET_FILE="$DIR_NAME/fleet.ks"
+
+mv "$ISO_FILE" "$DIR_NAME/fleet_source.iso"
+chmod 775 "$DIR_NAME"
+# extract fleet.ks to directory
+xorriso -osirrox on -indev "$DIR_NAME/fleet_source.iso" -extract /fleet.ks "$DIR_NAME/fleet.ks"
```
martinpitt, line 69: You can also do this with `isoinfo -x`, that tool is already in the tasks container.

images/scripts/rhel4edge.bootstrap
```diff
@@ -55,8 +55,33 @@ if [ "$STATUS" != "SUCCESS" ]; then
...
+# this is important to do for automatic installation
+sed -i "/network --boot/a rootpw --lock" "$DIR_NAME/fleet.ks"
+
+curl -L "https://raw.githubusercontent.com/loadtheaccumulator/fleet-iso-util/main/fleetkick.sh" -o "$DIR_NAME/fleetkick.sh"
+chmod a+x "$DIR_NAME/fleetkick.sh"
+# change workdir in script (it is used inside podman, so it expects specific path)
+sed -i "s@BASEDIR=.*@BASEDIR='$DIR_NAME'@" "$DIR_NAME/fleetkick.sh"
+mkdir "$DIR_NAME/workdir"
+
+$DIR_NAME/fleetkick.sh -w "$DIR_NAME/workdir"
```
martinpitt, line 79: Just about everything about this is wrong: Conceptually we are not testing the built rhel4edge image, but a heavily hacked rebuild of it. Operationally, this is untrusted arbitrary code execution inside the RH network with access to secrets.

Customizations should happen either via a kickstart file (which then runs *inside* the built VM), which can be served over local http instead of using the existing `fleet.ks` on the image; or as virt-install arguments in images/scripts/virt-install-rhel4edge; or perhaps edge-request.json has some customization options here?

martinpitt (changes requested):
Can you please give some high-level description of what you are trying to do here? It's not clear from the code changes, they seem unrelated to fixing the anaconda hang.

Thanks for looking into it!

jscotka:
replaced by version with custom kickstart, seems to work well, at least for now https://github.com/cockpit-project/bots/pull/5825

PR #5850, opened 2024-01-30
image-trigger: Refresh old RHEL images less often

martinpitt:
RHEL 7 and the non-current RHEL 8.y/9.y hardly change at all, only for targeted security fixes. It's enough to refresh them every month instead of every week.


jelly:
Having happily clicked "rebase & merge" on small RHEL image updates, this change seems good to me! Thanks @martinpitt

jelly (approved):

PR #5804, opened 2024-01-18
Image refresh for arch

cockpituous:
Image refresh for arch
 * [x] image-refresh arch

PR #5861, opened 2024-02-02
Image refresh for arch (for fsinfo)

cockpituous:
Let's get a working image with fsinfo inside of it.

 * [x] image-refresh arch


cockpituous:
image-refresh arch done: https://github.com/cockpit-project/bots/commits/image-refresh-arch-20240202-202818

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/arch-20240202-202828.log

allisonkarlitskaya:
```
  cockpit (308-1 -> 310.1-1)
```

How very convenient. :)

allisonkarlitskaya:
Great.  More `curl` upgrade fallout:

```
[root@archlinux ~]# curl --head http://172.27.0.15:9090
HTTP/1.1 301 Moved Permanently
Content-Type: text/html
Location: https://172.27.0.15:9090/

curl: (8) Weird server reply
```

and indeed:

```c
sendto(5, "HEAD / HTTP/1.1\r\nHost: 172.27.0.15:9090\r\nUser-Agent: curl/8.6.0\r\nAccept: */*\r\n\r\n", 80, MSG_NOSIGNAL, NULL, 0) = 80
recvfrom(5, "HTTP/1.1 301 Moved Permanently\r\nContent-Type: text/html\r\nLocation: https://172.27.0.15:9090/\r\n\r\n<html><head><title>Moved</title></head><body>Please use TLS</body></html>\r\n", 102400, 0, NULL, NULL) = 171
```

That looks distinctly like a body sent in reply to a HEAD request.  This one would have been coming from `cockpit-tls`.  That's going to be more "interesting" to fix: `cockpit-tls` doesn't read past the first byte on the incoming message, and it certainly doesn't go about trying to parse the HTTP request.  I think our best bet here would probably be to never send a body.

martinpitt:
The two cockpit PRs landed, retrying. I didn't yet look into the [two machines failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-5861-20240202-210947-bb54e614-arch-cockpit-project-cockpit-machines/log.html#18-2), but we've accrued quite a number of flakes there, so retrying for comparison as well.

martinpitt:
Ah, there is still the [testRaidRepair crash](https://cockpit-logs.us-east-1.linodeobjects.com/pull-5861-20240202-211002-bb54e614-arch-storage-cockpit-project-cockpit/log.html), which already fixed in the other pending refresh in PR #5804 . I suppose this is for @mvollmer.

martinpitt:
The same regression now landed in Fedora updates-testing, see https://github.com/cockpit-project/cockpit/issues/19937 . Smells like a kernel regression?

allisonkarlitskaya:
I wouldn't be surprised if it's related to the one in #5793...

jelly:
> I wouldn't be surprised if it's related to the one in #5793...

If you have a link to the patch I can easily build a kernel / verify

martinpitt:
The ubuntu one was a rather different area though (partition vs. RAID repair)

mvollmer:
I'll have a look as well.

The first failure is interesting, the other two with /dev/loop10 is just a failed cleanup.

mvollmer:
> testRaidRepair does indeed look like https://bugzilla.redhat.com/show_bug.cgi?id=2256432,

Yes, lvconvert hangs with the exact same kernel stack trace.

mvollmer:
Blocked on https://github.com/cockpit-project/cockpit/pull/19940

allisonkarlitskaya:
Looking into the machines failure.  All of the noise about

```
> error: Scrollbar test exception: TypeError: Cannot read properties of null (reading 'appendChild')
> log: osinfo-detect command failed:  (process:1624): GLib-GIO-WARNING **: 06:32:25.473: Can't find module 'gvfs' specified in GIO_USE_VFS
  File "<string>", line 28, in <module>
gi.repository.GLib.GError: osinfo-tree-error: URL protocol is not supported (1)
> log: osinfo-detect command failed:  (process:1769): GLib-GIO-WARNING **: 06:32:35.476: Can't find module 'gvfs' specified in GIO_USE_VFS
  File "<string>", line 22, in <module>
gi.repository.GLib.GError: osinfo-media-error: No volume descriptors (0)
> error: Failed when connecting: Connection closed (code: 1000)
> info: Connection lost:  {"isTrusted":"false","detail":"Object","type":"disconnect","target":"null","currentTarget":"null"}
> error: Tried changing state of a disconnected RFB object
> error: spawn 'vm creation' returned error: "{"problem":null,"exit_status":1,"exit_signal":null,"message":"ERROR    Domain not found: no domain with matching uuid 'e8d716ac-835b-4a5c-b0ba-8f720bab6b01' (subVmTestCreate8)\nDomain installation does not appear to have been successful.\nIf it was, you can restart your domain by running:\n  virsh --connect qemu:///system start subVmTestCreate8\notherwise, please restart your installation.\nTraceback (most recent call last):\n  File \"<string>\", line 361, in <module>\n  File \"<string>\", line 263, in create_vm\n  File \"/usr/lib/python3.11/subprocess.py\", line 466, in check_output\n    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/usr/lib/python3.11/subprocess.py\", line 571, in run\n    raise CalledProcessError(retcode, process.args,\nsubprocess.CalledProcessError: Command '['virt-install', '--connect', 'qemu:///system', '--quiet', '--os-variant', 'fedora28', '--memory', '128', '--name', 'subVmTestCreate8', '--check', 'path_in_use=off', '--wait', '-1', '--noautoconsole', '--disk', 'none', '--graphics', 'vnc,listen=127.0.0.1', '--graphics', 'spice,listen=::1', '--cdrom', 'https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/28/Server/x86_64/os/images/boot.iso']' returned non-zero exit status 1."}"
```

is just noise.  I see the same thing in a local run on main with the old image (which ends successfully).

martinpitt:
c-machines itself, in particular its machine_install.py, doesn't call curl. But of course that could be caused by qemu/block-curl.so calling it the wrong way. The test fakes fedoraproject.org with test/files/mock-range-server.py , but that doesn't use curl at all (it's the server side).

allisonkarlitskaya:
Downgrading curl (using 8.5.0 from https://archive.archlinux.org/packages/c/curl/) fixes the issue.

allisonkarlitskaya:
> Blocked on [cockpit-project/cockpit#19940](https://github.com/cockpit-project/cockpit/pull/19940)

Landed.

mvollmer:
> Blocked on [cockpit-project/cockpit#19940](https://github.com/cockpit-project/cockpit/pull/19940)

Merged.

naughty/arch/5881-curl-ssl-errno-0
```diff
@@ -0,0 +1,3 @@
+Traceback (most recent call last):
+  File "*/check-machines-create", line *, in testCreateUrlSource
+    runner.createTest(TestMachinesCreate.VmDialog(self, sourceType='url',
```
mvollmer, line 3: Can we make that more specific somehow? Otherwise this would match whenever anything is broken with `createTest`, not just when it is curls fault.

mvollmer: E.g., there is also "CURL: Error opening file: OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 0" in the browser log, a few lines above.

mvollmer:
`arch/other` probably just needs a rebase to go green.

naughty/arch/5881-curl-ssl-errno-0
allisonkarlitskaya: OK.

PR #5873, opened 2024-02-05
machines: Make ruff happy again

mvollmer:


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [tool.ruff]
-select = [
+lint.select = [
```
allisonkarlitskaya, line 2: can you rather move these to a new `[tool.ruff.lint]` section?

test/run
```diff
@@ -19,9 +19,9 @@ test_mypy() {
 }
 
 test_ruff
-test_mypy
+#test_mypy
```
allisonkarlitskaya, line 22: Is this for testing?  (and x2 more below)

machine/machine_core/ssh_connection.py
```diff
@@ -32,7 +32,7 @@
 # HACK: some projects directly import ssh_connection before adjusting sys.path; add bots root dir
 sys.path.insert(1, os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__)))))
 
-from lib.constants import TEST_DIR  # noqa: E402
+from lib.constants import TEST_DIR
```
allisonkarlitskaya, line 35: According to my understanding, this noqa should still beq required.  Or did ruff (very) recently get smarter about this?

allisonkarlitskaya (changes requested):

test/run
mvollmer: Ouch, yes.

pyproject.toml
mvollmer: Done.

mvollmer:
I can't access the "unit-tests" results, the URL is `https://localhost.localdomain:9000/logs/pull-5873-20240205-124243-664ff424-unit-tests/log.html`.

allisonkarlitskaya:
> I can't access the "unit-tests" results, the URL is `https://localhost.localdomain:9000/logs/pull-5873-20240205-124243-664ff424-unit-tests/log.html`.

The `cockpituous` workflow does a test of CI by running a tasks container locally and having it process this PR as if it were any other.  The `unit-tests` status that results from that is pointing back to the inside of that container.

allisonkarlitskaya (approved):
Fixed with an update to the tasks container — it needed the new version of ruff in order to agree with your changes.  Let's land this.

PR #5875, opened 2024-02-05
noop (check for pixels changes in all the projects after tasks container refresh)

allisonkarlitskaya:


allisonkarlitskaya:
Actually, let's just trigger on `main`.

PR #5885, opened 2024-02-08
naughty: RAID repair kernel hang now affects Fedora

martinpitt:
See https://github.com/cockpit-project/cockpit/issues/19967

---

Tested locally against the failure output in https://cockpit-logs.us-east-1.linodeobjects.com/pull-0-20240208-012335-d2a0a971-fedora-39-daily/log.html

I had to modify the pattern slightly as on Fedora the "[dm-mod]" prefix doesn't appear.

### DIFF
--- a/naughty/fedora-38/5874-lvconvert-hangs-in-kernel
+++ b/naughty/fedora-38/5874-lvconvert-hangs-in-kernel
@@ -1,0 +1,6 @@
+[<0>] md_stop+*
+*
+[<0>] dev_suspend+*
+*
+  File "check-storage-lvm2", line *, in testRaidRepair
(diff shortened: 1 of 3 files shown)


naughty/fedora-38/5874-lvconvert-hangs-in-kernel
```diff
@@ -0,0 +1,6 @@
+[<0>] md_stop+*
```
mvollmer, line 1: Fedora 38 as well?

martinpitt: That's more proactive. If it goes into F39, it's likely to go into F38 as well.

mvollmer (approved):

PR #5891, opened 2024-02-09
debian.setup: copy our pmlogger hack from fedora

allisonkarlitskaya:
We've started to see long shutdown hangs on pmlogger on Debian and Ubuntu and well, so copy the hack over.

 * [x] https://github.com/cockpit-project/cockpit/pull/20032


cockpituous:
image-refresh ubuntu-2204 done: https://github.com/cockpit-project/bots/commits/image-refresh-ubuntu-2204-20240209-084744

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-logs/ubuntu-2204-20240209-084747.log

allisonkarlitskaya:
Lots of

> `warning: Error creating partition on /dev/loop12: Failed to inform OS about changes on the '/dev/loop12' device (Partition(s) 1 on /dev/loop12 have been written, but we have been unable to inform the kernel of the change, probably because it/they are in use.  As a result, the old partition(s) will remain in use.  You should reboot now before making further changes.)`

Which is an improvement from when it listed off partitions 1 through 63, but it's still not fixed...

allisonkarlitskaya:
@mvollmer would love if you could render your opinion here...

mvollmer:
> Lots of
> 
> > `warning: Error creating partition on /dev/loop12: Failed to inform OS about changes on the '/dev/loop12' device (Partition(s) 1 on /dev/loop12 have been written, but we have been unable to inform the kernel of the change, probably because it/they are in use.  As a result, the old partition(s) will remain in use.  You should reboot now before making further changes.)`
> 
> Which is an improvement from when it listed off partitions 1 through 63, but it's still not fixed...

This happens because of this failed syscall:

```
ioctl(3, BLKPG, {op=BLKPG_ADD_PARTITION, flags=0, datalen=152, data={start=1048576, length=9437184, pno=1, devname="/dev/loop2p1", volname="primary"}}) = -1 EINVAL (Invalid argument)
```

and this syscall probably fails because loop devices have the GENHD_FL_NO_PART_SCAN flag set.

Reproducer:

```
# parted -s /dev/loop2 mktable gpt
# parted -s /dev/loop2 mkpart primary ext2 1M 10M
Error: Partition(s) 1 on /dev/loop2 have been written, but we have been unable to inform the kernel of the change, probably because it/they are in use.  As a result, the old partition(s) will remain in use.  You should reboot now before making further changes.
# cat /sys/block/loop2/capability 
240
```

Other kernels that I have looked at don't have that flag set for loopback devices.

mvollmer:
> This happens because of this failed syscall:

Setting up our loopback devices with `losetup -P` will fix all this.

jelly (approved):

PR #5916, opened 2024-02-14
run-queue: Add loop mode and wrapper script

martinpitt:
Prerequisites for moving our CI to single-shot tasks containers.

I tested `--loop` in `tasks/run-local.sh -i`, that's obvious.

For testing `--wrapper`: In a second container shell (`podman exec -it cockpituous-tasks bash`) I queued this test command:
```
echo 'echo $((1+1)); echo this is my test' | ./publish-queue --amqp $AMQP_SERVER -q public
```

and in the primary, the "normal" direct consumption:

```
bash-5.2$ ./run-queue --amqp $AMQP_SERVER
INFO:root:Consuming message with command: echo $((1+1)); echo this is my test
2
this is my test
```

and with my wrapper:
```sh
cat <<EOF > /tmp/wrapper.sh
#!/bin/sh -ex
echo "got test from queue \$TASK_QUEUE" >> /tmp/log
echo "test command: \$@" >> /tmp/log
"\$@"
EOF
chmod 755 /tmp/wrapper.sh
```

running 
```
./run-queue --amqp $AMQP_SERVER --wrapper /tmp/wrapper.sh
```

it also works:
```
+ echo 2
2
+ echo this is my test
this is my test
```

and /tmp/log is as expected:
```
got test from queue public
test command: echo $((1+1)); echo this is my test
```

### DIFF
--- a/run-queue
+++ b/run-queue
@@ -22,6 +22,7 @@ import json
 import logging
 import os
 import random
+import signal
 import smtplib
 import socket
 import subprocess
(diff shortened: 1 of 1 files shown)


run-queue
```diff
@@ -191,11 +195,15 @@ def main():
...
 
             if isinstance(cmd, list):
-                cmd_str = ' '.join(cmd)
+                cmd_str = shlex.join(cmd)
             else:
                 cmd_str = cmd
             logging.info("Consuming message with command: %s", cmd_str)
-            p = subprocess.Popen(cmd, shell=isinstance(cmd, str))
+            if opts.wrapper:
+                cmd = [opts.wrapper, queue_name, cmd_str]
```
allisonkarlitskaya, line 203: I think I'd prefer to get the separate command arguments rather than the `shlex.join`-ed blob here.

I'm not sure the queue name is super useful as a way to help choose which container to use, particularly if we expand to a situation where we want to have per-git-repository task container tags.  I see a couple of options:

 - have the scanner side determine which container to use and include that in the queue somehow.  This would have the benefit of being able to be run a single time when doing the scanning (since all tasks queued on a given PR would use the same tasks container).
 - some kind of intelligence in the wrapper script — but this seems hard.  We actually don't even really know what git repository we're dealing with since the command is an opaque blob.
 - make the command to run a bit more structured — something like a git repository plus the command to run inside of that repository or something.  Then the wrapper script would have an easier job of figuring out what to do... but it would also be less like a wrapper and more like a beefed up `tests-invoke`...

Let's talk at breakfast :)

martinpitt: Yeah, I don't know what I was thinking -- I had the same thought this night. I want to expose the additional information via env, like `$TASK_QUEUE`. I do want to retain the option of having a staging deployment to test infra or container changes without breaking production.

We can look into what it takes to wrap the queue contents into JSON, between tests-scan and run-queue. Shouldn't be hard.

martinpitt: Wait, it already is a JSON blob. We just need to add the repo and branch name there. I'll do that in a separate PR.

martinpitt: See #5917

martinpitt: Yeah, I don't really like passing JSON to the wrapper script that much. I suppose we can make sure `jq` is installed, but I think env variables are a bit simpler. `run-queue` can pass `$TASK_REPO` and `$TASK_BASE` as env vars. However, if you prefer json we can also do that (feed it via stdin or `argv[1]`)? No strong opinion.

run-queue
```diff
@@ -142,51 +143,73 @@ Subject: cockpituous: run-queue crash on %s
...
+                cmd, delivery_tag = consume_task_queue(
+                    channel,
+                    opts.amqp,
+                    q.declare_results['public'],
+                    q.declare_results['rhel'],
+                    q.declare_results['statistics'])
+            if not cmd:
+                if opts.loop:
+                    logging.info("All queues are empty, sleeping")
+                    time.sleep(opts.loop_sleep)
```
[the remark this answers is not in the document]
martinpitt: I was going to put that onto the TODO list. We've been doing sleep-based polling for many years, and that isn't in the critical path.

run-queue
martinpitt: Discussed IRL: Let's go with argv and env vars.

run-queue
```diff
@@ -142,49 +143,71 @@ Subject: cockpituous: run-queue crash on %s
...
 def main():
-    parser = argparse.ArgumentParser(description='Bot: read a single test command from the queue and execute it')
+    parser = argparse.ArgumentParser(description='Bot: read and execute task command(s) from the queue')
     parser.add_argument('--amqp', default=distributed_queue.DEFAULT_AMQP_SERVER,
                         help='The host:port of the AMQP server to consume from (default: %(default)s)')
+    parser.add_argument('--loop', action='store_true',
+                        help='Run forever until SIGTERM; by default, run a single task or exit '
+                             'immediately if queue is empty')
+    # this exists mostly to speed up integration tests; for production, 1 min is fine
+    parser.add_argument('--loop-sleep', type=int, default=60, metavar='SECONDS',
```
allisonkarlitskaya, line 156: Please drop this argument.

run-queue
```diff
@@ -191,11 +196,22 @@ def main():
...
+                cmd = ['sh', '-ec', cmd_str]
+            env = os.environ.copy()
             logging.info("Consuming message with command: %s", cmd_str)
-            p = subprocess.Popen(cmd, shell=isinstance(cmd, str))
+
+            if opts.wrapper:
+                cmd.insert(0, opts.wrapper)
+                env['TASK_QUEUE'] = queue_name
+                try:
+                    env['TASK_CONTAINER'] = body['container']
```
allisonkarlitskaya, line 210: `body['container']` may well be `None`, which is not a valid value for an environment variable.

How about more like:

```python
if container := body.get('container'):
    env['TASK_CONTAINER'] = container
```

or more defensively:

```python
container = body.get('container')
if isinstance(container, str):
    env['TASK_CONTAINER'] = container
```

allisonkarlitskaya (changes requested):

run-queue
martinpitt: Are you ok with a `LOOP_SLEEP` env var? One minute sucks for testing (both interactively and also in the unit test)

run-queue
martinpitt: I take it you mean it may be `null`?

martinpitt: Done.

martinpitt:
The `--loop` part got folded into the job-runner branch, and we don't want the wrapper part.

PR #5918, opened 2024-02-15
tests-scan: add container information to the queue

allisonkarlitskaya:
When scanning tasks on a particular PR, look at the `.cockpit-ci/container` file to determine which container image (if any) the project wishes to be tested against.
    
This will allow us to have gated roll-outs of new tasks containers.

### DIFF
--- a/tests-scan
+++ b/tests-scan
@@ -23,6 +23,7 @@ import logging
 import shlex
 import sys
 import time
+from typing import Iterable, NamedTuple
 
 from lib import ALLOWLIST, testmap
 from task import distributed_queue, github, labels_of_pull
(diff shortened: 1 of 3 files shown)


tests-scan
```diff
@@ -107,18 +108,31 @@ def main():
...
+    github_context: str
+
+
 # Prepare a human readable output
-def tests_human(priority, name, number, revision, ref, context, base, repo, bots_ref, github_context, options):
-    if not priority and not options.force:
+def tests_human(job: Job, options: argparse.Namespace) -> str:
+    if not job.priority and not options.force:
         return ""
     return "{name:11} {context:25} {revision:10} {priority:2}{repo}{bots_ref}{branches}".format(
```
martinpitt, line 129: I'm surprised that you didn't convert to f-string :grin:

martinpitt (approved):
This looks good to me, thanks! CI is green, so I'm willing

tests-scan
allisonkarlitskaya: I was tempted.... :)

PR #5921, opened 2024-02-16
Add type annotations for s3-streamer

allisonkarlitskaya:
... and fix the two-and-a-half bugs we found along the way.


martinpitt (approved):
Nice, thanks!

PR #5938, opened 2024-02-21
Fix unit test, move to actions/checkout@v4

martinpitt:
[Oops](https://github.com/cockpit-project/bots/actions/runs/7984922035/job/21802523366?pr=5937)! I fixed all our projects (including session-recording and composer), but forgot about bots, sorry.


mvollmer (approved):
Danke!

PR #5939, opened 2024-02-21
testmap: move composer testing to fedora-39

jelly:


martinpitt (approved):
composer is green, and the others were triggered accidentally.

PR #5941, opened 2024-02-21
testmap: test podman on fedora-40

jelly:


martinpitt (approved):
Looking good so far.. :crossed_fingers:

PR #5945, opened 2024-02-21
testmap: move cockpit-composer to rhel-8-10 and rhel-9-4

jelly:
@jkozol Needs to be moved with: https://github.com/osbuild/cockpit-composer/pull/1967


martinpitt (approved):
Cheers!

PR #5960, opened 2024-02-22
tests-invoke: Drop has_open_prs() call for issue filing

martinpitt:
That API is broken as-is. But we also don't need it here at all, as we already know via our CLI arguments whether or not we run for a PR.

Fixes #5958

PR #5961, opened 2024-02-22
Revert "task: Fix has_open_prs() return value"

martinpitt:
This function is broken beyond repair -- it needs to be called on the PR *source* repo (i.e. the fork), the GitHub API endpoint will always return `[]` for tests which run for a PR for a fork.

There is no easy way to fix this -- the necessary information gets lost in translation, and e.g. in tests-scan we can't rely on always having a `--pull-number`: When we retry a test with `./tests-trigger`, that will become a statuses event which has no idea any more whether it belongs to a PR or not.

So for the time being, stop the firehose of bogus "Nightly test failed" issues by reverting commit 7c4f06c2b359e2984458711d51677ea10e8fb446.

This breaks test_amqp_sha(), so disable it for now.

Fixes #5960

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -402,7 +402,7 @@ class GitHub(object):
         pulls = self.get(f"commits/{sha}/pulls")
         if pulls:
             return any((pull['state'] == 'open' for pull in pulls))
-        return False
+        return True
 
     def getHead(self, pr):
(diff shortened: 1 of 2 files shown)


jelly (approved):
Thanks for the quick fix!

PR #5966, opened 2024-02-22
tests-invoke: Drop has_open_prs() call for issue filing

martinpitt:
That API is broken beyond repair. But we also don't need it here at all, as we already know via our CLI arguments whether or not we run for a PR.

---

This was originally #5960, but I can't reopen that for some reason.

It wasn't bad after all -- my original worry was that we wouldn't be able to determine the PR number if we do `bots/tests-trigger NNN` in a PR to retry failures. In https://github.com/cockpit-project/starter-kit/pull/574 I just triggered a completely new "fedora-40" status, and the [log shows](https://cockpit-logs.us-east-1.linodeobjects.com/pull-574-20240222-155208-a92a4b77-fedora-40/log) that we indeed get tests-invoke with `--pull-number 574`, so all is well.


tests-invoke
```diff
@@ -96,7 +96,9 @@ def main():
                 time.sleep(int(os.getenv("TEST_INVOKE_SLEEP", "60")))
         return_code = p.returncode
         sys.stderr.write(f"Test run finished, return code: {return_code}\n")
-        if not api.has_open_prs(opts.revision) and not collision_detected and return_code != 0:
+        # --pull-number is absent for nightly runs; but we special-case "pull-0",
+        # so if someone passes "0" that also doesn't count
+        if not opts.pull_number and not collision_detected and return_code != 0:
```
jelly, line 101: Then we should check for `not opts.pull_number` and `opts.pull_number != 0`

jelly (changes requested):

tests-invoke
martinpitt: I'm confused. We want to run the "then" part for nightly runs without a PR, but never for PRs. So if `opts.pull_number` is truthy (a number > 0), `not opts.pull_number` is False, and the `if` is skipped. But if it's absent (None) or 0, then `not opts_pull_number` is truthy.

I.e. do you think it's wrong, or you just prefer it to be more explicit?

`if (opts.pull_number is None or opts.pull_number == 0) and ...`

?

jelly: Aaah, it's time to recharge. Right, I thought we checked for `is None` originally so the check is correct.

martinpitt:
I looked at this in more details. Dropping the API call from tests-invoke is correct and a good thing to do, so @jelly can you please re-review?

I'm still figuring out `tests-scan`: Re-triggering a failed test run in a PR with us doing `./tests-trigger 1234` does some expensive magic which only happens to work because of a combination of bugs and quirks. I'll capture this in a unit test and try to clean it up.

martinpitt:
I just found two disabled tests-invoke unit tests which cover this (in I think the wrong way). But let's land the tests first, i.e. PR #5977.

jelly (approved):
Thanks!

PR #6002, opened 2024-03-02
image-trigger: Disable Fedora ELN images

### LINKED ISSUE #5823: Image refresh for fedora-eln-boot
Image refresh for fedora-eln-boot
 * [ ] image-refresh fedora-eln-boot

### LINKED ISSUE #5911: Image refresh for fedora-eln
Image refresh for fedora-eln
 * [ ] image-refresh fedora-eln

martinpitt:
They have been broken since January, and the Anaconda team does not currently have the capacity to maintain/use them. They are already disabled in the testmap.

Closes #5823
Closes #5911

### DIFF
--- a/image-trigger
+++ b/image-trigger
@@ -46,8 +46,9 @@ REFRESH = {
     "fedora-39": {},
     "fedora-40": {},
     "fedora-coreos": {},
-    "fedora-eln": {},
-    "fedora-eln-boot": {},
+    # TODO: fix and re-enable once Anaconda team has time and energy to maintain these again
+    # "fedora-eln": {},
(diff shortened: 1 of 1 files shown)


KKoukiou (commented):
I would completely remove these from the list. https://github.com/cockpit-project/bots/pull/6028#issuecomment-1981241902

It's not about not having time at this point, but we actually want to test RHEL10, and this is already ahead of that.

KKoukiou (approved):

martinpitt:
OK, thanks. Right, ELN is "future RHEL 11" now. So I'll land this, and remove the old images tomorrow.

PR #6010, opened 2024-03-04
task: distributed_queue: add 'localhost' mode

allisonkarlitskaya:
If the --amqp parameter of `run-queue`, `inspect-queue`, `tests-scan`, etc. is given as exactly the string `"localhost"` then don't try to load certificates -- instead try to connect to localhost on port 6572 using plain text credentials (guest:guest).  That's the default configuration for the docker image.

This means that you can do this for simplified local testing:

```
$ podman run --net=host --rm docker.io/rabbitmq
$ ./run-queue --amqp localhost
```

### DIFF
--- a/task/distributed_queue.py
+++ b/task/distributed_queue.py
@@ -92,19 +92,25 @@ class DistributedQueue(object):
                 if os.path.isdir(user_secrets):
                     secrets_dir = user_secrets
 
-        try:
-            host, port = amqp_server.split(':')
-        except ValueError:
-            raise ValueError('Please format amqp_server as host:port')
-        context = ssl.create_default_context(cafile=os.path.join(secrets_dir, 'ca.pem'))
-        context.load_cert_chain(keyfile=os.path.join(secrets_dir, 'amqp-client.key'),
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
If you need this and prefer it over `run-local.sh -i`, then no objection.

allisonkarlitskaya:
I don't *need* this, but it makes my life easier for random local/heavily-interactive testing outside of the framework of `run-local.sh`.  I could still be convinced that I'm "doing it wrong"...

martinpitt:
Yeah, you *really* want `-i` for that, that's what it's for :grin:

allisonkarlitskaya:
> Yeah, you _really_ want `-i` for that, that's what it's for 😁

So for example, when I first do that, it complains about missing `openssl`.  I install it, then it complains about:

`mc: <ERROR> Unable to initialize new alias from the provided credentials. Get "https://localhost.localdomain:9000/probe-bsign-ise2c9rty4fwdhus6uzqf19xoizcz0/?location=": http: server gave HTTP response to HTTPS client.`

PR #6016, opened 2024-03-05
task: add a COCKPIT_GITHUB_TOKEN_FILE envvar

allisonkarlitskaya:
Similar to COCKPIT_S3_KEY_DIR this lets us point to secrets in non-standard locations (eg: /run/secrets/github-token for the tasks container).

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -130,7 +130,7 @@ class GitHub(object):
         self.token = None
         self.debug = False
         try:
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Cheers! I'll make use of that in https://github.com/cockpit-project/cockpituous/pull/585

PR #6019, opened 2024-03-05
lib: add a COCKPIT_IMAGE_STORES_FILE envvar

allisonkarlitskaya:
Similar to the others, we provide this file when testing from run-local.sh (and it's the only one for which we don't yet have an override).

### DIFF
--- a/lib/stores.py
+++ b/lib/stores.py
@@ -31,7 +31,7 @@ REDHAT_STORES = [
 
 # local stores
 try:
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Nice, thanks! The test fails because #6018, but that's unrelated. Happy to work that into cockpituous tomorrow, together with JOB_RUNNER_CONFIG

allisonkarlitskaya:
> `Error: unable to connect to Podman socket: server API version is too old. Client "4.0.0" server "3.4.4"`

:eyes:

PR #6020, opened 2024-03-05
issue-scan: add a Job blob to the queued command

allisonkarlitskaya:
We're going to need this eventually but probably we need to figure out a way to test it....

I feel like we might like to get `mock-github-pr` into the bots repo and figure out a way to hook it up to tests here.  It's already pretty similar to the code already in `task/test-task` and `task/test-tests-scan`... It would be nice to be able to unit-test this stuff without creating yet another copy of it in a new test file...


issue-scan
```diff
@@ -134,31 +134,43 @@ def output_task(command, issue, repo, verbose):
...
     }
-    queue = 'rhel' if contains_any(result, REDHAT_TASKS) else 'public'
+
+
+def queue_task(channel, body):
+    queue = 'rhel' if contains_any(channel['job']['context'], REDHAT_TASKS) else 'public'
     channel.basic_publish('', queue, json.dumps(body),
                           properties=pika.BasicProperties(priority=distributed_queue.MAX_PRIORITY))
-    logging.info("Published issue task with command: '%s'", result)
+    logging.info("Published issue task with command: '%s'", channel['command'])
```
martinpitt, line 173: `channel['command']` looks like "I did this at 23:00" :grin: 

How about just printing the entire `body`, and `s/ with command//` ? Useful for debugging.

martinpitt (changes requested):
If mock-github-pr helps for this, then by all means let's move it into bots. Dependency and conceptually wise it is a better fit anyway. cockpituous manually copies it into the bots.git checkout anyway.

martinpitt:
... but for this use case I'd propose to add some unit tests to tasks/test-tests-scan and rename that accordingly. It already covers `tests-invoke` as well. Or we move `GITHUB_DATA` and `Handler` into a shared module between test-{tests,issue}-scan?

issue-scan
martinpitt: #6026 spots this, and also brings the scaffolding for validating the `job` JSON. So let's merge that first, rebase this and update the tests?

issue-scan
```diff
@@ -134,42 +138,50 @@ def output_task(command, issue, repo, verbose):
...
     }
-    queue = 'rhel' if contains_any(result, REDHAT_TASKS) else 'public'
+
+
+def queue_task(channel, body):
+    queue = 'rhel' if contains_any(body['job']['context'], REDHAT_TASKS) else 'public'
     channel.basic_publish('', queue, json.dumps(body),
                           properties=pika.BasicProperties(priority=distributed_queue.MAX_PRIORITY))
-    logging.info("Published issue task with command: '%s'", result)
+    logging.info("Published issue task with command: '%s'", body['command'])
```
martinpitt, line 175:
```suggestion
 logging.info("Published issue task: '%s'", body)
```
so that we see the JSON too. Grabbing them from the log and stuffing them into `job-runner` feels useful

test/test_issue_scan.py
```diff
@@ -172,20 +172,40 @@ def test_scan_ghapi_amqp(self, mock_queue):
         self.assertEqual(channel.basic_publish.call_args_list[0][0][1], "public")
         request = json.loads(channel.basic_publish.call_args_list[0][0][2])
 
-        self.assertEqual(request, {
+        assert request == {
```
martinpitt, line 175: TBH I'd prefer setting `self.maxDiff = None` here, as it pretty much breaks running with unittest (the failure output is useless). But no strong opinion.

martinpitt (changes requested):
Thanks! One small issue and a bikeshed :grin:

issue-scan
allisonkarlitskaya: Already did that, but with `%r` and without quotes.

PR #6023, opened 2024-03-06
workflows: Drop obsolete git safe.directory config

martinpitt:
actions/checkout@v4 does this by default.


martinpitt:
... or it doesn't!? https://github.com/cockpit-project/bots/actions/runs/8166816971/job/22326199214?pr=6023

Odd, I've seen that work in other projects. Nevermind.

PR #6024, opened 2024-03-06
workflows: Replace deprecated :set-output

martinpitt:
This causes a "will soon be removed" warning, see
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

----

This is also included in https://github.com/cockpit-project/cockpituous/pull/587 where it covers the "positive" case of a changed repo with a container rebuild. Here it should *not* run cockpituous test.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,6 +42,7 @@ jobs:
         id: changes
         run: |
           git log --exit-code --stat HEAD --not origin/${{ github.event.pull_request.base.ref }} -- \
+              ':!.github/workflows' \
               ':!README.md' \
               ':!HACKING.md' \
               ':!images' \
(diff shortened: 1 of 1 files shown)


martinpitt:
Haha, I involuntarily tested the [positive case](https://github.com/cockpit-project/bots/actions/runs/8166916254/job/22326459267?pr=6024) here, as we currently consider workflow changes to affect the deployment. That is nonsense, I'll fix that in a separate commit.

martinpitt:
With the second commit,  [it doesn't run cockpituous](https://github.com/cockpit-project/bots/actions/runs/8166981821/job/22326627666?pr=6024), so we have coverage for both  cases.

jelly (approved):

PR #6025, opened 2024-03-06
Move all tests to test/, eliminate test-bots

martinpitt:
It's strange to have tests in task/, even more so for programs which
don't even live in task/. Move to a more standard layout where all tests
live in test/ and thus become discoverable.

The tests can now be run  with both `pytest` and `python3 -m unittest discover`.
Use `pytest` in our CI, as we already have it installed in the tasks container.


test/run
```diff
@@ -24,4 +24,4 @@ test_mypy
 # run static code checks like pyflakes and pep8
 flake8 $PYEXEFILES
 
-./test-bots
+python3 -m unittest discover -s test -v
```
allisonkarlitskaya, line 27: tasks container has pytest already.....

allisonkarlitskaya (approved):

PR #6027, opened 2024-03-06
images: Add "foonux" mock image

martinpitt:
This will be used by cockpituous integration tests. The "cirros" image requires kvm to build, and we also don't want to suffer through the download. It's also better to keep a separate name to ensure that we don't accidentally push a broken cirros image anywhere public.

### DIFF
--- a/images/scripts/foonux.bootstrap
+++ b/images/scripts/foonux.bootstrap
@@ -1,0 +1,4 @@
+#!/bin/sh
+# mock image which doesn't need any qemu/kvm; used by cockpituous integration tests
+echo fakeimage > $1
+date >> $1


images/scripts/foonux.bootstrap
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+# mock image which doesn't need any qemu/kvm; used by cockpituous integration tests
+echo fakeimage > $1
```
allisonkarlitskaya, line 3: I feel like it might be interesting to inject a timestamp here so we get a different hash each time.

martinpitt: Good idea!

martinpitt:
I confirmed that two image refreshes in a row now give a different SHA:

```diff
--- images/foonux
+++ images/foonux
@@ -1 +1 @@
-foonux-a6d4ae0f46e6136c1737eee712c2bb657b22b1e56c899ba0e20235f918f4e0e8.qcow2
\ No newline at end of file
+foonux-03ea4e1c8115aaf816fbdd470fc53d900f5d8a6b115466e91dd9768438c38785.qcow2
\ No newline at end of file
```

allisonkarlitskaya (approved):

PR #6029, opened 2024-03-06
pyproject.toml: add config for pytest and coverage

allisonkarlitskaya:
Add a config for pytest which enables coverage reporting (pytest --cov).

This also properly defines the project root, enabling us to do stuff like:

  pytest test/test_issue_scan.py

in order to run single files.


martinpitt (approved):
Thanks!

PR #6035, opened 2024-03-07
Convert .format() to f-strings

martinpitt:
ruff 0.3.1 (which we will get with the next tasks container refresh) starts complaining about some .format() usages.

---

See [this test run](https://github.com/cockpit-project/cockpituous/actions/runs/8186987774/job/22386529483?pr=586#step:7:361), which was running against a locally (in the workflow) refreshed tasks container.

Our integration test already exercises PRs, image refresh etc., but I still triggered two tests for good measure.

### DIFF
--- a/image-trigger
+++ b/image-trigger
@@ -125,7 +125,7 @@ def scan(api, force, verbose):
         if perform:
             text = f"Image refresh for {image}"
             issue = task.issue(text, text, "image-refresh", image)
-            sys.stderr.write("#{0}: image-refresh {1}\n".format(issue["number"], image))
+            sys.stderr.write(f'#{issue["number"]}: image-refresh {image}\n')
 
 
 if __name__ == '__main__':
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -218,7 +218,7 @@ class VirtNetwork:
         for remote, local in result["forward"].items():
             local = self._lock(int(local) + result["number"])
             result["forward"][remote] = f"127.0.0.2:{local}"
-            forwards.append("hostfwd=tcp:{}-:{}".format(result["forward"][remote], remote))
+            forwards.append(f"hostfwd=tcp:{result['forward'][remote]}-:{remote}")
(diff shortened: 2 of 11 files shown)


allisonkarlitskaya (commented):
This particular failure is extremely well-timed and I'd like to avoid wasting the opportunity it presents.

How do you feel about:

 - separate PR introducing the `.cockpit-ci/container` file to pin us back to the old version (making things green again)
 - a lock-step single commit to update to the new version while fixing the issues here?

Just to prove the concept, basically...


The thing is, I guess this wouldn't actually work properly currently, since the failing test is a workflow.  We'd need to up our game a bit there.

If you'd prefer to take the "omg main is red we need to fix this now" tack, I'm not going to insist.

martinpitt:
@allisonkarlitskaya you probably meant that commit for https://github.com/cockpit-project/cockpit/pull/20149 ?  that might also fail with a new tasks container, but it wouldn't help us here indeed. And in cockpit it's the "tox" workflow, which isn't pinnned. A bit "meh".. But I'll think about it at some later time (this day has been long enough)

martinpitt:
@allisonkarlitskaya TBH I'd like to wait with introducing .cockpit-ci/container until we have an `update-ci-container` bot. That shouldn't be super hard, a little bit of skopeo and our tasks API. It would be nice to use dependabot for that, but it only supports [k8s yaml files](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#docker). It may be worth checking if we can support .cockpit-ci/container.yml with a three-line mock yaml, see https://github.com/dependabot/dependabot-core/issues/5819#issuecomment-1984010295

allisonkarlitskaya (approved):
> @allisonkarlitskaya : c-machines is a perfect example to start that. It fails with ruff 0.3.1, so how about we test-drive it there, and use that as a first opportunity to automate container updates?
> 
> See [cockpit-project/cockpit-machines#1484](https://github.com/cockpit-project/cockpit-machines/pull/1484)

Deal.

PR #6039, opened 2024-03-07
lib: Add $COCKPIT_IMAGES_DATA_DIR env variable

martinpitt:
Similarly to commit 85b0db4a0bfd5c and commit 6b7864a1e6d76b, this lets us configure the image cache volume in one place (job-runner.toml), instead of having to hardcode it in the tasks Containerfile. This will replace
https://github.com/cockpit-project/cockpituous/blob/accb1fdc6a/tasks/Containerfile#L73

### DIFF
--- a/lib/directories.py
+++ b/lib/directories.py
@@ -61,9 +61,11 @@ def get_images_data_dir() -> str:
     global _images_data_dir
 
     if _images_data_dir is None:
-        _images_data_dir = get_git_config('--type=path', 'cockpit.bots.images-data-dir')
-
+        _images_data_dir = os.getenv('COCKPIT_IMAGES_DATA_DIR')
         if _images_data_dir is None:
-            _images_data_dir = xdg_cache_home('cockpit-images')
+            _images_data_dir = get_git_config('--type=path', 'cockpit.bots.images-data-dir')
+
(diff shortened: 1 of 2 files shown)


lib/directories.py
```diff
@@ -61,9 +61,11 @@ def get_images_data_dir() -> str:
     global _images_data_dir
 
     if _images_data_dir is None:
-        _images_data_dir = get_git_config('--type=path', 'cockpit.bots.images-data-dir')
-
+        _images_data_dir = os.getenv('COCKPIT_IMAGES_DATA_DIR')
         if _images_data_dir is None:
-            _images_data_dir = xdg_cache_home('cockpit-images')
+            _images_data_dir = get_git_config('--type=path', 'cockpit.bots.images-data-dir')
```
allisonkarlitskaya, line 66: Does anyone still use this?

martinpitt: I don't know -- we've documented it for many years, and I had even had it myself for a long time (when I didn't have enough space in my $HOME). For human users this may feel better than setting an env var.

allisonkarlitskaya: OK.

allisonkarlitskaya (approved):

PR #6041, opened 2024-03-08
job-runner: Run job container with --init

### LINKED ISSUE #6036: Let's test job-runner
- [ ] image-refresh fedora-39

martinpitt:
We run arbitrary commands in these job containers. These don't expect to be PID 1 and be responsible for reaping zombies. This leads to e.g. image-refresh timeouts as when qemu finishes, the process remains around as 'Z' indefinitively, and waiting for it to finish times out.

Fix this by using `podman run --init`, which inserts a minimal pid 1.

Fixes #6036

 * [ ] FAIL: image-refresh debian-stable

### DIFF
--- a/lib/aio/job.py
+++ b/lib/aio/job.py
@@ -84,6 +84,8 @@ async def run_container(job: Job, subject: Subject, ctx: JobContext, log: LogStr
 
         args = [
             *ctx.container_cmd, 'run',
+            # we run arbitrary commands in that container, which aren't prepared for being pid 1; reap zombies
+            '--init',
             *ctx.container_run_args,
             f'--cidfile={cidfile}',
             *(f'--env={k}={v}' for k, v in job.env.items()),


cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-debian-stable-3b811cf2-20240308-032318/log.html

martinpitt:
Oh, shucks -- `job-runner` PRs are self-testing :cry: So I suppose you just have to trust me on that.

allisonkarlitskaya:
> Oh, shucks -- `job-runner` PRs are self-testing 😢 So I suppose you just have to trust me on that.

I guess you mean that this is still failing because job-runner gets picked up by cockpituous out of the main branch, and not the PR?

lib/aio/job.py
```diff
@@ -84,6 +84,8 @@ async def run_container(job: Job, subject: Subject, ctx: JobContext, log: LogStr
 
         args = [
             *ctx.container_cmd, 'run',
+            # we run arbitrary commands in that container, which aren't prepared for being pid 1; reap zombies
+            '--init',
```
allisonkarlitskaya, line 88: The comment is too long, and in the wrong tone.

```python
'--init',  # make sure zombies get reaped
```

would have done it :)

allisonkarlitskaya (approved):
Thanks for the catch!  I was thinking about doing this anyway, but sort of liked the novel idea of `test/run` ending up as pid1.  I agree that this is better, though!

martinpitt:
> I guess you mean that this is still failing because job-runner gets picked up by cockpituous out of the main branch, and not the PR?

Correct, yes.

lib/aio/job.py
martinpitt: Yeah, I wanted to point out that we don't control the entry point (or at least, it's not a single one), so we can't add reapers there.

allisonkarlitskaya:
> > I guess you mean that this is still failing because job-runner gets picked up by cockpituous out of the main branch, and not the PR?
> 
> Correct, yes.

Can we do something about it, for testing?

martinpitt:
@allisonkarlitskaya run-local.sh, i.e. the "cockpituous" test *does* use the branch. But this particular bug didn't manifest on a trivial image refresh like "cirros" or "foonux", that's why it didn't catch it. We can't run libvirt in that workflow (no kvm).

What we can do is to add a check that zombies get cleaned up. It's not quite "image refreshes on job-runner PRs use that changed code", that'd require some more involved magic that needs some deeper thought.

PR #6044, opened 2024-03-08
images: Drop Fedora ELN

martinpitt:
This isn't interesting right now for the Anaconda team, as ELN already moved to Fedora 41 (i.e. it doesn't represent RHEL 10 any more). We'll bring that back as needed, until then clean up the obsolete images.


martinpitt:
I discussed that with @KKoukiou in chat yesterday.

jelly (approved):
Sure!

PR #6045, opened 2024-03-08
tests-scan: don't 'job'-ify other-repo jobs

allisonkarlitskaya:
job-runner doesn't support running jobs from other repositories, so don't emit `'job': { ... }` objects for those.

### DIFF
--- a/tests-scan
+++ b/tests-scan
@@ -225,8 +225,9 @@ def queue_test(job: Job, channel, options: argparse.Namespace) -> None:
             "sha": job.revision,
             "ref": job.ref,
             "name": job.name,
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks, that's a good bandaid for not having to rush a solution. It of course blocks all cleanup.

allisonkarlitskaya:
> Thanks, that's a good bandaid for not having to rush a solution. It of course blocks all cleanup.

I think we should hold off a bit on the cleanup anyway.  Let's use this thing for a week or two to make sure we're sure we're completely happy with it before we start ripping out the old bits.

PR #6062, opened 2024-03-11
tests-scan: Add "image-download" secret

martinpitt:
So that job containers can download private images (RHEL). This secret gets defined in https://github.com/cockpit-project/cockpituous/pull/594

### DIFF
--- a/test/test_tests_scan.py
+++ b/test/test_tests_scan.py
@@ -282,7 +282,7 @@ class TestTestsScan(unittest.TestCase):
                 "slug": f"pull-{self.pull_number}-abcdef-20240102-030405-fedora-nightly",
                 "target": "stable-1.0",
                 "container": "supertasks",
-                "secrets": ["github-token"],
+                "secrets": ["github-token", "image-download"],
                 "env": {
                     "BASE_BRANCH": "stable-1.0",
                     "COCKPIT_BOTS_REF": "main",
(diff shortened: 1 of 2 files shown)


test/test_tests_scan.py
```diff
@@ -282,7 +282,7 @@ def test_amqp_pr(self, mock_queue, _mock_strftime):
                 "slug": f"pull-{self.pull_number}-abcdef-20240102-030405-fedora-nightly",
                 "target": "stable-1.0",
                 "container": "supertasks",
-                "secrets": ["github-token"],
+                "secrets": ["github-token", "image-download"],
```
jelly, line 285: Is this JSON schema defined anywhere? Half of it looks like a Github event and the other part of the job looks like something Cockpit test specific?

jelly: So the secret is `image-download`? I found that a bit confusing as `image-download` is also something I often use :) So I almost wonders if `secrets` was tied to an invocation.

jelly (approved):

test/test_tests_scan.py
martinpitt: That secret is defined in https://github.com/cockpit-project/cockpituous/pull/594 . I named it that so that it matches the already existing "image-upload" secret. I.e. We declare set of abstract secret names, assign them in the generated tasks ("image-upload", "image-download", "github-token"), and define their "implementation" in the job config.

martinpitt:
https://github.com/cockpit-project/cockpituous/pull/594 landed and deployed. So I'll land this and retry all the failed rhel* tests that happened over the weekend and today.

Issue #5979, opened 2024-02-26
virtinterfaced SIGABRT in udevInterfaceLookupByMACString due to memory corruption


martinpitt:
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2266014

cockpituous:
fedora-39
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testHostDevicesList (__main__.TestMachinesHostDevs.testHostDevicesList)
*** You are running in headless mode.
*** You are running in headless mode.
[GFX1-]: glxtest: libGL.so.1 missing
WebDriver BiDi listening on ws://127.0.0.1:9866
[GFX1-]: RenderCompositorSWGL failed mapping default framebuffer, no dt
DevTools listening on ws://127.0.0.1:9866/devtools/browser/2410c05d-f1e1-4749-b18e-a834d73c8f44
> warning: This error page has no error code in its security info
CDP: {"source":"javascript","level":"warning","text":"Sending message that cannot be cloned. Are you trying to send an XPCOM object?","timestamp":1709915006877,"url":"resource://gre/modules/LoginManagerChild.sys.mjs","lineNumber":2431}
CDP: {"source":"other","level":"error","text":"The resource from “http://127.0.0.2:9391/cockpit/@localhost/*/po.manifest.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).","timestamp":1709915008772,"url":"http://127.0.0.2:9391/machines","lineNumber":0}
CDP: {"source":"other","level":"error","text":"The resource from “http://127.0.0.2:9391/cockpit/@localhost/*/po.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).","timestamp":1709915008774,"url":"http://127.0.0.2:9391/machines","lineNumber":0}
CDP: {"source":"other","level":"error","text":"The resource from “http://127.0.0.2:9391/cockpit/@localhost/*/po.manifest.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).","timestamp":1709915008890,"url":"http://127.0.0.2:9391/machines","lineNumber":0}
CDP: {"source":"other","level":"warning","text":"Loading failed for the <script> with source “http://127.0.0.2:9391/cockpit/@localhost/*/po.manifest.js”.","timestamp":1709915008890,"url":"http://127.0.0.2:9391/machines","lineNumber":13}
[...]

PR #6073, opened 2024-03-12
naughty: Add variation of #5979

martinpitt:
This can also crash in udevConnectListAllInterfaces() instead of udevInterfaceLookupByMACString(), but it's the same root cause.

Fixes https://github.com/cockpit-project/cockpit-machines/issues/1490


allisonkarlitskaya (approved):
I'm not sure if having different naughty patterns per distro is something we do, but I'm fairly sure that if anyone would know about that, it's you :sweat_smile: 

:+1:

martinpitt:
@allisonkarlitskaya how do you mean? THese three files are exact copies, they just applied to three different Fedoras. THe already existing file (without -2) refers to a different stack trace.

PR #6083, opened 2024-03-13
tests/issue-scan: Drop old shell output and tests-invoke/make-checkout/s3-streamer

martinpitt:
Our job-runner infra has settled down, so we don't need the old s3-streamer/make-checkout/tests-invoke output anymore.


tests-scan
```diff
@@ -462,7 +401,7 @@ def scan_for_pull_tasks(api, contexts, opts, repo):
         results.sort(reverse=True, key=str)
         return [tests_human(job, options=opts) for job in results]
     if not opts.amqp:
-        return [tests_invoke(job, options=opts) for job in results]
+        return [queue_test(job, None, options=opts) for job in results]
```
martinpitt, line 404: Note: This isn't super-nice or elegant. But this beast is due for a rewrite anyway, I just want to shed some dead weight first.

test/test_tests_scan.py
```diff
@@ -108,147 +117,101 @@ def setUp(self):
...
-        args = ["--dry", "--repo", self.repo, "--pull-number", self.pull_number,
-                "--context", self.context]
-        proc, output, stderr = self.run_tests_scan(args)
+        return code, mock_stdout.getvalue(), mock_stderr.getvalue()
 
-        self.assertEqual(proc.returncode, 0)
-        expected_output = self.expected_command()
-        self.assertRegex(output, expected_output)
-        self.assertIsNone(stderr)
+    def run_success(self, args, expected_output, repo=None):
```
allisonkarlitskaya, line 141: (volume low, but gradually increasing) I would appreciate it if we type-annotated new code...

tests-scan
```diff
@@ -144,110 +143,50 @@ def is_internal_context(context):
...
-            env["COCKPIT_BOTS_REF"] = job.bots_ref
-        if job.base:
-            env["BASE_BRANCH"] = job.base
-
-        body = {
-            # old-style command line jobs
-            "command": command,
-            "type": "test",
+    body = {
+        "job": {
```
allisonkarlitskaya, line 164: I wonder if you'd like to add a `"human"` field here, like in `issue-scan`.

lib/s3-html/log.html
```diff
@@ -478,7 +478,7 @@
     content += await fetch_from(filename, bytes);
     set_content(content);
 
-    console.log('Thank you for using s3-streamer.  Have a nice day.');
+    console.log('Thank you for using s3streamer.  Have a nice day.');
```
allisonkarlitskaya, line 481: lolwat

On account of that's the name of the library? :)

allisonkarlitskaya (approved):
This can go as is.  Gotta say, though: a bit upset that I didn't get to score these -1000 lines!!

If you're going to do fixups, note: the "issue-scan" commit message says "...similar to we did in issue scan" when I think you mean "tests-scan".

lib/s3-html/log.html
martinpitt: Yeah, "s3-streamer" as such isn't a thing any more.

martinpitt:
I fixed the commit message, too.

tests-scan
martinpitt: Is this ok for a follow-up? Running out of time today

test/test_tests_scan.py
```diff
@@ -18,16 +18,16 @@
...
 import importlib
+import io
 import json
 import os
 import shutil
-import subprocess
 import tempfile
 import unittest
 import unittest.mock
+from typing import Tuple
```
allisonkarlitskaya, line 28: This isn't cool anymore — just use `tuple` directly.

allisonkarlitskaya (approved):

PR #6087, opened 2024-03-14
cockpit-container-update: automate ci container update workflow

jelly:
Similar to cockpit-lib-update the CI container can be periodically updated using a GitHub workflow (or something similar).

---

Ran it on my machines fork and it a PR!

https://github.com/jelly/cockpit-machines/actions/runs/8279777037
https://github.com/jelly/cockpit-machines/pull/2

### DIFF
--- a/cockpit-container-update
+++ b/cockpit-container-update
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2024 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+# Update .cockpit-ci/container to the latest tasks container tag automatically
+
+import json
+import os
+import sys
+import urllib.request
+
+import task
+from lib.constants import BASE_DIR
+
+sys.dont_write_bytecode = True
+
+
+URL = 'https://quay.io/api/v1/repository/{}/tag/?onlyActiveTags=true'
+
+
+def fetch_tags(url):
+    req = urllib.request.Request(url)
+    with urllib.request.urlopen(req) as response:
+        return json.loads(response.read().decode().strip())
+
+
+def run(context, verbose=False, **kwargs):
+    container = os.path.join('.cockpit-ci', 'container')
+    container_path = os.path.join(BASE_DIR, container)
+
+    with open(container_path) as fp:
+        content = fp.read()
+
+    container_uri, _, current_tag = content.strip().partition(':')
+
+    if not container_uri.startswith("quay.io"):
+        print("CI container not hosted on quay.io, skipping")
+        return
+
+    url = URL.format(container_uri.removeprefix("quay.io/"))
+    tags = fetch_tags(url)
+
+    # Strip latest tags
+    tags = [t for t in tags['tags'] if t['name'] != 'latest']
+    latest_image = sorted(tags, key=lambda x: x['start_ts'])[-1]
+    latest_tag = latest_image['name']
+    print(f"Latest discovered tag: '{latest_tag}'")
+
+    if current_tag == latest_tag:
+        print("CI container is already up to date, nothing to do")
+        return
+
+    with open(container_path, 'w') as fp:
+        fp.write(f"{container_uri}:{latest_tag}\n")
+
+    if not kwargs['dry']:
+        title = f".cockpit-ci: Update CI container to {latest_tag}"
+        branch = task.branch('cockpit-ci-container', title, pathspec=container, **kwargs)
+        kwargs["title"] = title
+        task.pull(branch, **kwargs)
+
+
+if __name__ == '__main__':
+    task.main(function=run, title="Update CI container for cockpit projects")


cockpit-container-update
```diff
@@ -0,0 +1,83 @@
...
+URL = 'https://quay.io/api/v1/repository/cockpit/tasks/tag/?onlyActiveTags=true'
+
+
+def fetch_tags(url):
+    try:
+        req = urllib.request.Request(url)
+        with urllib.request.urlopen(req) as response:
+            return json.loads(response.read().decode().strip())
+    except (urllib.error.HTPPError, ConnectionResetError, socket.error, SSLEOFError) as e:
+        print("Transient error during quay.io request %s", e)
```
jelly, line 45: Kindly borrowed this from our github.py code but maybe it is a bit over the top :)

martinpitt: This should either be a retry loop and hard fail after a few retries, or just fail right away. The latter is fine IMHO -- we are going to run it every day anyway. Let's not silently ignore it, we won't be able to tell the difference between "the job didn't find something new" and "the job is broken".

TL/DR: Please drop all error handling here.

cockpit-container-update
```diff
@@ -0,0 +1,83 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+# Update .cockpit-ci/container to the latest tasks container tag automatically
+
+import json
+import os
+import socket
+import sys
+import urllib.parse
```
martinpitt, line 26: Not used anywhere -- didn't your vim complain? :grin:

cockpit-container-update
```diff
@@ -0,0 +1,83 @@
...
+            return json.loads(response.read().decode().strip())
+    except (urllib.error.HTPPError, ConnectionResetError, socket.error, SSLEOFError) as e:
+        print("Transient error during quay.io request %s", e)
+
+    return None
+
+
+def run(context, verbose=False, **kwargs):
+    container = context or os.path.join('.cockpit-ci', 'container')
+    container_path = os.path.join(BASE_DIR, container)
```
martinpitt, line 52: `context` is strange here. This will always be .cockpit-ci/container, so it may just as well hardcode that.

cockpit-container-update
```diff
@@ -0,0 +1,83 @@
...
+
+    with open(container_path) as fp:
+        content = fp.read()
+
+    current_tag = content.strip().partition(':')[-1]
+    tags = fetch_tags(URL)
+    if tags is None:
+        return
```
martinpitt, line 60: In theory, projects could also use a different container. Let's not blindly overwrite that with quay/tasks, but actually read the image name and put it into the URL. And exit with a message if it doesn't start with `quay.io`?

cockpit-container-update
```diff
@@ -0,0 +1,83 @@
...
+
+    # Strip latest tags
+    tags = [t for t in tags['tags'] if t['name'] != 'latest']
+    latest_image = sorted(tags, key=lambda x: x['name'])[-1]
```
martinpitt, line 64: Please sort by `start_ts`, it's a standard Unix timestamp.

cockpit-container-update
```diff
@@ -0,0 +1,83 @@
...
+    latest_tag = latest_image['name']
+
+    new_content = content.replace(current_tag, latest_tag)
```
martinpitt, line 67: This is buggy if the tags could be substrings of one another. Let's take the container image name from the original `.partition()` call and append the new tag.

cockpit-container-update
```diff
@@ -0,0 +1,83 @@
...
+
+    current_tag = content.strip().partition(':')[-1]
+    tags = fetch_tags(URL)
+    if tags is None:
+        return
+
+    # Strip latest tags
+    tags = [t for t in tags['tags'] if t['name'] != 'latest']
+    latest_image = sorted(tags, key=lambda x: x['name'])[-1]
+    latest_tag = latest_image['name']
```
martinpitt, line 65: Please print that, so that it's easier to follow what happens in the log.

martinpitt (changes requested):
Thanks!

cockpit-container-update
jelly: Ok, for cockpit-lib-update we made it flexible I guess it's not really useful here.

cockpit-container-update
jelly: Let's just fail and see how often it breaks.

cockpit-container-update
jelly: Right, that is a weird property, what does it even mean? Image start time? I couldn't find anything about it in the quay docs.

martinpitt: The time stamp is pretty much the same as `"last_modified"`, but much easier to use for sorting.

jelly:
Re-running my machines fork after the new push https://github.com/jelly/cockpit-machines/actions/runs/8279777037

cockpit-container-update
```diff
@@ -0,0 +1,80 @@
...
+def run(context, verbose=False, **kwargs):
+    container = os.path.join('.cockpit-ci', 'container')
+    container_path = os.path.join(BASE_DIR, container)
+
+    with open(container_path) as fp:
+        content = fp.read()
+
+    container_uri, _, current_tag = content.strip().partition(':')
+
+    if not container_uri.startswith("quay.io"):
```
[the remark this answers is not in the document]
martinpitt: Nah, this is fine.

cockpit-container-update
```diff
@@ -0,0 +1,80 @@
...
+    url = URL.format(container_uri.removeprefix("quay.io/"))
+    tags = fetch_tags(url)
+
+    # Strip latest tags
+    tags = [t for t in tags['tags'] if t['name'] != 'latest']
+    latest_image = sorted(tags, key=lambda x: x['start_ts'])[-1]
+    latest_tag = latest_image['name']
+    print(f"Latest discovered tag: '{latest_tag}'")
+
+    new_content = content.replace(f"{container_uri}:{current_tag}", f"{container_uri}:{latest_tag}")
```
martinpitt, line 64: Still not happy about using `.replace()` here. A "blind" substring match is potentially and conceptually wrong. Please just write `"{container_uri}:{latest_tag}"` into the file.

martinpitt (changes requested):
Almost, thanks!

cockpit-container-update
jelly: Ack! But with newline :)

martinpitt (approved):
Thanks! I'm happy, but please coordinate with @allisonkarlitskaya wrt. https://github.com/cockpit-project/cockpituous/pull/603

PR #6095, opened 2024-03-18
tasks-container-update: tasks container is on ghcr

allisonkarlitskaya:
Rename the script to tasks-container-update: the 'cockpit' in the name was probably copied from 'cockpit-lib-update' (which is actually about 'cockpit-lib') but the name is not accurate for what this thing updates.

Scan ghcr.io instead of quay.io for the container image.

Use the async GitHub API as a proof of concept.

Don't avoid calling tasks functions if --dry-run is given: they already know how to avoid acting in that case (and give nice output as well).

Unfortunately, in this change, we drop our ability to act as a tasks action. Fortunately, that's deeply theoretical (since the only supported task is image-refresh) and also decreasingly important (as more "tasks" functionality moves to job-runner).

One note: this script requires the `read:packages` capability on the GitHub token.


tasks-container-update
```diff
@@ -0,0 +1,76 @@
...
+    parser.add_argument('--file', default='.cockpit-ci/container')
+    parser.add_argument('--org', default='cockpit-project')
+    parser.add_argument('--package', default='tasks')
+    args = parser.parse_args()
+
+    if args.debug:
+        logging.basicConfig(level=logging.DEBUG)
+
+    async with GitHub.api(program='tasks-container-update') as api:
+        for version in await api.get_list(f'orgs/{args.org}/packages/container/{args.package}/versions'):
```
martinpitt, line 48: So is this the bit which needs authentication, which we discussed in https://github.com/cockpit-project/cockpituous/pull/604#issuecomment-2003103469 ? Indeed it would be quite nice to use `skopeo` for this, as it's universal for all registries and doesn't bind us to the GitHub API

allisonkarlitskaya: We don't have/want skopeo in the tasks container, but thanks to your helpful hit I looked under the hood what it's doing — it turns out there's an API to create a 60-seconds duration anonymous access token which you can use the access the other APIs.

I'm in the process of writing a small version of this in aiohttp which we can use.  I want to see if I can possibly get to a place where I can choose the tag based on which one is equal to `latest` — that's not possible with skopeo's output, either.

tasks-container-update
```diff
@@ -0,0 +1,79 @@
...
+    headers = {
+        'User-Agent': 'cockpit-project/bots (tasks-container-update)'
+    }
+
+    service, _, repository = args.image.partition('/')
+    if not service or not repository:
+        args.error(f'Invalid image name: {args.image}')
+
+    async with aiohttp.ClientSession(raise_for_status=True) as session:
+        async with session.get(URL.build(scheme='https', host=service, path='/token', query={
```
martinpitt, line 57: This could use a comment where this API came from (e.g. you got it from `skopeo --debug`, right?)

martinpitt (changes requested):
Thanks, this looks even simpler than the previous version! But I think this bit of the commit message is obsolete now:

> One note: this script requires the `read:packages` capability on the GitHub token.

.. as it gets its own token via the docker API?

tasks-container-update
```diff
@@ -0,0 +1,79 @@
...
+            token = get_str(typechecked(await response.json(), dict), 'token')
+            headers['Authorization'] = f'Bearer {token}'
+
+        async with session.get(f'https://{service}/v2/{repository}/tags/list', headers=headers) as response:
+            logging.debug('list response: %r', await response.json())
+            tags = get_strv(await response.json(), 'tags')
+
+    tag = max({*tags} - {'latest'})
+    logging.debug('Latest tag is %r', tag)
+    Path(args.file).write_text(f'ghcr.io/{args.org}/{args.package}:{tag}\n')
```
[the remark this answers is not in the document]
allisonkarlitskaya: I wonder, actually.  I feel like git will complain in that case (no changes to commit) but I guess we won't deal with it gracefully.  I'll test it.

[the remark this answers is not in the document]
allisonkarlitskaya: So `.branch()` checks for this and returns `None` in that case.  We just need to check for that before calling `.pull()` and it DTRT.

tasks-container-update
allisonkarlitskaya: I added a note to the commit message (as well as links to the official documentation).

allisonkarlitskaya:
This is obviously blocked on actually publishing the container on ghcr...

PR #6102, opened 2024-03-19
tests-scan: Restore "no-test" handling

martinpitt:
The dropped `tests_invoke()` in commit 23a5f798466360 also cared about
skipping tests with priority == 0, so dropping that function caused
"no-test" PRs to still get tested.

Factorize this check into the main scanning function, so that
tests_human() and queue_tests() don't need to duplicate it.

Also fix the typing: Consistently handle `priority` as `int`, and fix
the bogus `priority = github.NO_TESTING` (which is a string constant).
That was the reason why the old `tests_invoke()` had to ignore TypeError
on int conversion.

Add a no-test PR to the unit test. Without the above fix, this caused
three tests to fail due to the unexpected queueing of PR #3. But also
add explicit human and JSON unit tests for a no-test  PR.

-----

I also tested this on https://github.com/martinpitt/cockpit-podman/pull/11 with a local testmap modification:
```diff
--- lib/testmap.py
+++ lib/testmap.py
@@ -87,7 +87,7 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
             'rhel4edge',
         ],
     },
-    'cockpit-project/cockpit-podman': {
+    'martinpitt/cockpit-podman': {
         'main': [
             'arch',
             'rhel-8-10',
```

Now `./tests-scan --repo martinpitt/cockpit-podman -p 11` behaves correctly. Previously, it first said "skipping because it is no-test" and then queued them anyway.

### DIFF
--- a/test/test_tests_scan.py
+++ b/test/test_tests_scan.py
@@ -192,7 +192,8 @@ class TestTestsScan(unittest.TestCase):
         self.run_success_mock_pr(["--dry", "--context", self.context])
 
     def test_pull_number_human_readable(self):
-        self.run_success(["--dry", "-v", "--context", self.context], self.expected_human_output)
+        self.run_success(["--dry", "-v", "--context", self.context, "--pull-number", str(self.pull_number)],
+                         self.expected_human_output)
 
     def test_pull_data_human_readable(self):
         args = ["--dry", "-v", "--context", self.context,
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya (approved):

allisonkarlitskaya:
Thanks for looking into this.

PR #6105, opened 2024-03-19
tests-scan: remove concept of priority

allisonkarlitskaya:
This kills off the prioritize() function and makes a lot of related changes.

To my eye, the main logic of the program through the `todos` loop in `cockpit_tasks()` is a lot easier to understand now.

There is a ridiculous number of commits here, but I've been very careful about each commit either being a straight refactor or an intentional and obvious logic change (mostly to do with removing priorities).  I've verified that each commit passes `pytest`, and have made test adjustments in the "logic change" commits.

There's more work to do, but I wanted to subject this to a full CI run before I continue.

### DIFF
--- a/tests-scan
+++ b/tests-scan
@@ -250,8 +250,6 @@ def cockpit_tasks(api: github.GitHub, contexts, opts: argparse.Namespace, repo: 
         login = pull["head"]["user"]["login"]
         base = pull.get("base", {}).get("ref")  # The branch this pull request targets (None for direct SHA triggers)
 
-        allowed = login in ALLOWLIST
-
         logging.info("Processing #%s titled '%s' on revision %s", number, title, revision)
 
         labels = labels_of_pull(pull)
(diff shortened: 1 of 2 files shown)


tests-scan
```diff
@@ -181,28 +181,20 @@ def queue_test(job: Job, channel, options: argparse.Namespace) -> None:
         print(json.dumps(body['job']))
 
 
+def update_status(
+    api: github.GitHub, revision: str, last: Mapping[str, str], dry: bool, changes: Mapping[str, str] | None
+) -> bool:
+    if changes is None or {**last, **changes} == last or dry:
+        return True
```
martinpitt: The commit message claims "use task.would() to print a nice message in that case", but it doesn't?

martinpitt (changes requested):
Thanks! Let's just keep the commits, they are easy to follow (and if necessary, bisect/revert), and well described. This looks great, I just have one question (a very likely omission)

tests-scan
allisonkarlitskaya: It did but it broke the expected output of the tests so I took it back out.  I'll fix the commit message.

allisonkarlitskaya:
> While you are at it, it'd also be super useful to add comments to the `Job` named tuple. In particular, `revision`, `rev`, `base`, and the difference between `context` and `github_context`. We both learned what these mean _now_, but we'll forget again in a few weeks/months 😅 Possibly also some renaming, like `{test,report}_context` ?

The very next thing I'm going to do is to get rid of that NamedTuple :)

martinpitt (approved):
Ack, looking forward to it!

PR #6112, opened 2024-03-20
 next round of tests-scan cleanups

allisonkarlitskaya:
I'm not 100% on the direction here, but at least Job is gone...

I think we need to start thinking about how to split cockpit_tasks() up a bit more.


martinpitt (approved):
Thanks!

PR #6115, opened 2024-03-20
s3: use .raw_path for canonical request

allisonkarlitskaya:
When signing the canonical request for S3, we need to make sure that we escape the path properly.  That's `.raw_path`.

### DIFF
--- a/lib/aio/s3.py
+++ b/lib/aio/s3.py
@@ -134,7 +134,7 @@ def s3_sign(
         signing_key = hmac.new(signing_key, item.encode('ascii'), hashlib.sha256).digest()
 
     algorithm = 'AWS4-HMAC-SHA256'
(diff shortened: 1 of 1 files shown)


martinpitt:
> Error: unable to connect to Podman socket: server API version is too old. Client "4.0.0" server "3.4.4"

Mh? Ah, [temporary launchpad outage](https://github.com/cockpit-project/bots/actions/runs/8360993925/job/22887983938?pr=6115#step:7:40). Let's retry

martinpitt (approved):
Thanks! (Assuming green, but we don't cover a `@bots#` scenario in the tests)

martinpitt:
Yeah, https://ppa.launchpadcontent.net/ still sad. I locally ran `COCKPIT_BOTS_BRANCH=raw-path pytest`

allisonkarlitskaya:
I tested it on lislogs with manual `--slug`.  It was broken before, and after: https://lislogs.nl-ams-1.linodeobjects.com/x%23y/log.html

PR #6126, opened 2024-03-21
workflows: Adjust to cockpituous' new pytest

martinpitt:
`run-local.sh` was replaced with pytest. See https://github.com/cockpit-project/cockpituous/pull/606

----

The first version has an additional test commit e75b0f4525acfb8c24 which breaks `image-download`. The test [failed](https://github.com/cockpit-project/bots/actions/runs/8378382325/job/22942800873?pr=6126)  , proving that it actually uses the branch from the PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,6 +68,12 @@ jobs:
           repository: cockpit-project/cockpituous
           path: cockpituous
 
+      - name: Install test dependencies
+        if: steps.changes.outputs.changed
+        run: |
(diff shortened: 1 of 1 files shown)


martinpitt:
OK, so the tests [failed](https://github.com/cockpit-project/bots/actions/runs/8378382325/job/22942800873?pr=6126) in the right way, `test_images` and the mock image refresh stumbled over the crashing `image-download`. Dropping the test commit.

allisonkarlitskaya (approved):

PR #6136, opened 2024-03-26
testmap: Add manual opensuse-tumbleweed contexts for cockpit/machines/podman

martinpitt:
These are the next targets for enabling tests on openSUSE.


mvollmer (approved):
Welcome!

PR #6137, opened 2024-03-26
Add support for testing cockpit-podman for opensuse tumbleweed

Nykseli:


images/scripts/opensuse-tumbleweed.setup
```diff
@@ -78,6 +78,9 @@ rpm-config-SUSE \
 
 NETWORK_PACKAGES="\
 systemd-network \
+sssd \
+sssd-dbus \
```
martinpitt, line 82: Is this just "for later", when you start testing cockpit? c-podman shouldn't care about sssd.

images/scripts/opensuse-tumbleweed.setup
```diff
@@ -117,12 +120,17 @@ su builder -c "sudo mock --verbose -i $(/var/lib/testvm/build-deps.sh "${ID} ${V
 
 # Prevent SSH from hanging for a long time when no external network access
 echo 'UseDNS no' >> /etc/ssh/sshd_config
+# Pam is not enabled by default and is needed for systemctl --user to work
+echo 'UsePAM yes' >> /etc/ssh/sshd_config
```
martinpitt, line 124: Hmm, that worries me quite a lot -- we try to test OSes and adjust to them as they are, not how we'd like them to be. That means that when you SSH into a TW image, you don't get a user systemd and D-Bus? That will actually make things fail when you connect to TW in secondary sessions (host switcher) or with the flatpak.

Could this go into podman's test/vm.install for the time being? Then we can keep the pristine bots image as the real default config.

images/scripts/opensuse-tumbleweed.setup
```diff
@@ -117,12 +120,17 @@ su builder -c "sudo mock --verbose -i $(/var/lib/testvm/build-deps.sh "${ID} ${V
...
 
 # Audit events to the journal
 rm -f '/etc/systemd/system/multi-user.target.wants/auditd.service'
 rm -rf /var/log/audit/
 
 systemctl enable --now systemd-networkd.service
+# dbus-broker needs to be manually enabled on opensuse-tumbleweed
+systemctl enable --now dbus-broker.service
+systemctl --global enable dbus-broker.service
```
martinpitt, line 133: :eyes: There's no running D-Bus daemon on openSUSE by default? That is by far the most worrying thing of all then. Or is that "you specifically want dbus-broker instead of dbus-daemon"? If dbus-daemon is the default, then let's please use that.

martinpitt (changes requested):
Thanks! Before we go ahead with this I have some questions.

images/scripts/opensuse-tumbleweed.setup
martinpitt: The current image *does* have a running `dbus.service` (`dbus-daemon`), so let's please keep that. That feels like a distro choice, not something which we should change in our tests.

images/scripts/opensuse-tumbleweed.setup
```diff
@@ -89,7 +92,7 @@ printf '[main]\nno-auto-default=*\n' > /etc/NetworkManager/conf.d/noauto.conf
 echo "127.0.1.1 $(hostname)" >> /etc/hosts
 
 if ! getent passwd admin >/dev/null; then
-    useradd -c Administrator admin
+    useradd -c Administrator -G wheel admin
```
martinpitt, line 95: Curious -- I booted the current main image, and I *can* login as `admin@` and *can* already sudo, and the admin user already is in `wheel`. So this probably means that the cloud image has an admin user by default? But then, how did it get its "foobar" password? Something doesn't add up here..

images/scripts/opensuse-tumbleweed.setup
Nykseli: When the tests were failing due to dbus configuration, it complained about not finding sssd service so I decided to throw them in, just in case.

images/scripts/opensuse-tumbleweed.setup
Nykseli: There was no user systemd by default. And now that I think about it, it feels like a bug. I'll reach out the image maintainers and ask if it's intended behavior.

images/scripts/opensuse-tumbleweed.setup
Nykseli: There's D-Bus daemon by default, so this is "I prefer to use dbus-broker". But you're right. As dbus-daemon is the default. I should figure out a way to use it instead.

images/scripts/opensuse-tumbleweed.setup
Nykseli: That's interesting, I assumed based on the original code that admin is not created by default. But if it is and it's in wheel already, we can get rid of the whole if statement. Thanks for pointing this out!

martinpitt: Indeed the [most recent build log](https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-opensuse-tumbleweed-be5a86cc-20240323-053844/log.html) shows:
```
+ getent passwd admin
+ '[' opensuse-tumbleweed '!=' opensuse-tumbleweed ']'
```
I.e. admin is already present. You'd still need the `chpasswd` that sets the password, but I figure that's already taken care of by cloud-init. So indeed you can drop this  whole paragraph.

images/scripts/opensuse-tumbleweed.setup
martinpitt: > I should figure out a way to use it instead.

At least the system one does out of the box. I tried to do the `echo 'UsePAM yes' >> /etc/ssh/sshd_config` and I can do `busctl --user` -- `systemctl --user status dbus.socket` is running, so it spawns it on demand. That seems "good enough", and what the distro decided to do. So what needs fixing?

images/scripts/opensuse-tumbleweed.setup
Nykseli: So this was indeed a bug. Tumbleweed has the sshd_config in `/usr/etc/ssh/sshd_config` and creating `/etc/ssh/sshd_config` confused the system and accidentally disabled PAM

Nykseli: Also now that `sshd_config` is properly set up in bots, podman's `test/vm.install` doesn't need to be modified

martinpitt:
Thanks @Nykseli ! This looks good now, I opened #6174 to do the actual build. I need to go AFK now, but if/once it succeeded, please cherry-pick the refresh commit here.

martinpitt:
Image build succeeded, I pushed it here with a rebase.

martinpitt (approved):
I didn't test podman with this yet, but (1) as you already did, and (2) the commits here are at least necessary, and (3) it doesn't break starter-kit, let's land this.

In the best case, podman works now and we can enable it in the testmap, right?

martinpitt:
@Nykseli see #6175

PR #6164, opened 2024-04-01
Image refresh for rhel-9-4

cockpituous:
Image refresh for rhel-9-4
 * [x] image-refresh rhel-9-4
 * [ ] https://github.com/candlepin/subscription-manager-cockpit/pull/64


cockpituous:
image-refresh rhel-9-4 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-9-4-20240401-225350

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel-9-4-7a3c02e3-20240401-223733/log.html

martinpitt:
sub-man testSubAndInAndFail test is notoriously flaky, but it failed retries three times in a row. This  needs a closer look.

mvollmer:
> sub-man testSubAndInAndFail test is notoriously flaky, but it failed retries three times in a row. This needs a closer look.

It fails also with our current image.  https://github.com/candlepin/subscription-manager-cockpit hasn't seen an update in 7 months, hmm.

I'll investigate more.

mvollmer:
> I'll investigate more.

Seems to be a too short timeout.

martinpitt:
Note that I sent https://github.com/candlepin/subscription-manager-cockpit/pull/63 just last week, and it passed.

mvollmer:
> Note that I sent [candlepin/subscription-manager-cockpit#63](https://github.com/candlepin/subscription-manager-cockpit/pull/63) just last week, and it passed.

Yes, with warnings about getting very close to the timeout... :-)

martinpitt:
Oh right -- I faintly remember patching `run-tests` to show them even for successful tests :grin: https://github.com/cockpit-project/cockpit/commit/ebaea5592d000a41557cf73d75e16a7ae70fa844

mvollmer:
I was suspecting that we miss some notification, but it just takes long for insights-client to fail.

 - I see if I can implement a cleaner way for it to fail by having mock-insights return an error. This doesn't make the test faster, but is more realistic.
 - There is a avalanche of "[Errno -2] Name or service not known" alerts popping up. I'll investigate that as well.

mvollmer:
> * There is a avalanche of "[Errno -2] Name or service not known" alerts popping up. I'll investigate that as well.

No, I gave up on that. Nicht meine Baustelle.

mvollmer:
> * I see if I can implement a cleaner way for it to fail by having mock-insights return an error. This doesn't make the test faster, but is more realistic.

https://github.com/candlepin/subscription-manager-cockpit/pull/64

martinpitt:
The re-re-re-retry failed as well. Blocking for a while. However, let's time out in a week or so, we shouldn't block the refresh forever.

martinpitt (approved):

martinpitt:
Landing -- more RHEL images are coming in, and this isn't a regression.

PR #6168, opened 2024-04-03
images: Add Fedora ELN bootc

martinpitt:
This introduces an image which uses [bootc](https://containers.github.io/bootc/) for deploying OCI container images as OS.

The initial installation is a bit tricky as it requires converting a container image to a bootable qcow2 image. That happens with https://github.com/osbuild/bootc-image-builder which is very demanding: It needs both /dev/kvm and a `--privileged` container, so the only place where we can run this is in a VM on our own CI.

https://issues.redhat.com/browse/COCKPIT-1108

 * [x] image-refresh fedora-eln-bootc

### DIFF
--- a/image-trigger
+++ b/image-trigger
@@ -46,6 +46,7 @@ REFRESH = {
     "fedora-39": {},
     "fedora-40": {},
     "fedora-coreos": {},
+    "fedora-eln-bootc": {},
     "fedora-rawhide": {},
     "fedora-rawhide-boot": {},
     "fedora-rawhide-anaconda-payload": REFRESH_30,
--- a/images/scripts/fedora-eln-bootc.Containerfile
+++ b/images/scripts/fedora-eln-bootc.Containerfile
@@ -1,0 +1,18 @@
+FROM quay.io/centos-bootc/fedora-bootc:eln
+
+# pre-install the distro version, which is useful for testing extensions and manual experiments
+RUN \
+    dnf install -y --setopt install_weak_deps=False cockpit-system cockpit-networkmanager && \
+    dnf clean all
+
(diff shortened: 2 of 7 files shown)


cockpituous:
image-refresh fedora-eln-bootc done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-eln-bootc-20240403-170843

martinpitt:
https://github.com/osbuild/bootc-image-builder/issues/321 got fixed, so we can get rid of that expensive "local registry" hack :tada: But starter-kit doesn't work yet, this needs some further adjustments.

martinpitt:
I have a bigger stumbling block: On this image, running cockpit/ws fails:
```
# podman container runlabel RUN cockpit/ws -- --no-tls
Error: netavark: unable to append rule '! -d 224.0.0.0/4 -j MASQUERADE' to table 'nat': code: 4, msg: Warning: Extension MASQUERADE revision 0 not supported, missing kernel module?
iptables v1.8.10 (nf_tables):  RULE_APPEND failed (No such file or directory): rule in chain NETAVARK-1D8721804F16F
```

On our fedora-coreos image, `net/netfilter/xt_MASQUERADE.ko.xz` is shipped in kernel-modules-core-6.7.9-200.fc39.x86_64. But on ELN, kernel-modules-core-6.9.0-0.rc2.23.eln136.x86_64 does not contain this module, just some `net/netfilter/nft_masq.ko.xz`. So this is somewhere between an ELN kernel and a podman bug.

martinpitt:
Ah, needs to be filed from origin anyway. Closing then, then we have that whole PR as an archive.

PR #6171, opened 2024-04-04
images: Add CentOS 9 stream bootc

martinpitt:
This introduces an image which uses [bootc](https://containers.github.io/bootc/) for deploying OCI container images as OS. Don't include "stream" in the name, as that's the only kind of CentOS 9 that there is, and the name gets too long otherwise.

The initial installation is a bit tricky as it requires converting a container image to a bootable qcow2 image. That happens with https://github.com/osbuild/bootc-image-builder which is very demanding: It needs both /dev/kvm and a `--privileged` container, so the only place where we can run this is in a VM on our own CI.

The bootstrap mechanism is generic and also works for e.g. Fedora ELN, so put it into scripts/lib/.

Add manual test map entries for a few projects where we may want to start using this image.

https://issues.redhat.com/browse/COCKPIT-1108

----

I locally tested this against cockpit-podman and it passes all tests. It needs a little Makefile tweak though: https://github.com/cockpit-project/cockpit-podman/pull/1656


cockpituous:
image-refresh centos-9-bootc done: https://github.com/cockpit-project/bots/commits/image-refresh-centos-9-bootc-20240404-074827

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-centos-9-bootc-45eea2de-20240404-073139/log.html

cockpituous:
image-refresh centos-9-bootc done: https://github.com/cockpit-project/bots/commits/image-refresh-centos-9-bootc-20240404-080744

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-centos-9-bootc-09c19343-20240404-075004/log.html

lib/constants.py
```diff
@@ -18,7 +18,7 @@
 import os
 
 # Images which are OSTree based
-OSTREE_IMAGES = ["fedora-coreos", "rhel4edge"]
+OSTREE_IMAGES = ["centos-9-bootc", "fedora-coreos", "rhel4edge"]
```
jelly, line 21: We could make this a bit more vendor neutral and call it `IMMUTABLE_IMAGES`. But that might require changes in Cockpit as well

images/scripts/lib/bootc.bootstrap
```diff
@@ -0,0 +1,68 @@
...
+m.write('/root/config.json', json.dumps(config))
+
+m.execute('mkdir output')
+
+m.execute(['podman', 'run', '--rm', '-i', '--privileged', '--security-opt=label=type:unconfined_t',
+           # for --local
+           '--volume=/var/lib/containers/storage:/var/lib/containers/storage',
+           '--volume=./config.json:/config.json',
+           '--volume=./output:/output',
+           'quay.io/centos-bootc/bootc-image-builder:latest',
```
jelly, line 58: Will this also be the builder for a possible RHEL image?

jelly (approved):
Wow, that was some work! Conceptually I have some questions:

* Does this still work with `image-diff`? that is useful for refreshes imo

lib/constants.py
martinpitt: Yeah, lots of them. I kept it because (1) it's referenced in a lot of places, and (2) decorators like `@skipOstree` are applied for lots of different reasons: "/usr is not writable", or "our coreos image doesn't have PCP support", etc.

For now, bootc images are still ostree, and will be for a long time. It just changes the transport of images/updates and the CLI for applying them.

images/scripts/lib/bootc.bootstrap
martinpitt: For now, yes -- that's mostly just image-builder wrapped into a container, and it can build ELN, RHEL, and CentOS. There is a private RHEL-built container image as well, but it's hard to get to, and we really don't need it. We just need the functionality of "turn an OCI image into a qcow".

martinpitt:
> Does this still work with image-diff? that is useful for refreshes imo

Absolutely, yes. It's still RPM content, and `rpm -qa` works as always.

images/scripts/lib/bootc.bootstrap
jelly: No worries!

PR #6175, opened 2024-04-04
testmap: Enable opensuse-tumbleweed on cockpit-podman

martinpitt:


martinpitt:
@Nykseli FYI

martinpitt:
So this is good progress! Some failures, but lots of green already!

Nykseli:
Thanks!

The test are failing, and to me it seems to be an issue with cockpit-ws and not cockpit-podman. Is there a good workflow for running the podman test suite against locally built cockpit?

martinpitt:
@Nykseli How do you mean with "cockpit-ws"? That part seems to work fine. Do the failures look differently when you run the test locally?

Going through the failures:
 - `testCheckpointRestore` [screenshot](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6175-17dbadf4-20240404-145650-opensuse-tumbleweed-cockpit-project-cockpit-podman/TestApplication-testCheckpointRestore-opensuse-tumbleweed-127.0.0.2-2401-FAIL-1.png) says "crun checkpoint ... failed". Might be missing `criu`, or some other bug. Should be straightforward to reproduce on the command line. If tumbleweed simply doesn't have `criu`, then adjust `self.has_criu` in `def setUp()`  accordingly.
 - `testCreatePod{System,User}` and `testRunImage{System,User}`: all the same root cause. perhaps opensuse doesn't use SELinux ? If not, adjust `self.has_selinux` accordingly
 - `testLifecycleOperationsUser`: podman doesn't report memory usage, perhaps that cgroup controller isn't enabled? Needs CLI investigation.

So at first sight the failures are all quite straightforward?

Nykseli:
I saw that there was bunch of errors like this:
```
CDP: {"source":"network","level":"error","text":"Failed to load resource: the server responded with a status of 404 (ERROR)","timestamp":1712242957368.314,"url":"http://127.0.0.2:9291/cockpit/@localhost/*/po.js","networkRequestId":"12805.21"}
```
Which seemed like something wrong with cockpit-ws. But I guess those are not fatal errors?

Thanks @martinpitt for pointing me into the right direction. I'll look into fixing those issues!

martinpitt:
@Nykseli Ah, ignore these -- that just means it was looking for English translation  files and there aren't any. Indeed it's annoying, I suppose in the cockpit team we just all learned to not even "see" these any more :sweat_smile:

Nykseli:
Ah, so errors that are not really errors. Good to know. I'll stop seeing them as well! :smile:

martinpitt:
Closing this for now, it's stale. Let me know if/when you want to enable this project for opensuse.

PR #6182, opened 2024-04-05
README.md: Document how to create a new CI image

jelly:
* [ ] document how to build a package for a new image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,39 @@ progress and at the end there will be links to commits with the new
 images.  You can then include these commits into the pull request in
 any way you like.
 
+### Creating a new image
+
+Creating a new image from scratch requires some
+[images/scripts/](./images/scripts) files. For a new image called `tux` we need:
+
+- `tux.bootstrap`: Download or create an initial qcow2 image which boots and has SSH
+  acccess. If available, this should be the latest available cloud image,
+  but it may also invoke some other VM build tool or build service.
+  This script runs in the [cockpit/tasks](https://github.com/cockpit-project/cockpituous/tree/main/tasks/container)
+  container, and thus can only use the tools installed there. If necessary, add
+  them first.
+- `tux.setup`: The setup script runs inside the downloaded test image,
+  install all required build and test dependencies, and sets up an `admin`
+  user.
+
+For a new image it is recommended to follow the existing setup/bootstrap scripts,
+for example the `fedora` one.
+
+Run `./image-create -v tux` to build the image. If that succeeds, a new image
+is saved in `images` as `images/tux` (a symlink to the real qcow2 file).
+Test-boot it with `./vm-run tux`, and ensure you can:
+
+ - Log in with SSH as the `admin` and `root` user with our usual test SSH key
+ - Become root with `sudo` as admin works (password "foobar")
+
+To add that image to our CI, create a PR and follow the "Creating new
+images for a pull request" section. External contributors will need to ask a
+Cockpit team member to create a copy of the PR and follow this workflow.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,6 +43,29 @@ For running and debugging the images:
    overlay in test/images/ instead.
  - `vm-reset`: Remove all overlays from test/images/
 
+### Creating a new image
+
+Creating a new image from scratch requires a bootstrap and a setup script,
```
martinpitt, line 48: Technically the setup script is optional

README.md
```diff
@@ -43,6 +43,29 @@ For running and debugging the images:
...
+these scripts are located in `images/scripts`. In this example our new image is
+called `tux` and we run `./image-create -v tux` to create this image we
+expect two scripts:
+
+- `tux.bootstrap`: The bootstrap script downloads the latest available cloud image (qcow2)
```
martinpitt, line 53: While most of our images start with a cloud image, some don't (cirros, bootc, rhel-7.9, rhel4edge). Perhaps

> The bootstrap script has to produce a bootable qcow2 image file, e.g. by downloading a cloud image or invoking some image build service.

README.md
```diff
@@ -43,6 +43,29 @@ For running and debugging the images:
...
+for example the `fedora` one. 
+When `image-create` succeeds a new image is saved in `images` as `images/tux`
+(a symlink to the real qcow2 file).
+
+To add a new image to our CI, create a PR and following the `Creating new
+images for a pull request` workflow. Externals contributors will need to ask a
+Cockpit team member to create a copy of the PR and follow this workflow.
+
+For the initial PR it is recommended to add the new image to the `_manual`
+testmap of `starter-kit` to proof the created image is functional.
```
martinpitt, line 67: I think this should be expanded to explain how to boot-test the image, with `./vm-run`, trying to log in via SSH as `root` and `admin`, testing sudo on the latter. And then trying starter-kit against it locally to ensure it has a functioning package build environment.

martinpitt (commented):
Thanks for starting this! This needs some grammar/interpunction polish too, but let's first collect all the info which we want to add here.

README.md
jelly: Ah! TIL

README.md
jelly: Trying starter-kit requires a script to build packages right? We should also document that ideally and then `vm-customize`.

martinpitt: Hm, right, if it's not rpm. We don't even support .deb there yet. cockpit-ostree has an alternative approach which directly copies dist/ into the VM's /usr/local/share/ -- that is also a good start for a new distro that avoid blocking on packaging as the first step.

martinpitt (dismissed):
@jelly This is becoming stale. I reworked the wording a bit after the discussion above, and did some grammar/language fixes. Indeed the package build is still missing, but this is already an improvement. Please cross-check?

README.md
```diff
@@ -207,6 +207,36 @@ progress and at the end there will be links to commits with the new
...
 any way you like.
 
+### Creating a new image
+
+Creating a new image from scratch requires some
+[images/scripts/](./images/scripts) files. For a new image called `tux` we need:
+
+- `tux.bootstrap`: Download or create an initial qcow2 image which boots and has SSH
+  acccess. If available, this should be the latest available cloud image,
+  but it may also invoke some other VM build tool or build service.
```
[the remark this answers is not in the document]
martinpitt: Good idea! Done.

martinpitt (approved):

martinpitt:
Let's just land this. We can always adjust this easily.

PR #6250, opened 2024-04-21
mypy/ruff mega cleanup

allisonkarlitskaya:
This PR basically aims to get the typing/linting situation under control in `bots/`, improving many things along the way.

It enables almost all of the ruff rules from the main cockpit repository, and gets a majority of the code under `mypy --strict` (~7000 out of ~12000 lines).  The parts the remain are the `github` API, the main body of the `tasks` code, and many of the toplevel scripts.  Those remaining parts feel like they're within striking distance, with another PR or three...

Noteworthy is that `machine` and `lib` are fully-covered, which is nice, because those are the parts we use from `testlib`.

This all started with noticing that the `mypy` step of `test/static-code` in the Cockpit repo takes ~23 seconds, which is by far the longest step in that script.  A lot of research into how mypy actually works followed.  `mypy` isn't slow — we just use it in a way that makes it slow.

This work is going to be followed up with a matching set of changes in the main Cockpit repo (which were developed in parallel, but depend on this work landing first).


allisonkarlitskaya:
All cockpit-composer tests are failing with:

```
  File "/work/make-checkout-workdir/test/verify/check-blueprintList", line 8, in <module>
    import composerlib
  File "/work/make-checkout-workdir/test/verify/composerlib.py", line 6, in <module>
    import testlib
  File "/work/make-checkout-workdir/test/common/testlib.py", line 43, in <module>
    import testvm
  File "/work/make-checkout-workdir/bots/machine/testvm.py", line 27, in <module>
    from lib.constants import BOTS_DIR, DEFAULT_IMAGE, IMAGES_DIR, SCRIPTS_DIR, TEST_DIR, TEST_OS_DEFAULT
ModuleNotFoundError: No module named 'lib'
not ok 1 test/verify/check-blueprintList TestBlueprintList.testFilter [ND@0]
```

...and oh, wow:
https://github.com/osbuild/cockpit-composer/blob/a92e18f07f416d38603df3efe024b5bc57dbcaf4/Makefile#L91

```makefile
machine: bots
	rsync -avR --exclude="bots/machine/machine_core/__pycache__/" bots/machine/testvm.py bots/machine/identity bots/machine/cloud-init.iso bots/machine/machine_core bots/lib test
```

allisonkarlitskaya:
The timeout/execute changes appear to have broken something.  Looking into this.

martinpitt:
Heh, clearly a cold rainy weekend :grin: I had a cursory look, and this looks great! Please re-ask for review once this goes green, i.e. the tests with a custom m.execute() timeout (not sure if there are others).

allisonkarlitskaya:
> Heh, clearly a cold rainy weekend 😁 I had a cursory look, and this looks great! Please re-ask for review once this goes green, i.e. the tests with a custom m.execute() timeout (not sure if there are others).

:bullettrain_side: to Berlin.  :)

lib/constants.py
```diff
@@ -18,7 +18,7 @@
 import os
 
 # Images which are OSTree based
-OSTREE_IMAGES = ["centos-9-bootc", "fedora-coreos", "rhel4edge"]
+OSTREE_IMAGES = ("centos-9-bootc", "fedora-coreos", "rhel4edge")
```
martinpitt, line 21: OOI, is that really a typing issue, or  a random unrelated change? I know you have a preference for tuples, but I think conceptually they are the wrong data type for this kind of use case: It is really intended that all entries have the same type, which isn't the case for tuples.

task/__init__.py
```diff
@@ -312,7 +322,7 @@ def branch(context, message, pathspec=".", issue=None, push=True, branch=None, d
     return branch
 
 
-def pull(branch, body=None, issue=None, base=None, labels=['bot'], run_tests=True, dry: bool = False, **kwargs):
+def pull(branch, body=None, issue=None, base=None, labels=('bot',), run_tests=True, dry=False, **kwargs):
```
martinpitt, line 325: Here as well -- this is really conceptually a list, not a tuple/record. REST/JSON don't even have a concept of a tuple. (Also, it's bad practice to have non-scalar default values anyway, these should be `None` and set inside the function)

task/github.py
```diff
@@ -286,15 +286,15 @@ def get(self, resource=None):
             self.cache.write(qualified, response)
             return json.loads(response['data'] or "null")
 
-    def post(self, resource, data, accept=[]):
+    def post(self, resource, data, accept=()):
```
martinpitt, line 289: And again, this is a string list, not a tuple.. What does ruff complain about here?

github-upload-secrets
```diff
@@ -98,7 +98,7 @@ def upload_secrets(
     for name, secret in secrets.items():
```
martinpitt, line 98: The "ruff: enable preview mode" commit says "run ruff as mypy", it certainly means "ruff and mypy"

martinpitt (commented):
Thanks! This all looks pretty good, I just don't like the list → tuple conversions. I'd like to understand the motivation for these, and whether this is a bug in mypy/ruff somewhere?

machine/machine_core/constants.py
```diff
@@ -1,22 +0,0 @@
-# Backwards compatibility shim.  This will go away soon.
```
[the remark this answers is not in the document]
allisonkarlitskaya: I queued up a full slate of rhel-8 runs, but didn't check rhel-7...

lib/constants.py
allisonkarlitskaya: You can view this from two sides, I think.

The first is from the dynamic/pragmatic/implementation side.  In that case, a tuple is essentially exactly the same as an array except that:
 - all of the methods for modifying it are removed
 - `[] + ()` doesn't produce the same result as `[] + []` (since Python doesn't know what type to Pick, but you should probably write that as `[*[], *()] anyway.
Also, if you look around online for "immutable list Python" everyone says "use a tuple".  This is pervasive practice.

The second side is from the static/type-theoretical standpoint, which I think is the one you're arguing.  I agree that tuples and list are not semantically the same thing, and this part of it also looks odd to my eye, but Python has a very strong tradition of using tuples also as immutable lists.  Since the beginning it was always valid to write `Tuple[str, ...]` to mean "immutable list of strings" (and now, of course, it's also valid to write `tuple[str, ...]`).

Would you feel better if we annotated that as `tuple[str, ...]`?  We could also type it as a `Sequence` or `Collection`...  We could also change it to a `frozenset`.  For me, the main concern is immutability.

We've had actual bugs in our code before where someone does something like:

```python
  myvalue = SOME_CONSTANT
  if condition:
    myvalue += ['foo']
```

without realizing that they're modifying `SOME_CONSTANT`...

task/__init__.py
allisonkarlitskaya: The bad practice is not about non-scalar value — it's about non-immutable values, and indeed, you've hit upon exactly the reason it's better to write these as `()` instead of `[]` (and this is exactly the thing that ruff complains about).

Whether its better to use an immutable non-scalar type or something like `None` is more of a matter of opinion, but I prefer this style, since it's more succinct and we don't have to have an extra check for it later.

It conceptually checks too — the default is that we want a bots label.

Incidentally, there's another way: if you type the paramter with an immutable type (like `Sequence`) then even if you use a list, ruff will consider that the list is sufficiently protected against modification.  That would require annotating this function, though, which I'm trying to avoid doing at present...

All around, I prefer to geh auf Nummer sicher and use the `()`.

martinpitt (approved):
Ack, thanks for the explanation!

PR #6265, opened 2024-04-24
strict typing for `task.github`

allisonkarlitskaya:


task/github.py
```diff
@@ -129,12 +130,14 @@ def __init__(
...
         remote: str | None = None
     ):
-        self._remote = remote
-        self._repo = repo
-        self._base = base
-        self._url = None
-
-        self.conn = None
+        if repo is not None:
+            self.repo = repo
```
martinpitt, line 144: So do I understand this right? By default, `self.repo` is a `@property` method, but this conditionally replaces that method with a static value? That seems quite evil.. Duck typing wise it's fine of course.

TBH I'd prefer if that was a lot simpler: The constructor (this  __init__) could compute the default values and the accessors could go away completely. Like
```
self.repo = repo or os.environ.get("GITHUB_BASE", None) or get_repo()
``

Comment (allisonkarlitskaya) [reply task/github.py:144]: It *is* a bit evil, but it's designed to work this way: https://docs.python.org/3/library/functools.html#functools.cached_property

> The mechanics of [cached_property()](https://docs.python.org/3/library/functools.html#functools.cached_property) are somewhat different from [property()](https://docs.python.org/3/library/functions.html#property). A regular property blocks attribute writes unless a setter is defined. In contrast, a cached_property allows writes.

and even, the way `@cached_property` works internally is actually exactly the same:

> The cached_property decorator only runs on lookups and only when an attribute of the same name doesn’t exist. When it does run, the cached_property writes to the attribute with the same name. Subsequent attribute reads and writes take precedence over the cached_property method and it works like a normal attribute.

It's all a bit magic, but hey, this is Python... I figure if it's good enough for the standard library, then it's good enough for us.

The reason I'd lean against computing the values at the outset is the same reason they're cached in the first place: they're expensive, and sometimes involve shelling out.  That sort of thing doesn't belong in a constructor...

Comment (martinpitt) [reply task/github.py:144]: Why not? In the ctor it happens exactly once, as opposed to calling `api.repo` multiple times. The total cost is the same, just errors would happen earlier (which in general seems like a good thing).

Not a strict veto, but this all seems unnecessarily complicated.

Comment (martinpitt) [reply task/github.py:144]: Very much disagree about "extremely simple", it's the opposite IMHO. Just a single static assignment of `self.foo = ` in the ctor is as easy as possible. Adding properties, caches, etc. is making things *less* simple, it's optimization which I find questionable here.

But anyway, it seems you have a strong opinion about it, so I'll STFU.

**Review by martinpitt — APPROVED** [review]

**Comment by jelly** [comment]
FYI: I still get:
```
[jelle@t14s][~/projects/cockpit-bots]%mypy .
Found 1 error in 1 file (checked 44 source files)
```

PR #6266, opened 2024-04-24
tests-trigger, tests-scan: Add typing

martinpitt:
.. to get my feet wet. After a round of learning/reviews with @allisonkarlitskaya I'm happy to do some more in between waiting for test runs :grin: 

 - [x] Builds on top of PR #6265

### DIFF
--- a/task/__init__.py
+++ b/task/__init__.py
@@ -33,7 +33,7 @@ import traceback
 from collections.abc import Sequence
 from datetime import datetime, timezone
 
-from lib.aio.jsonutil import JsonObject, get_str, typechecked
+from lib.aio.jsonutil import JsonObject, get_dictv, get_str, typechecked
 from lib.constants import BASE_DIR
 
 from . import github
@@ -400,10 +400,13 @@ def label(issue, labels=('bot',)):
     return api.post(resource, labels)
 
 
-def labels_of_pull(pull):
-    if "labels" not in pull:
-        pull["labels"] = api.get(f"issues/{pull['number']}/labels")
-    return [label["name"] for label in pull["labels"]]
+def labels_of_pull(pull: JsonObject) -> Sequence[str]:
+    if "labels" in pull:
+        labels = get_dictv(pull, "labels")
+    else:
+        labels = api.get_objv(f"issues/{pull['number']}/labels")
+
+    return [get_str(label, "name") for label in labels]
 
 
 def comment(issue, comment, dry=False):
--- a/tests-scan
+++ b/tests-scan
@@ -22,7 +22,7 @@ import json
 import logging
 import sys
 import time
-from collections.abc import Iterable, Mapping, Sequence
+from collections.abc import Collection, Iterable, Mapping, Sequence
 from typing import Any, Required, TypedDict
 
 from lib import ALLOWLIST, testmap
@@ -38,12 +38,14 @@ try:
 except ImportError:
     no_amqp = True
 
+Policy = Mapping[str, Sequence[str]]
 
-def build_policy(repo, context):
+
+def build_policy(repo: str, requested_contexts: Collection[str]) -> Policy:
     policy = testmap.tests_for_project(repo)
-    if context:
+    if requested_contexts:
         short_contexts = []
-        for c in context:
+        for c in requested_contexts:
             short_contexts.append(c.split("@")[0])
         new_policy = {}
         for (branch, contexts) in policy.items():
@@ -78,7 +80,7 @@ class QueueEntry(TypedDict):
     human: str
 
 
-def is_internal_context(context):
+def is_internal_context(context: str) -> bool:
     # HACK: CentOS CI doesn't like firefox, browser fails to start way too often
     for pattern in ["rhel", "edge", "vmware", "openstack", "/firefox"]:
         if pattern in context:
--- a/tests-trigger
+++ b/tests-trigger
@@ -5,16 +5,17 @@ import subprocess
 import sys
 
 from lib import ALLOWLIST, testmap
+from lib.aio.jsonutil import get_str
 from task import github
 
 sys.dont_write_bytecode = True
 
 
-def git(*args):
+def git(*args: str) -> str:
     return subprocess.check_output(('git', *args), encoding='utf-8').strip()
 
 
-def trigger_pull(api, opts):
+def trigger_pull(api: github.GitHub, opts: argparse.Namespace) -> int:
     if opts.target != '-':
         pull = api.get("pulls/" + opts.target)
 
@@ -37,7 +38,7 @@ def trigger_pull(api, opts):
 
     statuses = api.statuses(revision)
     if opts.context:
-        contexts = set()
+        contexts = set[str]()
         for cntx in opts.context:
             if cntx.startswith("image:"):
                 contexts.update(testmap.tests_for_image(cntx.split(':', 1)[1]))
@@ -75,7 +76,7 @@ def trigger_pull(api, opts):
         changes = {"state": "pending", "description": github.NOT_TESTED_DIRECT, "context": context}
 
         # Keep the old link for reference, until testing starts again
-        link = status.get("target_url", None)
+        link = get_str(status, "target_url", None)
         if link:
             changes["target_url"] = link
 
@@ -84,7 +85,7 @@ def trigger_pull(api, opts):
     return ret
 
 
-def main():
+def main() -> int:
     parser = argparse.ArgumentParser(description='Manually trigger CI Robots')
     parser.add_argument('-f', '--force', action="store_true",
                         help="Force setting the status even if the program logic thinks it shouldn't be done")
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,6 @@ module = [
     'tests-policy',
     'tests-scan',
     'tests-status',
-    'tests-trigger',
 ]
 
 [tool.ruff]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,6 @@ module = [
     'tasks-container-update',
     'test-failure-policy',
     'tests-policy',
-    'tests-scan',
     'tests-status',
 ]
 


tests-trigger
```diff
@@ -77,14 +77,15 @@ def trigger_pull(api, opts):
         # Keep the old link for reference, until testing starts again
         link = status.get("target_url", None)
         if link:
+            assert isinstance(link, str)
```
martinpitt, line 80: Without this:

> tests-trigger:80: error: Incompatible types in assignment (expression has type "Mapping[str, JsonValue] | Sequence[JsonValue] | float | bool", target has type "str")  [assignment]

This feels a bit strange to me -- is there a `| str` missing there?

allisonkarlitskaya: I don't see that warning.  I only get:

> tests-trigger:19: error: Call to untyped function "get" in typed context  [no-untyped-call]
> tests-trigger:38: error: Call to untyped function "statuses" in typed context  [no-untyped-call]
> tests-trigger:83: error: Call to untyped function "post" in typed context  [no-untyped-call]

(and you'll need to wait on #6265 for those...)

allisonkarlitskaya: Ah.  I fetched your branch approximately 30 seconds too soon.  Now I understand :)

allisonkarlitskaya: The reason for this is pretty simple, and it lies with the line above:

```python
changes = {"state": "pending", "description": github.NOT_TESTED_DIRECT, "context": context}
```
in that line, all the attributes that you add to the dictionary happen to have string values, so mypy deduces a type of `dict[str, str]`.  You can add `changes: dict[str, JsonValue] =` to help it.

allisonkarlitskaya: Another out would be to replace `link = status.get("target_url", None)` with `link = get_str(status, "target_url", None)`

allisonkarlitskaya: ...and yet another out would be for us to define a TypedDict subclass for status objects... but that's a whole other can of worms...

allisonkarlitskaya: actually, I don't like that one.  There's a bunch of other junk that GitHub returns in that object that we shouldn't be copying into the new one...

allisonkarlitskaya: The issue is that `status` came from fetching a JSON blob from GitHub.  The API enforces that it's an object, but doesn't do anything else.  That means that if you `.get()` something out of it, it could be any possible JSON value — not just a string, and that's what mypy is complaining about here.  That's why the assert (or `get_str()`) helps.

task/__init__.py
```diff
@@ -400,10 +400,11 @@ def label(issue, labels=('bot',)):
...
 
-def labels_of_pull(pull):
-    if "labels" not in pull:
-        pull["labels"] = api.get(f"issues/{pull['number']}/labels")
-    return [label["name"] for label in pull["labels"]]
+def labels_of_pull(pull: JsonObject) -> Sequence[str]:
+    if "labels" in pull:
+        return [get_str(label, "name") for label in get_dictv(pull, "labels")]
+    else:
+        return api.get(f"issues/{pull['number']}/labels")
```
allisonkarlitskaya, line 407: This looks suspicious.  I think this is going to end up returning a sequence of JSON objects instead of a sequence of strings.

Consider that the old code always did the unpacking step, also on the result of `.get()`.

The reason this didn't raise alarm bells is because `.get()` returns `Any` (ie: is unsafe).  Please try to use `.get_objv()` when fixing this.

tests-scan
```diff
@@ -38,12 +38,14 @@ try:
 except ImportError:
     no_amqp = True
 
+Policy = Mapping[str, Sequence[str]]
 
-def build_policy(repo, context):
+
+def build_policy(repo: str, requested_contexts: Iterable[str]) -> Policy:
```
allisonkarlitskaya, line 44: The use of `Iterable` looks suspicious here because you can't always do a truth compare on `Iterable` to find out of its empty (which the code does, below).

Probably you want `Collection`, `Sequence` would also likely be fine, but it's overkill.

In general, I try to avoid `Iterable` or `Iterator` on interfaces, because it means that the value you receive might possibly be destroyed by your interaction with it.  Note specifically that an `Iterator` is `Iterable` (which is the problem).  `Collection` is the type which means "durable collection of things that I can iterate over".

This stuff is pretty opaque, so a quick primer on some confusingly-overlapping abstract types, from strict to loose:

 - `Sequence`: implements `in`, has a size, has a specific order, and can be indexed.  This is basically a `tuple` or a `list`.  Only use this if the order actual matters.
 - `Container`: implements `in`, has a size, but no specific order and cannot be indexed.  `tuple` and `list` are `Container`, but `set` is too (whereas `set` is not a `Sequence`).  This one tends to be the one you want most of the time when the exact order isn't important.
 - `Collection`: implements `in` but is not iterable.  Not super useful.

run-queue
```diff
@@ -68,10 +72,10 @@ def consume_webhook_queue(channel, amqp):
         elif action == 'closed' and pull_request.get('merged', False):
             sha = pull_request['head']['sha']
             db = os.path.join(get_images_data_dir(), "test-results.db")
-            cmd = (f"./store-tests --db {db} --repo {repo} {sha} && "
-                   f"./prometheus-stats --db {db} --s3 {os.path.join(LOG_STORE, 'prometheus')}")
+            stat_cmd = (f"./store-tests --db {db} --repo {repo} {sha} && "
+                        f"./prometheus-stats --db {db} --s3 {os.path.join(LOG_STORE, 'prometheus')}")
             body = {
-                "command": cmd,
+                "command": stat_cmd,
```
allisonkarlitskaya, line 78: imho this is begging to be rolled out into

```python
body = {
    "command": fr"""set -e
      ./store-tests ...
      ./prometheus-stats ...
    """
}
```

but that's Geschmacksache...

task/distributed_queue.py
```diff
@@ -65,6 +66,15 @@
 }
 
 
+class PikaMethod:
+    message_count: int
+
+
+class DeclareResult:
+    # this is pika.spec.Basic.DeclareOk but that isn't exported
```
allisonkarlitskaya, line 74: I think this is `pika.spec.Queue.DeclareOk`, no?

allisonkarlitskaya (changes requested):

task/distributed_queue.py
martinpitt: Ah, right! Got confused by https://pika.readthedocs.io/en/stable/_modules/pika/spec.html#Queue.DeclareOk . Queue.* works, thanks!

task/__init__.py
martinpitt: Erk, yes, what nonsense.. Fixed, thanks!

tests-scan
martinpitt: Thanks for the primer! Collection and Container don't work here, as neither is iterable (and we use a `for` loop over it). As this isn't public API, I went with `Sequence` which is both fixed and iterable, the "overkill" doesn't matter here (it's basically a `list` and nothing else..)

allisonkarlitskaya: hilariously, I flipped `Collection` and `Container` in my own primer :facepalm: .  `Collection` is indeed iterable.

task/distributed_queue.py
```diff
@@ -65,6 +66,10 @@
 }
 
 
+class DeclareResult:
```
allisonkarlitskaya, line 69: This still isn't setting well with me.

I looked into this a bit more, and the actual type of the thing here is a `pika.frame.Method`, but as implemented in `pika`, this isn't marked `Generic` so we have no way to instantiate it like `pika.frame.Method[pika.spec.Queue.DeclareOk]`.

There is, however, https://pypi.org/project/types-pika/ which does have a `Generic` `pika.frame.Method`: https://github.com/baod-rate/types-pika/blob/ccde5ce04818c1e678ff46f701203cd5322f2683/pika-stubs/frame.pyi#L21

I'm not sure it's worth including that as a dependency, though, and I'm not sure mypy would correctly deal with the case that the base library was present and the type annotations were missing.

The problem with what you're doing here is that PEP484 is nominal typing and you're trying to do structural typing here.  I think I'd feel a lot better about this if you subclassed it from `typing.Protocol`: https://docs.python.org/3/library/typing.html#typing.Protocol

task/distributed_queue.py
```diff
@@ -82,7 +87,7 @@ def __init__(
         any extra arguments in **kwargs will be passed to queue_declare()
 
         the results of the result declarations are stored in
-        DistributedQueue.declare_results, a dict mapping queue name to result
+        DistributedQueue.declare_results, a dict mapping queue name to DeclareResult
```
allisonkarlitskaya, line 90: The fact that `DeclareResult` only appears in comments in this module strikes me as a bit strange.  It probably makes sense to annotate `declare_results` as something better than `dict[str, Any]`...

allisonkarlitskaya (changes requested):

tests-trigger
```diff
@@ -37,7 +38,7 @@ def trigger_pull(api, opts):
 
     statuses = api.statuses(revision)
     if opts.context:
-        contexts = set()
+        contexts: set[str] = set()
```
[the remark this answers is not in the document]
martinpitt: That looks awesome, but doesn't seem to work?
```
contexts: set[str]() # E: Invalid type comment or annotation  [valid-type]
```

tests-scan
```diff
@@ -38,12 +38,14 @@ try:
 except ImportError:
     no_amqp = True
 
+Policy = Mapping[str, Sequence[str]]
```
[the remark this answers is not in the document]
martinpitt: Yeah, I expected it to be used more often, but it seems declaring the type at the start is enough for mypy to track it through everything else. I'll keep or drop it as you recommend, I don't have a good gut feeling about this style yet.

run-queue
```diff
@@ -39,10 +39,14 @@ logging.basicConfig(level=logging.INFO)
 
 statistics_queue = os.environ.get("RUN_STATISTICS_QUEUE")
 
-# Returns a command to execute and the delivery tag needed to ack the message
+# as per pika docs
+DeliveryTag = int
```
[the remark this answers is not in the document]
martinpitt: Cool, thanks! From our POV this is pretty much an opaque type -- we receive a tag and pass it on. It's not even important that it is an `int`, we should make no assumptions about it. Would `Any` actually be better here? Or is there something like `Opaque`?

run-queue
```diff
@@ -39,10 +39,14 @@ logging.basicConfig(level=logging.INFO)
...
 
 
-def consume_webhook_queue(channel, amqp):
+# Returns a command argv to execute and the delivery tag needed to ack the message
+def consume_webhook_queue(
```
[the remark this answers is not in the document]
martinpitt: Roger!

tests-trigger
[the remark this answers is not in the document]
allisonkarlitskaya: `s/:/=/`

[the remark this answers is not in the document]
martinpitt: D'oh! "Don't do thinks while being in a meeting" :see_no_evil:

run-queue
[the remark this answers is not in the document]
allisonkarlitskaya: `NewType` is something like a one-way opaque type.  This is Python though — it's an extremely dynamic language and you can't actually ever definitively hide anything, and I don't know of a purely static construct for representing this idea.  At the end of the day, you have an object, and that object is of a particular dynamic type.

The closest you could do is have some kind of a wrapper object, which shields the actual data as a private instance variable.  You could then teach this object how to `basic_ack()` the tag.  I don't think you want to do this though...

For what it's worth, I just checked the pika types module, and it just uses `int`.  I think that's probably the most reasonable thing to do here — if pika ever adds types, it'll probably do it in a way that's compatible with that module.  The less stuff we "invent" here, the better.

run-queue
```diff
@@ -39,10 +40,15 @@ logging.basicConfig(level=logging.INFO)
...
+# as per pika docs
+DeliveryTag = NewType('DeliveryTag', int)
 
 
-def consume_webhook_queue(channel, amqp):
+# Returns a command argv to execute and the delivery tag needed to ack the message
+def consume_webhook_queue(
+    channel: pika.channel.Channel,
+    amqp: str
+) -> tuple[list[str] | None, DeliveryTag | None]:
```
allisonkarlitskaya, line 51: Sorry for not catching this before: does this actually have to be `list` or could it be `Sequence`?

If the intent is not for the caller to modify the returned data, `Sequence` is more appropriate.

tests-scan
martinpitt: Heh -- Collection it is then. Utterly unimportant here, but a good learning exercise.

run-queue
martinpitt: Indeed, here we have a case where order mattes, iteration matters, but the value should be read-only. So as per your list above it should be Sequence.

task/distributed_queue.py
martinpitt: Good point! Done.

task/distributed_queue.py
martinpitt: Wrt. your chat question:

> did you fail to get the type for the message_count field without doing the protocol for yourself? it seems like the kind of thing that mypy might have succeeded at scraping out of the source, even without proper types

I tried this:
```diff
-- task/distributed_queue.py
+++ task/distributed_queue.py
@@ -23,7 +23,7 @@ import os
 import ssl
 from collections.abc import Collection
 from types import TracebackType
-from typing import Any, Protocol, Self
+from typing import Any, Protocol, Self, TypeVar
 
 no_amqp = False
 try:
@@ -66,14 +66,13 @@ arguments = {
 }
 
 
-# conceptually these two are https://pika.readthedocs.io/en/stable/_modules/pika/spec.html#Queue.DeclareOk
+# conceptually this is https://pika.readthedocs.io/en/stable/_modules/pika/spec.html#Queue.DeclareOk
 # but we don't want to depend on types-pika, and are also only interested in one field
-class _PikaMethod(Protocol):
-    message_count: int
+T = TypeVar("T")
 
 
-class DeclareResult(Protocol):
-    method: _PikaMethod
+class DeclareResult(Protocol[T]):
+    method: T
```

and it doesn't work:
```
task/distributed_queue.py:130: error: Missing type parameters for generic type "DeclareResult"  [type-arg]
run-queue:106: error: Missing type parameters for generic type "DeclareResult"  [type-arg]
run-queue:107: error: Missing type parameters for generic type "DeclareResult"  [type-arg]
run-queue:108: error: Missing type parameters for generic type "DeclareResult"  [type-arg]
```

i.e. maybe I can specify some instance of T at the declaration, but IMHO that would be besides the point. So you most likely meant something else, but wihtout specifying `_PikaMethod`, what would `method`'s type be?

martinpitt: This is all awkward, and it's also pointed out by the impedance mismatch in the [test failure](https://github.com/cockpit-project/bots/actions/runs/8818339405/job/24206896506?pr=6266). Some things like `run-queue` just blindly assume that all queues exist, but `DistributedQueue` cannot ensure that. This needs more thinking and bigger redesign. I'll back out the top (run-queue) commit for now, and work on that separately.

martinpitt:
@allisonkarlitskaya Green now, let's land this separately. PTAL?

task/distributed_queue.py
allisonkarlitskaya: The correct type, if we could, would be 
`pika.frame.Method[pika.spec.Queue.DeclareOk]`.

Of course, we can't.  So, first attempt:

```python
from typing import Protocol, reveal_type

import pika.spec


class DeclareFrame(Protocol):
    method: pika.spec.Queue.DeclareOk


y: DeclareFrame = None  # type: ignore[assignment]
reveal_type(y.method.message_count)
```

pyright is smart enough to figure out that this is `int`, but mypy says `Any`.  So much for that.

So I'd go with this instead:

```python
from typing import Protocol, reveal_type


class DeclareOk(Protocol):
    message_count: int


class DeclareFrame(Protocol):
    method: DeclareOk


y: DeclareFrame = None  # type: ignore[assignment]
reveal_type(y.method.message_count)
```

and then they both agree that this is an `int`.

tests-trigger
```diff
@@ -75,7 +76,7 @@ def trigger_pull(api, opts):
         changes = {"state": "pending", "description": github.NOT_TESTED_DIRECT, "context": context}
 
         # Keep the old link for reference, until testing starts again
-        link = status.get("target_url", None)
+        link = get_str(status, "target_url", None)
```
allisonkarlitskaya, line 79: mental note: we're allowed to have walruses now `:=`

allisonkarlitskaya (approved):
Look good, thanks! :)

tests-trigger
martinpitt: Thanks for pointing out! done in #6278

PR #6267, opened 2024-04-24
Assorted ruff / documentation fixes

jelly:
This is a SyntaxError in Python, spotted by ruff 0.4 and introduced by a ruff autofix in a0223ff82.

### DIFF
--- a/tests-policy
+++ b/tests-policy
@@ -249,7 +249,8 @@ def update_known_issue(api, number, err, details, context, timestamp=None):
                         # only keep the last 10
                         if len(occurrences) > 10:
                             occurrences.pop(0)
-                        parts[part_idx] = f"{latest[0]}{latest_occurrences}{'\n'.join(occurrences)}"
+                        newline = '\n'
+                        parts[part_idx] = f"{latest[0]}{latest_occurrences}{newline.join(occurrences)}"
                         updated = True
                     break
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -31,8 +31,13 @@ There are static code and syntax checks which you should run often:
 
     $ test/run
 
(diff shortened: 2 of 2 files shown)


jelly:
FYI: from test/run I now only get:
```
[jelle@t14s][~/projects/cockpit-bots]%./test/run
```

martinpitt (approved):
Cheers! While github is broken, I ran the integration tests manually with `COCKPIT_BOTS_BRANCH=ruff-new pytest`.

tests-policy
```diff
@@ -249,7 +249,8 @@ def update_known_issue(api, number, err, details, context, timestamp=None):
                         # only keep the last 10
                         if len(occurrences) > 10:
                             occurrences.pop(0)
-                        parts[part_idx] = f"{latest[0]}{latest_occurrences}{'\n'.join(occurrences)}"
```
allisonkarlitskaya, line 252: This change looks weird...

martinpitt: Yeah, I'd also prefer to just build the whole `"\n".join()` as a separate value, but it felt too much like bikeshedding. We have lots of things in flight today..

jelly: I would be happy to do that honestly. Was a bit suprised it was merged :)

allisonkarlitskaya: Oh.

```
>>> f'{"\n".join(["a", "b"])}'
'a\nb'
```

I see.....

allisonkarlitskaya: So this came in relatively recently, in a0223ff82431603b6926569ea8d8165b47e5566a.  And indeed, the syntax is new in Python 3.12.

allisonkarlitskaya: See https://github.com/cockpit-project/bots/pull/6269

PR #6269, opened 2024-04-24
tell ruff that we target Python 3.12

allisonkarlitskaya:


pyproject.toml
```diff
@@ -61,6 +61,7 @@ exclude = [
 ]
 line-length = 118
 preview = true
+target-version = 'py312'
```
jelly, line 64: This feels a bit meh, yet another thing to keep track of in X projects. I'm updating to 3.12 this weekend so let's not do this :)

jelly: Ah, then maybe I like it more! Sounds good to me.

martinpitt: @jelly is your objection that 3.12 is too new? I could definitively understand that, e.g. 3.12 isn't even in Debian unstable yet, and probably not in RHEL 9 either. 3.11 seems like a saner choice to me indeed.

Or is your objection that declaration by itself? (That seems fine to me -- better explicit than implicit, as long as some type checker checks it)

jelly: I'll update on Friday and then have 3.12 so that is fine. My objection would be if this was required in all our projects which was a misunderstanding. Because then we have to update files, cockpit, etc. But that is not the case.

allisonkarlitskaya: There's quite a few things scattered around the bots which are 3.12-only, at this point (another that comes to mind is `@override`) so I think we probably don't want to go back on that...

Part of the effort of always running inside of the tasks container was that we'd be able to target a single Python version without having to worry about testing stuff against older ones.

I'd accept an argument that we should be slower about embracing new stuff in the future (ie: hold off on 3.13 stuff) but the way https://github.com/cockpit-project/bots/commit/a0223ff82431603b6926569ea8d8165b47e5566a happened by accident is sort of a demonstration of the difficulty of that approach, unless we get some sort of a tox setup going here to verify that old versions are working (and I'm against putting in the effort for that)...

martinpitt: Fine with me. We'll see if someone runs into trouble, and we can take it a bit slower with 3.13 indeed.

martinpitt (approved):
(but waiting for Jelle's review as well)

jelly (approved):

allisonkarlitskaya:
(straight-rebase force-push because GitHub got stuck trying to compute the merge)

PR #6116, opened 2024-03-20
images: Add rhel-10-0

mvollmer:
* [x] image-refresh rhel-10-0

PR #6270, opened 2024-04-24
Fedora 40 final

jelly:
* Tested the fedora-40.bootstrap locally. I don't want to refresh it yet. (Rather after a release)
* Rhel.setup is tested in the RHEL 10 PR https://github.com/cockpit-project/bots/pull/6116

### DIFF
--- a/images/scripts/fedora-40.bootstrap
+++ b/images/scripts/fedora-40.bootstrap
@@ -19,7 +19,7 @@
 
 set -eux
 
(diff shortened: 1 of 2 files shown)


jelly:
Could have combined those two commits..

martinpitt (approved):
Thanks! I trust your local/rhel-10 testing, so if you want to land as-is, and let the bots do the work please go ahead. (Even if this PR *should* break, it'll break in an orderly and debuggable way)

PR #6272, opened 2024-04-24
a bunch more typing

allisonkarlitskaya:
Some nice light "after the kids are in bed" hacking... :)

...now down to ~3k lines of non-strictly-typed code.

This represents all the scripts that don't depend on `import task` (which is going to be a bit of a fight to figure out properly), except:
 - run-queue: @martinpitt is working on this
 - store-tests: this is big and scary

Another trivial pytest conversion in there, too.


allisonkarlitskaya:
For a rough idea of "what's left" and what the next "easy targets" might be, I removed all of the exceptions from `pyproject.toml` and did this:

```
$ test/run 2>&1 | cut -f1 -d: | uniq -c | sort -nr
     78 store-tests
     63 task/__init__.py
     52 test/test_tests_scan.py
     41 test/test_issue_scan.py
     29 test/test_task.py
     29 test/test_github.py
     24 npm-update
     23 po-refresh
     13 tests-status
     10 test/test_test_failure_policy.py
     10 naughty-prune
      9 run-queue
      9 image-trigger
      8 test/test_policy.py
      7 cockpit-lib-update
      3 image-refresh
      2 tasks-container-update
```

task/github.py
```diff
@@ -423,6 +423,19 @@ def issues(
             result += issues
         return result
 
+    def issue_comments(self, number: int) -> Sequence[JsonObject]:
```
martinpitt, line 426: Hmm, I wonder if this belongs into task/__init__.py rather? That has the higher-level abstractions, while github.py is the "basic" API. The separation isn't totally clear-cut of course.

martinpitt (approved):
Cheers!

PR #6302, opened 2024-04-26
testmap: Drop c8s

martinpitt:
CentOS 8 stream will go end of life in a month [1], and we will
discontinue our CI image soon. Most of the entries were in `_manual`
anyway. Move starter-kit to c9s instead.

Also drop the obsolete rhel-8-10 manual entry from Cockpit's main
branch.

[1] https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/


martinpitt:
@allisonkarlitskaya That'll come next. Need to check if we still use it anywhere.

PR #6303, opened 2024-04-26
task: Change ellipsis type check for mypy 1.9.0 compatibility

### LINKED ISSUE #6271: mypy 1.9.0 still complains about `task.github`
```
[jelle@t14s][~/projects/cockpit-bots]%mypy .
Found 1 error in 1 file (checked 44 source files)
```

martinpitt:
mypy 1.9.0 started failing on this:

> task/github.py:297: error: Incompatible return value type (got "_DT | EllipsisType", expected "_T | _DT")  [return-value]

which is nonsense -- the `default is not ...` explicitly ensures that we are not returning an ellipsis.

Rewrite this using a type instead of a value check, which works.

Fixes #6271

### DIFF
--- a/task/github.py
+++ b/task/github.py
@@ -293,7 +293,7 @@ class GitHub:
             elif modified:
                 headers['If-Modified-Since'] = modified
         response = self.request("GET", resource, "", headers)
-        if response['status'] == 404 and default is not ...:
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
Better than what I would have done!  Thanks!

PR #6279, opened 2024-04-25
images: Drop rhel-7-9 and centos-7

martinpitt:
RHEL 7.9 reaches end-of-maintenance at the end of June [1]. We haven't actually done any updates in years, at this point it's just dead weight.

Move testing of the services image to cockpit's RHEL 8 branch, which is still active.

https://github.com/cockpit-project/cockpit/pull/20368 dropped centos-7 usage in cockpit's tests (it doesn't use rhel-7-9).

[1] https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/rhel-7-end-of-maintenance

PR #6304, opened 2024-04-26
tests-policy: drop this old script

allisonkarlitskaya:
Now that we no longer run tests from the rhel-7.9 branch of Cockpit, nothing else is using this.

 - [x] #6279


martinpitt (changes requested):
test/test_policy.py and the pyproject.toml entry as well, please (I expect this to fail)

allisonkarlitskaya:
lol thanks

martinpitt (approved):
*flush* :toilet:

PR #6314, opened 2024-04-29
testmap: Introduce cockpit-machines rhel-8 branch

martinpitt:
Due to synced RHEL releases, this will also receive updates for RHEL 9.4.

---

I just pushed the branch. The very first PR against it will/has to fix CI, so I didn't bother with `_manual`.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -132,6 +132,11 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
             'rhel-9-4',
             'rhel-9-5',
         ],
+        'rhel8': [
+            'rhel-8-10',
+            'rhel-9-4',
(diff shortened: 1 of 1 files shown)


jelly (approved):

jelly:
But this is blocked on https://github.com/cockpit-project/cockpit-machines/pull/1599 no?

martinpitt:
I thought it was the other way around, as my test requests weren't picked up -- but it does work, the queue was just full. It's not really blocked, but if you prefer I can postpone landing and manually request the tests until then.

jelly:
Seems it went green, so let's merge it? :+1:

martinpitt:
There's one regression which I still need to work out, but we can merge this -- it'll save me from having to trigger these tests manually on each iteration.

PR #6320, opened 2024-04-30
pika cleanups

allisonkarlitskaya:


run-queue
```diff
@@ -99,35 +99,29 @@ def consume_webhook_queue(channel: pika.channel.Channel, amqp: str) -> ConsumeRe
...
 # Returns a command to execute and the delivery tag needed to ack the message
-def consume_task_queue(
-    channel: pika.channel.Channel,
-    amqp: str,
-    public_count: int,
-    rhel_count: int,
-    stats_count: int,
-) -> ConsumeResult:
-    if statistics_queue and stats_count > 0:
+def consume_task_queue(dq: distributed_queue.DistributedQueue) -> ConsumeResult:
```
martinpitt, line 102: That's a great simplification!

task/distributed_queue.py
```diff
@@ -105,14 +106,15 @@ def __init__(
...
+        self.address = amqp_server
         self.connection = pika.BlockingConnection(params)
         self.channel = self.connection.channel()
         self.queue_counts: dict[str, int] = {}
 
         for queue in queues:
             try:
                 result = self.channel.queue_declare(queue=queue, arguments=arguments.get(queue, None), **kwargs)
-                self.queue_counts[queue] = result.method.message_count
+                self.queue_counts[queue] = result.method.message_count or 0
```
martinpitt, line 117: I that `or 0` necessary? It wasn't before, and it really shouldn't be `None`. If so, this requires an explanation (possibly in a separate commit?)

martinpitt (changes requested):
Thanks! Small issue, and somehow this breaks inspect-queue (see failing integration test). Reminder: `COCKPIT_BOTS_BRANCH=pika-pika pytest -k test_queue`

inspect-queue
```diff
@@ -29,6 +30,7 @@ def main() -> int:
     parser = argparse.ArgumentParser(description='Read and print messages from the queue without acknowleding them')
     parser.add_argument('--amqp', default=distributed_queue.DEFAULT_AMQP_SERVER,
                         help='The host:port of the AMQP server to consume from (default: %(default)s)')
+    parser.add_argument('--human', '-h', help="Print the 'human' field, if present")
```
martinpitt, line 33: So that `-h` is what breaks inspect-queue:
```
argparse.ArgumentError: argument --human/-h: conflicting option string: -h
```

task/distributed_queue.py
allisonkarlitskaya: So according to the spec:
 - https://www.rabbitmq.com/amqp-0-9-1-reference#queue.declare-ok.message-count
 - https://www.rabbitmq.com/amqp-0-9-1-reference#domain.message-count

There is no mention that this might be nullable.

Also, looking at the code:

```python
    class DeclareOk(amqp_object.Method):
        ...
        def decode(self, encoded, offset=0):
            ...
            self.message_count = struct.unpack_from('>I', encoded, offset)[0]
            ...
```

Which is to say, that no `DeclareOk` created from calling `.decode()` is ever going to have this field as `None`, but if `.decode()` hasn't been called, it might be.  This looks like a case of bad binding...

I feel like an assert is probably more appropriate here, indeed.

task/distributed_queue.py
```diff
@@ -26,6 +26,7 @@
 from typing import Any, Self
 
 import pika
+import pika.exceptions
```
martinpitt, line 29: Why is that necessary? It's not being used anywhere?

martinpitt (approved):
Thanks! One question, but as ruff is happy..

task/distributed_queue.py
allisonkarlitskaya: The `pika` module doesn't have anything called "exceptions" on it, but we say:

```python
            except pika.exceptions.ChannelClosedByBroker as e:
```

to be honest I'm not sure why that works without the explicit import.  Python's import system is a bit magic for my tastes...

martinpitt: Thanks! Ah, so it wasn't new in this PR, just fixing a complaint. Makes sense!

PR #6321, opened 2024-04-30
tests-scan: remove extra period in --help output

allisonkarlitskaya:
None of the other options have dots.

This is definitely going to need a full CI run to check that it doesn't introduce any regressions.


martinpitt (approved):
Lol. I think I know pretty much exactly what this is for, but I just sent the real failure reason to #6320

PR #6334, opened 2024-05-02
issue-scan: fix issues with empty bodies

### LINKED ISSUE #6333: issue-scan crashes
issue-scan is currently crashing when scanning issues with empty bodies.

This is easy to reproduce locally without any extra secrets or permissions:

```
$ ./issue-scan --repo cockpit-project/starter-kit
...
  File "/var/home/lis/src/bots/main/./issue-scan", line 110, in tasks_for_issues
    checklist = github.Checklist(get_str(issue, "body"))
                                 ^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/lis/src/bots/main/lib/aio/jsonutil.py", line 80, in get_str
    return _get(obj, lambda v: typechecked(v, str), key, default)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/lis/src/bots/main/lib/aio/jsonutil.py", line 68, in _get
    raise JsonError(obj, f"{target} {exc!s}") from exc
lib.aio.jsonutil.JsonError: attribute 'body': must have type str
```

Looking at the API:

https://api.github.com/repos/cockpit-project/starter-kit/issues/859

we can see that the `body` field is indeed `null` here.  That seems to be what github returns for an empty body.  That should never have worked: even without the typecheck, the function that receives this argument calls `.splitlines()` on it directly, which would have also crashed with `None`.

In any case, the fix is probably pretty simple: we can accept `None` as a valid value, but we either need to exit the loop in that case, or have an `or ''` clause to treat `None` as "empty body".

allisonkarlitskaya:
Fixes #6333

### DIFF
--- a/issue-scan
+++ b/issue-scan
@@ -106,7 +106,7 @@ def tasks_for_issues(issues_data: str, opts_repo: str | None) -> Sequence[tuple[
         # the bots can do, then the bots will ignore this issue
         # (below in output_task)
         #
-        checklist = github.Checklist(get_str(issue, "body"))
+        checklist = github.Checklist(get_str(issue, "body", None))
         for item, checked in checklist.items.items():
             if not checked:
                 results.append((item, issue, api.repo))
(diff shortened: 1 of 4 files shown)


martinpitt (approved):
Cheers!

PR #6348, opened 2024-05-07
images: Drop centos-8-stream

martinpitt:
Commit bc26f65a434 already dropped (almost) all testmap entries, as CentOS 8 Stream goes EOL at the end of this month. So there's no point any more in refreshing and keeping it. Further RHEL 8 investigations should happen directly on the rhel-8-10 image.

Keep the naughty symlink though, as at least for now we are still running some c8s tests through packit.


allisonkarlitskaya (approved):
Goodbye old friend.

PR #6349, opened 2024-05-07
images: Keep cloud-init on alpine

martinpitt:
We want to keep cloud-init so that we can actually test c-machine's
user/password/SSH setup in the image creation dialog. That will create a
custom cloud-init iso.

This also gets us closer to running the tests against an off-the-shelf
upstream (i.e. not cockpit bots) Alpine image.

 * [x] image-refresh alpine

-----

Motivation: Better test coverage of UEFI and aarch64 (in subsequent PRs). This is mostly just ensuring that there are no  regressions.


cockpituous:
image-refresh alpine done: https://github.com/cockpit-project/bots/commits/image-refresh-alpine-20240507-121253

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-alpine-e46492e1-20240507-121117/log.html

mvollmer:
> We want to keep cloud-init so that we can actually test c-machine's
> user/password/SSH setup in the image creation dialog. That will create a
> custom cloud-init iso.

Oh, did we not run these tests anymore after switching to Alpine?

mvollmer (approved):
Thanks!

martinpitt:
> Oh, did we not run these tests anymore after switching to Alpine?

AFAIK we never tested the guest, i.e. that we can *actually* connect to the guest with the provided user/password/key.

mvollmer:
> > Oh, did we not run these tests anymore after switching to Alpine?
> 
> AFAIK we never tested the guest, i.e. that we can _actually_ connect to the guest with the provided user/password/key.

Ah, I see.

PR #6350, opened 2024-05-07
images: fix quoting in candlepin container setup

allisonkarlitskaya:
We execute a here-document in the container, with a ``-expansion in it which is clearly meant to be running inside of the container, but is actually running on the host (where it fails).

Change it to a $() expansion and escape it.

 * [x] image-refresh services


martinpitt:
Cheers! Obviously "good", I'll give formal after you pushed the image and undrafted.

images/scripts/services.setup
```diff
@@ -126,7 +126,7 @@ systemctl stop tomcat
 # regenerate the certs used by Candlepin; this is done so the hostname
 # of the certificate is actually this container's hostname (which actually
 # is the hostname of the system, because of --uts=host)
-~candlepin/devel/candlepin/bin/deployment/gen_certs.sh --force --trust --hostname `hostname -f`
```
martinpitt, line 129: FTR, not that "obvious", I actually did mean this to run on the host, as the container often has some unpredictable/random name; I wanted this to have a sensible name "services". Perhaps `hostnamectl hostname`?

cockpituous:
image-refresh services done: https://github.com/cockpit-project/bots/commits/image-refresh-services-20240507-144344

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-services-ff6ccc2a-20240507-142412/log.html

images/scripts/services.setup
allisonkarlitskaya: The reason I thought it was obvious was that we explicitly go out of our way to install the `hostname` command in the container specifically so that we can call it...

allisonkarlitskaya: The issue with using `hostnamectl hostname` (or even `hostname` without `-f`) is that it doesn't have the fully-qualified name.... `hostname -f` seems to be trying to do a lookup, and failing...

martinpitt (approved):
Thanks! Bots have the last word obviously.

martinpitt:
If the [cockpit/rhel-8 branch run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6350-0eb52db6-20240507-150841-rhel-8-10-cockpit-project-cockpit-rhel-8/log.html) fails, https://github.com/cockpit-project/cockpit/pull/20441 should help. The sub-man-1.28 failure is known.

martinpitt:
[Three failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6350-0eb52db6-20240507-150841-rhel-8-10-cockpit-project-cockpit-rhel-8/log.html#268-2). One is #20441 (just landed), one is #20373, and the OOM in TestStorageMountingLUKS.testOverMounting is probably unstable. I'll retry anyway to mop up the c8s fix.

martinpitt:
Ack -- all remaining failures are accounted for.

PR #6351, opened 2024-05-07
images: Add alpine-efi image

martinpitt:
So that we can test EFI image boots in cockpit-machines. These aresupported in the UI and in libvirt, but not tested well right now. This also paves the way for testing c-machines on aarch64 (which only supports EFI).

 * [x] image-refresh alpine-efi

https://github.com/cockpit-project/cockpit-machines/pull/1619 makes use of that.


cockpituous:
image-refresh alpine-efi done: https://github.com/cockpit-project/bots/commits/image-refresh-alpine-efi-20240507-144350

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-alpine-efi-3f8e95ce-20240507-144205/log.html

allisonkarlitskaya (approved):
This all looks fine, but the usual few questions:
 - is there any harm in always providing an EFI firmware?  Shouldn't it fall back to BIOS if the image doesn't support EFI?  Shouldn't we just prefer EFI at this point for all of our images?  Would this cause regressions with our hardware support?
 - Do we need an extra alpine image, or can we just convert the existing one to EFI?
 - There's no `image-trigger` entry for either this image or the BIOS one.  Is that intentional?

martinpitt:
> is there any harm in always providing an EFI firmware?

*Most* of our images (I tested f39, debian-stable, rhel-8-10) get along fine, as they are prepared for an EFI world. The main offenders are `vm-run alpine` and `vm-run cirros`, which just sit in an EFI shell when  booting it with OVMF:
```
UEFI Interactive Shell v2.2
EDK II
UEFI v2.70 (EDK II, 0x00010000)
Mapping table
     BLK0: Alias(s):
          PciRoot(0x0)/Pci(0x1,0x1)/Ata(0x0)
     BLK1: Alias(s):
          PciRoot(0x0)/Pci(0x4,0x0)
Press ESC in 1 seconds to skip startup.nsh or any other key to continue.
Shell> 
```
so it does seem to require cooperation from the OS. Also, while EFI is a thing on real iron, and of course the ARM64 world, my impression (from our OpenStack, or EC2) is that most real-world VMs still boot with classic BIOS, probably because of momentum.

We could drop `cirros` soon (if/when we move c-machines rhel-8 branch to alpine) and *maybe* get rid of the bios `alpine` image -- but IMHO it's nice that we can also support the default QEMU bootloader (i.e. bios). But this isn't a decision that I want to entangle in this PR.

> Do we need an extra alpine image, or can we just convert the existing one to EFI?

c-machines/libvirt defaults to bios, and we want to test that case in c-machines. In fact, right now you have to jump through some hoops to make them boot with EFI (mostly because libosinfo doesn't have alpine), but libosinfo even defaults to BIOS (not EFI) boot for modern Fedora 40/Debian etc., and only forces EFI on Windows 11. This is all frustratingly lagging behind the times..

> There's no image-trigger entry for either this image or the BIOS one. Is that intentional?

Yes, "intentional" (in the sense of "not accidentally forgotten"): We didn't have one for cirros, @mvollmer didn't introduce one for `alpine`, so I followed suit. There's not that much point in updating it regularly, and the bootstrap scripts even hardcode a version, so these would need to be made dynamic first.

allisonkarlitskaya (approved):

machine/machine_core/machine_virtual.py
```diff
@@ -280,6 +282,7 @@ def __init__(
         else:
             console_file = None
         self.console_file = console_file
+        self.is_efi = "-efi" in image
```
mvollmer, line 285: I am not liking this...

PR #6362, opened 2024-05-09
images: rhel4edge - remove workaround and use osbuild.ks as main config file

### LINKED ISSUE #6356: Image refresh for rhel4edge
Image refresh for rhel4edge

jscotka:
Image refresh for rhel4edge
 * [x] image-refresh rhel4edge

Fixes #6356


cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-e35db370-20240509-133354/log.html

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-5b943cc0-20240510-111050/log.html

cockpituous:
image-refresh rhel4edge done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel4edge-20240510-131557

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-5b943cc0-20240510-122138/log.html

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-c8cb126e-20240513-110217/log.html

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-0f66a93b-20240513-123332/log.html

cockpituous:
image-refresh rhel4edge done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel4edge-20240514-084435

images/scripts/ostree.setup
```diff
@@ -28,14 +28,19 @@ if [ "$IMAGE" == "fedora-coreos" ]; then
...
+    restorecon -r /var/home/admin/.ssh
+
     # Setup password for admin user as edge management only sets ssh key
     echo foobar | passwd --stdin admin
 
     # Setup password for root user
     echo foobar | passwd --stdin root
 
-    # HACK: Fix .ssh ownership, see https://issues.redhat.com/browse/THEEDGE-3033
-    chown admin:admin /var/home/admin/.ssh
```
jelly, line 38: Fedora-coreos uses the same script so is this not needed there?

jelly:
Triggered the tests.

images/scripts/ostree.setup
jscotka: Hi @jelly I've completely changed use creation, this issue is already resolved so could be removed. 
there is some issue inside how admin user inside is created, so that I'm now trying to workaorund the issue (some tests fails with strange reasons, some change in sudo for user)

cockpituous:
image-refresh rhel4edge done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel4edge-20240515-080225

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-ebd04d1b-20240515-071355/log.html

jscotka:
the issue inside rhel4edge/other@cockpit-project/cockpit  may be caused by different defaults for home directories folder from my PoV, not sure if really, just guessing now. But question is if we should fix the test, or try to add another workaround for this into image regeneration.

images/scripts/ostree.setup
```diff
@@ -37,6 +37,11 @@ if [ "$IMAGE" == "rhel4edge" ]; then
     # HACK: Fix .ssh ownership, see https://issues.redhat.com/browse/THEEDGE-3033
     chown admin:admin /var/home/admin/.ssh
 
+    # change how sudo is configured inside edge images via osbuild.ks, do not use passwordless wheel group
+    # but use passwordless admin user directly.
+    rm -f /etc/sudoers.d/%wheel
+    echo -e "admin\tALL=(ALL)\tNOPASSWD: ALL" >> /etc/sudoers
```
martinpitt, line 43: What's the motivation here? I don't like this -- we should stick to what the OS does, i.e. test images that are as close to reality as possible. If it's a workaround for a bug, then it needs to be reported, and this should get a "# HACK: url" comment.

martinpitt (changes requested):
Thanks for investigating!

That Terminal failure smells like some /home/admin vs. /var/home/admin confusion? Did the `passwd` entry change, or the mechanics of creating the initial user? That feels like some fallout from the fleet.ks change?

Where does osbuild.ks come from? Is that buggy perhaps?

martinpitt:
Thanks @jscotka for the examination in https://github.com/cockpit-project/cockpit/issues/20479 ! IMHO this is a bug in creating the `admin` user -- /etc/default/useradd's `HOME=/var/home` is at odds with /home/admin being the home dir -- it is /var/home/admin everywhere else on OSTree. So my suggestion would be to fix /etc/passwd after the fact in the setup script.

images/scripts/ostree.setup
jscotka: Hi @martinpitt, It is the issue as we've seen that some sudo tests failed. and we do not know why. 
The root cause were that in previous version by kickstart just admin user were set as passwordless sudo user. In current images whole wheel group is in sudoers and as I've seen inside cockpit tests, they expect to have set `admin` user inside sudoers.

So as you said, it is hack, but works, the most proper solution would be:
 - to avoid using `wheel` group anywhere inside tests, and create own group, because some tests uses adding users into wheel to check it is able to do admin task but with password. 
 - and also be more robust how sudo for `admin` user is configured. do not rely that just `admin` is passwordless sudoer.
 
So it seems like bigger refactoring. My gut feeling here is  that it will need more work and changes inside lots of (>10 tests, maybe much more) to make it more robust. So I prefer to do some Hackish solution and maybe create issue to refactor and robustify using users and groups inside cockpit tests.

jscotka:
> Thanks @jscotka for the examination in [cockpit-project/cockpit#20479](https://github.com/cockpit-project/cockpit/issues/20479) ! IMHO this is a bug in creating the `admin` user -- /etc/default/useradd's `HOME=/var/home` is at odds with /home/admin being the home dir -- it is /var/home/admin everywhere else on OSTree. So my suggestion would be to fix /etc/passwd after the fact in the setup script.

Sure @martinpitt not sure if it is a bug, or just little bit different workflow how the users are created.  But you are right it could be issue inside `anaconda`, that it does not respect this setting of HOME or some missing param to anaconda in case this `HOME` var for `adduser` command is added to system later.

My feeling here is, that as SSH behaves correctly, as it does not resolve symlink, so when SSH to machine, the `~` is displayed, and it is switched to `/home/admin`, not resolved symlink dir, although Terminal plugin what cockpit uses probably just make additional resolving of directory, what is probably not good option.

Whoever use symlinks for users, will end up with this issues, that `~` is not displayed as the first command prompt. So from that perspective it is not issue of rhel4edge, but common issue when using symlinks for user home dirs.

My idea is to fix it as you wish with improve home dir for `admin` user and let opened the issue, to force somehow the `Terminal cockpit plugin` to behave more correctly and do not resolve the dir.

martinpitt:
That's not specific to SSH. It's just fallout from OSTree's choice of putting home dirs into /var/home/, and happens with any login where an user has their home dir set (in /etc/passwd) to /home/user.

images/scripts/ostree.setup
martinpitt: > do not rely that just admin is passwordless sudoer.

We don't actually *mean* to assume that -- all our images are supposed to have "admin:foobar" and the same password for sudo. But indeed I just checked that the current rhel4edge image also has passwordless sudo, and an explicit entry in `/etc/sudoers`.

So let's keep that for now, and let's clean this up in a follow-up. This is a needless complication, but indeed not the fault of this PR. Sorry, didn't check before!

martinpitt:
Previous run timed out, trying again.

jscotka:
> That's not specific to SSH. It's just fallout from OSTree's choice of putting home dirs into /var/home/, and happens with any login where an user has their home dir set (in /etc/passwd) to /home/user.

I understand it. I wanted to say, that SSH behaves correctly from that perspective, what is initial dir, no matter where are home dirs stored and how are symlinked. I understand that there is now difference between OSTree choice and rhel4edge, or better to say, from my perspective just cosmetic, as probably also inside OSTree there will be this symlink `/home -> var/home`, so that user dirs are stored on same places everywhere, just changed if there is used full less-standard path or symlinked more common path for home dirs.

jscotka:
> @jscotka Timed out two times in a row -- is there a way to get some logs of the image build process? It apparently still reported "BUILDING" not fail/error or so. But maybe it's hanging somewhere?

Theoretically yes, I think that possible via logging into console.rh.com with bots credentials. and look into status of building process or with bots token and getting logs via cmd line. Maybe it was just stucked and took longer time than 120m, it is my feeling, but unable to look into it.

jscotka:
I've created the issue into edge project: https://issues.redhat.com/browse/THEEDGE-3955

jscotka:
@martinpitt seems currently there is some outage, Now trying it locally, and build still in progress for long time, than it is typical: https://console.redhat.com/edge/manage-images/3433 so I'll see later. My feel is just some glitch inside infra. We'll see later.

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-af5b9839-20240524-040806/log.html

martinpitt:
@jscotka aha, a "proper" failure, that may at least be more useful for debugging? I still don't see the logs yet, though :(

jscotka:
> @jscotka aha, a "proper" failure, that may at least be more useful for debugging? I still don't see the logs yet, though :(

Hi Martin, there is no proper failure as I though as well. I've looked into web UI and there is nothing more useful, no log, just error state for the image

jscotka:
edge image building should be already fixed, trying to reschedule it.

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-af5b9839-20240528-154438/log.html

cockpituous:
Failed. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-af5b9839-20240528-172721/log.html

cockpituous:
image-refresh rhel4edge done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel4edge-20240528-201338

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel4edge-5e5c7183-20240528-191601/log.html

martinpitt:
Nice, thanks! I pushed the image and triggered tests.

martinpitt (approved):
Yay green! Thanks @jscotka ! :star2:

PR #6380, opened 2024-05-14
setup-deploy-keys: Add weblate deploy key for cockpit-files

jelly:


setup-deploy-keys
```diff
@@ -74,6 +74,16 @@ deploy_to osbuild/cockpit-composer-weblate \
     --deploy-from \
         osbuild/cockpit-composer/cockpit-composer-weblate/DEPLOY_KEY
 
+# cockpit-files
+deploy_to cockpit-project/cockpit-files \
+    --deploy-from \
+        cockpit-project/cockpit-files/npm-update/SELF_DEPLOY_KEY \
```
allisonkarlitskaya, line 80: I think this is a separate issue, yes?  (the whole node_modules/reposchutz business...)

jelly: Yes

PR #6422, opened 2024-05-24
Fix debian-testing image refresh

### LINKED ISSUE #6377: Image refresh for debian-testing
 * [ ] FAIL: image-refresh debian-testing

martinpitt:
Fixes #6377

 * [x] Requires https://github.com/cockpit-project/cockpit/pull/20502


cockpituous:
image-refresh debian-testing done: https://github.com/cockpit-project/bots/commits/image-refresh-debian-testing-20240524-051432

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-debian-testing-ac17cf87-20240524-050101/log.html

martinpitt:
I have the other failures fixed locally, but [this one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6422-2a46a68f-20240524-051916-debian-testing-other-cockpit-project-cockpit/log.html#52-2) remains. Connecting to the RHEL 8 machine now takes rather long, but does eventually succeed.

This is independent of cockpit -- `ssh -vv 10.111.113.5` also hangs for a long time. This seems to be due to an attempted kerberos auth:

```
debug1: Next authentication method: gssapi-with-mic

# 5 seconds pause

debug1: No credentials were supplied, or the credentials were unavailable or inaccessible
No Kerberos credentials available (default cache: FILE:/tmp/krb5cc_0)

# 10 seconds pause

debug1: No credentials were supplied, or the credentials were unavailable or inaccessible
No Kerberos credentials available (default cache: FILE:/tmp/krb5cc_0)
debug2: we did not send a packet, disable method
debug1: Next authentication method: publickey
```

Curiously it works fast with cockpit-ssh, but this is ferny ssh. The test itself isn't involved -- I added a sit() right at the top.

As the previous image refresh was already 3 weeks ago, [tons of packages changed](https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-debian-testing-ac17cf87-20240524-050101/log.html). The most plausible ones are openssh-client (but not much happened, no difference with downgrade) and sssd (but it's not running, and still happens with removing all "sss" from nsswitch.conf). 

So I started with the current image, and upgraded packages step by step. But even after a full upgrade that is fast.

Finally I figured out that installing krb5-config fixes this. ssh is just slow without /etc/krb5.conf.

cockpituous:
image-refresh debian-testing done: https://github.com/cockpit-project/bots/commits/image-refresh-debian-testing-20240524-081242

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-debian-testing-665d7989-20240524-075858/log.html

allisonkarlitskaya (approved):
Why do we require the tzdata-legacy package?  will chrony not work properly without it (and if so, why isn't it a dependency)?  do our tests depend on the old names?

I am surprised that timezone data has anything at all to do with keeping the system (UTC) time in sync...

Everything else looks quite good and is a welcome clean-up.

martinpitt:
> Why do we require the tzdata-legacy package? will chrony not work properly without it (and if so, why isn't it a dependency)? do our tests depend on the old names?

It *is* a dependency. But `apt-get download` doesn't download dependencies, only the specified package. I had an intermediate version with `apt-get install --download-only`, but that doesn't work properly either as chrony is already installed by default on some images.

PR #6437, opened 2024-05-28
Add support for cockpit-machines testing for tumbleweed

Nykseli:


images/scripts/opensuse-tumbleweed.setup
```diff
@@ -120,6 +129,12 @@ echo 'PasswordAuthentication yes' >> /usr/etc/ssh/sshd_config
 rm -f '/etc/systemd/system/multi-user.target.wants/auditd.service'
 rm -rf /var/log/audit/
 
+# Make sure virtinterfaced is enabled
+systemctl enable --now virtinterfaced
```
martinpitt, line 133: This is a bit strange for a generic image, and non-c-machines tests don't need that. Let's rather do that in c-machines `test/vm.install`.

Also, are you sure that you want to enable the *service*? Fedora and Debian have a `.socket` unit to avoid running the service all the time.

martinpitt (changes requested):
Thanks! Let's clean this up a bit, then I'm happy to do the usual dance to build the image.

images/scripts/opensuse-tumbleweed.setup
Nykseli: It's now moved to vm.install on https://github.com/cockpit-project/cockpit-machines/pull/1655

martinpitt:
@Nykseli Building in #6438 -- feel free to cherry-pick the commit once it's ready.

PR #6442, opened 2024-05-29
testmap: Enable opensuse-tumbleweed for cockpit-machines

martinpitt:
Tests got fixed in https://github.com/cockpit-project/cockpit-machines/pull/1655

---

But needs to land that PR first, so blocked.


martinpitt:
c-machines PR landed. Triggering test here again for the fun of it.

@SlugeGirl, @Nykseli Nice work! :partying_face:

mvollmer (approved):
Yay!

PR #6450, opened 2024-05-31
images: Add libcriu2 dependency for tumbleweed

Nykseli:
Add a missing libcriu2 dependency for tumbleweed that's required for podman testing


martinpitt:
Done in #6451

PR #6470, opened 2024-06-05
setup-deploy-keys: cockpit-files access to node-cache

allisonkarlitskaya:
We just merged the post-dependabot workflow for updating the node-cache, so add its environment (`npm-update`) plus add the `node-cache` environment as well — we don't use it yet, but we will use it once we start making releases, hopefully soon.

 - [x] run it

### DIFF
--- a/setup-deploy-keys
+++ b/setup-deploy-keys
@@ -77,6 +77,7 @@ deploy_to osbuild/cockpit-composer-weblate \
 # cockpit-files
 deploy_to cockpit-project/cockpit-files \
     --deploy-from \
+        cockpit-project/cockpit-files/npm-update/SELF_DEPLOY_KEY \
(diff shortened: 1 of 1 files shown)


setup-deploy-keys
```diff
@@ -77,6 +77,7 @@ deploy_to osbuild/cockpit-composer-weblate \
 # cockpit-files
 deploy_to cockpit-project/cockpit-files \
     --deploy-from \
+        cockpit-project/cockpit-machines/npm-update/SELF_DEPLOY_KEY \
```
jelly, line 80: that should be files.

jelly, line 80: that should be files.

allisonkarlitskaya: Thanks!  Fixed.

jelly (approved):

PR #6478, opened 2024-06-07
setup-deploy-keys: fix cockpit-files node-cache keys

allisonkarlitskaya:
These are in the correct environment, but with bogus names.  It's currently preventing the release from running properly.


jelly (approved):

allisonkarlitskaya:
I ran the script.

PR #6553, opened 2024-06-27
tests-scan: prioritise /devel scenario

allisonkarlitskaya:
Add back a very simple priority scheme: any test that hsa '/devel' in the name runs first.  We want to do this because the '/devel' scenarios take longer, so we want to get them running ASAP.


martinpitt (approved):
Nice! And for just a few :heavy_dollar_sign:  more you can get MAX+1!

Honestly I have no idea whether priorities actually work, or whether they need changes in `run-queue`. But even if that's the case, that's a good start.

PR #6559, opened 2024-06-28
testmap: Enable cockpit-podman on RHEL 10

martinpitt:
We are over the initial "everything is broken" hump, and start to regularly land our releases in RHEL 10. Let's shift left the gating failures.

---

@jelly I don't have a strong feeling about this either way. I mostly lean towards "let's enable it if it works, and pull the plug again if it becomes too much of a burden". WDYT?

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -110,6 +110,7 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
             'rhel-8-10',
             'rhel-9-4',
             'rhel-9-5',
+            'rhel-10-0',
             'rhel4edge',
             'ubuntu-2204',
             'ubuntu-stable',
@@ -118,7 +119,6 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
             'centos-10',
             'fedora-rawhide',
             'opensuse-tumbleweed',
-            'rhel-10-0',
         ],
     },
     'cockpit-project/cockpit-machines': {


jelly:
```
Error: netavark: unable to append rule '! -d 224.0.0.0/4 -j MASQUERADE' to table 'nat': code: 4, msg: Warning: Extension MASQUERADE revision 0 not supported, missing kernel module?
iptables v1.8.10 (nf_tables):  RULE_APPEND failed (No such file or directory): rule in chain NETAVARK-1D8721804F16F
```

Seems this is still broken and rings a bell. (We had this issue before but it was fixed in CentOS but not in RHEL). So from me this is a nack for merging.

martinpitt:
Ah, our [centos-10 image](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6559-996deb3e-20240628-101734-centos-10-cockpit-project-cockpit-podman/log.html#1-2) still runs into the "old netavark/iptables" issue. Curiously the rhel-10-0 image [seems to work](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6559-996deb3e-20240628-101803-rhel-10-0-cockpit-project-cockpit-podman/log.html)!? Well then, let's start with that one.

martinpitt:
@jelly As I said, please absolutely feel free to say "no" here. This is more of a gut check.

martinpitt:
@jelly yes, that triggers a very strong déjà vù, and indeed I filed this as https://bugzilla.redhat.com/show_bug.cgi?id=2273078 already. We still have the workaround in cockpit, but not in podman yet. I'll file a CentOS 10 issue and send a c-podman PR to apply the workaround there as well.

martinpitt:
This time there's nothing wrong with the individual packages, C10 composes/mirror just has been broken for over a month. :exploding_head:  I reported https://issues.redhat.com/browse/CS-2309

martinpitt:
In other words, centos-10 is rather useless right now, and rhel-10-0 is a better testing target. I'll look at enabling c-10 again once https://issues.redhat.com/browse/CS-2309 gets fixed.

jelly (approved):

PR #6557, opened 2024-06-27
Image refresh for rhel-9-2

cockpituous:
Image refresh for rhel-9-2
 * [x] image-refresh rhel-9-2
 * [x] #6585

PR #6585, opened 2024-07-05
images: Drop rhel4edge

### LINKED ISSUE #6584: Image refresh for rhel4edge
Image refresh for rhel4edge

martinpitt:
This is a high-maintenance vs. low-gain image: The infrastructure is very brittle and mostly abandoned, and fails on the majority of refreshes. It's also structurally abandoned, bootc is the designated successor. It also never really found any issues which were specific to rhel4edge, we can cover everything with our normal rhel* and the fedora-coreos images, and have centos-9-bootc as well.

Fixes #6584

It also unblocks #6557.


martinpitt:
@croissanne FYI, are you d'accord with this?

martinpitt:
@marusak this is more of a strategic decision. We need to shed some piloting load, and this is IMHO a good candidate.

croissanne:
> @croissanne FYI, are you d'accord with this?

Absolutely, :axe: it!

PR #6609, opened 2024-07-12
image-trigger: Auto-refresh rhel-10-0

martinpitt:
Forgotten in commit 1459ce23bba0f.


martinpitt:
In the meantime I triggered a refresh manually in #6610. Our current image is already two weeks old.

tomasmatus (approved):

PR #6630, opened 2024-07-18
Image refresh for centos-9-bootc

cockpituous:
Image refresh for centos-9-bootc
 * [x] https://github.com/cockpit-project/cockpit-ostree/pull/663

### DIFF
--- a/images/centos-9-bootc
+++ b/images/centos-9-bootc
@@ -1,1 +1,1 @@
-centos-9-bootc-45933c33faa108ab9b1ed7761e10406afd579e49bcf086b9aa0c897ab6b3af48.qcow2
+centos-9-bootc-d80f0721282e49fec497b9cdd9831e96681839ffecda183da4ee535faa47c119.qcow2


cockpituous:
image-refresh centos-9-bootc done: https://github.com/cockpit-project/bots/commits/image-refresh-centos-9-bootc-20240718-035449

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-centos-9-bootc-f6db7494-20240718-033103/log.html

martinpitt:
This [failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6630-69533d7a-20240718-035558-centos-9-bootc-cockpit-project-cockpit-ostree/log.html#4-2) is a mystery to me. With both the current and the new image the order is like this:

image

That's correct -- we just `bootc switch`ed to the new (local) "cockpit1" branch, and are currently running the bootc:latest branch. This coincides with
```
# rpm-ostree status
State: idle
Deployments:
  ostree-unverified-registry:localhost:5000/bootc:cockpit1
                   Digest: sha256:517591117df964bd7c5f57dc788e856599b5e9cadb4940bda3a27a8f04180bc9
                  Version: stream9.20240717.0 (2024-07-18T08:00:47Z)
                     Diff: 1 removed

● ostree-unverified-registry:localhost/bootc:latest
                   Digest: sha256:ea525d1f7db0f06cc8bb894fc1b59d147ca6e4465cfdd39ab61e1616dc08b932
                  Version: stream9.20240717.0 (2024-07-18T03:35:29Z)
```
(`bootc status` says the same, in more verbose json form). i. newest (in terms of creation date) deployment first.

And that's also what the test tries to do:

> wait_deployment_prop(b, 1, "Branch", new_repo_branch)

which translates to

> #available-deployments > tbody:nth-child(2) td[data-label=Branch]","ostree-unverified-registry:localhost:5000/bootc:cockpit1"

There's no visual nor HTML structure difference between "current" and "pr", not even with `TEST_SHOW_BROWSER=1` (i.e. it's the primary browser, I didn't attach to an already existing session, which would do a reload of the page from scratch).

`wait_deployment_prop()` deliberately increases the given index by one, presumably because to account for the `<thead>` (which is the first sibling of the parent element). There are just *two* tbody's. 

I have a local patch to clean this up, but it still fails afterwards. But at least in a less confusing way.

martinpitt:
The screenshot only shows the expanded first deployment -- this is annoying, as you can't see the second one. But the HTML dump confirms that, there's just one tbody+deployment. 

I get this locally with all four of {chromium,firefox} x {headless, interactive}, at least most of the time. I wonder if it's because rpm-ostree shuts down too early, so the page is missing events afterwards?

```
bootc[1164]: Image contains non-ostree compatible file paths: run: 2
systemd[1]: systemd-hostnamed.service: Deactivated successfully.
systemd[1]: rpm-ostreed.service: Deactivated successfully.
```

The "comaptible file paths" warning happens on the current image as well, but rpm-ostreed timed out a little later, so that the event was still sent. So this does  seem to be a timing problem, and on the new image the bootc deployment just takes a little longer.

martinpitt (approved):

PR #6644, opened 2024-07-20
Image refresh for services

cockpituous:
Image refresh for services
 * [x] image-refresh services
 * [x] https://github.com/cockpit-project/cockpit/pull/20788
 * [x] https://github.com/cockpit-project/cockpit/pull/20789


cockpituous:
image-refresh services done: https://github.com/cockpit-project/bots/commits/image-refresh-services-20240720-225717

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-services-5b58ef51-20240720-223718/log.html

martinpitt:
The [failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6644-a5246c84-20240720-225839-fedora-40-other-cockpit-project-cockpit/log.html#28-2) happens because the new grafana version turned the `<div>`s for the panels into `<section>`s. That's a good move, but now we have to write the test in a way to stay compatible with both, as there is also a nested `<button>` with the same `data-testid` attribute.

martinpitt:
Fixed in https://github.com/cockpit-project/cockpit/pull/20788 , blocking on that.

martinpitt (approved):

PR #6671, opened 2024-07-29
vm-run: don't pass graphics argument

jelly:
Commit 7331ff9da77eaa986593f removed the useless graphics option, but didn't remove it from vm-run.


martinpitt (approved):
Thanks! I figure the main effect is starting virt-viewer indeed.

PR #6681, opened 2024-08-01
testmap: updates to sub-man & sub-man-cockpit

ptoscano:
Long-awaited and almost forgotten updates for subscription-manager and
subscription-manager-cockpit:
- sub-man/main: this branch now tracks the work for RHEL 10.x, so drop
  almost the RHEL 9 OSes (leaving the most recent one to help for
  backporting); rhel-10-0 will be added later when a fixed image is
  available
- sub-man/1.28: drop EOL & out of EUS OSes from there; add 8.10, which
  is what the branch tracks now
- sub-man/1.28.29: drop, as it is out of EUS
- sub-man/1.29: new branch that tracks the work for RHEL 9.x, so add the
  supported RHEL 9 OSes to it
- sub-man/1.29.26: drop, as it is out of EUS
- sub-man/1.29.40: new branch to track the work for RHEL 9.4.z
- sub-man-cockpit/main: add the latest version of RHEL 9.x, and modifiy
  CS 9 to use the right upstream branch for EL9
- drop almost all the manual OSes for sub-man & sub-man-cockpit,
  leaving RHEL 10, as every OS there is now in supported branches

Update also the triggered versions for the "service" image: bump to the latest versions of RHEL 9 & RHEL 8.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -178,47 +178,40 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
     },
     'candlepin/subscription-manager': {
         'main': [
-            'rhel-9-0',
-            'rhel-9-2',
-            'rhel-9-3',
-            'rhel-9-4',
+            'rhel-9-5',
             'fedora-39',
             'fedora-40',
         ],
         'subscription-manager-1.28': [
-            'rhel-8-4',
-            'rhel-8-6',
             'rhel-8-8',
-            'rhel-8-9',
-        ],
-        'subscription-manager-1.28.29': [
-            'rhel-8-6',
+            'rhel-8-10',
         ],
         'subscription-manager-1.28.36': [
             'rhel-8-8',
         ],
-        'subscription-manager-1.29.26': [
-            'rhel-9-0',
+        'subscription-manager-1.29': [
+            'rhel-9-2',
+            'rhel-9-4',
+            'rhel-9-5',
         ],
         'subscription-manager-1.29.33': [
             'rhel-9-2',
         ],
+        'subscription-manager-1.29.40': [
+            'rhel-9-4',
+        ],
         '_manual': [
-            'rhel-8-10',
-            'rhel-9-5',
             'rhel-10-0',
         ],
     },
     'candlepin/subscription-manager-cockpit': {
         'main': [
-            'centos-9-stream',
-            'rhel-9-3',
-            'rhel-9-4',
+            'centos-9-stream/subscription-manager-1.29',
+            'rhel-9-5/subscription-manager-1.29',
             'fedora-39',
             'fedora-40',
         ],
         '_manual': [
-            'rhel-9-5',
             'rhel-10-0',
         ],
     },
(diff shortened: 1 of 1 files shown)


ptoscano:
AFAIAA (As Far As I Am Aware), all the added/modified configurations ought to be working; please schedule jobs for them, and in case there are issues I'll happily fix them.

Thanks!

jelly:
This is a big change :eyes: maybe it's worth to split things up but for now I'll try to trigger some tests.

lib/testmap.py
```diff
@@ -178,48 +178,42 @@ def contexts(image: str, *scenarios: Iterable[str], repo: str | None = None) ->
...
-            'rhel-9-5',
-            'rhel-10-0',
         ],
     },
     'candlepin/subscription-manager-cockpit': {
         'main': [
             'centos-9-stream',
-            'rhel-9-3',
-            'rhel-9-4',
+            'centos-10-stream',
```
jelly, line 210: The image is named `centos-10`

jelly (changes requested):
Seeing how much needs to be triggered, I would favor splitting this up in parts

lib/testmap.py
ptoscano: ooops, will fix, thanks

jelly:
rhel-9-4@candlepin/subscription-manager/subscription-manager-1.29.40  stands on its own, maybe split this off? It's also green :)

ptoscano:
the el10 jobs fails most likely because of old `insights-client` in the el10 images; hence I'm dropping them for now, leaving only changes for rhel8 & rhel9 (which seem to work fine)

is this still big enough?

lib/testmap.py
```diff
@@ -178,47 +178,40 @@ def contexts(image: str, *scenarios: Iterable[str], repo: str | None = None) ->
...
             'rhel-10-0',
         ],
     },
     'candlepin/subscription-manager-cockpit': {
         'main': [
-            'centos-9-stream',
-            'rhel-9-3',
-            'rhel-9-4',
+            'centos-9-stream/subscription-manager-1.29',
+            'rhel-9-5/subscription-manager-1.29',
```
jelly, line 210: Is this correct?

ptoscano: yes, the 1.29 branch targets rhel 9.x -- see also the branch map above for sub-man

jelly: But that is different, it lists the branches under the branch name. This would require to trigger:

```
centos-9-stream/subscription-manager-1.29@candlepin/subscription-manager-cockpit
```
? Doesn't seem to make sense compared to the previous triggers

ptoscano: sub-man-cockpit implements the "context" as branch of upstream sub-man to use for testing, defaulting to "main"; the value "system" means to use whatever sub-man is available on the system

jelly: I triggered it but have a hard time wondering if it will work. The other changes look more trivial :)

ptoscano: it worked, as expected:
```
git -C subscription-manager fetch --quiet --depth=1 origin subscription-manager-1.29
git -C subscription-manager checkout --quiet FETCH_HEAD
```

if you look at the history of the mappings for sub-man-cockpit, there used to contexts like this in the past

jelly:
@ptoscano I think I triggered all that changed now.

ptoscano:
> I think I triggered all that changed now.

thanks! and all LGTM :)

jelly (approved):
This allows us to drop some rhel images, which is nice :partying_face:

PR #6688, opened 2024-08-02
test-failure-policy: Ignore `window.` prefix for ph_* test helper functions

martinpitt:
We will soon move cockpit's test-functions.js helpers to the `window` object, so that they become compatible with Webdriver's script preloading API. This affects all naughty patterns which include any of the `ph_*()` function calls (i.e. most patterns). Until this all lands and settles down, avoid mass-changing all the patterns and just remove the `window.` prefix for the comparison.

Note that the naughties also apply to old project branches.

---

I tested locally that this works. Once this lands, I'll send a cockpit PR to actually do that move -- it's intrusive but simple, and I want it out of the way. Then https://github.com/cockpit-project/cockpit/pull/20832 should become much greener.

### DIFF
--- a/test-failure-policy
+++ b/test-failure-policy
@@ -146,6 +146,9 @@ def normalize_traceback(trace: str) -> str:
     trace = re.sub(r'audit\([0-9.:]+\)', 'audit(0)', trace)
     trace = re.sub(r'\b(pid|ino)=[0-9]+ ', r'\1=0 ', trace)
 
+    # testlib API change: move ph_*() test-functions to window object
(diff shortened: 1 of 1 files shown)


jelly (approved):

martinpitt:
[This run](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6688-776ac08a-20240802-070932-ubuntu-2204-cockpit-project-cockpit-podman/log.html) shows that naughties with a `ph_` are still caught correctly.

PR #6690, opened 2024-08-02
Drop unused RHEL versions

ptoscano:
Drop unused and no more supported (at least until EUS) RHEL versions:
- `rhel-8-4`
- `rhel-8-6`
- `rhel-8-9`
- `rhel-9-0`
- `rhel-9-3`

While I'm there cleaning things up, I dropped a couple of symlinks for naughties that are no more used:
- `rhel-8-7`
- `rhel-9-1`


martinpitt (approved):
Whoo, nice Friday cleanup! Thanks @ptoscano !

PR #6708, opened 2024-08-07
services: expose also port 8080 from Candlepin container

ptoscano:
tomcat exposes an HTTP server that listens on port 8080, serving the yum/dnf repositories with the content of the test data. Hence expose it when running the container, so it can be accessed by tools.

Verify that the port is actually open by checking the RPM GPG key of the repositories can be fetched.

 * [x] image-refresh services


martinpitt:
@ptoscano I set the image build in motion. If it succeeds, feel free to cherry-pick the change and push it here. Otherwise I can help tomorrow morning.

cockpituous:
image-refresh services done: https://github.com/cockpit-project/bots/commits/image-refresh-services-20240807-183931

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-services-a1340333-20240807-181901/log.html

martinpitt (approved):
Thanks!

PR #6712, opened 2024-08-08
testmap: add _manual rhel-9-5 to cockpit-composer

croissanne:
Want to trigger these manually before adding them to each run.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -164,10 +164,10 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
             'rhel-8-10',
             'rhel-9-4',
         ],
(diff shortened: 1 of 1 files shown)


lib/testmap.py
```diff
@@ -164,10 +164,10 @@ def contexts(image: str, *scenarios: Iterable[str], repo: str | None = None) ->
             'rhel-8-10',
             'rhel-9-4',
         ],
-        # These can be triggered manually with bots/tests-trigger
```
martinpitt, line 167: A bit odd to remove that comment, but I don't mind..

martinpitt (approved):

lib/testmap.py
croissanne: figured it was a bit out of place ._.

PR #6713, opened 2024-08-08
machine_core: Fix and clean up start_cockpit()

martinpitt:
Explicitly `reset-failed` our two main cockpit units. Globbing does not always work, and tests like `TestLogin.testClientCertAuthentication` which restart cockpit in fast succession still ran into "Failed with result 'start-limit-hit'."

Stop the service before daemon-reload, which is more robust.

Since commit b7d0edba5d026 we don't need the `&&` any more, and they are in fact actively harmful. Drop them.

---

This should fix the [downstream gating failure](https://dashboard.osci.redhat.com/#/details/brew-build-63204955?focus=osci.brew-build.tier0.functional) (sorry RH internal link). The [TF journal artifact](https://artifacts.osci.redhat.com/testing-farm/ee0b95d4-1f20-465f-8d12-3b20d9ec050b/work-main6dzrwp30/plans/all/main/execute/data/guest/default-0/test/browser/main-1/data/TestLogin-testClientCertAuthentication-rhel-10-0-10.88.0.1-22-FAIL.log.gz) gives it away.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -260,24 +260,26 @@ class Machine(ssh_connection.SSHConnection):
             self.wait_for_cockpit_running(atomic_wait_for_host or "localhost")
         elif tls:
             self.execute("""
-            rm -f /etc/systemd/system/cockpit.service.d/notls.conf &&
-            systemctl reset-failed 'cockpit*' &&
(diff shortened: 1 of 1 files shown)


mvollmer (approved):
Thanks!

PR #6725, opened 2024-08-11
Some naughty adjustments for BiDi transition

martinpitt:
See https://github.com/cockpit-project/cockpit/pull/20832#discussion_r1712928431 and detailed commit messages.


martinpitt:
That cockpituous failure is unrelated -- *sigh* we just haven't done any real bots or cockpituous changes in weeks, so this hasn't run in a while. I put it onto the pilot board.

allisonkarlitskaya (approved):
Happy to see this change, but I can't help but think maybe we should run *all* the tests here to make sure nothing broke...

martinpitt:
[This result](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6725-ecbd2c14-20240811-140213-rhel-9-5-storage-cockpit-project-cockpit/log.html) shows the naughty  matching still works, and the `window.*` change was never deployed.

PR #6734, opened 2024-08-14
testmap: add el10 testing to cockpit-composer

croissanne:


martinpitt (approved):
Wohoo, thanks! Would be nice to remove rhel-10-0 from `_manual`, but it doesn't really hurt -- we can clean this up later.

PR #6741, opened 2024-08-16
naughty: Update fedora-rawhide to fedora-42

martinpitt:
Fedora 41 branched off. Copy the rawhide naughties, and update the rawhide symlink accordingly.


mvollmer (approved):
Thanks!

PR #6759, opened 2024-08-19
testmap: Test cockpit on fedora-41

mvollmer:


martinpitt:
https://github.com/cockpit-project/cockpit/pull/20911 landed, so this should work now.

martinpitt (approved):

PR #6760, opened 2024-08-19
testmap: Run c-files on Fedora 41 and RHEL 9/10

martinpitt:
We ship cockpit-files in RHEL 9.5/10 beta now, so we should test it there.

Also enable Fedora 41 (branched off from Rawhide recently).


martinpitt:
I'll look at the [rhel-9-5](https://cockpit-logs.us-east-1.linodeobjects.com/pull-6760-6cb82890-20240819-124109-rhel-9-5-cockpit-project-cockpit-files/log.html) failure separately, in a c-files PR. Reducing this for now to drop rhel-9-5.

jelly (approved):
sure!

PR #6761, opened 2024-08-19
testmap: Run sub-man-cockpit on Fedora 41

martinpitt:
- [x] https://github.com/candlepin/subscription-manager-cockpit/pull/79

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -177,6 +177,7 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
             'rhel-10-0',
             'fedora-39',
             'fedora-40',
+            'fedora-41',
         ],
         'subscription-manager-1.28': [
             'rhel-8-8',
@@ -206,9 +207,9 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
             'rhel-10-0',
             'fedora-39',
             'fedora-40',
+            'fedora-41',
         ],
         '_manual': [
-            'fedora-41',
         ],
     },
     'rhinstaller/anaconda-webui': {


martinpitt:
@ptoscano FYI

martinpitt:
@ptoscano : This fails because sub-man checks out an ancient cockpit lib (based on version 292). This failure happens because Fedora 41 moved to dnf 5, and packagelib.py was fixed accordingly in https://github.com/cockpit-project/cockpit/commit/c71ae28e557ae79210537bd0af8f6fcf5c9ad20c .

This is going to be a huge jump (including the CDP → BiDi transition, which is imperative if you want to test on Firefox), but apparently becomes pressing now.

lib/testmap.py
```diff
@@ -214,9 +214,9 @@ def contexts(image: str, *scenarios: Iterable[str], repo: str | None = None) ->
             'rhel-10-0',
             'fedora-39',
             'fedora-40',
+            'fedora-41',
```
martinpitt, line 217: This needs to be applied to sub-man above as well

ptoscano:
> @ptoscano : This fails because sub-man checks out an ancient cockpit lib (based on version 292). This failure happens because Fedora 41 moved to dnf 5, and packagelib.py was fixed accordingly in [cockpit-project/cockpit@c71ae28](https://github.com/cockpit-project/cockpit/commit/c71ae28e557ae79210537bd0af8f6fcf5c9ad20c) .

I see; we have few occurrences of dnf too:
```
$ PAGER= git grep -nw dnf
.github/workflows/stylish.yml:17:        dnf --setopt install_weak_deps=False install -y \
README.md:22:    $ sudo dnf install gettext nodejs npm make
test/vm.install-sub-man:4:dnf install -y cockpit-ws python3-devel openssl-devel intltool gcc
```
- `stylish.yml` uses the fedora:stable container; once f41 is released then we'll need to adapt to it to dnf5 (may be nothing to do)
- `README.md` doesn't matter
- `vm.install-sub-man` should work fine just fine with dnf5?

> This is going to be a huge jump (including the CDP → BiDi transition, which is imperative if you want to test on Firefox), but apparently becomes pressing now.

I'm not sure what you are talking about; do you have any references/links about it?

martinpitt:
@ptoscano This is the place that checks out cockpit's test/common: https://github.com/candlepin/subscription-manager-cockpit/blob/630d2319a5a638958d04b2d8c84ccc97b382da1e/Makefile#L225  d9426e07a261c182d61163900740fd74d02fe99b is the cockpit commit. The comment "292 + sizzle workaround for Chromium 113" describes it.

ptoscano:
> @ptoscano This is the place that checks out cockpit's test/common: https://github.com/candlepin/subscription-manager-cockpit/blob/630d2319a5a638958d04b2d8c84ccc97b382da1e/Makefile#L225 d9426e07a261c182d61163900740fd74d02fe99b is the cockpit commit. The comment "292 + sizzle workaround for Chromium 113" describes it.

Oh that I know :D I was referring to your comments about the "huge jump" in the version requirement, and which other changes in cockpit may require changes in sub-man-cockpit.

martinpitt:
@ptoscano You don't import pkg/lib, so there shouldn't be node-modules changes (other than that you can drop `chrome-remote-interface`). But it's still a big jump, I'd be surprised if it "just works". If there's something unclear, please feel free to @ me in the sub-man PR.

ptoscano:
Bumping the cockpit test version seemed doable, with a couple of easy changes outside the test themselves:
https://github.com/candlepin/subscription-manager-cockpit/pull/79
it seems to be working fine!

martinpitt:
Yay, all green, thanks @ptoscano ! Welcome to Fedora 41.

ptoscano:
> Yay, all green, thanks @ptoscano ! Welcome to Fedora 41.

nice, thank you!

tomasmatus (approved):

PR #6768, opened 2024-08-20
naughty: Copy #4769 to fedora-41

mvollmer:


martinpitt (approved):

martinpitt:
Ah, we missed that as we copied f-41 naughties from rawhide, not from 40. That's conceptually correct, but we don't run the destructive tests in TF, so we never noticed that one. Thanks!

mvollmer:
The pattern has matched correctly, the rest of the failures will be addressed in https://github.com/cockpit-project/cockpit/pull/20911

PR #6706, opened 2024-08-07
Image refresh for fedora-40

cockpituous:
Image refresh for fedora-40
 * [x] image-refresh fedora-40
 * [x] https://github.com/cockpit-project/cockpit/pull/20850

PR #6770, opened 2024-08-20
naughty: New 6769 for fedora-41

mvollmer:


mvollmer:
The naughty pattern has matched, the rest are known bugs.

mvollmer:
> Also, this sounds an awful lot like #6706 

Ah! It's exactly the same. Unfortunately (or fortunately), increasing RAM doesn't help. (It only helps when we increase it so much that the crashkernel setting goes into its next interval.)

So this is a bug in kdump then: on fedora-41 it produces initrds that are too large for it's smallest `crashkernel` value, at least when dumping to NFS.

I think a naughty is appropriate, instead of a workaround.

mvollmer:
> That pattern feels a little thin -- can you please restrict that to at least a test class, and also an exception what goes wrong?

Yes, good point.

martinpitt (approved):
Thanks! Sorry for misleading you then -- it really felt like the same issue :cry: Well, either that, or we didn't fully resolve the original one, and 1 → 1.5 GiB RAM just somehow papered over it. This is all something which we really shouldn't have to know :cry:

mvollmer:
> and 1 → 1.5 GiB RAM just somehow papered over it.

I think that's what happened. No idea how.

mvollmer:
> > and 1 → 1.5 GiB RAM just somehow papered over it.
> 
> I think that's what happened. No idea how.

In general, I think this is another instance of "nobody has only 1 GiB of RAM" and we are running into kinda unsupported scenarios.

martinpitt:
RAM is is still the most expensive bit in cloud environments, so I wouldn't bet on that. On real iron, sure.

mvollmer:
> RAM is is still the most expensive bit in cloud environments, so I wouldn't bet on that. On real iron, sure.

Yes, convinced. Let's reduce the NFSAnsible machine back to 1G and add this naughty also to fedora-40. It has actually happened: https://cockpit-logs.us-east-1.linodeobjects.com/pull-20911-8d5b7527-20240821-093006-fedora-40-devel/log.html#389-2

PR #6772, opened 2024-08-21
testmap: Run firefox scenario for cockpit-podman

martinpitt:


martinpitt:
Ours did -- but what is up with github actions?

PR #6808, opened 2024-08-29
testmap: add centos-10 to sub-man

ptoscano:
Test also centos-10 for the "main" branches of sub-man and sub-man-cockpit, which already test on rhel-10-0. This way, we can spot changes coming from CentOS Stream, and compare the behaviour of centos-10 with rhel-10-x.


martinpitt (approved):
Yay!

Issue #6806, opened 2024-08-29
ssh - listener socket -  too long name  for Unix domain socket


jscotka:
Hi,
I've tested the cockpit with some machine with long hostname and I've met the issue 
```
17:35:47                 out: # testBasic (__main__.TestApps.testBasic)
17:35:47                 out: Starting machine1 xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.example.com-22
17:35:47                 out: Assuming machine is already running
17:35:47                 out: + ! test -f /run/nologin && cat /proc/sys/kernel/random/boot_id
17:35:48                 out: unix_listener: path "/tmp/.cockpit-test-resources/ssh-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.example.com-22-root-1757212.f7IbxcOE2AKZBgW6" too long for Unix domain socket
17:35:48                 out: ssh: connection refused, trying again
```
seems We've met the issue: 
https://stackoverflow.com/questions/35970686/ansible-ssh-error-unix-listener-too-long-for-unix-domain-socket

We should probably use `%C` in ssh config, to avoid to exceed this control path file name or maybe possible to use some cmd line option and use own customized path without hostname inside path: 
```
Host *
  ControlPath ~/.ssh/control/%C
  ControlMaster auto
```

Thanks&Regards
Honza

martinpitt:
Indeed I figure the reason is the excessively long host name. Unix socket paths are limited to something like 118 characters or so

PR #6809, opened 2024-08-29
avoid to use long ssh control path when there is long hostname - FIX COMMIT MESSAGE ON MERGE

jscotka:
according to: https://stackoverflow.com/questions/35970686/ansible-ssh-error-unix-listener-too-long-for-unix-domain-socket using `%C` should create hash of hostname and timestamt to avoid to exceed limit of control path.

fixes: https://github.com/cockpit-project/bots/issues/6806

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -170,7 +170,7 @@ class SSHConnection:
     def _start_ssh_master(self) -> None:
         self._kill_ssh_master()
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Excellent, thanks @jscotka! Our test queue is rather long, and I want to wait for the two example test runs to make sure this doesn't break.

jscotka:
> Excellent, thanks @jscotka! Our test queue is rather long, and I want to wait for the two example test runs to make sure this doesn't break.

Sure, thanks, I've also randomly pick one, to prove it works somehow :-)

PR #6838, opened 2024-09-07
Image refresh for services

cockpituous:
Image refresh for services
 * [x] image-refresh services
 * [x] https://github.com/cockpit-project/cockpit/pull/21019


cockpituous:
image-refresh services done: https://github.com/cockpit-project/bots/commits/image-refresh-services-20240907-230026

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-services-08623b91-20240907-224006/log.html

jelly:
Legit bidi driver crash in our code. The body element contains something which makes `unpack_value` unhappy.

jelly:
Issue being in:

```
x = [["__zone_symbol__state", {"type": "boolean", "value": True}], ["__zone_symbol__value", {"type": "undefined"}]]
```

The undefined type has no "value".

jelly:
With DEBUG protocol log:

```
INFO:bidi.command:← script.evaluate({'expression': 'ph_wait_cond(() => ph_is_present("body"), 15000, null)', 'awa
itPromise': True, 'target': {'context': '3ef59319-5baf-4a4c-9d6c-3096ad7c6be6'}}) [id 6]
DEBUG:bidi.proto:ws TEXT → {'type': 'success', 'id': 6, 'result': {'realm': '3e50188f-4ff6-4fd1-b006-315962db1458
', 'type': 'success', 'result': {'type': 'object', 'value': [['__zone_symbol__state', {'type': 'boolean', 'value'
: True}], ['__zone_symbol__value', {'type': 'undefined'}]]}}}
DEBUG:bidi.proto:ws_reader: resolving pending command 6
DEBUG:bidi.proto:[id 6] unpacking raw result {'type': 'object', 'value': [['__zone_symbol__state', {'type': 'bool
ean', 'value': True}], ['__zone_symbol__value', {'type': 'undefined'}]]}
INFO:bidi.command:← browsingContext.captureScreenshot('quiet') [id 6]
```

jelly:
After hacking `unpack_value` to ignore `undefined` this still fails:

```
Exiting due to channel error.
  File "/home/jelle/projects/cockpit/main/./test/verify/check-metrics", line 1494, in testBasic
    bg.click("button:contains('Log in')")
  File "/home/jelle/projects/cockpit/main/test/common/testlib.py", line 575, in click
    self.mouse(selector + ":not([disabled]):not([aria-disabled=true])", "click")
  File "/home/jelle/projects/cockpit/main/test/common/testlib.py", line 568, in mouse
    self.bidi("input.performActions", context=self.driver.context, actions=keys_pre + actions + keys_post)
  File "/home/jelle/projects/cockpit/main/test/common/testlib.py", line 324, in bidi
    raise Error(str(e)) from None
testlib.Error: invalid argument: Expected "element" to be a SharedReference, got: [object Object]
```

jelly:
```
INFO:bidi.command:← script.evaluate({'expression': 'ph_wait_cond(() => ph_is_visible("button:contains(\'Log in\')
:not([disabled]):not([aria-disabled=true])"), 15000, null)', 'awaitPromise': True, 'target': {'context': '3964f4f
7-f26a-4497-aaa1-ae68b2d4a0a7'}}) [id 32]
DEBUG:bidi.proto:ws TEXT → {'type': 'success', 'id': 32, 'result': {'realm': '8801c256-32fc-46fe-999e-5ef54f4a744
6', 'type': 'success', 'result': {'type': 'object', 'value': [['__zone_symbol__state', {'type': 'boolean', 'value
': True}], ['__zone_symbol__value', {'type': 'undefined'}]]}}}
DEBUG:bidi.proto:ws_reader: resolving pending command 32
DEBUG:bidi.proto:[id 32] unpacking raw result {'type': 'object', 'value': [['__zone_symbol__state', {'type': 'boo
lean', 'value': True}], ['__zone_symbol__value', {'type': 'undefined'}]]}
INFO:bidi.command:→ {'realm': '8801c256-32fc-46fe-999e-5ef54f4a7446', 'type': 'success', 'result': {'__zone_symbo
l__state': True}}
INFO:bidi.command:← script.evaluate({'expression': 'ph_find_scroll_into_view("button:contains(\'Log in\'):not([di
sabled]):not([aria-disabled=true])")', 'awaitPromise': True, 'target': {'context': '3964f4f7-f26a-4497-aaa1-ae68b
2d4a0a7'}}) [id 33]
DEBUG:bidi.proto:ws TEXT → {'type': 'success', 'id': 33, 'result': {'realm': '8801c256-32fc-46fe-999e-5ef54f4a744
6', 'type': 'success', 'result': {'type': 'object', 'value': [['__zone_symbol__state', {'type': 'null'}], ['__zon
e_symbol__value', {'type': 'array', 'value': []}]]}}}
DEBUG:bidi.proto:ws_reader: resolving pending command 33
DEBUG:bidi.proto:[id 33] unpacking raw result {'type': 'object', 'value': [['__zone_symbol__state', {'type': 'nul
l'}], ['__zone_symbol__value', {'type': 'array', 'value': []}]]}
INFO:bidi.command:→ {'realm': '8801c256-32fc-46fe-999e-5ef54f4a7446', 'type': 'success', 'result': {'__zone_symbo
l__value': []}}
INFO:bidi.command:← input.performActions({'context': '3964f4f7-f26a-4497-aaa1-ae68b2d4a0a7', 'actions': [{'id': '
pointer-34', 'type': 'pointer', 'parameters': {'pointerType': 'mouse'}, 'actions': [{'type': 'pointerMove', 'x':
0, 'y': 0, 'origin': {'type': 'element', 'element': {'__zone_symbol__value': []}}}, {'type': 'pointerDown', 'butt
on': 0}, {'type': 'pointerUp', 'button': 0}]}]}) [id 34]
```

martinpitt:
I'll take a look at this.

martinpitt:
The first crash is easy and my fault, I have a local fix. The SharedReference one is a real brain teaser, though -- Normally a `ph_find()` should return an actual [HTML Element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element), which looks like

    {'type': 'element', 'element': {'type': 'node', 'sharedId': '123....'}}

but here we get this `__zone_symbol__state` crap instead, which is of course not at all a (serialized) element node.

The same happens for filling out the user/password fields, but there we don't use the returned (bogus) value.

This stuff has a few google hits, but nothing too enlightening. https://github.com/angular/zone.js/issues/385 seems to suggest that it's part of Angular. 

I debugged `ph_select()`, and it does get regular elements for `input[name='user']` and password and also for the "Log in" button (via sizzle), which firefox/bidi then serializes to that junk.

This is weirdly related to `ph_find_scroll_into_view()`: If I return `el` directly it works (i.e. equivalent to calling `ph_find()`). The return value is correct when directly returning the element, but not when  returning it through a promise. `eval_js()` already sets `awaitPromise=True`.

With this:
```diff
--- test/common/testlib.py
+++ test/common/testlib.py
@@ -515,6 +515,8 @@ class Browser:
         # For Firefox and regular clicks, use the BiDi API, which is more realistic -- it doesn't
         # sidestep the browser
         element = self.call_js_func('ph_find_scroll_into_view', selector)
+        if element.get("type") != "node":
+            raise Error(f"Invalid serialized element: {element}")
 
         # btn=2 for context menus doesn't work with ph_mouse(); so translate the old ph_mouse() API
         if event == "contextmenu":
```

I get more info:

```
Invalid serialized element: {'__zone_symbol__state': True, '__zone_symbol__value': {'type': 'node', 'sharedId': 'e1d3df14-335e-49af-8fc3-3ff96e2c64eb', 'value': {'nodeType': 1, 'localName': 'button', 'namespaceURI': 'http://www.w3.org/1999/xhtml', 'childNodeCount': 1, 'attributes': {'class': 'css-1b7vft8-button', 'type': 'submit', 'data-testid': 'data-testid Login button', 'aria-disabled': 'false'}, 'shadowRoot': None}}}
```

So the inner value is actually correct, it's just wrapped in that weird zone thing. I'll try to unwrap this. It's not entirely trivial as sometimes it looks like this, i.e. an empty array:

```
{'type': 'object', 'value': [['__zone_symbol__state', {'type': 'null'}], ['__zone_symbol__value', {'type': 'array', 'value': []}]]}
```

jelly:
```
I debugged ph_select(), and it does get regular elements for input[name='user'] and password and also for the "Log in" button (via sizzle), which firefox/bidi then serializes to that junk.
```

Why would that go through sizzle?

```
document.querySelector("input[autocomplete='username']")
<input id="login-user-input" class="form-control" type="text" autocorrect="off" autocapitalize="none" autofocus="true" autocomplete="username">
```

I thought we only did sizzle when we have `:contains`:

```
window.ph_select = function(sel) {
    if (sel.includes(":contains(")) {
        if (!window.Sizzle) {
            throw new Error("Using ':contains' when window.Sizzle is not available.");
        }
        return window.Sizzle(sel);
    } else {
        return Array.from(document.querySelectorAll(sel));
    }
};
```

martinpitt:
I tried the unwrapping with

```diff
--- test/common/webdriver_bidi.py
+++ test/common/webdriver_bidi.py
@@ -151,6 +151,14 @@ def unpack_value(raw: Any) -> Any:
         obj = {}
         for k, v in raw["value"]:
             obj[k] = unpack_value(v)
+        # unwrap zoned results
+        if "__zone_symbol__state" in obj:
+            if obj["__zone_symbol__state"]:
+                log_proto.debug("unwrapping zoned value: %r", obj)
+                obj = obj["__zone_symbol__value"]
+            else:
+                log_proto.debug("ignoring invalid __zone_symbol__state %r", obj)
+                obj = None
         return obj
     raise ValueError(f"Unknown type in {raw!r}")
 
```

but that just moves the error -- the promise result just doesn't get serialized properly. I'll apply a different hack.

martinpitt:
@jelly The "Log in" (and other) buttons goes via sizzle. But I'm :100: certain this isn't related to sizzle.

martinpitt:
Ah! I bet Grafana is now using https://www.npmjs.com/package/zone.js (or something similar). This *patches* the global `Promise` object, and that's likely what's confusing the webdriver serialization.

jelly:
> Ah! I bet Grafana is now using https://www.npmjs.com/package/zone.js (or something similar). This _patches_ the global `Promise` object, and that's likely what's confusing the webdriver serialization.
> 
> In a JS console in Grafana:
> 
> ```
> >> p = new Promise((res, rej) => { pres = res; prej = rej});
> ← Object { __zone_symbol__state: null, __zone_symbol__value: [] }
> ​```
> 
> as opposed to the regular one:
> ```
> 
> > > p = new Promise((res, rej) => { pres = res; prej = rej});
> > > ← Promise { : "pending" }

Ugh custom Promises, who would ever do that!

PR #6904, opened 2024-09-23
opensuse tumebleweed: don't install libvirt-daemon-driver-interface

Lunarequest:
The maintainers of libvirt in opensuse have with the latest libvirt unconditionally disable building the interface driver this means the package `libvirt-daemon-driver-interface` is no longer in the repos. image-refresh is able to update the image with this change


Lunarequest:
I've run the test suite and https://github.com/cockpit-project/cockpit-machines/pull/1782 will need to be merged due to the dropping of `interfaced`

martinpitt:
Thanks @Lunarequest ! Copied as PR #6909 (because technical reasons..)

Issue #6792, opened 2024-08-26
SELinux denies swtpm access to /var/log/swtpm/libvirt/qemu/*.log


mvollmer:
Upstream: https://bugzilla.redhat.com/show_bug.cgi?id=2307853

martinpitt:
Is that a duplicate of #5944 aka. https://bugzilla.redhat.com/show_bug.cgi?id=2307853 ? Our naughty was closed just two weeks ago, but perhaps circumstances changed (like testing on RHEL 10?)

mvollmer:
> Is that a duplicate of #5944 aka. https://bugzilla.redhat.com/show_bug.cgi?id=2265346 ? 

I checked that and it doesn't mention /var/log/swtpm/libvirt/qemu, but maybe I was too quick...

mvollmer:
> > Is that a duplicate of #5944 aka. https://bugzilla.redhat.com/show_bug.cgi?id=2265346 ?
> 
> I checked that and it doesn't mention /var/log/swtpm/libvirt/qemu, but maybe I was too quick...

No, I think this here is new, and maybe a result of the fixes for [2265346](https://bugzilla.redhat.com/show_bug.cgi?id=2265346).

cockpituous:
fedora-41
Ooops, it happened again<hr>
```
# ----------------------------------------------------------------------
# testConfigureBeforeInstallBiosTPM (__main__.TestMachinesCreate.testConfigureBeforeInstallBiosTPM)
Starting ChromeDriver 127.0.6533.99 (f31af5097d90ef5ae5bd7b8700199bc6189ba34d-refs/branch-heads/6533@{#1910}) on port 52805
Only local connections are allowed.
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
ChromeDriver was started successfully.
[...]

PR #6920, opened 2024-09-26
testmap: Enable rhel-10-0 for c-machines

martinpitt:
The "wild west" days are over, and RHEL 10 is nearing its beta release. We need to properly test c-machines there as we release to it.


martinpitt:
Ack, needs #6792 naughty copied.

jelly (approved):
Thanks!

PR #6921, opened 2024-09-26
testmap: some update for sub-man & sub-man-cockpit

ptoscano:
- drop `rhel-9-5` from sub-man/main, as that branch tracks RHEL 10, and there is already `rhel-9-6` to check for backports
- add `rhel-9-6` to sub-man/1.29, as that branch tracks RHEL 9.6
- drop `rhel-9-5` from sub-man-cockpit/main, as there is already `rhel-9-6`


martinpitt (approved):
Thank you!

PR #6934, opened 2024-10-01
images: Drop libssh from Arch image

martinpitt:
Build and runtime dependency was dropped in
https://github.com/cockpit-project/cockpit/pull/19441

 * [x] image-refresh arch


cockpituous:
image-refresh arch done: https://github.com/cockpit-project/bots/commits/image-refresh-arch-20241001-050117

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-arch-cff7e34c-20241001-044931/log.html

martinpitt:
Note that it didn't actually drop yet from this image refresh, as the current cockpit package still pulls it in as a dependency. That's fine and expected. Once 326 lands, libssh will actually drop.

jelly (approved):

PR #6941, opened 2024-10-02
machine_core: toolbox compatibility with host session virtqemud

martinpitt:
If the user has libvirt installed on the host OS and the user virtqemud is already running, then launching a VM inside of toolbox doesn't work as the "outside" virtqemud cannot see the toolbox'es /var/tmp (that directory isn't shared with the host). Nevertheless, it's the right place for these files.

To fix that, go via the /run/host symlink to put the files on the host's /var/tmp. Conveniently, this path is also valid on the host (via a tmpfiles.d snippet shipped by toolbox), and there just points back to /.

-----

I don't have libvirt on the host, and it's not entirely easy for me to install, so I only regression-tested it. However, @jkonecny12 did a quick test with setting `$TEST_OVERLAY_DIR`, to confirm that it's really *only* that directory issue that broke it.

@ashley-cui ran into this problem as well a few weeks ago. I think this should reproduce it with main: Run this on the host (not toolbox):

```
pkill virtqemud
virsh list  # this auto-launches the session virtqemud
# ensure this finds something
pgrep -u $UID virtqemud
```

Now enter the cockpit toolbox, go to a bots checkout, and try `./vm-run fedora-40` (or any other image, doesn't matter). In theory this should fail for the two of you, with that "cannot find /var/tmp/blabla.qcow2" error. Now please check out this branch, and try again -- it should work now.

I'll trigger some tests to make sure this works on our CI.

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -305,7 +305,18 @@ class VirtMachine(Machine):
 
         Machine.__init__(self, image=image, **kwargs)
 
-        self.run_dir = os.path.join(os.getenv("TEST_OVERLAY_DIR", "/var/tmp"), "bots-run")
+        overlay_dir = os.getenv("TEST_OVERLAY_DIR")
+        if not overlay_dir:
+            # toolbox compatibility: /tmp is shared with the host, but may be too small for big overlays (tmpfs!)
+            # $HOME is shared, but we don't want to put our junk there (NFS, backups)
+            # /var/tmp is not shared with the host but the right place; just in case session libvirtd is already
+            # running, use the shared path so that the daemon can actually see our overlay.
+            # But this only makes sense if the host also has /run/host set up (toolbox ships a tmpfiles.d)
+            if os.path.exists("/run/host/var/tmp") and os.path.exists("/run/host/run/host"):
+                overlay_dir = "/run/host/var/tmp"
+            else:
+                overlay_dir = "/var/tmp"
+        self.run_dir = os.path.join(overlay_dir, "bots-run")
         os.makedirs(self.run_dir, 0o700, exist_ok=True)
 
         self.virt_connection = self._libvirt_connection(hypervisor="qemu:///session")


machine/machine_core/machine_virtual.py
```diff
@@ -305,7 +305,18 @@ def __init__(
...
 
-        self.run_dir = os.path.join(os.getenv("TEST_OVERLAY_DIR", "/var/tmp"), "bots-run")
+        overlay_dir = os.getenv("TEST_OVERLAY_DIR")
+        if not overlay_dir:
+            # toolbox compatibility: /tmp is shared with the host, but may be too small for big overlays (tmpfs!)
+            # $HOME is shared, but we don't want to put our junk there (NFS, backups)
+            # /var/tmp is not shared with the host but the right place; just in case session libvirtd is already
+            # running, use the shared path so that the daemon can actually see our overlay.
+            # But this only makes sense if the host also has /run/host set up (toolbox ships a tmpfiles.d)
+            if os.path.exists("/run/host/var/tmp") and os.path.exists("/run/host/run/host"):
```
jkonecny12, line 315: Maybe it is not an issue but I see these paths even on the host system. I guess it's just fine.

jkonecny12 (approved):
Tested and it works great! Thanks a lot!

machine/machine_core/machine_virtual.py
martinpitt: Yes, that's intentional. If you have toolbox installed, you should have a `/run/host → ../` symlink (and thus you can do infinite recursion :grin: ). Without toolbox installed, /run/host doesn't exist, but then you also can't run cockpit in toolbox.

ashley-cui:
Works for me!

machine/machine_core/machine_virtual.py
jkonecny12: Hmm, interesting. I wonder if the same applies with `distrobox`?

PR #6946, opened 2024-10-03
cockpit-lib-update: show the git shortlog in pull request body

jelly:
Show the git shortlog of Cockpit changes in the PR body, it would be nice to make this collapsible but that involves adding tons of HTML, see for example: https://github.com/cockpit-project/cockpit/pull/21053

---

Example PR https://github.com/jelly/cockpit-navigator/pull/1


martinpitt (approved):
This is nice! This could do with some formatting in the future.

PR #6974, opened 2024-10-10
machine_core: Add Machine nspawn clone API

martinpitt:
There are tests which require two independent machines in terms of networking, services, packages, and file systems -- this can cover e.g. the shell's host switching, or cockpit-machine's VM migration or remote terminal functionality.

Previously we used destructive tests with `provision`ing multiple VMs; but this is expensive and prevents us from running these tests in tmt. But for the above tests which don't require kernel modules/functionality, we would be quite happy with the auxiliary machines being a container.

systemd-nspawn supports booting its own parent OS/file system with a transient overlay. Wrap that functionality into a convenient `Machine.clone_container()` API which returns a new Machine (NspawnMachine subclass) with the usual methods. Teach `SSHConnection` about a jump host (using `ProxyCommand`) to directly connect from the outside test to the inner nspawn container.

This first version only supports one cloned container at a time. Supporting multiple would be nice, but requires some reengineering of the SSHConnection master.

---

I tested this locally, and also sent a Cockpit PR to exercise this: https://github.com/cockpit-project/cockpit/pull/21105

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -67,7 +67,8 @@ class Machine(ssh_connection.SSHConnection):
         identity_file: str | None = None,
         arch: str = "x86_64",
         ssh_port: int | str = 22,
-        web_port: int | str = 9090
+        web_port: int | str = 9090,
+        jumphost: str | None = None,
     ):
         identity_file_old = identity_file
         identity_file = identity_file or DEFAULT_IDENTITY_FILE
@@ -84,7 +85,7 @@ class Machine(ssh_connection.SSHConnection):
         if not label and image != "unknown":
             label = f"{image}-{ssh_address}-{ssh_port}"
 
-        super().__init__(user, ssh_address, ssh_port, identity_file, verbose=verbose, label=label)
+        super().__init__(user, ssh_address, ssh_port, identity_file, verbose=verbose, label=label, jumphost=jumphost)
 
         self.arch = arch
         self.image = image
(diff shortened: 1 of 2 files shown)


martinpitt:
So there are several problems with this -- [RHEL does not have networkd](https://cockpit-logs.us-east-1.linodeobjects.com/pull-21105-61d36c35-20241010-154738-centos-10-networking/log.html), and many/most OSes don't have a snapshot-capable file system, e.g. [debian-stable](https://cockpit-logs.us-east-1.linodeobjects.com/pull-21105-61d36c35-20241010-154726-debian-stable-networking/log.html). There, `systemd-nspawn --ephemeral` basically copies the entire file system and kills the VM. So we can't use that, but have to do something with `--overlay`, `--volatile`, etc. However,

```
# systemd-nspawn --machine=cockpit1 --volatile=overlay --network-veth --directory=/ --boot --bind /home
Spawning container on root directory is not supported. Consider using --ephemeral, --volatile=yes or --volatile=state.
```

`--volatile=yes` isn't supported properly by Debian (or any distro to date really), and `state` doesn't work either. The documentation of `--overlay` also says that this isn't supported with full OS.

martinpitt:
Next step -- more legwork with manual overlays, which should work fine on ext4 now. I triggered another round of testing. Remaining big issue is that neither RHEL nor CoreOS have systemd-networkd, which makes configuring the inner and outer veth awkward. Need to find a way to do this with NM or other tools.

machine/machine_core/machine.py
```diff
@@ -357,3 +358,93 @@ def curl(self, *args: str, headers: Mapping[str, str] = {}) -> str:
...
+                                        [ -z "$A" ] || break;
+                                        sleep 0.5;
+                                    done;
+                                    echo $A""", timeout=20).strip()
+
+        # wait until SSH is up, so that the ProxyCommand doesn't get stuck
+        parent.execute(f"until timeout 1 nc -z {address} 22; do sleep 1; done")
+
+        super().__init__(
+            address, parent.image, verbose, parent.label + "-" + label,
```
mvollmer, line 426: Looks like `self.verbose` should be initialized earlier, in case the code above throws an error.

martinpitt:
Some more ideas:

 - Unfortunately ssh [still cannot connect to a Unix socket](https://github.com/openssh/openssh-portable/pull/435). But we could play some tricks with socat to forward container port 22 to a host port via stdin/out or a unix socket?
 - Re-attempt to set up the veth pair manually with `ip` and a static IP, and figure out why the guest OS removes the IP. NM shouldn't touch veths.

martinpitt:
This would have been really neat, but not being able to run this on RHEL sensibly is a dealbreaker. So I'm admitting defeat here.

Issue #6992, opened 2024-10-16
firefox: NS_ERROR_FAILURE with Terminal canvas


mvollmer:
TestPages.testHistory seems to run into https://bugzilla.mozilla.org/show_bug.cgi?id=941146 on Firefox.

Example: https://cockpit-logs.us-east-1.linodeobjects.com/pull-21061-0bdc1495-20241015-154309-fedora-40-firefox-expensive/log.html#51-2

Browser console just shows

```
> error: NS_ERROR_FAILURE: 
```

Backtrace when run with TEST_SHOW_BROWSER:

```
    _fillCharTrueColor http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:8
    _renderBlockCursor http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:8
    _render http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:8
    value http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:8
    _animationFrame http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:8
    restartBlinkAnimation http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:8
    handleGridChanged http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:8
    renderRows http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:8
    _renderRows http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:12
    _renderDebouncer http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:12
    _innerRefresh http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:12
    _animationFrame http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:12
    refresh http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:12
    refreshRows http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:12
    refresh http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:12
    V http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:12
    fire http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:14
    parse http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:14
    _writeBuffer http://127.0.0.2:9091/cockpit/@localhost/system/terminal.js:14
[...]

PR #6993, opened 2024-10-16
naughty: Add pattern for #6992

mvollmer:


martinpitt (approved):
Ah, I thought it would fail reliably. That's fine then, thanks!

PR #6994, opened 2024-10-16
naughty: Fix pattern for #6992

mvollmer:
The text that it is matched against does not include the test name header...

This is not great, now we catch all Firefox internal errors that don't otherwise make the test fail...

### DIFF
--- a/naughty/fedora-40/6992-firefox-hidden-canvas-bug
+++ b/naughty/fedora-40/6992-firefox-hidden-canvas-bug
@@ -1,5 +1,3 @@
-# testHistory (__main__.TestPages.testHistory)
-*
 > error: NS_ERROR_FAILURE:
 *
 AssertionError: Cockpit shows an Oops


martinpitt:
That feels wrong -- why wouldn't the original match? https://cockpit-logs.us-east-1.linodeobjects.com/pull-21061-54a3ccd7-20241016-121730-fedora-40-firefox-expensive/log.html#62-2 does include the test name header.

martinpitt:
I copied the test output to /tmp/out and checked with `bots/test-failure-policy -o fedora-40 < /tmp/out`. It's not related to the header, but that there is apparently some invisible whitespace after the NS_ERROR_FAILURE. Try appending a `*` after it:


```
# testHistory (__main__.TestPages.testHistory)
*
> error: NS_ERROR_FAILURE:*
*
AssertionError: Cockpit shows an Oops
```

I sent that to #6995.

mvollmer:
> That feels wrong

Yes, you are right. I thought the "# testHistory ..." part would not be part of the test output, but it is. It comes from TapRunner.runOne.

PR #7007, opened 2024-10-18
recreate-dependabot-pr: re-create conflicting dependabot pr's

jelly:
This script should be run in a GitHub push workflow on main where it would iterate over all open conflicting dependabot pull requests and recreate the most recent conflicting one by commenting to let dependabot recreate the pull request. Additional the `no-test` label ensures we don't run unneeded tests as re-creating still requires us to update the `node_modules` git submodule.

The script has to retry the pull request detail endpoint as the `mergeable` state of a pull request can be `null` indicating GitHub is still determining if the pull request can be merged.

---

This automates one step which we do manually now, but does not yet set the `node_modules` label as for that we have to when a dependabot PR conflicts:

* Run this CI which recreates the PR
* Add a `node_modules` and remove a `no-test` label (which creates the new node_modules submodule and runs tests)

For this flow I already added a new label `dependabot-recreate` which we could use with another github pull_request workflow or the existing dependabot one to check if we need to add the `node_modules` label

### DIFF
--- a/recreate-dependabot-pr
+++ b/recreate-dependabot-pr
@@ -1,0 +1,82 @@
+#!/usr/bin/python3
+
+# This file is part of Cockpit.
+#
+# Copyright (C) 2024 Red Hat, Inc.
+#
+# Cockpit is free software; you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# Cockpit is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+
+# Update COCKPIT_REPO_COMMIT to cockpit HEAD automatically, defaults to
+# Makefile as input optionally the full path can be provided. (For example
+# Anaconda uses ui/webui/Makefile.am).
+
+import os
+import sys
+import time
+from typing import Optional
+
+import task
+from lib.aio.jsonutil import get_dict, get_int, get_str
+
+sys.dont_write_bytecode = True
+
+
+def main() -> None:
+    assert os.getenv('GITHUB_BASE') is not None, 'GITHUB_BASE must be set'
+
+    api = task.github.GitHub()
+    for pull in api.pulls(state="open"):
+        number = get_int(pull, 'number')
+        if number is None:
+            continue
+
+        user = get_dict(pull, 'user')
+        if get_str(user, 'login') != 'dependabot[bot]':
+            continue
+
+        pull_details = api.get(f'pulls/{number}')
+
+        # Ignore dependabot PR's with multiple commits, they might be work in progress.
+        if get_int(pull_details, 'commits') > 1:
+            print(f'Skipping pull {number}, it is being worked on')
+            continue
+
+        mergeable: Optional[bool] = pull_details['mergeable']
+        # State is unknown, retry with a timeout
+        if mergeable is None:
+            for retry in range(5):
+                pull_details = api.get(f'pulls/{number}')
+                mergeable = pull_details['mergeable']
+                if mergeable is not None:
+                    break
+
+                print(f'Retrying to obtain mergeable status for pull={number}, retry={retry}')
+                time.sleep(60)
+            else:
+                print(f'Reached timeout mergeable status still unknown for pull={number}')
+                return
+
+        if mergeable:
+            print(f'Skipping pull {number}, it is in a mergeable state')
+            continue
+        else:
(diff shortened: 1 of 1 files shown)


recreate-dependabot-pr
```diff
@@ -0,0 +1,78 @@
...
+
+# Update COCKPIT_REPO_COMMIT to cockpit HEAD automatically, defaults to
+# Makefile as input optionally the full path can be provided. (For example
+# Anaconda uses ui/webui/Makefile.am).
+
+import sys
+import time
+from typing import Optional
+
+import task
```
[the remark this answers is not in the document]
jelly: This is a bit pedantic for my taste

jelly:
This approach is currently blocked as when trying this out in my own fork here https://github.com/jelly/cockpit-navigator/pull/3#issuecomment-2422548303 dependabot won't recreate the PR as only users who have push permissions can. So do we need to comment as cockpituous?

jelly:
As we can't `recreate` the PR, what about we just set a label `dependabot-recreate` and then execute a workflow which does:

```
git reset HEAD^
git rebase --autostash main
./tools/node-modules install
./tools/node-modules push
git commit $orig-commit-msg
```

With this we directly solve the whole problem in one step, so this seems more elegant/beneficial.

martinpitt:
@jelly do you still want to go ahead with this, or is that superseded by https://github.com/cockpit-project/cockpit-files/pull/798 ?

jelly:
> @jelly do you still want to go ahead with this, or is that superseded by [cockpit-project/cockpit-files#798](https://github.com/cockpit-project/cockpit-files/pull/798) ?

We can still go ahead with this by adding a `dependabot` label if it starts to conflict automatically. Doesn't help us much but would if we fully automated approval + landing.

pyproject.toml
```diff
@@ -77,6 +77,7 @@ select = [
     "YTT",     # flake8-2020
 ]
 
+
```
martinpitt, line 80: spurious newline

recreate-dependabot-pr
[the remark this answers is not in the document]
martinpitt: Still, I agree. Let's use fully qualified form *or* specific imports, not both. Too confusing otherwise.

recreate-dependabot-pr
```diff
@@ -0,0 +1,81 @@
...
+
+
+def main() -> None:
+    assert os.getenv('GITHUB_BASE') is not None, 'GITHUB_BASE must be set'
+
+    api = github.GitHub()
+    for pull in api.pulls(state="open"):
+        number = get_int(pull, 'number')
+        if number is None:
+            continue
```
martinpitt, line 43: :eyes: how can this happen?

recreate-dependabot-pr
```diff
@@ -0,0 +1,81 @@
...
+        # State is unknown, retry with a timeout
+        if mergeable is None:
+            for retry in range(3):
+                pull_details = api.get(f'pulls/{number}')
+                mergeable = pull_details['mergeable']
+                if mergeable is not None:
+                    break
+
+                print(f'Retrying to obtain mergeable status for pull={number}, retry={retry}')
+                time.sleep(5)
```
martinpitt, line 66: Is 15 seconds enough for this?

To accomodate a timeout, can you please add an `else:` to the `for` loop that fails the script with some "timeout" error?

recreate-dependabot-pr
```diff
@@ -0,0 +1,81 @@
...
+
+        if mergeable:
+            print(f'Skipping pull {number}, it is in a mergeable state')
+            continue
+        else:
+            # Not mergeable and a dependabot PR, add a `no-test` label and re-create the PR.
+            task.label(pull_details, ('no-test', 'dependabot-recreated'))  # type: ignore[no-untyped-call]
+            task.comment(pull_details, '@dependabot recreate')  # type: ignore[no-untyped-call]
```
martinpitt, line 74: This should now be replaced with simply adding a `node_modules` label, right? then our workflow will do the rebasing and pushing.

martinpitt (changes requested):
Thanks! This is actually quite nice and mostly done.

recreate-dependabot-pr
jelly: It can't be we are going full mypy so we have to check for this.

martinpitt: I see!

recreate-dependabot-pr
jelly: I am not sure if 15 seconds is enough, it would run on every push so we can make it longer? Any preferred time?

recreate-dependabot-pr
jelly: Indeed!

martinpitt (approved):
Thanks, looks good! As it's just a new script, it's harmless by itself. We may need some more cleverness at some point, such as adding `no-test` if it's only a devDependency

jelly:
> Thanks, looks good! As it's just a new script, it's harmless by itself. We may need some more cleverness at some point, such as adding `no-test` if it's only a devDependency

FYI, my local patch: 
[0001-workflows-add-dependabot-recreate-pull-request-workf.patch.txt](https://github.com/user-attachments/files/17617904/0001-workflows-add-dependabot-recreate-pull-request-workf.patch.txt)

PR #7026, opened 2024-10-24
cockpit-lib-update: Restrict changelog to shared files

martinpitt:
Only show the changes that concern the files that are relevant for external projects.

### DIFF
--- a/cockpit-lib-update
+++ b/cockpit-lib-update
@@ -62,7 +62,8 @@ def run(context, verbose=False, **kwargs):
         subprocess.check_call(['git', 'clone', cockpit_repo_url, clone_dir], cwd=tmpdir)
         git_describe = subprocess.check_output(['git', 'describe'], cwd=tmpdir / clone_dir).decode().strip()
         git_head = subprocess.check_output(['git', 'rev-parse', 'HEAD'], cwd=tmpdir / clone_dir).decode().strip()
-        git_shortlog = subprocess.check_output(['git', 'shortlog', f'{commit}...'],
+        git_shortlog = subprocess.check_output(['git', 'shortlog', f'{commit}...', '--',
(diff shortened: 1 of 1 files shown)


cockpit-lib-update
```diff
@@ -62,7 +62,7 @@ def run(context, verbose=False, **kwargs):
         subprocess.check_call(['git', 'clone', cockpit_repo_url, clone_dir], cwd=tmpdir)
         git_describe = subprocess.check_output(['git', 'describe'], cwd=tmpdir / clone_dir).decode().strip()
         git_head = subprocess.check_output(['git', 'rev-parse', 'HEAD'], cwd=tmpdir / clone_dir).decode().strip()
-        git_shortlog = subprocess.check_output(['git', 'shortlog', f'{commit}...'],
+        git_shortlog = subprocess.check_output(['git', 'shortlog', f'{commit}...', '--', 'pkg/lib', 'test/common'],
```
mvollmer, line 65: We copy pkg/lib, test/common, test/static-code, and tools/node-modules in cockpit-machines and cockpit-podman, and probably in all other external projects. _Technically_, this list is determined by the Makefiles of each external project...

We could make it possible to specify the list as arguments to this script and then change all cockpit-lib-update.yml to pass their own. But that's probably too much work. This here is only informational anyway, and pkg/lib and test/common is the most important part.

mvollmer (dismissed):
Thanks!

cockpit-lib-update
martinpitt: You are right -- I added test/static-code (as that's very relevant and we change that often) and tools/node-modules (doesn't change often, but meh). Indeed that's still just an approximation, but:

mediocrates

jelly (approved):
Thanks!

PR #7035, opened 2024-10-25
Image refresh for rhel-9-6

cockpituous:
Image refresh for rhel-9-6
 * [x] image-refresh rhel-9-6


cockpituous:
image-refresh rhel-9-6 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-9-6-20241025-225642

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel-9-6-cd6f1d25-20241025-224033/log.html

martinpitt:
[kdump failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7035-03cd81e7-20241025-225812-rhel-9-6-expensive-cockpit-project-cockpit/log.html) which is notoriously brittle. Retrying once for comparison.

martinpitt:
I confirmed that it really is #6769, I copied the naughty. This often fails with the current image as well, so it's not a clean regression, just apparently got a bit worse  -- enough to not get green any more.

jelly (approved):
Seen kdump fail a bunch of times too :(

mvollmer:
> Seen kdump fail a bunch of times too :(

Should we dump kdump? Srsly, I don't think we should lure people into thinking it's easy to setup by giving them some happy-clicky UI. It's not a suitable foundation to build a good user experience on...

martinpitt:
https://bugzilla.redhat.com/show_bug.cgi?id=2305968 is indeed a rather big dealbreaker. But it would break CLI or Ansible based setups just as well, no? I.e. the bug isn't really in our UI, but in the kdump runtime.

That said, I don't have a good gut feeling how many users/customers actually use the kdump page. We could certainly start to deprecate (i.e. drop) it in rawhide and RHEL 11 if we think it's unfixable.

PR #7036, opened 2024-10-26
machine_core: Explicitly reset cockpit.socket in start_cockpit()

martinpitt:
testClientCertAuthentication(), with its rapid start/stop of cockpit.socket, still runs into start-limit-hit errors. This is unfortunately wildly unpredictable [1].

Commit 2a94280e67a4 attempted to fix this, but it was wrong and reverted in commit 314032a66d2.

To be more robust, combine the current and 2a94280e67a4 approach and use both the glob (covering all loaded units) and also explicitly cockpit.socket (covering the unloaded case).

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2303828

-----

This still [keeps failing in RHEL 10 gating tests](https://artifacts.osci.redhat.com/testing-farm/deb4a51e-c7f1-440d-9033-f89b21cafb51/), but *nowhere else*, so this is a nuisance to debug. This commit is now very defensive and at least should not break anything. We'll see if it actually works in the next downstream release.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -265,6 +265,8 @@ class Machine(ssh_connection.SSHConnection):
             systemctl stop --quiet cockpit.service
             rm -f /etc/systemd/system/cockpit.service.d/notls.conf
             systemctl reset-failed 'cockpit*'
+            # unpredictable, see https://bugzilla.redhat.com/show_bug.cgi?id=2303828
+            systemctl reset-failed cockpit.socket 2>/dev/null || true
             systemctl daemon-reload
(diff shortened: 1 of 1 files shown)


jelly (approved):

martinpitt:
This was unfortunately still not sufficient, it [failed again](https://artifacts.osci.redhat.com/testing-farm/f4a46ac7-49ff-4b38-846f-957d5dcfbd98/). I think we just need to give up and treat this particular test specially, and reset the heck out of everything.

martinpitt:
Reverted in #7085

PR #7049, opened 2024-10-30
Enable VM console capturing by default

martinpitt:
Commit d60055284 added an opt-in for console capturing. However, we keep forgetting to specify that, so that we end up with VM boot failures without knowing why. https://github.com/cockpit-project/cockpit/issues/21165

Flip the default to "True" and opt out of console capturing in `vm-run`. The latter is the only situation where we care about an interactive console.

---

This should shed some light on [these failures](https://cockpit-logs.us-east-1.linodeobjects.com/pull-0-e6db1ad8-20241030-013154-fedora-40-daily/log.html#75), see the referenced issue.

### DIFF
--- a/image-create
+++ b/image-create
@@ -37,7 +37,6 @@ parser.add_argument('-v', '--verbose', action='store_true', help='Display verbos
 parser.add_argument('-s', '--sit', action='store_true', help='Sit and wait if setup script fails')
 parser.add_argument('-n', '--no-save', action='store_true', help='Don\'t save the new image')
 parser.add_argument('-u', '--upload', action='store_true', help='Upload the image after creation')
-parser.add_argument('--capture-console', action='store_true', help='Capture VM console to stderr')
 parser.add_argument('--no-build', action='store_true', dest='no_build',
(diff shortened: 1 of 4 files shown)


martinpitt:
I triggered two "full" cockpit runs. [this](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7049-2a2ac73f-20241030-075907-fedora-40-cockpit-project-cockpit/log.html) was (un)lucky and didn't have any flakes, but [this one](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7049-2a2ac73f-20241030-085356-fedora-40-updates-testing-cockpit-project-cockpit/log.html#88) reproduces the nightly failure and now shows the boot log. This clearly shows the OOM killer, which is useful to track down the actual bug.

mvollmer (approved):
Thanks!

mvollmer:
"Ensure branch was proposed from origin" has failed...

martinpitt:
Yes, I forgot to push this to origin instead, but this only affects tests..

PR #7062, opened 2024-11-03

cockpituous:

Issues

* [ ] pixel test changes in [cockpit-ostree](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7062-ac5e211c-20241104-110817-fedora-coreos-cockpit-project-cockpit-ostree/pixeldiff.html#OstreeRestartCase-testOstree-status)
* [x] Login issues in Cockpit


cockpituous:
image-refresh fedora-coreos done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-coreos-20241103-224908

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-fedora-coreos-d44635dd-20241103-224105/log.html

martinpitt:
Ah, coreos moved to Fedora 41, so we need to update the build image correspondingly.

jelly:
It's Monday, I accidentally triggered `image:fedora-41` so that's how fedora41@candlepin ended up there.

martinpitt:
Ugh:

> beiboot.py: error: argument --remote-bridge: invalid choice: 'supported' (choose from 'auto', 'never', 'always')

That feels like it's really running against an old cockpit-bridge package version instead of code from main

jelly:
Needs to be landed with https://github.com/cockpit-project/cockpit/pull/21217

jelly:
We can also land the "restart related" changes separately and then we can land things in sections

machine/machine_core/machine.py
```diff
@@ -290,8 +290,10 @@ def restart_cockpit(self) -> None:
         """Restart Cockpit.
         """
         if self.ws_container:
-            self.execute(f"podman restart {self.get_cockpit_container()}")
-            self.wait_for_cockpit_running()
+            cockpit_container = self.get_cockpit_container()
```
martinpitt, line 293: TBH this doesn't make sense to me -- why/how was it working before? This smells more like an assumption that you can't call  `restart_cockpit()` before you started it once, but that shouldn't change between the two ostree versions?

jelly: This was working before as Fedora-41 had multihost enabled and enabling it for non-default enabled machines n  `test/verify/check-shell-host-switching` calls `enable_multihost`:
```
            machine.write("/etc/cockpit/cockpit.conf",
                          '[WebService]\nAllowMultiHost=yes\n')
            machine.restart_cockpit()
```

We restart while the test never started the container:

```
        b = self.browser
        m1 = self.machines['machine1']
        m2 = self.machines['machine2']
        m3 = self.machines['machine3']

        self.enable_multihost(m1)
```

martinpitt (approved):
Thanks! Please land the two PRs simultaneously, and then send a c-ostree PR for adjusting the pixel test.

PR #7064, opened 2024-11-04
naughty: Add variant of #6792 for new test

martinpitt:
https://github.com/cockpit-project/cockpit-machines/pull/1796 adds TPM testing to TestMachinesSettings.testMultipleSettings, which also fails due to #6792. Add a corresponding pattern.

---

Tested locally and via additional contexts in the above PR.


jelly:
Waiting on https://cockpit-logs.us-east-1.linodeobjects.com/pull-1796-b176cdc5-20241104-143848-rhel-10-0-bots%237064/log.html to succeed

PR #7086, opened 2024-11-11
machine: don't try to restart a non-existing container

jelly:
Now that Fedora-Cores rebased on Fedora 41 no longer enables multihost support by default. Enabling this and blanket restarting the container in TestMultiMachines.testEdit fails as the ws container is not running yet.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -287,8 +287,10 @@ class Machine(ssh_connection.SSHConnection):
         """Restart Cockpit.
         """
         if self.ws_container:
-            self.execute(f"podman restart {self.get_cockpit_container()}")
-            self.wait_for_cockpit_running()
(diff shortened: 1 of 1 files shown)


machine/machine_core/machine.py
```diff
@@ -287,8 +287,10 @@ def restart_cockpit(self) -> None:
         """Restart Cockpit.
         """
         if self.ws_container:
-            self.execute(f"podman restart {self.get_cockpit_container()}")
-            self.wait_for_cockpit_running()
+            cockpit_container = self.get_cockpit_container()
+            if cockpit_container != "":
```
martinpitt, line 291: empty string comparison instead of just truthy test feels a bit brittle, but ok

martinpitt (approved):
Thanks! I triggered affected tests.

PR #7090, opened 2024-11-12
naughty: add nfs-server hang naughties for cockpit-machines

jelly:
Known issue #7088
Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2325556


martinpitt (approved):
Thanks! So these are not the ones breaking C10s then?

PR #7141, opened 2024-11-25
Image refresh for centos-9-bootc

cockpituous:
Image refresh for centos-9-bootc


cockpituous:
image-refresh centos-9-bootc done: https://github.com/cockpit-project/bots/commits/image-refresh-centos-9-bootc-20241125-230732

martinpitt:
Ugh, that breaks login with the ws container. Taking a look

martinpitt:
This trips over `cockpit-session` not being in the container any more. Seems like fallout from https://github.com/cockpit-project/cockpit/pull/21291 ? It succeeded there in fedora-coreos, so something is subtly different in centos-9-bootc?

One obvious difference is that our fcos image doesn't have any cockpit packages preinstalled, while c9-bootc has cockpit-{system,bridge}. But I don't believe that, as it doesn't even get that far, it seems to already fail during the initial opportunistic kerberos authentication. 

I.e. it feels like the cockpit/ws container was built with https://github.com/cockpit-project/cockpit/commit/035dc2d9be846320bbe4b2be75c3227793735ed8 but *without* https://github.com/cockpit-project/cockpit/commit/d7cb5617c007e5730742cd36d424af2d5bf46213 -- i.e. old  cockpit packages. But `/usr/libexec/cockpit-ws --version` says "329" :thinking:

martinpitt:
Oh, of course! These two commits are *not* yet in release 329. But our workflow to rebuild the container always uses the Dockerfile from main.

I reverted the :latest tag back to the previous version, so I'll just re-do this image refresh. I'll send a cockpit PR to build the container with the correct Dockerfile.

martinpitt:
Trivially reproducible with `sudo podman container runlabel RUN quay.io/cockpit/ws:329`. :327 works.

martinpitt:
Fixed in https://github.com/cockpit-project/cockpit/pull/21331

PR #7143, opened 2024-11-26
Add testing support for OpenSUSE Tumbleweed

SludgeGirl:
This pairs with cockpit-project/cockpit#21335 

This is aimed to add most of the support for testing Tumbleweed, let me know if you need any extra changes made to it

### DIFF
--- a/images/scripts/opensuse-tumbleweed.setup
+++ b/images/scripts/opensuse-tumbleweed.setup
@@ -98,7 +98,11 @@ if [ "${IMAGE%-i386}" != "$IMAGE" ]; then
 fi
 
 zypper dup -y
-# kernel-default has more kernel modules that are required for testing
+
+# HACK: There are a number of kernel modules needed for testing, for
+# instance scsi_debug which is needed for most if not all of the disk
+# based tests. Since the cloud image ships a slimmed down set of modules
+# We need to install the full set, this is still the same kernel
 zypper remove -y kernel-default-base
 zypper install -y kernel-default
 
--- a/images/scripts/opensuse-tumbleweed.setup
+++ b/images/scripts/opensuse-tumbleweed.setup
@@ -21,11 +21,15 @@ printf 'dictcheck = 0\nminlen = 6\n' >> /etc/security/pwquality.conf
 COCKPIT_DEPS="\
 criu \
 libcriu2 \
+crypto-policies-scripts \
 device-mapper \
 glibc-locale \
 glib-networking \
 json-glib \
 kexec-tools \
+krb5 \
+lastlog2 \
+libssh2-1 \
 libvirt-daemon-config-network \
 libvirt-daemon-driver-qemu \
 libvirt-daemon-driver-network \
@@ -38,8 +42,10 @@ libvirt-dbus \
 openssl \
 PackageKit \
 pcp \
-pcp-devel \
+pkexec \
+polkit \
 python3-pip \
+python3-pcp \
 qemu-block-curl \
 qemu-chardev-spice \
 qemu-hw-usb-host \
@@ -51,17 +57,23 @@ virt-install \
 virtiofsd \
 udisks2 \
 libudisks2-0_lvm2 \
+libudisks2-0_btrfs \
+setroubleshoot-server \
+wtmpdb \
 "
 
 TEST_PACKAGES="\
 acl \
 ansible-core \
+bind-utils \
+certmonger \
 clevis-luks \
 cryptsetup \
 firewalld \
 gdb \
 gettext \
 gvfs \
+krb5-client \
 libvirt-daemon-driver-storage-iscsi \
 libvirt-daemon-driver-storage-iscsi-direct \
 libvirt-daemon-driver-storage-logical \
@@ -69,8 +81,9 @@ libxml2-tools \
 ltrace \
 nginx \
 nfs-kernel-server \
+patterns-base-fips \
 podman \
-redis \
+valkey \
 socat \
 strace \
 targetcli \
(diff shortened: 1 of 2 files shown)


images/scripts/opensuse-tumbleweed.setup
```diff
@@ -114,6 +115,11 @@ fi
 
 zypper dup -y
 
+# cloud-init setups the wheel group incorrectly so we need to recreate it
+groupdel wheel
+zypper install -y system-group-wheel
+sudo usermod -a -G wheel admin
```
martinpitt, line 130: This also feels like working around the OS, and should get a bug reference. In what way is "wheel" wrong? I.e. in a few months time how can developers decide if this can be dropped?

images/scripts/lib/build-deps.sh
```diff
@@ -37,7 +37,7 @@ echo "$spec" | rpmspec -D "$OS_VER_NO_VARIANT" -D 'version 0' -D 'enable_old_bri
 # - nodejs for starter-kit and other projects which rebuild webpack during RPM build
 case "$OS_VER" in
     *suse*)
-        EXTRA_DEPS="appstream-glib rpmlint gettext-runtime desktop-file-utils nodejs-default"
+        EXTRA_DEPS="appstream-glib rpmlint gettext-runtime desktop-file-utils nodejs-default system-group-wheel"
```
martinpitt, line 40: This is for `configure.ac` to auto-detect `--with-admin-group=` presumably?

images/scripts/opensuse-tumbleweed.setup
```diff
@@ -145,6 +145,18 @@ useradd -c Builder -G mock builder
...
 
+# Setup redis
+sudo cp /etc/redis/default.conf.example /etc/redis/redis.conf
+sudo chown redis:redis /etc/redis/redis.conf
+
+mkdir -p /etc/systemd/system/redis.target.d
+cat > /etc/systemd/system/redis.target.d/70-redis-targets.conf <<EOF
+[Unit]
+Wants=redis@redis.service
+EOF
```
martinpitt: Is that documented in TW somehow? That's all stuff that should either happen by default (to get an useful experience out of the box) or in Cockpit's metrics page. It doesn't help users if tests go green with "it works against a hacked up image", but not against a default install.

images/scripts/opensuse-tumbleweed.setup
```diff
@@ -145,6 +145,18 @@ useradd -c Builder -G mock builder
...
+
+systemctl enable --now redis.target
```
martinpitt: I'd advise against that. We don't enable it in Fedora/RHEL images by default, and even disable it in Debian. Most tests don't need it, and Cockpit's metrics page starts it if/when necessary.

images/scripts/opensuse-tumbleweed.setup
```diff
@@ -112,6 +112,9 @@ wireguard-tools \
 mkdir -p /etc/NetworkManager/conf.d/
 printf '[main]\nno-auto-default=*\n' > /etc/NetworkManager/conf.d/noauto.conf
 
+# use unpredictable network interface names
+ln -s /dev/null /etc/systemd/network/99-default.link
```
martinpitt, line 116: This is another case of "please test the real-world default OS, not some hack". We fixed our projects to get along without hardcoded ethernet names a while ago, this really shouldn't be necessary.

martinpitt (changes requested):
Thanks! Initial review -- none of that is a veto of course, in the end this is "your" image, but let's at least discuss this somewhat.

Can you please squash this? Like, put all the new dependencies into one commit, and squash the redis/valkey commits, and such.

SludgeGirl:
Hey @martinpitt thanks for the review! I'll get the history tidied up and I'll take another look into everything you've mentioned here and see what I can get handled properly. Thanks for your time on this!

SludgeGirl:
Morning! Sorry for the wait, so thanks for the push back, I get a little more the more correct ways of handling this. I've been a touch ruthless and pretty much dropped most of what you mentioned above that didn't break double figures of tests with the exception of switching over to kernel-default. This is the same kernel just with more kernel modules (specifically a few needed ones for testing), so it's a hack we'll have to bare for now

In terms of redis/valkey handling, it doesn't get shipped with a default conf, I'm going to have some chats my side and get some further information/see what we can do about it, so I've dropped it for now. For all of the other points I agree with opinions on it, I'll drop it and see if I can find different fixes in the future none of these broke *that many* tests

images/scripts/opensuse-tumbleweed.setup
```diff
@@ -21,11 +21,15 @@ printf 'dictcheck = 0\nminlen = 6\n' >> /etc/security/pwquality.conf
 COCKPIT_DEPS="\
 criu \
 libcriu2 \
+crypto-policies-scripts \
```
martinpitt, line 24: Please add "... on openshift-tumbleweed" to the "fix missing dependencies" commit.

martinpitt:
I did the above two suggested changes and build the image in #7170. Once that's done, I'll push it here.

martinpitt (approved):
Works, and LGTM. @SludgeGirl please cross-check my changes, and if you :+1: I'll land.

martinpitt:
@Nykseli Just saw that these are actually your commits, so if you want to cross-check, please do. Thanks!

martinpitt:
Let's not get this stale. I'll land it now, and of course we can do further adjustments at any time. But hopefully this already unblocks step 2 in https://github.com/cockpit-project/cockpit/pull/21335#pullrequestreview-2467200934 ?

SludgeGirl:
Your changes look good! Sorry just hopped back on

PR #7158, opened 2024-11-29
lib: Move default OS to Fedora 41

martinpitt:
- [x] https://github.com/cockpit-project/cockpit/pull/21358


martinpitt:
Argh, there is heavy interference from the "cockpituous" test (which we wouldn't even need here..) Need to let that finish first before triggering integration tests. Anyway, I let the remaining ones run to spot errors.

martinpitt:
[This failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7158-a3fe3717-20241129-101230-fedora-41-firefox-other-cockpit-project-cockpit/log.html#87-2) looks relevant, everything else works. It failed in two different ways on CI, and locally it fails in a third way. I'll have a look.

jelly (approved):

PR #7182, opened 2024-12-05
testmap: Add 'efi' firmware test scenario for anaconda-webui

KKoukiou:
Introduced in https://github.com/rhinstaller/anaconda-webui/pull/542


martinpitt (approved):
Trivial, but please only land after https://github.com/cockpit-project/cockpit/pull/21386 goes green and lands.

PR #7196, opened 2024-12-08
Two ruff fixes

martinpitt:
The `abc` one now [happens with latest tasks container](https://github.com/cockpit-project/bots/actions/runs/12216771504/job/34080229123?pr=7193). I took the liberty to look a bit further ahead and fix the things with ruff 0.8 as well (`sudo pip install ruff`).

This also depends on bringing back pytest-cov in https://github.com/cockpit-project/cockpituous/pull/632


jelly (approved):
BTW, the regex raw string thing will cause major fallout in Cockpit

PR #7211, opened 2024-12-11
naughty: Move #6992 to Fedora 41

martinpitt:
This doesn't depend on the guest OS, but on which OS we run Firefox tests. Since commit a8e378fbe86 this is Fedora 41.

---

See [this failure](https://cockpit-logs.us-east-1.linodeobjects.com/pull-21419-c7bdffce-20241211-103032-fedora-41-firefox-expensive/log.html#58)


mvollmer (approved):
Thanks!

PR #7257, opened 2024-12-26

cockpituous:

### DIFF
--- a/images/debian-testing
+++ b/images/debian-testing
@@ -1,1 +1,1 @@
-debian-testing-058cd7b6a10d185ecdbf15a46eabd985c8ad6432f039ce723be2341d25b41818.qcow2
+debian-testing-88d6b3f54644f7e04be2927d5b769608a5faf8af93dcd6b27d8fb08c4276f01e.qcow2
--- a/naughty/debian-testing/7284-mdadm-bitmap
+++ b/naughty/debian-testing/7284-mdadm-bitmap
@@ -1,0 +1,5 @@
+Traceback (most recent call last):
+  File "test/verify/check-storage-mdraid", line *, in testBitmap
+    self.assertIn("Consistency Policy : bitmap", m.execute("mdadm --misc -D /dev/md/md0"))
+*
+AssertionError: 'Consistency Policy : bitmap' not found in *Consistency Policy : resync*


cockpituous:
image-refresh debian-testing done: https://github.com/cockpit-project/bots/commits/image-refresh-debian-testing-20241226-225335

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-debian-testing-4a54b2bb-20241226-223923/log.html

martinpitt:
[This regression/change](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7257-cc098834-20241226-225502-debian-testing-storage-cockpit-project-cockpit/log.html) smells real, but in lazy holiday mode let's do a retry first and compare.

jelly:
Same issue again, so needs work  investigation.

martinpitt:
* Still fails with latest testing
 * Not a race, stays at "Consistency Policy : resync" indefinitely
 * Updating *only* mdadm on the current image to 4.3+20241202-1.1 is enough to trigger this; kernel etc. are unrelated.

Reproducer:
```sh
truncate --size=110GB /var/tmp/loop1
truncate --size=110GB /var/tmp/loop2
losetup -P --show --find /var/tmp/loop1
losetup -P --show --find /var/tmp/loop2
mdadm --create md0 --level=1 --assume-clean --run --raid-devices=2 /dev/loop0 /dev/loop1

mdadm --misc -D /dev/md/md0

# clean up
mdadm --stop /dev/md/md0
```

However, the manpage still says

> When  creating an array on devices which are 100G or larger, mdadm automatically adds an internal bitmap as it will usually be beneficial.

And creating with  `--bitmap=internal` still works.

Likely candidate is https://github.com/md-raid-utilities/mdadm/commit/581ba1341017943d740ed8163d68f04671c2fdde

Reported to https://github.com/md-raid-utilities/mdadm/issues/138 and https://bugs.debian.org/1092137. Added naughty tracker #7284, adding pattern here now.

mvollmer:
I am afraid the change in mdadm behavior might be intentional: https://github.com/md-raid-utilities/mdadm/pull/133/commits/3c5d13e3673251904b843e67846d68958f9800d2#diff-00ce713d52a49e6029b8af31cdc42d04b9062d1040d43c7fce63ef373e8167a6

    Instead of auto-forcing bitmap only for large arrays, it is more
    reasonable to let user do the chooice if bimtap is not set.

I don't agree with the reasoning, fwiw, (and the man page needs to be updated) but we might need to adapt to this.

I think we can keep Cockpit's current behavior and just change the expectations in the test.

mvollmer (commented):
Thanks!

mvollmer (approved):
Thanks!

PR #7271, opened 2025-01-01
images: Drop Fedora 39

martinpitt:
It reached its end of live in November 2024 already: https://fedorapeople.org/groups/schedule/f-39/f-39-all-tasks.html


jelly (approved):
New year, new cleanups? :)

PR #7318, opened 2025-01-17
run-queue: Automatically prune the test results database

martinpitt:
If test-results.db gets too big (> 25 MB), prune it to the last 14 days. We are occasionally doing this manually, but not often enough -- it routinely grows to > 100 MB, which makes the weather report take ages to load.

The database shrinks to about 15 MB after being freshly pruned, so 25 MB gives enough margin to avoid running the pruning too often (as it's a bit expensive).

---

I tuned this with getting https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/test-results.db (currently 170 MB). Running `./prometheus-stats --db /tmp/test-results.db --prune 14` shrinks it to 15 MB.

I just made a backup at https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/test-results.db.20250117 in case anything goes wrong.

### DIFF
--- a/run-queue
+++ b/run-queue
@@ -74,8 +74,12 @@ def consume_webhook_queue(dq: distributed_queue.DistributedQueue) -> ConsumeResu
             sha = pull_request['head']['sha']
             db = os.path.join(get_images_data_dir(), "test-results.db")
             body = {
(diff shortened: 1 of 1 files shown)


jelly (approved):
Thanks!

PR #7332, opened 2025-01-21
chore: echo userpass for remote access

Venefilyn:
This adds username and password in the remote message that is outputted when running something like `bots/vm-run`. We already have this listed in the documentation but it doesn't hurt to have in the terminal.

See https://github.com/cockpit-project/cockpit/pull/21541

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -36,7 +36,9 @@ SSH ACCESS
   $ ssh -p {ssh_port} -i bots/machine/identity {ssh_user}@{ssh_address}
 
 COCKPIT
-  http://{web_address}:{web_port}
+  https://{web_address}:{web_port}
+  Username: admin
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! Not sure what the "chore:" is supposed to mean, it should be "machine_core:" (or "machine:" perhaps). But we can change this with squash&merge.

Venefilyn:
> Thanks! Not sure what the "chore:" is supposed to mean, it should be "machine_core:" (or "machine:" perhaps). But we can change this with squash&merge.

@martinpitt I've been doing my messages with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) so far. Missed that there is a [different usage in Cockpit](https://github.com/cockpit-project/cockpit/wiki/Workflow#git-related--merging-conventions) my bad :pray:

PR #7361, opened 2025-01-28
machine_core: Add Machine.ws_container runtime detection

martinpitt:
We want to introduce proper bootc image testing for cockpit. In that scenario we want to install the actual rpm, not use the cockpit/ws container. This conflicts with cockpit-ostree's current assumption that the centos-9-bootc image always uses the ws container.

Thus make this check dynamic: If cockpit-ws package is installed (check for `/usr/share/cockpit/static` as that's the same on all distros and a cheap test), use it, otherwise use the ws container. Keep the `TEST_SCENARIO=ws-container` check, at least until we update cockpit-ostree's image preparation to remove cockpit-ws.rpm.

We can't yet run `.execute()` in the Machine constructor, so delay the detection by making `ws_container` a (cached) property.

---

This will be used in https://github.com/cockpit-project/cockpit/pull/21559 . I triggered a set of ostree/bootc/standard RPM tests here to check for regressions.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -20,6 +20,7 @@ import os
 import re
 import subprocess
 from collections.abc import Collection, Mapping, Sequence
+from functools import cached_property
 
 from lib.constants import BOTS_DIR, DEFAULT_IDENTITY_FILE, OSTREE_IMAGES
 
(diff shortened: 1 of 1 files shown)


martinpitt:
Cleaned up the obsolete cached value (replaced by functools)

jelly (approved):

PR #7380, opened 2025-01-31
testmap: Enable centos-9-bootc for cockpit

martinpitt:
Added in https://github.com/cockpit-project/cockpit/pull/21559


mvollmer (approved):
Thanks!

PR #7419, opened 2025-02-11
testmap: Drop fedora-40 from our projects, enable fedora-42 on c-files and c-podman

martinpitt:
We stopped releasing to Fedora 40, so we can also stop testing.


allisonkarlitskaya (approved):
Thanks!  Not go back to bed!  :sweat_smile:

PR #7424, opened 2025-02-11
lib/testmap: drop redundant fedora-42 _manual entries

allisonkarlitskaya:
f583e4966bc0 ("testmap: Enable fedora-42 on c-files and c-podman") added `fedora-42` to `main` but forgot to drop it from `_manual`.


martinpitt (approved):
Whoops, thanks!

PR #7441, opened 2025-02-14
test: Give anaconda tests an additional fedora-wiki secret

martinpitt:
This secret was added in
https://github.com/cockpit-project/cockpituous/pull/642

### DIFF
--- a/tests-scan
+++ b/tests-scan
@@ -241,6 +241,10 @@ def cockpit_tasks(api: github.GitHub, contexts: Sequence[str], opts: argparse.Na
                 env["BASE_BRANCH"] = branch
             if number:
                 env["TEST_PULL"] = f'{number}'
+            # for updating naughty trackers and downloading private images
+            secrets = ["github-token", "image-download"]
+            if api.repo == "rhinstaller/anaconda-webui":
+                secrets.append("fedora-wiki")
 
             yield {
                 "job": {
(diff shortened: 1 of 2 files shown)


tests-scan
```diff
@@ -241,6 +241,10 @@ def cockpit_tasks(api: github.GitHub, contexts: Sequence[str], opts: argparse.Na
                 env["BASE_BRANCH"] = branch
             if number:
                 env["TEST_PULL"] = f'{number}'
+            # for updating naughty trackers and downloading private images
+            secrets = ["github-token", "image-download"]
+            if api.repo == "rhinstaller/anaconda-webui":
```
allisonkarlitskaya, line 246: Maybe also try to limit this only to nightlies?

allisonkarlitskaya (commented):
It's a bit ad hoc, but it looks fine to me.

tests-scan
martinpitt: It is a bit ad hoc, right. https://github.com/rhinstaller/anaconda-webui/pull/641 is still figuring out the details. I propose we start with that, and once @KKoukiou is happy with how it works, we further limit the secret by scenario or by branch name (only handing it out for testing the `main` branch -- can we do that? absent `number`?)

KKoukiou (approved):

PR #7442, opened 2025-02-14
machine: add option to set which bus is used to connect VM disk

tomasmatus:
Needed in order to test SMART disk info in CI
https://github.com/cockpit-project/cockpit/pull/19103


martinpitt (approved):
Looks good, thanks! I start a few tests to make sure this doesn't regress.

PR #7450, opened 2025-02-17
testmap: Enable fedora-42 for subscription-manager-cockpit

martinpitt:


martinpitt:
@ptoscano FYI

martinpitt:
@ptoscano That smells like a hack.. Why is the sub-man test moving binaries around? (Fedora 42 merged /sbin and /bin, so that won't work any more).

```
+ mv /usr/bin/subscription-manager /usr/sbin
mv: '/usr/bin/subscription-manager' and '/usr/sbin/subscription-manager' are the same file
```

ptoscano:
> That smells like a hack.. Why is the sub-man test moving binaries around?

Because distutils only installs binaries to "$prefix/bin"; since the "subscription-manager" executable is an administration utilities, it needs to go do "sbin", and thus the build system takes care of that:
https://github.com/candlepin/subscription-manager/blob/96b120fcbf052c8ce203ed6a6b7e6a9cdccfd08d/Makefile#L203-L205

This required changes, yes: https://github.com/candlepin/subscription-manager/pull/3498

Back to the actual issue: https://github.com/candlepin/subscription-manager-cockpit/pull/90 should easily fix this issue. I triggered it on `fedora-42` as well, and it works there too.

ptoscano:
My PR for sub-man-cockpit was merged, so please retrigger the tests :)

martinpitt:
Excellent, thanks @ptoscano ! Retriggered.

ptoscano:
Ah, almost there: https://github.com/candlepin/subscription-manager/pull/3513 fixes the same script in sub-man itself.

ptoscano:
My PR for sub-man was merged, so please retrigger the tests :)

(the message above was brought to you by your friendly copypasta service)

martinpitt:
Nice! Re-re-rebased and triggered.

ptoscano (approved):
thanks!

allisonkarlitskaya (approved):
Thanks!

PR #7468, opened 2025-02-22
Image refresh for rhel-9-6

cockpituous:
Image refresh for rhel-9-6
 * [x] image-refresh rhel-9-6


cockpituous:
image-refresh rhel-9-6 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-9-6-20250222-225720

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel-9-6-747c7e86-20250222-223825/log.html

martinpitt:
[breaks criu](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7468-5e35705a-20250222-225848-rhel-9-6-cockpit-project-cockpit-podman/log.html) due to "executable not found". Indeed it fell out of the image, but I'm not sure why. `crun` recommends criu, and it didn't change. Still the same version on the new image. Some dnf change? :thinking:

mvollmer:
> crun recommends criu, and it didn't change

Both podman and crun are already installed in the qcow2 that we download, so our dnf invocations never process the "recommended" dependencies of podman or crun. Since we test with it, I think we should install is explicitly.

martinpitt:
Fixed in #7473

PR #7472, opened 2025-02-24
images: Don't install selinux-policy-default on the Debians

mvollmer:
We don't enable SELinux and it can't currently be installed in debian-testing.

- [ ] image-refresh debian-testing
- [ ] image-refresh ubuntu-stable

### DIFF
--- a/images/scripts/debian.setup
+++ b/images/scripts/debian.setup
@@ -99,6 +99,11 @@ wireguard-tools \
 xfsprogs \
 "
 
+# These are the packages we don't want.
+DENIED_PACKAGES="\
+selinux-policy-default-
+"
+
 # older libvirt have this builtin
 if [ "$IMAGE" = "debian-stable" ] || [ "$IMAGE" = "ubuntu-2204" ] ; then
     TEST_PACKAGES="${TEST_PACKAGES/virtiofsd/}"
@@ -214,7 +219,7 @@ $APT dist-upgrade
 systemctl restart systemd-networkd.service || true
 
 # install our dependencies
-$APT install $TEST_PACKAGES $COCKPIT_DEPS $IPA_CLIENT_PACKAGES
+$APT install $TEST_PACKAGES $COCKPIT_DEPS $IPA_CLIENT_PACKAGES $DENIED_PACKAGES
 
 # download package sets
 


cockpituous:
image-refresh debian-testing done: https://github.com/cockpit-project/bots/commits/image-refresh-debian-testing-20250224-091514

images/scripts/debian.setup
```diff
@@ -99,6 +99,11 @@ wireguard-tools \
 xfsprogs \
 "
 
+# These are the packages we don't want.
+DENIED_PACKAGES="\
+selinux-policy-default-
```
martinpitt, line 104: Could we limit this to debian-testing? It doesn't get in the way on the other OSes, so we shouldn't modify more than necessary.

mvollmer: Sure, but it might get in the way in the future... less selinux is more, no? :-)

cockpituous:
image-refresh debian-testing done: https://github.com/cockpit-project/bots/commits/image-refresh-debian-testing-20250224-095414

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-debian-testing-2cae63a4-20250224-094058/log.html

images/scripts/debian.setup
martinpitt: OK, not a strong opinion. I even wonder what brings it in.

images/scripts/debian.setup
```diff
@@ -153,6 +157,9 @@ if [ "$IMAGE" = "debian-testing" ]; then
     COCKPIT_DEPS="${COCKPIT_DEPS/libpcp3 /}"
     COCKPIT_DEPS="${COCKPIT_DEPS/pcp /}"
     COCKPIT_DEPS="${COCKPIT_DEPS/python3-pcp /}"
+
+    # https://bugs.debian.org/XXX
```
martinpitt, line 161: Filed: https://bugs.debian.org/1098793

You already triggered tests here on 913b18dced11f7fbcaaf3c69cbaaac0ded5bfd16 and they are green. So I'd say  push the comment update without re-running them again.

martinpitt (changes requested):
Thanks!

martinpitt:
@mvollmer When you force-push, please also add "Fixes #7454" to the commit message.

images/scripts/debian.setup
mvollmer: Done!

martinpitt (approved):
Cheers!

PR #7536, opened 2025-03-13
machine: correct add_disk typing

jelly:
add_disk's size argument is a string and not an integer as we pass `10G` directly to `qemu` and not the amount of bytes.


martinpitt (approved):

martinpitt (approved):
Checked our usage in tests, and it's indeed always a string. Thanks!

PR #7538, opened 2025-03-13
testmap: track new stable branches of sub-man

ptoscano:
- use `rhel-9-6` for `subscription-manager-1.29.45`, as that branch tracks 9.6
- use `rhel-10-0` for `subscription-manager-1.30.6`, as that branch tracks 10.0
- add `centos-9-stream` for `subscription-manager-1.29`, as that branch tracks future 9.x versions


martinpitt:
Thanks! I triggered the new contexts.

PR #7543, opened 2025-03-14
machine: find OVMF on Ubuntu systems

allisonkarlitskaya:
We hardcode the path to the OVMF UEFI BIOS for -efi images to the location on Fedora/RHEL systems, but Ubuntu has it installed in a different location.

This will let us run VMs on GitHub Actions runners without having to create a symlink.


martinpitt (approved):
Sure! I triggered webui's efi scenario to make sure it doesn't break. Please let that go green (or at least sufficiently far, I know they are currently very flaky) before landing.

PR #7579, opened 2025-03-25
allowlist: Add Peter Koprda sub-man team member

martinpitt:
@pkoprda FYI

See e.g. https://github.com/candlepin/subscription-manager-cockpit/pull/92

I have met with @pkoprda and confirm he's on the sub-man team, taking over from @ptoscano .

### DIFF
--- a/lib/allowlist.py
+++ b/lib/allowlist.py
@@ -45,4 +45,5 @@ ALLOWLIST = {
 
     # candlepin team + contributors
     'ptoscano',
+    'pkoprda',
 }


ptoscano (approved):
LGTM, thanks!

Also now I understand why the cockpit CI jobs in PRs of @pkoprda do not start automatically, even if his account is part of the @candlepin organization: rather than branches of the organization repositories, he creates PRs from his forks. Nothing problematic with that, simply that I got the explanation on why that happened :)

martinpitt:
Right, that's what I told @pkopda in the internal chat, too. We can make this work with forks.

PR #7590, opened 2025-03-27
testmap: Move c-machines to RHEL 9.7/10.1

martinpitt:
9.6/10.0 is done from a developer's PoV.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -142,8 +142,8 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
             f'{TEST_OS_DEFAULT}/firefox',
             'opensuse-tumbleweed',
             'rhel-8-10',
-            'rhel-9-6',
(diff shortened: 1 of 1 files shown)


jelly (commented):
Do we want to drop rhel-8-10, we still have it in podman and it is our lowest supported Python version. I recall this should be switched to a ws-container scenario?

lib/testmap.py
```diff
@@ -141,18 +141,16 @@ def contexts(image: str, *scenarios: Iterable[str], repo: str | None = None) ->
             f'{TEST_OS_DEFAULT}/devel',
             f'{TEST_OS_DEFAULT}/firefox',
             'opensuse-tumbleweed',
-            'rhel-8-10',
-            'rhel-9-6',
             'rhel-10-0',
```
jelly, line 144: This keeps rhel-10-0 while cockpit-files does not

jelly (approved):

PR #7591, opened 2025-03-27
testmap: Move sub-man to RHEL 9.7/10.1

martinpitt:
9.6/10.0 is done from a developer's PoV.

---

@pkoprda @ptoscano Can you please check this? There are already stable branches set up for 9-6 and 10-0, but I don't know if you need something else here.

Also, I'd like to clean up rhel-9-5 (non-EUS release, and 9.6 is around the corner) and fedora-40 (soon EOL) from the test map and the images from our CI. Any objections?

Thanks!

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -190,8 +190,8 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
     'candlepin/subscription-manager': {
         'main': [
             'centos-10',
-            'rhel-9-6',
-            'rhel-10-0',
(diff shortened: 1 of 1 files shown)


ptoscano (approved):
This makes sense, thanks!

> Also, I'd like to clean up rhel-9-5 (non-EUS release, and 9.6 is around the corner)

This makes sense too.

> and fedora-40 (soon EOL)

Yes, please by default do add new Fedora images when they are created also to sub-man/main and sub-man-cockpit/main, and remove EOLed Fedora images too.

jelly (approved):

PR #7593, opened 2025-03-27
testmap: add rhel-8-10/ws-container to c-machines

jelly:


martinpitt:
You don't need this -- you can trigger arbitrary scenarios with the current test map already. I suggest to send a podman PR with enabling the scenario, test it there, and enable it for real here.

martinpitt:
https://github.com/cockpit-project/cockpit-machines/pull/2072 landed, so now this is actually urgent -- "rhel-8-10" context stopped working.

martinpitt (approved):
Moved to production, thanks!

PR #7597, opened 2025-03-28
testmap: Drop cockpit-composer from Fedora

martinpitt:
cockpit-composer was replaced with
https://github.com/osbuild/image-builder-frontend, see https://github.com/osbuild/cockpit-composer/ README.

Fedora 40 is EOL soon, and c-composer won't be updated there any more.

---

Just talked this over with @ochosi and @croissanne in chat, but will let them confirm anyway. After this lands, we can drop the fedora-40 image.


croissanne (approved):
Thank you!

Issue #7646, opened 2025-04-11
libvirt: segfault when attach-disk source is a directory


tomasmatus:
upstream: https://gitlab.com/libvirt/libvirt/-/issues/765

cockpituous:
debian-testing
Ooops, it happened again<hr>
```
Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
ChromeDriver was started successfully on port 34433.
Failed to try-restart sssd.service: Unit sssd.service not found.
bash: line 1: rpmquery: command not found
> error: Scrollbar test exception: TypeError: Cannot read properties of null (reading 'appendChild')
> warn: Failed to get current crypto policy: not-found ; falling back to /etc/crypto-policies/config
> warn: cockpit.format_{bytes,bits}[_per_sec](..., MiB, [object Object]) is deprecated.
> warn: could not execute head --bytes=16 head: error reading '/tmp/': Is a directory
Process 7107 (libvirtd) of user 0 terminated abnormally with signal 11/SEGV, processing...
Process 7107 (libvirtd) of user 0 dumped core.
Wrote JS log to TestMachinesDisks-testAddDiskCustomPath-debian-testing-127.0.0.2-3001-FAIL.js.log
Journal extracted to TestMachinesDisks-testAddDiskCustomPath-debian-testing-127.0.0.2-3001-FAIL.log.gz
Core dumps downloaded to /work/make-checkout-workdir/TestMachinesDisks-testAddDiskCustomPath-debian-testing-127.0.0.2-3001-FAIL.core
Stopping 'libvirtd.service', but its triggering units are still active:
libvirtd.socket, libvirtd-ro.socket, libvirtd-admin.socket
Traceback (most recent call last):
  File "/work/make-checkout-workdir/test/machineslib.py", line 573, in tearDown
    super().tearDown()
    ~~~~~~~~~~~~~~~~^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 1990, in tearDown
    self.check_journal_messages()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 2253, in check_journal_messages
    raise Error(UNEXPECTED_MESSAGE + "journal messages:\n" + first)
testlib.Error: FAIL: Test completed, but found unexpected journal messages:
Process 7107 (libvirtd) of user 0 terminated abnormally with signal 11/SEGV, processing...

[...]

PR #7647, opened 2025-04-11
naughty: Add patterns for #7646

tomasmatus:


jelly:
This usually goes into the refresh PR of the affected image

Issue #7648, opened 2025-04-11
libvirt: unable to restore snapshots


tomasmatus:
upstream: https://bugzilla.redhat.com/show_bug.cgi?id=2358626

cockpituous:
arch
Ooops, it happened again<hr>
```
File "/work/make-checkout-workdir/test/common/testlib.py", line 817, in wait_js_cond
    raise Error(f"timeout\nwait_js_cond({cond}): {last_error.msg}") from None
testlib.Error: timeout
wait_js_cond(!ph_is_present("#vm-subVmTest1-snapshot-1-current")): Error: condition did not become true

Wrote screenshot to TestMachinesSnapshots-testSnapshotRevert-arch-127.0.0.2-2201-FAIL.png
Wrote HTML dump to TestMachinesSnapshots-testSnapshotRevert-arch-127.0.0.2-2201-FAIL.html
Wrote JS log to TestMachinesSnapshots-testSnapshotRevert-arch-127.0.0.2-2201-FAIL.js.log
Journal extracted to TestMachinesSnapshots-testSnapshotRevert-arch-127.0.0.2-2201-FAIL.log.gz
> warn: transport closed: disconnected
Stopping 'libvirtd.service', but its triggering units are still active:
libvirtd-ro.socket, libvirtd.socket, libvirtd-admin.socket
Traceback (most recent call last):
  File "/work/make-checkout-workdir/test/check-machines-snapshots", line 462, in testSnapshotRevert
    b.wait_not_present("#vm-subVmTest1-snapshot-1-current")
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 829, in wait_not_present
    self.wait_js_func('!ph_is_present', selector)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 820, in wait_js_func
    self.wait_js_cond("%s(%s)" % (func, ','.join(map(jsquote, args))))
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 817, in wait_js_cond
[...]

PR #7628, opened 2025-04-07
arch image: remove stratis

tomasmatus:
* [x] image-refresh arch

Stratis was removed from archlinux repositores.

fixes: #7624

PR #7649, opened 2025-04-11
naughty: add pattern for #7648

tomasmatus:


tomasmatus:
I'll add this on the image refresh PR instead #7628

Issue #7474, opened 2025-02-24
bots get stuck in job-runner on failing podman


martinpitt:
Our test queue has become very long and slow recently. Many of them often don't do any work any more. The container log usually ends with something like
```
+ timeout 12h ./run-queue
INFO:root:Consuming message with command: ./job-runner json {"repo": "rhinstaller/anaconda-webui", ...
INFO:lib.aio.job:Log: https://cockpit-logs.us-east-1.linodeobjects.com/pull-657-9f15d9d1-20250224-025303-fedora-42-boot-efi-other/log.html
Error: reading CIDFile: open /tmp/tmpkdsgqcpp/cidfile: no such file or directory
```

and there is either no corresponding podman container on the host *at all* (cockpit-tasks-1 is the only one), or it has exited long ago, like

```
c8d82bebd818  ghcr.io/cockpit-project/tasks:2025-02-22  python3 -c #!/usr...  4 hours ago  Exited (1) 4 hours ago              adoring_bose
```

I found some as old as 7 hours, which is when I mass-rebooted all bots.

martinpitt:
I haven't seen this any more in the last two days -- this was somehow magically triggered by https://github.com/rhinstaller/anaconda-webui/pull/655 . But now disappeared just as mysteriously as it arrived.

martinpitt:
Infra is fine. Let's shelve this.

martinpitt:
Reopening, this keeps happening these days. Somehow podman gets stuck, throws this CIDFile error, and `job-runner` doesn't notice that the process died and gets stuck forever.

jelly:
Today I found this container which is up for a long time:

```
3a92118d0790  ghcr.io/cockpit-project/tasks:2025-03-29  python3 -c #!/usr...  7 hours ago  Up 7 hours              heuristic_agnesi
```

`podman logs` end with:

```
rsync --recursive --perms --copy-links -e 'ssh -p 2201 -o ControlPath=/tmp/.cockpit-test-resources/ssh-%C-3360' --verbose /work/make-checkout-workdir/dist/playground 'root@[127.0.0.2]:/usr/local/share/cockpit/'
sending incremental file list
created directory /usr/local/share/cockpit
playground/
playground/exception.css.gz
playground/exception.html
playground/exception.js.gz
```

So there is nothing interesting to report there and he tasks container reports the same issue:

```
Error: reading CIDFile: open /tmp/tmp8a1hp8r4/cidfile: no such file or directory
```

[...]

PR #7655, opened 2025-04-14
Job runner debug

jelly:
See https://github.com/cockpit-project/bots/issues/7474 I am not sure if we can't stop containers reliably anymore or if our timeouts are kaput.

### DIFF
--- a/lib/aio/job.py
+++ b/lib/aio/job.py
@@ -151,6 +151,7 @@ async def run_container(job: Job, subject: Subject, ctx: JobContext, log: LogStr
 
             finally:
                 await run([*ctx.container_cmd, 'rm', '--force', '--time=0', f'--cidfile={tmpdir}/cidfile'],
+                          stderr=asyncio.subprocess.STDOUT,
                           stdout=asyncio.subprocess.DEVNULL)  # don't show container ID output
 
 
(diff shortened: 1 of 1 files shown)


lib/aio/job.py
```diff
@@ -151,6 +151,7 @@ async def run_container(job: Job, subject: Subject, ctx: JobContext, log: LogStr
 
             finally:
                 await run([*ctx.container_cmd, 'rm', '--force', '--time=0', f'--cidfile={tmpdir}/cidfile'],
+                          stderr=asyncio.subprocess.STDOUT,
```
martinpitt, line 157: This *hides* stderr (as stdout goes to /dev/null). The default is that stderr goes to the terminal or log file, i.e. you should currently see stderr in `podman logs`.

lib/aio/job.py
```diff
@@ -150,6 +152,7 @@ async def run_container(job: Job, subject: Subject, ctx: JobContext, log: LogStr
                     raise Failure(f'Container exited with code {returncode}')
 
             finally:
+                logger.debug('force removing container')
```
martinpitt, line 155: Ah, is it that removal which fails? I suggest to also log if `cidfile` exists at that point.

PR #7693, opened 2025-04-25
images: Pre-install tlog on bootc

justin-stephenson:
Dependency needed for cockpit-session-recording tests on bootc.


martinpitt:
Thanks! Because reasons I do the actual refresh in #7694. I'll shepherd this.

martinpitt (approved):
Cheers!

PR #7702, opened 2025-04-28
image-create: Remove old image before installing new one

### LINKED ISSUE #7680: Image refresh for fedora-42-boot
Image refresh for fedora-42-boot
 * [ ] FAIL: image-refresh fedora-42-boot

mvollmer:
Otherwise, shutil.move might try to overwrite it and fail with a confusing permission error.  This happens when the new image is the same as the old image, a "payload" or "iso", and /var/tmp and /cache are on different filesystems.

Fixes #7680

 * [x] image-refresh fedora-42-boot

### DIFF
--- a/image-create
+++ b/image-create
@@ -165,6 +165,10 @@ class MachineBuilder:
 
         name = self.machine.image + "-" + sha + self.suffix
         data_file = os.path.join(data_dir, name)
+        if os.path.exists(data_file):
+            # shutil.move has trouble when the destination exists but
+            # has too restrictive permissions to be overwritten.
+            os.unlink(data_file)
         shutil.move(self.machine.image_file if (self.iso or self.payload) else rebuild, data_file)
 
         if not self.iso and not self.payload:


cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-fedora-42-boot-f7497b94-20250428-113444/log.html

mvollmer:
This fixes the immediate bug, but the behavior might not be what we want. No git commit is created (because there are no changes), and things might be equally confusing...

martinpitt (approved):
Thanks! This supposedly fixes #7680, so I added that to the description. Please trigger the anaconda tests before merging though, to make sure the image doesn't break.

mvollmer:
Oops, triggered some "fedora-42" tests instead of the "fedora-42-boot" ones...

mvollmer:
Anything to do with Cockpit Storage in Anaconda is broken right now, I think. The rest still works.

PR #7715, opened 2025-05-01
po-refresh: bump requirement to 80%

jelly:
The default 50% is rather low for the amount of strings we have in for example Cockpit (~2500).

---

There might be some fallout in removing languages, for cockpit for example norwegian (bokmal) would be removed


martinpitt (approved):
I agree. 50% looks rather sad. We'll see the fallout on the next po-refresh run.

This just bumps a number and the integration test doesn't cover po-refresh, so we can let the failed integration test slip (due to this not being an origin branch).

Issue #7716, opened 2025-05-01
checkpoint restore fails with criu error


jelly:
Downstream issue https://issues.redhat.com/browse/RHEL-89354

cockpituous:
fedora-41
Ooops, it happened again<hr>
```
CONTAINER ID  IMAGE                          COMMAND     CREATED         STATUS         PORTS       NAMES
8d16660cad16  localhost/test-busybox:latest  sh          19 seconds ago  Up 17 seconds              swamped-crate
---- 8d16660cad16 ----
/ # Traceback (most recent call last):
  File "/work/make-checkout-workdir/test/check-application", line 1612, in testCheckpointRestore
    b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in NOT_RUNNING)
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 779, in wait
    raise Error('timed out waiting for predicate to become true')
testlib.Error: timed out waiting for predicate to become true

Wrote screenshot to TestApplication-testCheckpointRestore-fedora-41-127.0.0.2-2901-FAIL.png
Wrote HTML dump to TestApplication-testCheckpointRestore-fedora-41-127.0.0.2-2901-FAIL.html
Wrote JS log to TestApplication-testCheckpointRestore-fedora-41-127.0.0.2-2901-FAIL.js.log
Journal extracted to TestApplication-testCheckpointRestore-fedora-41-127.0.0.2-2901-FAIL.log.gz
Connection to 127.0.0.2 closed by remote host.
> warn: transport closed: disconnected
Failed to reset failed state of unit podman.service: Unit podman.service not loaded.
Failed to reset failed state of unit podman.socket: Unit podman.socket not loaded.
Traceback (most recent call last):
  File "/work/make-checkout-workdir/test/check-application", line 1612, in testCheckpointRestore
    b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in NOT_RUNNING)
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 779, in wait
[...]

Issue #7692, opened 2025-04-25
ubuntu-stable 25.04: criu is broken


mvollmer:
Upstream: not yet reported

Criu produces these errors in cockpit-podman TestApplication.testCheckpointRestore in Ubuntu 25.04:

```
(00.140849) Error (criu/vdso.c:381): vdso: Unexpected rt vDSO area bounds
(00.140852) Error (criu/vdso.c:613): vdso: Failed to fill self vdso symtable
(00.140854) Error (criu/kerndat.c:1906): kerndat_vdso_fill_symtable failed when initializing kerndat.
```

cockpituous:
ubuntu-stable
Ooops, it happened again<hr>
```
CONTAINER ID  IMAGE                          COMMAND     CREATED         STATUS         PORTS       NAMES
c7d7b8d2575b  localhost/test-busybox:latest  sh          19 seconds ago  Up 17 seconds              swamped-crate
---- c7d7b8d2575b ----
/ # Traceback (most recent call last):
  File "/work/make-checkout-workdir/test/check-application", line 1612, in testCheckpointRestore
    b.wait(lambda: self.getContainerAttr("swamped-crate", "State") in NOT_RUNNING)
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 775, in wait
    raise Error('timed out waiting for predicate to become true')
testlib.Error: timed out waiting for predicate to become true

Wrote screenshot to TestApplication-testCheckpointRestore-ubuntu-stable-127.0.0.2-2501-FAIL.png
Wrote HTML dump to TestApplication-testCheckpointRestore-ubuntu-stable-127.0.0.2-2501-FAIL.html
Wrote JS log to TestApplication-testCheckpointRestore-ubuntu-stable-127.0.0.2-2501-FAIL.js.log
Journal extracted to TestApplication-testCheckpointRestore-ubuntu-stable-127.0.0.2-2501-FAIL.log.gz
Connection to 127.0.0.2 closed by remote host.
> warn: transport closed: disconnected
Failed to reset failed state of unit podman.service: Unit podman.service not loaded.
Failed to reset failed state of unit podman.socket: Unit podman.socket not loaded.
Traceback (most recent call last):
[...]

PR #7685, opened 2025-04-23
Image refresh for ubuntu-stable

cockpituous:
- [ ] https://github.com/cockpit-project/cockpit/pull/21945
 - [ ] https://github.com/cockpit-project/cockpit-podman/pull/2106
 - [x] naughty for failing criu

Image refresh for ubuntu-stable
 * [x] image-refresh ubuntu-stable

PR #7717, opened 2025-05-01
naughty: add CentOS 10 CRIU issue

jelly:
Known issue #7716

---

Happens on testing farm https://artifacts.dev.testing-farm.io/7f36e00e-c995-4d4f-aa08-12628371a567/


mvollmer:
I have the feeling this happens _all_ _the_ _time_, and there is something fundemantally broken about how criu and the kernel are synchronized.

naughty/rhel-10/7716-checkpoint-restore-failure
```diff
@@ -0,0 +1,7 @@
+Traceback (most recent call last):
```
mvollmer, line 1: The same things is happening in ubuntu-stable currently: https://github.com/cockpit-project/bots/issues/7692

We now have the criu debug dump in the test output, and we can match against it, like in  https://github.com/cockpit-project/bots/pull/7685/commits/3defe4a34fe0ffab95769e50f73f774780fcaee9. I think we should do this here, too.

mvollmer (changes requested):

mvollmer (approved):
Thanks, looks good to me.

PR #7764, opened 2025-05-14
images: Drop fedora-42-{boot,anaconda-payload}

martinpitt:
Fedora 42 is released, and neither image will get updates any more. We can stop testing the installer on Fedora 42 now. Once 43 branches off, we'll reintroduce this accordingly.

Also drop the duplicate fedora-rawhide-boot test map entry.


image-trigger
```diff
@@ -44,7 +44,6 @@ REFRESH = {
     "debian-stable": {},
     "fedora-41": {},
     "fedora-42": {},
-    "fedora-42-boot": {},
```
martinpitt, line 47: FTR, I didn't forget to drop 42-anaconda-payload here. It was never here in the first place.

KKoukiou (commented):
There will not be any respins so there is not point in testing Fedora 42 ISO anymore.

jkonecny12:
There was a discussion for respin but it seems to me that it won't happen.
However, even if it would happen the change set should be minimal, so we can do the testing manually.

PR #7774, opened 2025-05-16
naughty: copy checkpoint restore issue to Fedora-rawhide

jelly:
Should have done that as well when copying to fedora-42/41 ... now it affects our reverse dependency testing in podman https://artifacts.dev.testing-farm.io/d8c8fd27-db60-412f-933f-88aee05596ec/


Venefilyn (approved):
LGTM

Related to
https://github.com/containers/podman/pull/26115

PR #7776, opened 2025-05-16
naughty: copy kdump issues to rhel-10

jelly:
The crashkernel=2G issue also affects rhel-10.

---

See https://github.com/cockpit-project/cockpit/pull/22005


mvollmer (approved):
More naughty, more better.

PR #7781, opened 2025-05-19
lib: Move default OS to fedora-42

tomasmatus:


martinpitt (approved):
Thanks! I first thought "pixel tests", but they use their own reference image file. That should be moved as the next step.

PR #7782, opened 2025-05-19
Avoid rate limits by using the contents API for reading the .cockpit-ci/container file

jelly:
* [x] https://github.com/cockpit-project/cockpituous/pull/654

### DIFF
--- a/lib/aio/github.py
+++ b/lib/aio/github.py
@@ -85,15 +85,15 @@ class GitHub(Forge, contextlib.AsyncExitStack):
 
         return await retry(post_once)
 
-    async def get(self, resource: str) -> JsonValue:
+    async def get(self, resource: str, parameters: Mapping[str, str] | None = None) -> JsonValue:
         async def get_once() -> JsonValue:
             headers = {**self.session.headers}
             cache_entry = self.cache.get(resource)
             if cache_entry is not None:
                 headers.update(cache_entry.conditions)
 
-            logger.debug('get %r %r', resource, cache_entry)
-            async with self.session.get(self.api / resource, headers=headers) as response:
+            logger.debug('get %r %r %r', resource, parameters, cache_entry)
+            async with self.session.get(self.api / resource % parameters, headers=headers) as response:
                 condition_map = {'etag': 'if-none-match', 'last-modified': 'if-modified-since'}
                 conditions = {c: response.headers[h] for h, c in condition_map.items() if h in response.headers}
 
@@ -110,8 +110,8 @@ class GitHub(Forge, contextlib.AsyncExitStack):
 
         return await retry(get_once)
 
-    async def get_obj(self, resource: str) -> JsonObject:
-        return typechecked(await self.get(resource), dict)
+    async def get_obj(self, resource: str, parameters: Mapping[str, str] | None = None) -> JsonObject:
+        return typechecked(await self.get(resource, parameters), dict)
 
     async def check_pr_changed(self, repo: str, pull_nr: int, expected_sha: str) -> str | None:
         try:
--- a/lib/aio/github.py
+++ b/lib/aio/github.py
@@ -14,6 +14,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import asyncio
+import base64
 import contextlib
 import json
 import logging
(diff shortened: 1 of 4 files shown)


jelly:
The mockserver in cockpituous needs to be adjusted

lib/aio/github.py
```diff
@@ -85,15 +85,15 @@ async def post_once() -> JsonValue:
...
         async def get_once() -> JsonValue:
             headers = {**self.session.headers}
             cache_entry = self.cache.get(resource)
             if cache_entry is not None:
                 headers.update(cache_entry.conditions)
 
-            logger.debug('get %r %r', resource, cache_entry)
-            async with self.session.get(self.api / resource, headers=headers) as response:
+            logger.debug('get %r %r %r', resource, parameters, cache_entry)
+            async with self.session.get(self.api / resource % parameters, headers=headers) as response:
```
allisonkarlitskaya, line 98: I guess `% None` is well-defined?

lib/aio/github.py
```diff
@@ -133,17 +135,11 @@ async def open_issue(self, repo: str, issue: JsonObject) -> None:
...
-                    return await response.text()
-            except aiohttp.ClientResponseError as exc:
-                if exc.status == 404:
-                    return None
-                raise
-
-        return await retry(read_once)
+        content = await self.get_obj(f'repos/{subject.repo}/contents/{filename}', {'ref': subject.sha})
+        try:
+            return base64.b64decode(str(content['content'])).decode().strip()
```
allisonkarlitskaya, line 145: Can you jsonutils this a bit?

lib/aio/github.py
```diff
@@ -133,17 +135,11 @@ async def open_issue(self, repo: str, issue: JsonObject) -> None:
...
+        except binascii.Error:
```
allisonkarlitskaya, line 146: This could easily toss UnicodeError too: you're requesting a blob, so the thing you get could well be binary...

But ...

lib/aio/github.py
```diff
@@ -133,17 +135,11 @@ async def open_issue(self, repo: str, issue: JsonObject) -> None:
...
+            return None
```
allisonkarlitskaya, line 147: ...it kinda seems like it should just be fatal, to be honest?

lib/aio/github.py
```diff
@@ -133,17 +135,16 @@ async def open_issue(self, repo: str, issue: JsonObject) -> None:
...
-                    return await response.text()
-            except aiohttp.ClientResponseError as exc:
-                if exc.status == 404:
-                    return None
-                raise
+        try:
+            content = await self.get_obj(f'repos/{subject.repo}/contents/{filename}', {'ref': subject.sha})
+        except aiohttp.ClientError as exc:
+            logger.warning('Error when reading file %s#%s: %r', subject.repo, filename, exc)
+            return None
```
allisonkarlitskaya, line 142: The previous code only returned `None` on 404, letting other errors bubble...

lib/aio/github.py
jelly: That I have to test honestly.

lib/aio/github.py
jelly: There is no get_bytes tho so that seems unlikely?

lib/aio/github.py
jelly: Seems fine:

```
>>> import yarl
>>> yarl.URL('https://compose.rs') / 'flatpak' % None
URL('https://compose.rs/flatpak')
>>> str(yarl.URL('https://compose.rs') / 'flatpak' % None)
'https://compose.rs/flatpak'
>>> str(yarl.URL('https://compose.rs') / 'flatpak' % {'download': 1 })
'https://compose.rs/flatpak?download=1'
>>> yarl.URL('https://compose.rs') / 'flatpak' % {'download': 1 }
URL('https://compose.rs/flatpak?download=1')
```

lib/aio/github.py
jelly: Everything is also heavily string based so changing this is a bit more work, ie. the default argument for `_get` is a string etc.

jelly: I misunderstood my code, honestly get_str() makes it kinda worse but here it is.

allisonkarlitskaya (commented):
One minor comment, but otherwise looks good.  Not approving because of the `DONTNOTMERGE`, although, I gotta say, if you yell at me to don't not merge something I kinda feel like you definitely want me to make sure that I immediately merge it, no? ;)

allisonkarlitskaya (commented):
One more thing: there ought to be some dead code here now, right?  Specifically the setting of the `raw.` endpoint that's now unused?  Please clean that up.

PR #7784, opened 2025-05-19
logs: Always show init section

Venefilyn:
Init section is important no matter which type of tests you are looking
for. Lets make it show for all tabs no matter the filtering.

### DIFF
--- a/lib/s3-html/log.html
+++ b/lib/s3-html/log.html
@@ -24,10 +24,10 @@
 
         /* Filters start */
         /* Only show failed */
-        body.c-filter-only-failed .test-entry:not(.failed:not(.skipped,.retried)) {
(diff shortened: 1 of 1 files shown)


martinpitt (commented):
Hm, it's important for build failures, but these are relatively rare. I think this would make more sense:
 - Keep the current behaviour of only listing it in "Show all"
 - Automatically switch to "Show all" and expand the initialization if there are no actual tests, i.e. the build failed.

Venefilyn:
> * Automatically switch to "Show all" and expand the initialization if there are no actual tests, i.e. the build failed.

@martinpitt how do we know if the build failed? If the total tests is 0?

martinpitt:
Right, there's no initial `1..` TAP marker.

PR #7786, opened 2025-05-19
logs: Improve filtering for pixel tests

Venefilyn:
- Can now toggle failed tests.
- Failed tests is filtered to by default if any are found.
- If no failed tests are there, always turn the checkbox off.
- Filtering is now stored in `localStorage`.
- Default filter selection won't reset anymore on chunked logs fetching.

### DIFF
--- a/lib/s3-html/log.html
+++ b/lib/s3-html/log.html
@@ -24,9 +24,6 @@
 
         /* Filters start */
         /* Only show failed */
-        body.c-filter-only-failed .test-entry:not(.failed:not(.skipped,.retried)) {
-            display: none;
-        }
         body.c-filter-only-failed-retried .test-entry:not(.failed:not(.skipped), .retried) {
             display: none;
         }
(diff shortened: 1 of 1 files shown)


lib/s3-html/log.html
```diff
@@ -165,7 +165,7 @@
                   </button>
                 </div>
                 <div class="pf-v6-c-toggle-group__item">
-                    <button class="pf-v6-c-toggle-group__button pf-m-selected" type="button" data-filter="c-filter-only-failed-retried">
+                    <button class="pf-v6-c-toggle-group__button" type="button" data-filter="c-filter-only-failed-retried">
```
martinpitt, line 168: I'm not sure what this does, but does that mean that you can't "select" the "Show all" and friends buttons? I find the current behaviour very confusing: You can select and unselect them, but unselecting does nothing. Also, "only show failed' is a bit useless -- one often needs to compare the retries, and there shouldn't be so many failures. Can the whole thing please just collapse to a Switch that says "Show only failures"?

Venefilyn: Ack will do

martinpitt (approved):
Ah, misread, thanks!

lib/s3-html/log.html
```diff
@@ -589,24 +587,34 @@
...
-            }
-        )
-    });
+    const failedFilter = testingToolbar.querySelector("#switch-for-failed-filtering");
+
+    function handleFilterSelection(checkbox) {
+        // Apply the clicked filter and add it to localStorage
+        if (checkbox.checked) {
+            document.body.classList.add(selectorTestRetried);
+            localStorage.setItem(selectorTestRetried, checkbox.checked);
```
martinpitt, line 599: This still rubs me the wrong way. This *really* should not be "configuration", this is more like temporary state. If you look at two different test logs, your choice is independent for both of them.

Hence, `sessionStorage` feels a bit less sticky and more appropriate here. TBH I think a fragment value would be even better.

martinpitt (changes requested):

PR #7803, opened 2025-05-21
machines: Give our test VMs a TPM device

mvollmer:
We can use that to test Clevis.

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -100,6 +100,12 @@ TEST_DOMAIN_XML = """
     <rng model='virtio'>
       <backend model='random'>/dev/urandom</backend>
     </rng>
+    <tpm model='tpm-crb'>
+      <backend type='emulator' version='2.0'>
+        <profile name='default-v1'/>
+      </backend>
+      <alias name='tpm0'/>
+    </tpm>
   </devices>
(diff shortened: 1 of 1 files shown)


machine/machine_core/machine_virtual.py
```diff
@@ -100,6 +100,12 @@ def stdchannel_redirected(stdchannel: TextIO, dest_filename: str) -> Iterator[No
     <rng model='virtio'>
       <backend model='random'>/dev/urandom</backend>
     </rng>
+    <tpm model='tpm-crb'>
+      <backend type='emulator' version='2.0'>
```
jelly, line 104: Do we need to give a version?

jelly (approved):
Makes sense, on UEFI virt-install automatically adds a TPM.

This would require swtpm in our tasks container which luckily is already installed :)

machine/machine_core/machine_virtual.py
mvollmer: No idea... I copied this snippet from one of my own VMs after using c-machines to add TPM to it...

mvollmer:
Proof that it works: https://cockpit-logs.us-east-1.linodeobjects.com/pull-21927-e6e6cc81-20250521-085138-fedora-42-storage-bots%237803/log.html

PR #7806, opened 2025-05-21
setup-deploy-keys-anaconda: fix typo

KKoukiou:


martinpitt (approved):
Hehe -- perhaps run the script from your branch first? It doesn't have to be on main for running.

PR #7832, opened 2025-05-29
WIP: 0b75852ee434: [no-test] Image refresh for fedora-coreos

cockpituous:
 * [ ] image-refresh fedora-coreos


cockpituous:
image-refresh fedora-coreos done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-coreos-20250529-225327

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-fedora-coreos-74909ee4-20250529-224430/log.html

Venefilyn:
Should we wait 2 weeks on this then? Or whenever the next refresh of coreos would be.

However, if this goes back to how it was before I am fine with this change. Feels like they corrected a regression in that case, but should probably find out before blindly merging. I'll double-check this week :pray:

jelly:
> error: not-found

That sounds like something from a Cockpit channel.

PR #7890, opened 2025-06-17
images: rhel: move stratisd-tools to test packages

jelly:
This package is used to create V1 pools on a stratis version which defaults to V2 pools and only used in our tests.

### DIFF
--- a/images/scripts/rhel.setup
+++ b/images/scripts/rhel.setup
@@ -195,7 +195,6 @@ subscription-manager \
 sos \
 stratis-cli \
 stratisd \
-stratisd-tools \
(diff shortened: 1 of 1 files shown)


images/scripts/rhel.setup
```diff
@@ -234,6 +233,7 @@ nmap-ncat \
 qemu-kvm \
 socat \
 strace \
+stratisd-tools \
```
martinpitt, line 236: Thanks! However, you also need to adjust line 250.

martinpitt (changes requested):

PR #7891, opened 2025-06-18
logs: Re-add inline links

Venefilyn:
This separates the title so that the test ID becomes an anchor link to
be able to link or go to specific entries.


martinpitt (approved):
Thanks!

Issue #7785, opened 2025-05-19
Update workflows to ubuntu-latest


jelly:
Try to update and check if we can resolve this workaround.

```
      # HACK: Ubuntu 22.04 has podman 3.4, which isn't compatible with podman-remote 4 in our tasks container
      # This PPA is a backport of podman 4.3 from Debian 12; drop this when moving `runs-on:` to ubuntu-24.04
      - name: Update to newer podman
        if: steps.changes.outputs.changed
        run: |
          sudo add-apt-repository -y ppa:quarckster/containers
          sudo apt install -y podman
          systemctl --user daemon-reload
```

@tomasmatus maybe something for you?

PR #7899, opened 2025-06-19
github: workflows: update test workflow to ubuntu-latest

jelly:
The latest ubuntu image is Ubuntu 24.04 which should have a recent enough podman version.


jelly:
Fixes https://github.com/cockpit-project/bots/issues/7785

martinpitt:
@jelly nice, thanks! Please drop the test commit and let's land.

PR #7914, opened 2025-06-24
lib: kill podman run --init if podman rm fails

jelly:
In CI `podman run` sometimes fails to stop a container with:

Error: reading CIDFile: open /tmp/tmpw4b91s99/cidfile: no such file or directory

The job-runner is then stuck in `gather_and_cancel` waiting on `task.cancel()` for `create_container` as this is stuck due to a bug kill the process.

`create_container` itself hangs on:

futex(0x5575a9ff4580, FUTEX_WAIT_PRIVATE, 0, NULL) = 0 futex(0xc00077c148, FUTEX_WAKE_PRIVATE, 1) = 1

---

See https://github.com/cockpit-project/bots/issues/7474#issuecomment-3000816973 for more detail. We need to open a new podman bug report.

### DIFF
--- a/lib/aio/job.py
+++ b/lib/aio/job.py
@@ -150,8 +150,11 @@ async def run_container(job: Job, subject: Subject, ctx: JobContext, log: LogStr
                     raise Failure(f'Container exited with code {returncode}')
 
             finally:
-                await run([*ctx.container_cmd, 'rm', '--force', '--time=0', f'--cidfile={tmpdir}/cidfile'],
-                          stdout=asyncio.subprocess.DEVNULL)  # don't show container ID output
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Thanks for your great debugging! Let's give this a try. I trust our integration tests enough for this.

I suppose you'll have to manually restart the bots one more time to mop up the remaining stragglers?

jelly:
> Thanks for your great debugging! Let's give this a try. I trust our integration tests enough for this.
> 
> I suppose you'll have to manually restart the bots one more time to mop up the remaining stragglers?

Right but other bots should just pull the latest bots by itself right?

martinpitt:
@jelly Correct. I meant the ones which are already stuck.

PR #7920, opened 2025-06-26
naughty: add targetcli Python 3.14 exception

jelly:
Doesn't hurt that much, but not that hard to get rid off example of the failure:

https://artifacts.dev.testing-farm.io/895de1ad-921f-434f-89c9-6b23d5fd016f/#artifacts-/plans/all/basic

### DIFF
--- a/naughty/fedora-43/7919-targetcli-exception
+++ b/naughty/fedora-43/7919-targetcli-exception
@@ -1,0 +1,1 @@
+ModuleNotFoundError: No module named 'rtslib_fb.utils'; 'rtslib_fb' is not a package


jelly:
> Well, wait -- this needs to include the actual error, not just the location. I.e. match on
> 
> ```
> ModuleNotFoundError: No module named 'rtslib_fb.utils'; 'rtslib_fb' is not a package
> ```
> 
> You can drop all the squiggly lines, and just one `targetcli` line should be enough.

Erm, but not all failing tests have that exception

jelly:
Ah, ok every failing test has the ModuleNotFound error but I just read over it as the whole test traceback is so big :)

martinpitt (approved):
Cheers!

PR #7942, opened 2025-07-01
Image refresh for rhel-8-10

cockpituous:
Image refresh for rhel-8-10
 * [x] image-refresh rhel-8-10
 * [x] https://github.com/cockpit-project/cockpit/pull/22161
 * [x] https://github.com/cockpit-project/cockpit-machines/pull/2211


cockpituous:
image-refresh rhel-8-10 done: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-8-10-20250701-232704

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-rhel-8-10-369d1e59-20250701-224459/log.html

jelly:
> The [rhel 8 machines branch](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7942-ca1a84a1-20250702-075252-rhel-8-10-cockpit-project-cockpit-machines-rhel-8/log.html) also seems to have broken consistently - it seems the OS list does not contain "CentOS 7" any more? But we still do that on main as well. Needs actual domain knowledge.

So CentOS 7 went EOL, so likely it no longer shows by default. Let me investigate.

jelly:
I can't reproduce this issue locally, was it fixed by the fake mouse events? Worth a retry imo.

jelly:
> > was it fixed by the fake mouse events
> 
> Definitively not, as that is the rhel-8 branch. This is either a time bomb (reached EOL date?) or a change in the image.

Can't be osinfo though as we also run:

[rhel-8-10/ws-container@cockpit-project/cockpit-machines](https://cockpit-logs.us-east-1.linodeobjects.com/pull-7942-ca1a84a1-20250702-073301-rhel-8-10-ws-container-cockpit-project-cockpit-machines/log.html)

martinpitt:
This reproduces just fine for me locally, investigating.

martinpitt:
This reproduces just fine for me locally, investigating. Indeed there is no CentOS 7 in the dropdown, only 9 (and not even 10 -- this osinfo is OLD!)

And indeed /usr/share/osinfo/os/centos.org/centos-7.0.xml has
```xml
    <release-date>2014-07-07</release-date>
    <eol-date>2024-06-30</eol-date>
```

and since this old version of c-machines shows OSes up to 1 year after EOL, that deadline just passed a few days ago. So, time bomb.

martinpitt:
Rebased, should work now.

PR #7946, opened 2025-07-02
lib: constants: define default memory size for our vm

jelly:
Move the hardcoded 1400 constant from Cockpit's "test/common/run-tests" to bots.


martinpitt (approved):
`REVIEW = true`

martinpitt:
Sorry, I was a bit too quick here:

> machine/machine_core/machine_virtual.py:MEMORY_MB = 1152

This is the current default size for destructive VMs, while [here](https://github.com/cockpit-project/cockpit/pull/22163/commits/ee271fe888e2cd4392a60091a16acad1d0993218) you are changing the nondestructive size. We bumped the latter for the LUKS tests, as the KDF doesn't get along with just 1.1 GB. So can you please rename replace the above constant with `DEFAULT_MACHINE_MEMORY_SIZE`, change that to 1152, and make the new constant `DEFAULT_NONDESTRUCTIVE_MACHINE_MEMORY_SIZE`?

martinpitt:
Yes, it is confusing; it's a trade-off that we need to make so that we can run most of the LUKS tests as nondestructive, and still don't waste memory in multi-provision tests. I.e. if you do a destructive LUKS test, you have to specify `memory_mb=1400` explicitly in `provision`.

I'm fine with adding both constants here as well, or keep the non size in run-tests only -- if you have an opinion about that, let's go with yours.

PR #7951, opened 2025-07-03
lib: constants: re-define default memory size

jelly:
The 1400 memory size is for non-destructive machines while destructive (the default) use 1152. Also use this constant in our machine class now and remove the "non-public" MEMORY_MB definition.


martinpitt (approved):
Cheers! I triggered three tests, just to be sure.

jelly:
Thanks!

PR #7953, opened 2025-07-03
images: fedora: add dnf5daemon-server as new Cockpit dep

jelly:
The plan is to move from PackageKit to dnf5daemon-server and dnf5daemon-server is not automatically installed on Fedora systems.

 * [x] image-refresh fedora-42

---

Still a bit early but this will eventually make test draft PRs easier and now is a good calm time to refresh Fedora :)


images/scripts/fedora.setup
```diff
@@ -50,6 +50,7 @@ udisks2-iscsi \
 
 COCKPIT_DEPS="\
 criu-libs \
+dnf5daemon-server \
```
martinpitt, line 53: I *think* that even exists on Fedora 41, it's just not being used yet (as the default dnf is still 4)

martinpitt (approved):

PR #7961, opened 2025-07-07
used-tasks-tags: Print tags comma separated

martinpitt:
That way the output fits [1], and for human consumption it does not matter much.

[1] https://github.com/marketplace/actions/ghcr-io-cleanup-action


jelly (approved):
Thanks!

PR #7991, opened 2025-07-15
machine.py: allow provisioning machine with an IPv6

tomasmatus:
Adds an `address6` field to provision object during machine setup.


martinpitt (approved):
Yay, thanks! Triggering a cockpit test to make triple sure this doesn't break anything. Please wait for these.

PR #8025, opened 2025-07-23
various machines/ssh_connection cleanups

allisonkarlitskaya:
This is mostly cleanups related to resolving the weird relationship between the `SSHConnection` and `Machine` classes: `Machine` derives from `SSHConnection` but we never ever create an instance of `SSHConnection` which is not also a `Machine`.

This batch of changes is related to cleaning up and removing redundancy from our handling of building `ssh` command-lines for our various APIs (spawn, execute, write, upload, download, etc.) and is a nice bunch of cleanups in any case.

I'm mostly sending these in now to get them tested and reviewed while I continue with further work.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -38,6 +38,7 @@ from lib.constants import TEST_DIR
 
 class SSHConnection:
     ssh_default_opts = (
+        "-F", "none",
         "-o", "BatchMode=yes",
         "-o", "IdentitiesOnly=yes",
         "-o", "PKCS11Provider=none",
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -45,7 +45,7 @@ class SSHConnection:
         "-o", "StrictHostKeyChecking=no",
         "-o", "UserKnownHostsFile=/dev/null",
     )
-    ssh_master: str | None
+    ssh_control_path: str | None
     ssh_process: subprocess.Popen[bytes] | None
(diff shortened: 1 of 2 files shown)


machine/machine_core/ssh_connection.py
```diff
@@ -343,24 +343,14 @@ def upload(self, sources: Sequence[str], dest: str, relative_dir: str = TEST_DIR
         if not self.__ssh_direct_opt_var():
             self._ensure_ssh_master()
 
-        if use_scp:
```
martinpitt, line 346: This is not so much "backwards" and more about supporting images/targets which don't have `rsync` installed. This might well still be the case on our fedora-coreos or other images, and it also breaks `--machine` cases. Does that hurt in any way? I'm very hesitant to drop this and turn rsync into a hard requirement.

martinpitt (commented):
Thanks! This looks mostly good, except that I'm sceptical about turning rsync into a dependency. To be honest I'd rather drop it completely and always `scp` to reduce the number of code paths.

Note that it's less important to do 50 different tests on fedora-42 and more to cover all our images. I triggered a few more (The debian/ubuntu ones all have rsync installed explicitly). But this e.g. doesn't cover gating tests in Testing Farm.

allisonkarlitskaya:
The issue is that the `/usr/local/share/cockpit/` directory doesn't exist in advance.  Hmm.  `scp` needs a `mkdir -p` mode...

martinpitt (approved):
Nice, thanks! I love it.

PR #8052, opened 2025-07-30
logs: round overall test progress percentage

jelly:
Round the progress percentage down to a non-fractioned number.

Fixes:

Screenshot From 2025-07-30 13-14-03


jelly:
https://github.com/cockpit-project/bots/pull/8053

PR #8053, opened 2025-07-30
logs: round overall test progress percentage

jelly:
Round the progress percentage down to a non-fractioned number.


Venefilyn (approved):

martinpitt:
@jelly did you test this? Now I ssee

image

i.e. no percentage at all.

PR #8068, opened 2025-08-04
Image refresh for ubuntu-2404

cockpituous:
Image refresh for ubuntu-2404
 * [x] image-refresh ubuntu-2404


cockpituous:
Success. Log: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/image-refresh-ubuntu-2404-0b9b90e8-20250804-043625/log.html

martinpitt:
S3 mirror fixed. Still a regression, needs investigation.

Venefilyn:
I also get the issue on `fedora-42` so seems unrelated to the image refresh. Triggered here just in case.

However, if we forcefully remove a drive, why would the logical volume show the error when it is more of a volume group issue when the disk is removed? Or is it supposed to show for both the vg and lv? Gosh this is confusing,

image

PR #8085, opened 2025-08-07
log.html: Adjust pixel test regexp

martinpitt:
https://github.com/cockpit-project/cockpit/commit/76cd6019b9c504646b changed the string, adjust the pattern accordingly.

While at it, also allow `_` as part of the file name.

----

See https://github.com/cockpit-project/cockpit/pull/22287#discussion_r2256122138

### DIFF
--- a/lib/s3-html/log.html
+++ b/lib/s3-html/log.html
@@ -314,7 +314,7 @@ Mustache.parse(link_template);
 let link_patterns = [
     {
         "label": "changed pixels",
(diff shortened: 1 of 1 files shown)


martinpitt:
> Thanks! So we need to roll out the updated `test/common` to all projects asap?

Right, good point -- I just triggered lib-update PRs for all our projects.

> What is this copy here in bots for?

That's the static copy in log.html , it can't/shouldn't include anything from cockpit.git.

PR #8101, opened 2025-08-13
stores: designate one of our S3 buckets "private"

allisonkarlitskaya:
Our attempts to mitigate scraper bots by restricting all access to images is causing significant grief for new contributors: having to privately message a few designated members of the Cockpit team in order to get an S3 token to be able to start testing your first contribution isn't particularly friendly or scalable.

One solution would be to generate a well-known "public" secret key for downloading images.  The idea is that the scraper bots would not be smart enough to use this key, but casual contributors would still be able to come across it easily enough.  One issue with this approach is that Linode's S3 ACLs are not sufficiently powerful to allow us to specify that this key can only download the public (read: not RHEL) images: we can only control access at the bucket level.

As a workaround, we can take advantage of the fact that we currently have two more-or-less equal S3 buckets and arbitrarily designate the EU one as "private" and the US one as "public".  We can upload the RHEL images to only the "private" EU mirror and then give "public" access to the US mirror for all the other images.

What happens with the new "public" secret key is currently up for debate.  For the time being, let's keep it as a well-known secret passed around Cockpit developers and shared freely with anyone who asks.  At some point we may also consider to publish it in a README or even integrate it into the downloader script directly.

### DIFF
--- a/image-download
+++ b/image-download
@@ -47,7 +47,7 @@ from lib import s3
 from lib.constants import IMAGES_DIR
 from lib.directories import get_images_data_dir
 from lib.network import get_curl_ca_arg, redhat_network
-from lib.stores import PUBLIC_STORES, REDHAT_STORES
+from lib.stores import PRIVATE_STORES, PUBLIC_STORES, REDHAT_STORES
 from lib.testmap import get_test_image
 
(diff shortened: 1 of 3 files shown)


mvollmer (approved):
Thanks!

PR #8123, opened 2025-08-18
images/scripts: add debian-trixie

allisonkarlitskaya:
Let's see how this goes...

 * [x] image-refresh debian-trixie


mvollmer:
> Many of the test failures can probably be fixed by making `multihost_enabled` False for "debian-stable". The plan is to make a new image called "debian-trixie" instead, which will make this kind of transition easier.

Which has just happened! :konfetti:

cockpituous:
image-refresh debian-trixie done: https://github.com/cockpit-project/bots/commits/image-refresh-debian-trixie-20250819-100503

PR #8129, opened 2025-08-19
tests-trigger: add support for wildcard matches

allisonkarlitskaya:
This can be helpful if you want to trigger something like '*/networking@*', for example.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -331,12 +331,12 @@ def projects() -> Iterable[str]:
 
 
 def get_default_branch(repo: str) -> str:
+    """Get the default branch.  Currently this is always 'main' but it might
+    change if we add repositories with a branch like 'devel' or so."""
     branches = REPO_BRANCH_CONTEXT[repo]
     if 'main' in branches:
         return 'main'
-    if 'master' in branches:
-        return 'master'
-    raise ValueError(f"repo {repo} does not contain main or master branch")
+    raise ValueError(f"repo {repo} does not contain main branch")
 
 
 def tests_for_project(project: str) -> Mapping[str, Sequence[str]]:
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
So I did this from the standpoint of running it from the bots repository itself, similar to `image:`.  Maybe that's not right.  I think maybe we also want to consider that it's run from the individual target repositories and then limit it to the scenarios defined on that repo?

lib/testmap.py
```diff
@@ -374,6 +374,24 @@ def tests_for_image(image: str) -> Sequence[str]:
...
+    """Return all contexts which match the given pattern."""
+
+    # Iterate all scenarios on all repositories, excluding _manual
+    tests = set()
+    for repo, branch_contexts in REPO_BRANCH_CONTEXT.items():
+        for branch, contexts in branch_contexts.items():
+            if branch.startswith('_'):
+                continue
+            for context in contexts:
+                c = context + '@' + repo
```
tomasmatus, line 387: like you say this probably isn't right. I think this should work on the target repository so when I do
```
./tests-trigger --repo cockpit-project/cockpit 22330 '*/networking'
```
I'd like to trigger all images that run cockpit tests and their `/networking` scenario

allisonkarlitskaya:
> I'd say it's fine to only support wildcard match for image variants and maybe contexts for specific image? So basically `*/networking` or `fedora-42/*` and the latter kinda only makes sense for cockpit. What do you think?

I kinda feel like doing it at the string level is actually easier... it might be "too powerful" in some sense, but it's also just kinda the easiest way to do it...

But probably we do need to do something about the launch context: if we're in a repo other than bots we probably want to only refer to the contexts present in that repo.  I'll redo this a bit...

PR #8163, opened 2025-08-26
images: Clean up old packages earlier on Fedora images

martinpitt:
dnf auto-removes alleged "unused dependencies" even if they got manually installed [1]. This removed `netcat` from our fedora-43 image, which we need for testing.

Fix that by removing cloud-init before installing our test deps, and also before the `dnf upgrade` (which makes sense anyway -- no point in updating packages which we'll remove afterwards).

While at it, remove old kernels earlier as well (but after the system upgrade).

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2390986

----

 * [x] image-refresh fedora-43


cockpituous:
image-refresh fedora-43 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-43-20250826-074434

cockpituous:
Success. Log: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/image-refresh-fedora-43-86415776-20250826-073111/log.html

PR #8169, opened 2025-08-27
naughty: Add alternative pattern for #6178

martinpitt:
systemd-coredump shows both messages ("dumped core" and "terminated abnormally"). Let's not predict which one happens first, so just add both patterns.

---

Addresses our current top flake, [example](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-22359-640655e3-20250825-105056-ubuntu-stable-other/log.html)

image

Unfortunately this doesn't happen *at all* locally, I tried 50 iterations.


mvollmer (approved):
Thanks!

PR #8182, opened 2025-09-01
WIP: 40374a13f18a: [no-test] Image refresh for fedora-rawhide-boot

cockpituous:
Image refresh for fedora-rawhide-boot
 * [ ] image-refresh fedora-rawhide-boot

Issue #8059, opened 2025-07-31
mdadm fails with "Unable to initialize sysfs" since 6.17.0 rc0


jelly:
In rawhide the [testNotRemovingDisks test fails](https://artifacts.dev.testing-farm.io/20ab7624-6194-44d7-abc6-52544e40cf16/)

```
> warn: Error starting RAID array: Process reported exit code 1: mdadm: Unable to initialize sysfs
mdadm: No arrays found in config file or automatically

> error: Error starting RAID array: Process reported exit code 1: mdadm: Unable to initialize sysfs
mdadm: No arrays found in config file or automatically
```

This clearly comes from `mdadm`, via a udisks dbus call (which is called in libblockdev).

I can't figure out the exact `mdadm` reproducer but I gathered strace output: [mdadm-strace.log](https://github.com/user-attachments/files/21534476/mdadm-strace.log)

Downstream report: https://bugzilla.redhat.com/show_bug.cgi?id=2385871

cockpituous:
rhel-9-8
Ooops, it happened again<hr>
```
mdadm: error opening /dev/md/ARR: No such file or directory
/dev/sda:   11593         
umount: /dev/sda: not mounted.
Traceback (most recent call last):
  File "/work/make-checkout-workdir/test/verify/check-storage-mdraid", line 319, in testNotRemovingDisks
    b.wait_text(self.card_desc("MDRAID device", "State"), "Running")
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 904, in wait_text
    self.wait_js_cond("ph_text_is(%s,%s)" % (jsquote(selector), jsquote(text)),
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                      error_description="() => 'actual text: ' + ph_text(%s)" % jsquote(selector))
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/make-checkout-workdir/test/common/testlib.py", line 842, in wait_js_cond
    raise Error(f"timeout\nwait_js_cond({cond}): {last_error.msg}") from None
[...]

PR #8183, opened 2025-09-01

KKoukiou:
This reverts commit 7616a91b2df7813fa6f4e8355efa3de39f9cdefd.

We can only close this once the fix is present also in the Rawhide ISO, which is always a few days later than the package composes.

See https://github.com/cockpit-project/bots/pull/8182


martinpitt (approved):
Oops! I also re-opened the known issue #8059. It would have been easier to push that commit into #8182, but you can rebase it now.

PR #8199, opened 2025-09-03
images: arch: switch to geo based mirror

jelly:
The America sponsored mirror was retired, but instead we can switch to a better global mirror automatically returns the closest available mirror fur us (based on geoip).

 * [x] image-refresh arch
 * [ ] https://github.com/cockpit-project/cockpit/pull/22398


martinpitt:
Retracting review request, as this still needs to be fixed. Put it onto the pilot board instead. I'll get to it, after the umpteen other things.

PR #8207, opened 2025-09-04
lib: move coreos package building to Fedora 42

jelly:


martinpitt (approved):
Thanks! Assuming this works. After that we can also just remove f41.

PR #8208, opened 2025-09-04
images: drop Fedora-41 image

jelly:
Only subscription-manager was left over in the testmap, with the impending EOL of Fedora-41 and subscription-manager not really being a thing on Fedora drop it from the testmap.


martinpitt (approved):
:broom:  thanks!

PR #8219, opened 2025-09-09
images: Install systemd-experimental

Lunarequest:
/usr/lib/systemd/system-generators/systemd-ssh-generator is required by the testing framework without it the guest never listens to the vsock meaning we can never connect to it. On opensuse it is now provided by the systemd-experimental package.

### DIFF
--- a/images/opensuse-tumbleweed
+++ b/images/opensuse-tumbleweed
@@ -1,1 +1,0 @@
-opensuse-tumbleweed-dc9c38579160cc3fdc38a7bcd79f2f55476661ee344ae27078cc24da86859de8.qcow2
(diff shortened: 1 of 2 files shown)


martinpitt:
@allisonkarlitskaya FYI, should make more of your table green or yellow?

martinpitt:
Building this in #8220, because technical reasons. Sorry for the noise. Thanks!

allisonkarlitskaya:
> @allisonkarlitskaya FYI, should make more of your table green or yellow?

Yes.  I already tried the image from the other PR and it moves opensuse to the yellow category.  Ephemeral support is still missing, though...

PR #8222, opened 2025-09-09
images: work around ssh's "UNKNOWN" DNS lookups

allisonkarlitskaya:
Fedora and Debian already carry patches to help mitigate this, but Arch and OpenSUSE are affected: when connecting from vhost, ssh incorrectly tells PAM and libaudit that the remote host is called "UNKNOWN".  Both of those components then attempt to perform lookups on this address and fail.

We can work around that in our images by adding an `/etc/hosts` entry for UNKNOWN to shortcut the lookup.  My first impulse was to use `::` here but `/etc/hosts` is also used for reverse DNS lookups and this might change the meaning of `::` for somebody, so let's use a completely bogus address that would never appear anywhere else.

 - [ ] image-refresh arch

### DIFF
--- a/images/scripts/arch.setup
+++ b/images/scripts/arch.setup
@@ -184,6 +184,11 @@ pacman -R --noconfirm cloud-init
 # https://github.com/cockpit-project/bots/issues/3901#issuecomment-1260579703
 systemctl disable systemd-time-wait-sync.service
 
+# https://gitlab.archlinux.org/archlinux/packaging/packages/openssh/-/issues/16
+# https://github.com/openssh/openssh-portable/pull/388
+# https://github.com/openssh/openssh-portable/pull/593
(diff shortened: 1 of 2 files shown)


cockpituous:
image-refresh opensuse-tumbleweed done: https://github.com/cockpit-project/bots/commits/image-refresh-opensuse-tumbleweed-20250909-070937

cockpituous:
image-refresh arch done: https://github.com/cockpit-project/bots/commits/image-refresh-arch-20250909-072200

cockpituous:
Success. Log: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/image-refresh-arch-2445f442-20250909-070943/log.html

allisonkarlitskaya:
Arch is good too.

martinpitt (approved):
Cheers!

PR #8228, opened 2025-09-10
job-runner: add support for forgejo/multiple forges

allisonkarlitskaya:
...this is enough to let you do something like:

```sh
./job-runner --debug run --env=XDG_RUNTIME_DIR=/tmp --env=TEST_OS=arch --forge codeberg --target cockpit-ci lis/test.thing
```

(which will fail for reasons that are specific to `test.thing`).


job-runner.toml
```diff
@@ -47,12 +47,20 @@ dir = '~/.cache/cockpit-dev/job-runner-logs'
 link = 'http://127.0.0.1:8000/'
 
 [forge]
-driver='github'
+default = 'github'
 
 [forge.github]
+driver = 'github'
```
martinpitt, line 53: I *think* with the merging we can leave https://github.com/cockpit-project/cockpituous/blob/main/ansible/roles/tasks-systemd/tasks/main.yml#L84 as-is, right?

job-runner.toml
```diff
@@ -47,12 +47,20 @@ dir = '~/.cache/cockpit-dev/job-runner-logs'
...
 clone-url = 'https://github.com/'
 api-url = 'https://api.github.com/'
 post = true  # whether to post statuses, open issues, etc.
 user-agent = 'job-runner (cockpit-project/bots)'
 # (at least) one of `token` or `post = false` must be set
 # token = 'ghp_XXX'
+
+[forge.codeberg]
```
martinpitt, line 61: ... but we eventually need to get a cockpituous (-ish) user on Codeberg and add a token for that to cockpituous.

lib/aio/jobcontext.py
```diff
@@ -105,19 +106,33 @@ async def __aenter__(self) -> Self:
                     }
                 self.default_image = get_str(container, 'default-image')
 
-            with get_nested(self.config, 'logs') as logs:
-                driver = get_str(logs, 'driver')
+            with get_nested(self.config, "logs") as logs:
+                driver = get_str(logs, "driver")
```
martinpitt, line 110: Not that it matters, but I'm curiuos: You have been very consistently converted our code from double to single quotes. Why the sudden change?

lib/aio/jobcontext.py
```diff
@@ -55,7 +55,8 @@
 class JobContext(contextlib.AsyncExitStack):
     config: JsonObject = {}  # noqa:RUF012  # JsonObject is immutable
     logs: LogDriver
-    _forge: Forge
+    _forges: dict[str, Forge]
```
martinpitt, line 58: I'd like to attach this comment to line 45, but can't. Doesn't "codeberg" need to be added to `FORGES` above?

job-runner.toml
allisonkarlitskaya: Probably.  Right now I only plan to use this for "command subjects" which are limited to:
 - trivial api requests for things like "what is the sha of your main branch?" and "what is the content of your `.cockpit-ci/container` file?".  Those are working fine without a token at present
 - `git clone`

I consider a potential topic of discussion if we should have Codeberg configured here or not (ie: or leave it to the user's config), but we configure GitHub, so...

lib/aio/jobcontext.py
allisonkarlitskaya: This was accidental.  The main story is that I have ` f` on muscle memory from working with Rust (where absolutely everyone is using `rust-fmt`) and `test.thing` (where I use `ruff format`).  I keep accidentally pressing ` f` to fix line wrapping when working on other projects but of course this accidentally reformats the whole file.  I clearly missed backing this change out.

lib/aio/jobcontext.py
martinpitt: `FORGE_DRIVERS` sounds good!

martinpitt (approved):

PR #8245, opened 2025-09-14
fix .cockpit-ci/container scanner for cross-forge

allisonkarlitskaya:
Fixes https://github.com/cockpit-project/cockpituous/issues/670

I'm not 100% sure on the approach: maybe just want to teach the script how to create codeberg GitHub() instances instead...


test/test_aio.py
```diff
@@ -135,7 +135,7 @@ def service() -> Iterator[GitHubService]:
 
 @pytest.fixture
 async def api(service: GitHubService) -> AsyncIterator[GitHub]:
-    async with GitHub(service.config) as github:
+    async with GitHub('github', service.config) as github:
```
martinpitt, line 138: "fakehub"?

lib/aio/github.py
```diff
@@ -66,8 +69,19 @@ def __init__(self, name: str, config: JsonObject) -> None:
 
     async def __aenter__(self) -> Self:
         headers = {}
-        # token is mandatory if `post = true`
-        if token := get_str(self.config, 'token', *((None,) if self.dry_run else ())):
```
martinpitt, line 69: > The most obvious fix (adjusting the config file to `post=false`) is not an option at present because our CI deployment doesn't contain an explicit `forge.github.post = true`

This is easy to change -- if that makes things any simpler, go right ahead.

test/test_aio.py
allisonkarlitskaya: I mean, I guess...

martinpitt (approved):
Cheers!

pyproject.toml
```diff
@@ -50,6 +50,9 @@ line-length = 118
 preview = true
 target-version = 'py312'
 
+[tool.ruff.format]
+quote-style = "preserve"
```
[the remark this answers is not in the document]
allisonkarlitskaya: It's not about having a preference for quote style.  It's about the fact that I keep making massive amounts of unintended changes.  If we want to do this, then I say let's do it, but let's make it intentional...

PR #8255, opened 2025-09-18
jira-tasks: Set severity for rebase issues

martinpitt:
Required for moving an issue to "In progress" or later.

---

I'm doing this manually for the existing issues, but let's automate for next time, it's annoying.

### DIFF
--- a/jira-tasks
+++ b/jira-tasks
@@ -48,6 +48,8 @@ class JiraClient:
             "summary": summary,
             "description": description,
             "priority": {"name": "Normal"},
+            # Severity
+            "customfield_12316142": {"value": "Moderate"},
             # Story Points
             "customfield_12310243": 2,
             # Preliminary Testing


jira-tasks
```diff
@@ -48,6 +48,8 @@ class JiraClient:
             "summary": summary,
             "description": description,
             "priority": {"name": "Normal"},
+            # Severity
+            "customfield_12316142": {"value": "Moderate"},
```
martinpitt, line 52: Coming from here:
```
❱❱❱ ./jira-tasks find-field Severity
Found 1 field(s) matching 'Severity':
  ID: customfield_12316142
  Name: Severity
  Custom: True
```

allisonkarlitskaya: Might have been nice as a comment :)

allisonkarlitskaya (approved):

martinpitt:
@allisonkarlitskaya The explanation is in the original e6ef91453ca27d6c29f87e7e4d6de4beb64e0722, which introduced a few other custom fields already, so nothing new. But heh, ok.. :)

PR #8256, opened 2025-09-18
images: remove Live ISO image

KKoukiou:
The Live ISO image was never used in the CI tests of anaconda-webui. It was previously added for manual testing, but turned out not to be very useful. Manual testing against Live ISOs can be done directly with the official images instead.


martinpitt (approved):
:broom: ephraristo!

PR #8296, opened 2025-09-29

KKoukiou:
This partitally reverts commit 21af6a0f5f924d83ab63f10e0b3c94d10a88ff9c.

The `naughty/fedora-X/8059-mdraid-init-sysfs` file was not put back, since the bug is affecting only anaconda tests and this naughty was only relevant to cockpit storage tests.


Venefilyn (approved):
Ack, thanks!

PR #8300, opened 2025-10-01
images: extend existing images for serving payload for Web UI installation tests

KKoukiou:
* [x] image-refresh fedora-rawhide-anaconda-payload
 * [x] image-refresh fedora-43-anaconda-payload

### DIFF
--- a/images/fedora-43-anaconda-payload
+++ b/images/fedora-43-anaconda-payload
@@ -1,0 +1,1 @@
+fedora-43-anaconda-payload-95dbff6ace6f01662b9e48637a969e0c0b7e2d1b420e352dd5f0b8d9d5b3e494.tar.gz


cockpituous:
image-refresh fedora-rawhide-anaconda-payload done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-rawhide-anaconda-payload-20251001-071549

cockpituous:
Success. Log: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/image-refresh-fedora-rawhide-anaconda-payload-a06215aa-20251001-070350/log.html

cockpituous:
image-refresh fedora-43-anaconda-payload done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-43-anaconda-payload-20251001-072910

images/scripts/create-anaconda-dnf-payload
```diff
@@ -0,0 +1,112 @@
...
+
+def build_dnf_payload(image: str, output: str) -> None:
+    fedora_version = image.split('-')[1]
+
+    run([os.path.join(BOTS_DIR, "image-download"), image])
+    machine = testvm.VirtMachine(image=image, memory_mb=4096)
+    try:
+        machine.start()
+        machine.wait_boot()
+        machine.execute("dnf install -y dnf5-plugins createrepo_c python3 rsync curl", timeout=300)
```
Venefilyn, line 66: This mixes `dnf` and `dnf5`. Intentional?

images/scripts/fedora-43-anaconda-payload.bootstrap
```diff
@@ -3,4 +3,9 @@ set -eux
 
 OUTPUT="$1"
 
-PYTHONPATH=. python3 images/scripts/create-anaconda-payload --output=$OUTPUT --image fedora-43
+WORKDIR=$(mktemp -d)
+PYTHONPATH=. python3 images/scripts/create-anaconda-liveimg-payload --output="$WORKDIR" --image fedora-43
+PYTHONPATH=. python3 images/scripts/create-anaconda-dnf-payload --output="$WORKDIR" --image fedora-43
```
Venefilyn, line 8: Not entirely sure how this works, does it set any environment variables here? If so they might collide on some variable names given the two scripts have similar functionality

images/scripts/create-anaconda-dnf-payload
KKoukiou: dnf executable defaults to dnf5 since few fedora versions.

images/scripts/fedora-43-anaconda-payload.bootstrap
KKoukiou: Hm, not sure what do you mean? Which variable do you have in mind as problematic?

images/fedora-43-anaconda-payload
martinpitt, line 1: Deleting the links here is strange, but as the refreshes bring it back, it's ok.

images/scripts/fedora-43-anaconda-payload.bootstrap
martinpitt, line 8: Each of these scripts does a new reboot/shutdown, which slows down the build. Not a biggie, just a bit of an inconvenience.

images/scripts/lib/anaconda-comps.xml
```diff
@@ -0,0 +1,117 @@
+<?xml version="1.0" encoding="utf-8"?>
```
martinpitt, line 1: I'm fairly sure this is copied from somewhere. Like that it's hard to maintain, you don't know the original source nor how it was modified, nor how to keep it up to date with future changes. Preferably replace this with some `curl` call to download it from some git, or at least add a comment with an URL?

images/scripts/lib/anaconda-comps.xml
```diff
@@ -7,51 +7,24 @@
     <default>true</default>
     <uservisible>true</uservisible>
     <packagelist>
-      <packagereq type="mandatory">audit</packagereq>
```
martinpitt: Ah sorry, didn't see this next commit yet. So I suppose the original file in the previous commit was a straight copy. How does it decide which are required and which can go? Will the tests fail if an "important" package is missing, or gets renamed etc.?

martinpitt (commented):
Not a blocker, but worried about maintainability. But ultimately this is your playground :grin:

images/scripts/lib/anaconda-comps.xml
martinpitt: I think minimizing the set is fine, it both shields you from unnecessary noise and also makes the images smaller and tests faster. I mostly just wondered where it came from, i.e. adding that URL as a comment (`<!---`) to the XML helps our future selves.

images/scripts/fedora-43-anaconda-payload.bootstrap
Venefilyn: No variables are problematic right now but I think the behavior makes it fragile. Can easily create a variable in one that exists in the other, which would be overridden or override the other if one is not careful of the variable naming

images/scripts/lib/anaconda-comps.xml
KKoukiou: OK I will add a comment for now, and send a follow up next week to try the curl approach for the comps file, and see if it severelly affects test execution times.

images/scripts/create-anaconda-dnf-payload
KKoukiou: Right :) I think most of these dependencies are actually not needed anymore, after some cleanups. Trying it out with keeping only createrepo_c which I still use

cockpituous:
image-refresh fedora-rawhide-anaconda-payload done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-rawhide-anaconda-payload-20251002-111816

cockpituous:
image-refresh fedora-43-anaconda-payload done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-43-anaconda-payload-20251002-113021

KKoukiou:
@martinpitt I added the comment in the XML file and removed some unneeded RPM dependencies, that I was installing previously. Can I get another review please?

martinpitt (approved):
@KKoukiou sure! But please trigger the image tests, to make sure the refresh doesn't break anything.

PR #8301, opened 2025-10-01
images: opensuse-tumbleweed add quay to unqualified-search-registries

Lunarequest:
Some cockpit tests pull in quay.io/cockpit/ws. On opensuse-tumbleweed quay.io is not in the normal unqualified-search-registries. This results in the tests failing. To solve this we update /etc/containers/registries.conf to include quay

### DIFF
--- a/images/scripts/opensuse-tumbleweed.setup
+++ b/images/scripts/opensuse-tumbleweed.setup
@@ -181,4 +181,8 @@ zypper clean
 # c-podman tests
 /var/lib/testvm/podman-images.setup
 
+# cockpit tests pull quay.io/cockpit/ws which fails because 
+# quay.io is not present in the unqualified-search-registries in /etc/containers/registries.conf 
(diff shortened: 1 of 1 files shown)


martinpitt:
@Lunarequest To be honest I'd rather use the opportunity to fix our tests to use fully qualified names -- either `quay.io/` or `localhost/` depending on the case. Modifying OS images that way just papers over bugs. Where did you notice that? I'm happy to send a fix.

jelly:
> I noticed this with `test/verify/check-system-info` in cockpit. I couldn't find where it was being pulled in. I also would prefer to switch to fully qualified names. I think `quay.io` would be the easiest solution

I don't think it would be needed for Cockpit, but for cockpit-podman we do require searching. The Arch image is also guilty of configuring unqaulified-search-registries so I would be happy to help with that, afaik it is only needed for cockpit-podman tests.

Do you have an example of it not working in `test/very/check-system-info`? Which test fails?

Issue #8325, opened 2025-10-08
`selinux-policy` not working in `fedora-42/updates-testing`


Venefilyn:
Likely related to an upstream change for `selinux-policy` that causes this error
https://github.com/cockpit-project/cockpit/issues/22461

`sshd` with the `sshd_session_t` policy was changed 
https://github.com/fedora-selinux/selinux-policy/commit/1177a3241b430ef236fc12afa56f618f1cd45b1e

There was a change to fix it for Cockpit but doesn't seem to have fixed it, so likely unrelated
https://github.com/fedora-selinux/selinux-policy/commit/4c98210ff26f81d6f1f1db1c8bdce377d5235636

PR #8326, opened 2025-10-08
naughty: Add selinux for f42/updates-testing

Venefilyn:
Likely related to an upstream change for `selinux-policy` that
causes this error
https://github.com/cockpit-project/cockpit/issues/22461

`sshd` with the `sshd_session_t` policy was changed
https://github.com/fedora-selinux/selinux-policy/commit/1177a3241b430ef236fc12afa56f618f1cd45b1e

There was a change to fix it for Cockpit but doesn't seem to have
fixed it, so likely unrelated
https://github.com/fedora-selinux/selinux-policy/commit/4c98210ff26f81d6f1f1db1c8bdce377d5235636

Related-to: https://github.com/cockpit-project/bots/issues/8325


naughty/fedora-42/8325-testing-branch-selinux-policy
```diff
@@ -0,0 +1,5 @@
+Traceback (most recent call last):
```
martinpitt, line 1: Is that this particular message?

> type=AVC msg=audit(1758938947.568:397): avc:  denied  { read } for  pid=2423 comm="sshd-session" name="authorized_keys" dev="vda4" ino=110469 scontext=system_u:system_r:sshd_session_t:s0-s0:c0.c1023 tcontext=unconfined_u:object_r:httpd_exec_t:s0 tclass=file permissive=1

Can you please add it to the pattern, so that we don't give that test too much carte blanche?

martinpitt (changes requested):
Thanks! Can you please file a bugzilla about that and refer to it in the commit message and #8325 description?

PR #8398, opened 2025-10-30
testmap: add devel scenario for sub-man-cockpit

tomasmatus:
Add fedora-42/devel but also rhel-10-2/devel to test if code coverage runs on RHEL. Sub-man-c has some code that is only used on RHEL so it would be nice to collect test coverage of it.


tomasmatus:
running sub-man-c tests on https://github.com/cockpit-project/subscription-manager-cockpit/pull/136

PR #8403, opened 2025-10-31

KKoukiou:
This reverts commit a537d581bdab2647f3e4e0238a4a6e1400f9818b.

This issue should not have been closed, it still happens, see https://github.com/cockpit-project/bots/pull/8389#issuecomment-3472480158


KKoukiou:
Ok seems the issue is not the mdadm bug reported.

PR #8412, opened 2025-11-04
Drop fedora-43 ISO and payload used in anaconda tests

KKoukiou:
Fedora-43 is out and anaconda stopped releasing for Fedora-43 as we do not bother about updates testing.


jelly (approved):
Sure!

jelly (changes requested):
This removes all fedora-43 naughties while cockpit still tests on it

PR #8541, opened 2025-12-13
Image refresh for debian-trixie

cockpituous:
Image refresh for debian-trixie
 * [x] image-refresh debian-trixie


martinpitt:
Ignoring ruff failure, that fix just landed.

PR #8540, opened 2025-12-13
Image refresh for centos-9-stream

cockpituous:
Image refresh for centos-9-stream
 * [x] image-refresh centos-9-stream


martinpitt (approved):
Ignoring ruff failure, that fix just landed.

PR #8560, opened 2025-12-19
images: switch to fastly mirror for the Arch Linux image

jelly:
* [x] image-refresh arch


cockpituous:
image-refresh arch done: https://github.com/cockpit-project/bots/commits/image-refresh-arch-20251219-184636

PR #8562, opened 2025-12-20
Image refresh for rhel-9-8

cockpituous:
Image refresh for rhel-9-8
 * [x] image-refresh rhel-9-8


tomasmatus:
that looks like a sub-man issue, not the cockpit plugin 🤔

PR #8621, opened 2026-01-13
machine_core: Scale wait_[re]boot() by TEST_TIMEOUT_FACTOR

martinpitt:
Cockpit's testlib defines a `$TEST_TIMEOUT_FACTOR` to scale the timeouts for slow test bed runners. Apply this for VM (re)boots as well. This will help with avoiding boot timeouts on our OpenStack runners with busy/noisy neighbors, when our CI bots get descheduled. These functions measure wallclock time, not effective CPU time (and doing the latter isn't possible in Python).

----

This supports https://github.com/cockpit-project/cockpit/pull/22748 , in my vendetta to [fix this mess](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-22373-08af2109-20260109-101120-fedora-43-devel/log.html). Note that the current cockpit main testlib still defines TEST_TIMEOUT_FACTOR as an integer, but I'm changing that to a float.

### DIFF
--- a/machine/machine_core/ssh_connection.py
+++ b/machine/machine_core/ssh_connection.py
@@ -89,8 +89,8 @@ class SSHConnection:
         # If connected to machine, kill master connection
         self._kill_ssh_master()
 
-        start_time = time.time()
-        while (time.time() - start_time) < timeout_sec:
+        start_time = time.monotonic()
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (dismissed):
Looks good, minus some minor nits.

PR #8623, opened 2026-01-14
log.html: Add automatic dark mode

martinpitt:
Follow the system/browser preference for light/dark mode, and set PatternFly's dark mode class accordingly.

https://www.patternfly.org/developer-resources/dark-theme-handbook/#enabling-dark-theme

----

On http://localhost:8000/log.html it now follows my light/dark desktop switches, and in dark mode looks like this:

image

It's not *super* great, but then again that whole log viewer isn't great in light mode either. It at least stops stinging my eyes when I'm working at night :wink:

### DIFF
--- a/lib/s3-html/log.html
+++ b/lib/s3-html/log.html
@@ -96,7 +96,21 @@
             </span>
         </script>
         <script type="text/javascript">
-            addEventListener("click", (event) => { })
+            // Automatic dark mode based on system preference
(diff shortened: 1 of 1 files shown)


martinpitt:
@allisonkarlitskaya sadly not. PatternFly v6's dark theme is designed to be class-based (requires `.pf-v6-theme-dark` on the <html> element). CSS cannot add classes to elements. I've had Claude spit out some CSS code which sets `color-scheme`, but that doesn't work sadly.

PR #8624, opened 2026-01-14
log.html: Use template strings instead of mustache

martinpitt:
Define a lit-like html`..` tagged string to signal intent. It won't
actually make nvim etc. format the syntax properly, as this is still a
.html file, not a .js one. But it signals intent, and it can in theory
be done with TreeSitter.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

----

Looks and works the same. This was just a few minutes doodling with Claude, so if you don't like this, I don't have much attachment or a strong opinion about this.

### DIFF
--- a/lib/s3-html/log.html
+++ b/lib/s3-html/log.html
@@ -5,7 +5,6 @@
         <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <link href="https://cdn.jsdelivr.net/npm/@patternfly/patternfly/patternfly.css" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


lib/s3-html/log.html
```diff
@@ -589,7 +522,24 @@
...
+                />
+
+                <span class="pf-v6-c-switch__toggle"></span>
+
+                <span
+                    class="pf-v6-c-switch__label"
+                    id="switch-for-failed-filtering-text"
+                    aria-hidden="true"
+                >Show only failed</span>
+            </label>`;
```
jelly: Hmm there is nothing in this template? Why is it a template?

martinpitt: Right you are! I moved this to static inline HTML.

PR #8625, opened 2026-01-14
Dark mode colors

jelly:
The current retry /failure  color does not work well with dark mode as its not a "special PF token blah blah" color.

image

This PR changes the retry color to purple so it matches the label group and is more distinctly different from failures. Plus it switches failures to the same color as the failure label group:

image

As bonus marks "RETRY" to make it even more clear, we can also drop that.


martinpitt (approved):
Thanks!

PR #8631, opened 2026-01-16
images: fix asciidoctor build dependency on tumbleweed

jelly:
Tumbleweed is rolling release so these days its ruby 4.

 * [ ] FAIL: image-refresh opensuse-tumbleweed


cockpituous:
Failed. Log: https://logs-cockpit.apps.ocp.cloud.ci.centos.org/image-refresh-opensuse-tumbleweed-38c83a28-20260116-095554/log.html

PR #8688, opened 2026-02-03
Image refresh for centos-10

cockpituous:
Image refresh for centos-10
 * [x] image-refresh centos-10


jelly:
Reported https://issues.redhat.com/browse/RHEL-146213

mvollmer:
> > Yes, agreed. RHEL really isn't supported on < 1 GB RAM. I also changed `run-tests` to schedule tests with more `provision:`ed RAM correctly, so bumping RAM is what we should do.
> 
> We can have "minimum RAM", then we don't need to hunt down all the provisions. That could even be per OS. I'll try with a global 1G minimum.

As @martinpitt said elsewhere, let's hunt down all the provisions: https://github.com/cockpit-project/cockpit/pull/22853

mvollmer:
> As @martinpitt said elsewhere, let's hunt down all the provisions: [cockpit-project/cockpit#22853](https://github.com/cockpit-project/cockpit/pull/22853)

Landed, let's rerun tests.

PR #8692, opened 2026-02-05
testmap: enable Anaconda bootopts-net1 test scenario

rvykydal:
Related: INSTALLER-4524

Enabling the tests added in https://github.com/rhinstaller/anaconda-webui/pull/1087/commits/883f812b0edd9a31dd8de48e8e05194069372a33


rvykydal:
@KKoukiou - do you think we can add the scenario?
Currently It runs 3 network tests on a separate (specific network configuration boot options) VM.

PR #8740, opened 2026-02-19
images: Refresh fedora-rawhide

cockpituous:
To get the signature enforcing version of rpm.


mvollmer:
`image-customize` needs to sign packages now.

PR #8755, opened 2026-02-24
testmap: add explicit wscontainer relationship dependency

allisonkarlitskaya:
A recent update of the `fedora-43` container changed the contained `quay.io/cockpit/ws` container inside of it from being based on `fedora-42` to `fedora-43`.  Since we build the container for the `rhel-8-10/ws-container@cockpit-project/cockpit` scenarios on top of it, this changed the version of Fedora (and Python) that those scenarios run on.  We don't gate `fedora-43` updates on the `rhel-8-10/ws-container` scenarios, though, so we never caught this problem as we introduced the new image.

We're going to switch to only building the RPMs on the builder image and then copying them into the container image that's already cached in the `rhel-8-10` image.  That's Fedora 42 at the moment.

Let's start by making that relationship explicit in the bots so that we can test it.  We will follow this up with a commit in the cockpit repository that makes use of the new API we add here.

The idea in the future that we'll bump this when we build a rhel-8-10 image which contains a `cockpit/ws` container based on a new Fedora image.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -243,6 +243,13 @@ OSTREE_BUILD_IMAGE = {
     "fedora-coreos": "fedora-43",
 }
 
+# ws-container scenarios build RPMs for the cockpit/ws container on a different
+# image than the one being tested.  This must match the base OS version used in
+# the cockpit/ws container present in the given image.
+WSCONTAINER_BUILD_IMAGE = {
+    "rhel-8-10": "fedora-42",
+}
+
 # only put auxiliary images here; triggers for primary OS images are computed from testmap
 IMAGE_REFRESH_TRIGGERS = {
     "services": [
@@ -272,6 +279,10 @@ IMAGE_REFRESH_TRIGGERS = {
 # in fedora-X
 def get_build_image(image: str) -> str:
     return OSTREE_BUILD_IMAGE.get(image, image)
+
+
+def get_build_image_for_ws_container_inside_of(image: str) -> str | None:
+    return WSCONTAINER_BUILD_IMAGE.get(image)
 
 
 # some tests have suffixes that run the same image in different modes; map a
@@ -384,6 +395,11 @@ def tests_for_image(image: str) -> Sequence[str]:
         if image == i:
             tests.update(_direct_tests_for_image(a))
             break
+
+    # is this a build image for ws-container? then add those scenario tests
+    for test_image, build_image in WSCONTAINER_BUILD_IMAGE.items():
+        if image == build_image:
+            tests.update(_direct_tests_for_image(test_image, 'ws-container*'))
 
     return list(tests)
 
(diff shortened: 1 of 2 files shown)


lib/testmap.py
```diff
@@ -385,6 +396,11 @@ def tests_for_image(image: str) -> Sequence[str]:
             tests.update(_direct_tests_for_image(a))
             break
 
+    # is this a build image for ws-container? then add those scenario tests
+    for test_image, build_image in WSCONTAINER_BUILD_IMAGE.items():
+        if image == build_image:
+            tests.update(_direct_tests_for_image(test_image, 'ws-container*'))
```
martinpitt: Feels like that data structure is inside out 😁

test/test_testmap.py
```diff
@@ -100,3 +100,20 @@ def test_tests_for_image_ostree() -> None:
...
         assert ostree_tests, \
             f"{build_image} refresh should trigger {ostree_image} tests"
+
+
+def test_get_build_image_for_ws_container_inside_of() -> None:
+    # check that mapped images return their build image
+    for test_image, build_image in testmap.WSCONTAINER_BUILD_IMAGE.items():
+        assert testmap.get_build_image_for_ws_container_inside_of(test_image) == build_image
+    # images without ws-container scenarios return None
+    assert testmap.get_build_image_for_ws_container_inside_of("debian-testing") is None
```
martinpitt: TBH these are quite a shallow unit tests for a very shallow function. This is just data, and we shouldn't validate data by validating it against the same data.. Not a veto, but I think if we are hardcoding OSes anyway, then let's just keep the last line and add a line that ensures that we get "fedora-42" for "rhel-8-10". Yes, we'll have to update this along with the actual data structure, but it's either that, or not worth having at all IMHO.

lib/testmap.py
```diff
@@ -348,19 +349,27 @@ def tests_for_project(project: str) -> Mapping[str, Sequence[str]]:
...
             for context in contexts:
-                if context.split('/')[0].replace('-distropkg', '') == image:
-                    c = context + '@' + repo
-                    if branch != get_default_branch(repo):
-                        c += "/" + branch
-                    tests.add(c)
+                context_image, _, context_scenario = context.partition('/')
+                # -distropkg is a test mode suffix, not a separate image
+                if context_image.removesuffix('-distropkg') != image:
+                    continue
```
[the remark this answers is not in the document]
allisonkarlitskaya: It's really funny that you had *exactly* the same thought/misunderstanding about what this was doing, but it's not.  It's comparing the image name from the *context* to the image that we're looking for.  ie: the variable name on the left and right side are different.

lib/testmap.py
```diff
@@ -273,6 +281,10 @@ def get_build_image(image: str) -> str:
     return OSTREE_BUILD_IMAGE.get(image, image)
 
 
+def get_build_image_for_ws_container_inside_of(image: str) -> str | None:
+    return WSCONTAINER_BUILD_IMAGE.get(image)
```
martinpitt: So what's the point of splitting this out even? Nothing uses it.

PR #8769, opened 2026-03-02
Add Alma Linux image for local testing

jelly:
For https://github.com/cockpit-project/cockpit-files/issues/1328#issuecomment-3986256637

### DIFF
--- a/images/scripts/alma-9.bootstrap
+++ b/images/scripts/alma-9.bootstrap
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -ex
+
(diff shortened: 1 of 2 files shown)


jelly:
> So what's the intent, just keep the files for local `./image-create`? Or actually build/refresh/push it to S3?

Yes, only local for `./image-create`, I don't think we want to use additional CI capacity for Alma Linux.

> Does it actually make sense to call it `-9-7`, or can we perhaps just call it `-9`?

We can call it -9 imo, that works.

PR #8804, opened 2026-03-12
JRaaS (job-runner as a service, via testing-farm)

allisonkarlitskaya:
Add a test plan for receiving and running a `JOB_JSON` in Testing Farm.

Add an API in lib/testingfarm.py for sending jobs to this service.  This
uses the configuration serialization support from the previous commits
to serialize the job-runner configuration and send it to testing-farm
along with a job in JSON form.

Wire that up to `--to=testing-farm` on the `job-runner` CLI.

You can test this with something like
```
./job-runner --to=testing-farm run cockpit-project/bots
```
which will use your local job-runner configuration along with any
secrets that it contains or refers to.


allisonkarlitskaya:
The serialization of file content is maybe overwrought.  I originally wanted something like `{"": "content"}` for files and `{"a": "a content", "b": "b content"}` for subdirectories but I got worried that maybe we want to add permissions or such at some point and :robot: convinced me to do it in the current more-verbose way... I think it lost a lot of charm along the way, but meh...

Also the idea of building virtual secrets partially containing filesystem paths is a bit bonkers.  That fell out of the way the code recursively calls itself but I'm not sure if it's a feature that we actually want.

test/test_aio.py
```diff
@@ -424,3 +424,50 @@ async def test_secrets_conflict_error(tmp_path: Path) -> None:
...
+        serialized = ctx.serialize()
+
+    # Verify serialized format
+    secrets = serialized['secrets']
+    assert isinstance(secrets, dict)
+    assert secrets['external'] == {}
+    inline = secrets['inline']
+    assert isinstance(inline, dict)
+    assert inline['github-token'] == 'ghp_secret123'
+    assert inline['certs'] == {'ca.pem': '-----BEGIN CERT-----'}
```
martinpitt: This could just compare the full `serialized` object against the expected value. That makes sure you don't forget to check for some keys, or there aren't any unexpected keys there.

job-runner
```diff
@@ -79,7 +80,13 @@ async def main() -> None:
...
             assert_never(other)
 
-    async with JobContext(args.config_file, debug=args.debug) as ctx:
+    config: JsonObject | None
+    if config_json := os.environ.get('JOB_RUNNER_CONFIG_JSON'):
+        config = json.loads(config_json)
+    else:
+        config = None
+
+    async with JobContext(args.config_file, config=config, debug=args.debug) as ctx:
```
martinpitt: This is the "xor" assertion above: what happens if `JOB_RUNNER_CONFIG_JSON` is set and `--config-file` is given?

job-runner
```diff
@@ -86,8 +87,25 @@ async def main() -> None:
...
         config = None
 
-    async with JobContext(args.config_file, config=config, debug=args.debug) as ctx:
-        await run_job(job, ctx)
+    ctx = JobContext(args.config_file, config=config, debug=args.debug)
+
+    match args.to:
+        case 'testing-farm':
+            request_id = await submit_to_testing_farm(ctx, job_json)
+            artifacts_url = await wait_for_artifacts(request_id)
```
martinpitt: Please add a comment pointing to https://issues.redhat.com/browse/TFT-4379 . Hopefully this whole polling hack can disappear at some point.

lib/aio/testingfarm.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+
+    if git_url_ref is None:
+        git_url_ref = await get_git_upstream()
+
+    git_url, git_ref = git_url_ref
+    config_json = json.dumps(ctx.serialize())
+    job_json = json.dumps(job)
+
+    request = {
```
martinpitt: Please add a link to https://api.dev.testing-farm.io/docs#operation/request_a_new_test_v0_1_requests_post

martinpitt (changes requested):
Thanks! My gut doesn't really like all the serialization, file writing etc. - but I see where this is going, and as you did the work, let's get it in. I have some comments about details.

job-runner
allisonkarlitskaya: Environment wins.  Which is not optimal, I guess.  Normally commandline beats environment.

job-runner
allisonkarlitskaya: Sure.

job-runner
```diff
@@ -86,8 +87,25 @@ async def main() -> None:
...
+            if artifacts_url:
+                print(artifacts_url)
+            else:
+                sys.exit(f'Job failed to start: {request_id}')
```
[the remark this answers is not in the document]
allisonkarlitskaya: The API URL I guess you mean?

PR #8823, opened 2026-03-18
jira-tasks: make it work with new jira

tomasmatus:
Change base_url to new Jira, atlassian hosted Jira uses Basic Auth for ad-hoc api calls from scripts. See:
https://developer.atlassian.com/cloud/jira/platform/rest/v2/intro/#authentication

Authentication uses Base64 encoded 'email@redhat.com:API_KEY' so the jira token file should now contain two lines:
1. your_mail@redhat.com
2. API key generated in Jira

Custom fields are set to new values, these values I found eg. by `./jira-tasks find-field 'Severity'`. This was tested on https://redhat.atlassian.net/browse/RHEL-156977

### DIFF
--- a/jira-tasks
+++ b/jira-tasks
@@ -1,9 +1,12 @@
 #!/usr/bin/env python3
 
-# Generate a Jira token on https://issues.redhat.com/secure/ViewProfile.jspa → Personal Access Tokens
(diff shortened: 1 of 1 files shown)


jelly:
> @jelly the jira issue in the commit was made with this PR

Nice!

PR #8845, opened 2026-03-25
machine: handle kickstart_file_name in VirtMachine

KKoukiou:
kickstart_file_name is only for VirtInstallMachine (Class used by anaconda WebUI tests) for spawning VMs with specified KS file.


KKoukiou:
I realized I can pop that from kwargs before passing that to VirtMachine

PR #8907, opened 2026-04-14
README: include debian-testing dropped packages tip

jelly:
This happened a few times with PCP and target-cli.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,12 @@ control, when this occurs:
 2. Update `OSTREE_BUILD_IMAGE` to point to the Fedora release CoreOS
    uses.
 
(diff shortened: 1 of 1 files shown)


allisonkarlitskaya (approved):
I think this is probably fine but this README is getting a bit out of control.  We should probably have some "pilot knowledge" wikipage somewhere or something...

jelly:
> I think this is probably fine but this README is getting a bit out of control. We should probably have some "pilot knowledge" wikipage somewhere or something...

I can support this, today I made an overview of the [Test images](https://github.com/cockpit-project/cockpit/wiki/TestImagePolicy)

PR #8906, opened 2026-04-13
Image refresh for rhel-10-2

cockpituous:
Image refresh for rhel-10-2
 * [x] image-refresh rhel-10-2

PR #8874, opened 2026-04-02
Image refresh for debian-testing

cockpituous:
Image refresh for debian-testing
 * [x] image-refresh debian-testing

PR #8828, opened 2026-03-18
Image refresh for centos-9-stream

cockpituous:
Image refresh for centos-9-stream
 * [x] image-refresh centos-9-stream

PR #8909, opened 2026-04-14
Change image-trigger to use `/image-refresh` comment

allisonkarlitskaya:
...plus a bit of cleanup and a little quality of life enhancement.

This is designed to make the (*edit*)~~last~~second last commit as small as reasonably possible so we can revert it if things don't work out.  I triggered https://github.com/cockpit-project/bots/issues/8908 as a test but I don't want to do a full workflow dispatch from the branch until we're pretty sure we want to land this because otherwise the same issues will get recreated when the old workflow runs tonight and fails to find any checklist items.

### DIFF
--- a/image-trigger
+++ b/image-trigger
@@ -86,22 +86,23 @@ def main() -> int:
     return 0
 
 
-# Check if the given files that match @pathspec are stale
-# and haven't been updated in @days.
-def stale(days: int, pathspec: str, ref: str = "HEAD", verbose: bool = False) -> bool:
-    cmd = ["git", "log", "--max-count=1", "--pretty=format:%ct", ref, "--", pathspec]
+def last_update(pathspec: str, ref: str = "HEAD", verbose: bool = False) -> tuple[str, int] | None:
+    """Return (commit_sha, age_in_days) of the last commit touching pathspec, or None."""
+    cmd = ["git", "log", "--max-count=1", "--pretty=format:%ct %H", ref, "--", pathspec]
     if verbose:
         sys.stderr.write("+ " + shlex.join(cmd) + "\n")
     output = subprocess.check_output(cmd, cwd=BASE_DIR, text=True)
     if verbose:
         sys.stderr.write("> " + output + "\n")
 
+    parts = output.split(maxsplit=1)
     try:
-        timestamp = int(output)
-    except ValueError:
-        timestamp = 0
+        timestamp = int(parts[0])
+    except (ValueError, IndexError):
+        return None
+    sha = parts[1].strip() if len(parts) > 1 else ""
 
-    return timestamp < time.time() - days * 86400
+    return sha, int((time.time() - timestamp) / 86400)
 
 
 def scan(api: github.GitHub, force: str | None, verbose: bool, dry: bool = False) -> None:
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
One note: setting the `bot` label was left intentionally.  Removing it would make the `GitHub.issues()` scan fail to find the created issues/prs.  We can fix that at some point but since we're going for "minimal easy-to-revert patch" I didn't want to get into it.  There's also very little benefit, so let's leave it for now.

allisonkarlitskaya:
Okay, so after seeing the ubuntu refresh go green, I couldn't resist:

https://github.com/cockpit-project/bots/actions/runs/24419665216/job/71338256804

> #8828: image-refresh centos-9-stream (already open)
> #8874: image-refresh debian-testing (already open)
> #8911: image-refresh debian-trixie
> #8912: image-refresh rhel-8-10
> #8913: image-refresh rhel-9-4
> #8914: image-refresh rhel-9-6
> #8915: image-refresh rhel-10-0
> #8906: image-refresh rhel-10-2 (already open)

allisonkarlitskaya:
I'm glad I tested it because there's a problem: we create the issue with the github token, but there's a rule that github actions can't run in response to events triggered by the token in other github actions workflows.

So, the comment works, but only if it comes from a user other than the github-actions user...

allisonkarlitskaya:
A workflow run that correctly notices all of the open issues and does nothing:

https://github.com/cockpit-project/bots/actions/runs/24420428761/job/71340824900

> #8828: Image refresh for centos-9-stream (already open)
> #8874: Image refresh for debian-testing (already open)
> #8911: Image refresh for debian-trixie (already open)
> #8912: Image refresh for rhel-8-10 (already open)
> #8913: Image refresh for rhel-9-4 (already open)
> #8914: Image refresh for rhel-9-6 (already open)
> #8915: Image refresh for rhel-10-0 (already open)
> #8906: Image refresh for rhel-10-2 (already open)

Now I'll close one and try again.

allisonkarlitskaya:
That worked as expected:

https://github.com/cockpit-project/bots/actions/runs/24420519073/job/71341130989

> #8828: Image refresh for centos-9-stream (already open)
> #8874: Image refresh for debian-testing (already open)
> #8911: Image refresh for debian-trixie (already open)
> #8916: image-refresh rhel-8-10
> #8913: Image refresh for rhel-9-4 (already open)
> #8914: Image refresh for rhel-9-6 (already open)
> #8915: Image refresh for rhel-10-0 (already open)
> #8906: Image refresh for rhel-10-2 (already open)

allisonkarlitskaya:
...and the comment worked in #8916!

allisonkarlitskaya:
The trigger workflow now has access to more secrets than it strictly requires since it's using the environment of the image-refresh task (which can write to the image mirrors, etc).  We could add a separate environment that contains just the cockpituous token, but another thought occurred to me: at this point we could just have the trigger script actually *use* those extra secrets and directly queue the image refreshes in testing farm, avoiding the need for the comment, the roundtrip through GitHub, and the separate workflow run.

I'm not sure how I feel about that.  I sort of like the idea that the bot uses the same API as the humans and gets the same results...

allisonkarlitskaya:
I consider this proven to be working:
 - #8908 triggered locally using my personal key and the "force" mechanism to limit to a single image
 - #8912 part of the batch triggered from the action using the actions token, failing to trigger the comment workflow, so I closed it to try again
 - #8911 #8913 #8914 were all triggered in the first batch with #8912 and failed to run the comment action.  I could have closed them all but I left them open and just posted the comment myself to reduce churn.
 - #8916 was triggered from the action after I added the patch to use the `COCKPITUOUS_TOKEN` and worked completely without intervention

martinpitt (approved):
As discussed in chat. LGTM, especially the `environment:`.

image-trigger
```diff
@@ -114,28 +115,35 @@ def scan(api: github.GitHub, force: str | None, verbose: bool, dry: bool = False
...
-            days = options.get("refresh-days", DAYS)
-            perform = stale(days, os.path.join("images", image), "origin/main", verbose)
+        update = last_update(os.path.join("images", image), "origin/main", verbose)
 
-        if perform:
-            item = f"image-refresh {image}"
-            if item in issues:
-                sys.stderr.write(f'#{issues[item]["number"]}: {item} (already open)\n')
+        if force:
+            if image != force:
```
jelly, line 121: force being a string really tripped me up.

jelly (approved):

image-trigger
allisonkarlitskaya: Ya.  We could clean this script up a bit later.  It's pretty ancient...

PR #8918, opened 2026-04-15
testmap: stop testing RHEL images in starter-kit

jelly:
Historically we added this to verify that the image works and starter-kit is the easiest test for it. But lately `npm install` has been hanging an awful lot and been unreliable in our CI. As RHEL images are also tested in cockpit-files which is a fast, and simple project we already have a good alternative to starter-kit.

---

See also https://github.com/cockpit-project/starter-kit/issues/1394


jelly:
mistriggered....

allisonkarlitskaya (approved):
I'm sold.  And I agree — this is a pain.  But we're still gonna get exposure from the other images....

Thanks!

PR #8927, opened 2026-04-16
testmap: move subscriptions to latest RHEL image

jelly:


tomasmatus (approved):
niiice

lib/testmap.py
```diff
@@ -223,16 +225,14 @@ def contexts(image: str, *scenarios: Iterable[str], repo: str | None = None) ->
...
-            'rhel-10-2',
-            'rhel-10-2/devel',
+            'rhel-9-9/subscription-manager-1.29',
+            'rhel-10-3',
+            'rhel-10-3/devel',
             'fedora-42',
             'fedora-43',
         ],
         '_manual': [
             'fedora-44',
```
tomasmatus, line 235: we should also move this to main, but fine for a followup

jelly: Yes!

PR #8946, opened 2026-04-20
naughty: fedora-rawhide is fedora-45

mvollmer:
And start fedora-45 as a copy of fedora-43 since that is what used to work for us.

The 8929 naughty was erroneously added to fedora-43, via the old symlink. It only happens in the real fedora-rawhide, which is fedora-45 now.


mvollmer:
The 8929 pattern should have matched here: https://artifacts.dev.testing-farm.io/33283919-2a3e-4c60-b41b-91b77bc0137f/work-storage-extrai1hsq5f9/plans/all/storage-extra/execute/data/guest/default-0/test/browser/storage-extra-1/output.txt, but it didn't since $TEST_OS is "fedora-45" in fedora-rawhide runs on TF.

But note how the pattern was matched:

```
not ok 9 test/verify/check-storage-anaconda TestStorageAnaconda.testStratis [ND@-1] # NOTE Known issue #8929 in fedora-43
```

and that clued me in to the fact the fedora-rawhide is symlink to the wrong place.

jelly (approved):
Ack!

PR #8953, opened 2026-04-22
testmap: Test cockpit on fedora-44/storage and fedora-44/networking

mvollmer:
These are green. The rest come later.


jelly (approved):
Thanks!

PR #8955, opened 2026-04-22
services: drop curl to check if candlepin started

tomasmatus:
for testing atm


tomasmatus:
this is useless because it would need services VM rebuild, don't need this anyways I guess.

PR #8968, opened 2026-04-27
Image refresh for arch

cockpituous:
Last updated in ad297ee1bc37aa8471c895a277cfe7f482c4fd7c, 14 days ago.
 * [x] image-refresh arch

PR #8970, opened 2026-04-27
Image refresh for fedora-44

cockpituous:
Last updated in c56bb26a4767aa94be0836535f7e7c7e6ace0739, 7 days ago.

 - [x] Requires https://github.com/cockpit-project/cockpit-podman/pull/2520 or https://github.com/cockpit-project/cockpit-podman/pull/2522


cockpituous:
**Task scheduled:** issue-8970 image-refresh/fedora-44

Testing Farm link: https://artifacts.osci.redhat.com/testing-farm/18d4c8dd-d3be-4271-9155-5f17838d940e

<details>
<summary>Job JSON</summary>

```json
{
  "repo": "cockpit-project/bots",
  "sha": "bc1d5d79eaed18ad7096cd54e75344e018f0b048",
  "pull": null,
  "slug": "image-refresh-fedora-44-bc1d5d79-20260427-232743",
  "context": "image-refresh/fedora-44",
  "command": [
    "./image-refresh",
    "--verbose",
    "--issue=8970",
    "fedora-44"
  ],
  "secrets": [
    "github-token",
    "image-upload"
  ]
}
```
</details>

cockpituous:
image-refresh fedora-44 done: https://github.com/cockpit-project/bots/commits/image-refresh-fedora-44-20260427-235544

cockpituous:
Success. Log: https://cockpit-logs.us-east-1.linodeobjects.com/image-refresh-fedora-44-bc1d5d79-20260427-232743/log.html

jelly:
Seen [this](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-8970-de601db4-20260428-071050-fedora-44-cockpit-project-cockpit-podman/log.html) too in update-testing, and sometimes on Arch / OpenSUSE Tumbleweed. Something funky is going on.

martinpitt:
I can reproduce [this failure](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-8970-de601db4-20260428-071050-fedora-44-cockpit-project-cockpit-podman/log.html) locally. When that happens, it kills the entire VM, there is nothing on the VT, and neither ssh nor VT logins work any more.

But the test runs all the way to the end, it's the cleanup that kills it.  I added `testlib.sit()` as the last line, and when I just let it sleep a few seconds doing nothing before the cleanup, the test works.

Running through the `_terminate_sessions()` cleanup steps manually: `pkill -9 -u admin` is the thing that kills the VM. After that, running commands like `ps aux` or `loginctl` or `top` hang forever. There is no `dmesg -w` output for that.

Next steps: bisect the test to see how much I can reduce it, then gradually replace with shell commands.

I bisected the test to [here](https://github.com/cockpit-project/cockpit-podman/blob/772987d75008a93bc54497f630cbd7a213755f3f/test/check-application#L1615), i.e. starting the `swamped-crate` container.

martinpitt:
I reduced the test quite a bit, but as soon as I start replacing UI operations with shell, e.g.

```
         # Restart the container; there is no steady state change in the visible UI, so look for
         # a changed data-started-at attribute
-        old_start = self.getStartTime("swamped-crate", auth=auth)
-        b.wait_in_text(f'#containers-containers tr[data-started-at="{old_start}"]', "swamped-crate")
-        self.performContainerAction("swamped-crate", "Force restart", system=auth)
-        new_start = self.waitRestart("swamped-crate", old_start, auth=auth)
-        b.wait_in_text(f'#containers-containers tr[data-started-at="{new_start}"]', "swamped-crate")
-        self.waitContainer(container_sha, auth=auth, name='swamped-crate', image=IMG_BUSYBOX, state='Running')
+        self.execute("podman restart -t 0 swamped-crate", system=auth)
+        time.sleep(1)
```

the bug stops appearing. Same with the "force stop" further down. Same with

> b.click(".pf-m-expanded button:contains('Console')")

i.e. actually talking to the REST API and streaming console etc. is significant.

I was able to squeeze the test down to this, but from now on any attempt to de-cockpit-ify it further makes it pass:

```py
    def _testLifecycleOperations(self, *, auth: bool) -> None:
        b = self.browser

        if not auth:
            self.allow_browser_errors("Failed to start system podman.socket.*")

        self.login()
        self.filter_containers('all')

        # run a container
        self.execute(f"""
                podman run -d --name swamped-crate --stop-timeout 0 {IMG_BUSYBOX} sh -c 'echo 123; sleep infinity';
                podman stop swamped-crate""", system=auth)
        b.wait(lambda: self.execute("podman ps --all | grep -e swamped-crate -e Exited", system=auth))

        b.wait_visible("#containers-containers")
        time.sleep(1)
        # Start the container
        # replacing his with `self.execute("podman start swamped-crate", system=auth)` makes it pass
        self.performContainerAction("swamped-crate", "Start", system=auth)
        time.sleep(1)

        # Restart the container; there is no steady state change in the visible UI, so look for
        # a changed data-started-at attribute
        # replacing this with `self.execute("podman restart -t 0 swamped-crate", system=auth)` makes it pass
        self.performContainerAction("swamped-crate", "Force restart", system=auth)
        time.sleep(1)

        # Stop the container
        self.performContainerAction("swamped-crate", "Force stop", system=auth)
        time.sleep(1)

        # Check that console reconnects when container starts
        self.toggleExpandedContainer("swamped-crate", system=auth)
        b.click(".pf-m-expanded button:contains('Console')")
        # replacing this with `self.execute("podman start swamped-crate", system=auth)` makes it pass
        self.performContainerAction("swamped-crate", "Start", system=auth)
        time.sleep(2)
```

I was then able to replace all performContainerAction() calls with `curl` API calls, but this one puts up a fight:

> b.click(".pf-m-expanded button:contains('Console')")

That's the last browser action, but any attempt to replace this with e.g.

```diff
--- test/check-application
+++ test/check-application
@@ -1528,7 +1528,10 @@ Yaml={name}.yaml
 
         # Check that console reconnects when container starts
         self.toggleExpandedContainer("swamped-crate", system=auth)
-        b.click(".pf-m-expanded button:contains('Console')")
+        # Console connection via API: Variation 1 - exec → resize → start
+        exec_body = '{"AttachStderr":true,"AttachStdout":true,"AttachStdin":true,"Tty":true,"Cmd":["/bin/sh"]}'
+        exec_id = self.execute(f"curl -X POST 'http://localhost/v3.4.0/libpod/containers/{container_id}/exec' --unix-socket {socket} -H 'Content-Type: application/json' -d '{exec_body}' | jq -r .Id", system=auth).strip()
+        self.execute(f"curl -X POST 'http://localhost/v3.4.0/libpod/exec/{exec_id}/resize?h=24&w=206' --unix-socket {socket}", system=auth)
         # replacing this with `self.execute("podman start swamped-crate", system=auth)` makes it pass
         api_call("start")
         time.sleep(2)

```

makes the bug go away. I had claude do lots of variations (order changes, sleeps,  duplications).

martinpitt:
I tried to update the 6.9.13 kernel on this new image to the current 6.9.14, and I don't get the VM freeze any more. So with any luck, this was a regression between 6.9.10 → 6.9.13 which .14 fixed. Let's try to refresh.

martinpitt:
- 25 test runs in a row on current main image (6.9.10) work locally.
- [podman shows the same failure](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-8968-17137176-20260427-124255-arch-cockpit-project-cockpit-podman/log.html) in #8968 , and that's 6.9.14 already. Dang.. So it must be something else, and that it didn't happen yesterday was pure coincidence. This morning I also got it with 6.9.14.
- [weather report](https://ci-weather-cockpit.apps.ocp.cloud.ci.centos.org/tests.html?repo=cockpit-project%2Fcockpit-podman&days=7&test=%2Ftest%2Fcheck-application+TestApplication.testLifecycleOperationsUser) only shows this on Ubuntu 26.04 (kernel 7.0.0) and [tumbleweed](https://logs-cockpit.apps.ocp.cloud.ci.centos.org/pull-2506-870ab3d6-20260423-123014-opensuse-tumbleweed/log.html) (6.19.11). The latter is a very interesting data point.

So I've proven locally that this regression is triggered by updating kernel 6.9.10 → .13, and the above weather report data shows pretty conclusively that it's triggered by 6.9.10 → 11. This is enough for a naughty, but not yet enough for a bug report.

martinpitt:
# Kernel hang/deadlock in cgroup cleanup (6.9.11 regression)

## Summary

Kernel regression between 6.9.10 (working) and 6.9.11 (broken) causing complete system hang when cleaning up cgroups after podman container lifecycle operations. The hang manifests as:
- Unable to run commands that access `/proc` (ps, top, lsns, etc.)
- Unable to create new SSH sessions or VT logins
- System appears locked but existing VT sessions remain partially functional

## Environment

- **Last known working kernel**: 6.9.10
- **Bug observed on distributions**: Fedora 44 (6.9.13/6.9.14), Ubuntu 26.04 (7.0.0), OpenSUSE Tumbleweed (6.9.11)
- **Reproducer**: Podman REST API operations on rootless containers followed by user cleanup

## Root Cause (from sysrq-trigger analysis)

The system deadlocks in cgroup cleanup during user session termination. **systemd (pid 1)** gets stuck in `cgroup_drain_dying` trying to remove an empty cgroup:

```
[  207.958946] task:systemd         state:D stack:0     pid:1     tgid:1     ppid:0
[  207.959734] Call Trace:
[  207.960117]  <TASK>
[  207.960333]  __schedule+0x2b2/0x5d0
[  207.960603]  schedule+0x27/0x80
[  207.960945]  cgroup_drain_dying+0xef/0x1a0
[  207.961287]  ? __pfx_autoremove_wake_function+0x10/0x10
[  207.961639]  cgroup_rmdir+0x37/0x100
[  207.961945]  kernfs_iop_rmdir+0x6a/0xd0
[  207.962239]  vfs_rmdir+0x154/0x270
[  207.962486]  do_rmdir+0x201/0x280
[  207.962723]  __x64_sys_unlinkat+0x8c/0xd0
```

**Key observations:**
- **Empty cgroups**: `/sys/fs/cgroup/user.slice/user-1000.slice/cgroup.procs` and `cgroup.threads` were empty when checked, indicating all processes had been killed but the cgroup itself could not be removed
- **Secondary blocked task**: `kworker/0:1` (DRM framebuffer helper) also in D state, likely a side effect of the console being blocked
- **Zombie processes present**: Multiple zombie processes unable to be reaped, including user@1000.service systemd (pid 13549), podman processes, conmon processes, and systemd-hostnamed/systemd-timedated
- **RCU subsystem healthy**: `rcu_exp_gp_kthr` in S (sleeping) state, not blocked

The zombie processes suggest that systemd is unable to complete its cleanup duties while stuck in `cgroup_drain_dying`, preventing it from reaping terminated child processes.

### Alternative pattern: RCU namespace deadlock

In one early test run, a different hang pattern was observed where `rcu_exp_gp_kthr` (the RCU expedited grace period kernel thread) was in D state, and a process (`pool-spawner`) was stuck in `synchronize_rcu_expedited` during namespace cleanup (`nsproxy_free → namespace_unlock`). Multiple zombie processes were present, unable to complete exit. Unfortunately, the dmesg log from this hang was lost before it could be saved.

This variant has not reproduced in subsequent test runs despite trying various kernel parameters (`debug`, `rcupdate.rcu_expedited=1`, etc.). The cgroup cleanup deadlock appears to be the more common manifestation of the bug, though both variants occur after the same reproducer sequence, suggesting they may be related or share a common root cause.

## Reproducer

The hang is triggered by a specific sequence of podman REST API operations on rootless containers, followed by user cleanup.

**Test case**: https://github.com/martinpitt/cockpit-podman/blob/kernel-hang/test/check-application#L1486

The test performs:
1. Create and stop a rootless container as the `admin` user
2. Call podman REST API lifecycle operations: start → restart (t=0) → stop (t=0)
3. Create an exec session (console/TTY connection) via REST API
4. Start the container again via REST API
5. Post-test cleanup: `loginctl terminate-user admin`, `loginctl kill-user admin`, `pkill -9 -u admin`

The `pkill` command triggers the bug.

**Critical observation**: Using `podman` CLI commands (e.g., `podman start swamped-crate`) instead of the REST API does NOT trigger the hang. The issue is specific to the REST API code path, which goes through the podman service daemon.

Sadly, any attempt to move this out of the cockpit test suite into a standalone shell script reproducer has failed. The bug is way too timing sensitive for that.

## Flakiness

This is a timing-sensitive race condition. It can:
- Hang on the first iteration
- Pass 50+ iterations then suddenly hang
- Be stable for hours then start failing again

This suggests the deadlock depends on subtle timing in the cgroup cleanup path, likely related to when processes exit and when systemd attempts to clean up their cgroups.

## Collecting Additional Debug Info

### Kernel command line parameters

To enable RCU debugging and verbose logging:

```bash
grubby --update-kernel=ALL --args="debug ignore_loglevel rcupdate.rcu_cpu_stall_timeout=60 rcupdate.rcu_cpu_stall_suppress=0"
```

**Test results:**
- `dmesg-cgrouphang.txt`: Captured with debug parameters - shows cgroup cleanup deadlock
- Testing without debug parameters showed the identical cgroup cleanup deadlock pattern
- No RCU stall warnings were generated because the hang occurred within ~15-60 seconds, below the 60-second stall detection threshold

**Conclusion**: Debug kernel parameters did not influence which hang variant appeared - the cgroup cleanup deadlock appeared consistently.

To add expedited RCU (forces more expedited RCU operations, might trigger the RCU namespace variant more reliably):
```bash
grubby --update-kernel=ALL --args="rcupdate.rcu_expedited=1"
```

### Commands to run when system hangs

From a VT console (WARNING: most commands that touch `/proc` will hang):

```bash
dmesg > dmesg.txt
# Most important - task dump showing all stacks
echo t > /proc/sysrq-trigger
dmesg > hung-tasks.txt

# Memory and blocked tasks
echo m > /proc/sysrq-trigger
echo w > /proc/sysrq-trigger
dmesg > hung-full.txt

# Check for any RCU stall warnings or related messages
dmesg | grep -i "rcu\|stall\|hung\|blocked"

# Try to access cgroup state (may work, but systemctl usually hangs)
cat /sys/fs/cgroup/user.slice/user-1000.slice/cgroup.procs
systemctl status user@1000.service --no-pager
```

**Critical limitation**: Cannot read `/proc/<pid>/*` files when hung - all `/proc` access blocks, including `ps`, `top`, `lsns`, and reading individual process status/stack files.

## Full Evidence

**dmesg-cgrouphang.txt** (492K): Complete dmesg log from boot through the hang, captured with debug kernel parameters (`debug ignore_loglevel rcupdate.rcu_cpu_stall_timeout=60 rcupdate.rcu_cpu_stall_suppress=0`). Shows the cgroup cleanup deadlock with systemd stuck in `cgroup_drain_dying`. Includes all three sysrq triggers (t=show tasks, m=show memory, w=show blocked tasks). Contains verbose systemd debug output which provides additional context about the system state leading up to the hang.

This pattern reproduced consistently across multiple test runs with and without debug kernel parameters, and with `rcupdate.rcu_expedited=1`.

[dmesg-cgrouphang.txt](https://github.com/user-attachments/files/27195205/dmesg-cgrouphang.txt)

The above was incomplete. I reproduced the hang again, increased the buffer size with `grubby --update-kernel=ALL --args="log_buf_len=4M"` and captured a dmesg again: [dmesg.txt](https://github.com/user-attachments/files/27231725/dmesg-task-dump.txt)

mvollmer:
@martinpitt, I did some investigation and came up with a workaround: https://github.com/cockpit-project/cockpit-podman/pull/2520

mvollmer:
Yes, it's the `pkill -9 u 1000`.  When I let systemd do the session teardown all by itself, I could not observe the hang.

But of course, the system should not end up with a stuck PID 1, no matter what user processes are killed when.

Thanks a lot for the investigation!

martinpitt:
I reported the regression here: https://lore.kernel.org/regressions/afHNg2VX2jy9bW7y@piware.de/T/#u . Now I think I did due diligence, can reproduce this, so I'm fine with adding a workaround. Thanks @mvollmer !

martinpitt:
podman workaround landed, rebased.

martinpitt (approved):

PR #8984, opened 2026-04-30
lib: Allow overriding DEFAULT_MACHINE_MEMORY_MB via environment

KKoukiou:
Read DEFAULT_MACHINE_MEMORY_MB (MiB) from the DEFAULT_MACHINE_MEMORY_MB environment variable when set.
Default remains 1152 MiB when unset.


martinpitt:
we already have a run-tests option to change nondestructive RAM size, and destructive tests should use `provision: ` with bigger RAM for tests which need it? What's the use case for this third option?

KKoukiou:
Closing in favor of https://github.com/rhinstaller/anaconda-webui/pull/1285

PR #9044, opened 2026-05-18
machine_core: Don't create overlay for add_disk(path=)

martinpitt:
None of our tests use this with an explicit `path=`, they all use the "create temporary disk" behaviour. This is only being used with `vm-run --storage` to attach an extra disk. But for that it doesn't seem very useful to put things onto a temporary overlay, when the whole reason for this option is to provide some extra data or workspace to the VM.

Add a new `temporary` attribute to `self.disks`, so that `rem_disk()` knows whether or not it should remove the image file.

---

This is technically an API break, but I didn't find any of our projects which uses this. An alternative would be to add an extra "overlay" boolean, and plumb that through `vm-run` as extra argument. But you can always create an overlay with `qemu-img` yourself when using that, and it's actually desired..

### DIFF
--- a/machine/machine_core/machine_virtual.py
+++ b/machine/machine_core/machine_virtual.py
@@ -627,27 +627,23 @@ class VirtMachine(Machine):
         assert self._domain is not None
 
         index = len(self._disks)
+        temporary = False
 
-        if path:
-            fd, image = tempfile.mkstemp(suffix='.qcow2', prefix=os.path.basename(path), dir=self.run_dir)
-            os.close(fd)
-            subprocess.check_call(["qemu-img", "create", "-q", "-f", "qcow2",
-                                   "-o", f"backing_file={os.path.realpath(path)},backing_fmt=qcow2", image])
-
-        else:
+        if not path:
             assert self._domain is not None
             assert size is not None
             name = f"disk-{self._domain.name()}"
-            fd, image = tempfile.mkstemp(suffix='qcow2', prefix=name, dir=self.run_dir)
+            fd, path = tempfile.mkstemp(suffix='qcow2', prefix=name, dir=self.run_dir)
             os.close(fd)
-            subprocess.check_call(["qemu-img", "create", "-q", "-f", "raw", image, size])
+            subprocess.check_call(["qemu-img", "create", "-q", "-f", "raw", path, size])
+            temporary = True
 
         if not serial:
             serial = f"DISK{index}"
         dev = 'sd' + string.ascii_lowercase[index]
         extra = "<boot order='1'/>" if boot_disk else ""
         disk_desc = TEST_DISK_XML % {
-            'file': image,
+            'file': os.path.realpath(path),
             'serial': serial,
             'unit': index,
             'dev': dev,
@@ -659,13 +655,14 @@ class VirtMachine(Machine):
             raise Failure("Unable to add disk to vm")
 
         disk = {
-            "path": image,
+            "path": path,
             "serial": serial,
-            "filename": image,
+            "filename": path,
             "dev": dev,
             "index": index,
             "type": image_type,
             "extra": extra,
+            "temporary": temporary,
         }
 
         self._disks.append(disk)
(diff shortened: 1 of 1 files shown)


machine/machine_core/machine_virtual.py
```diff
@@ -628,26 +628,20 @@ def add_disk(
...
-            subprocess.check_call(["qemu-img", "create", "-q", "-f", "qcow2",
-                                   "-o", f"backing_file={os.path.realpath(path)},backing_fmt=qcow2", image])
-
-        else:
+        if not path:
             assert self._domain is not None
             assert size is not None
             name = f"disk-{self._domain.name()}"
-            fd, image = tempfile.mkstemp(suffix='qcow2', prefix=name, dir=self.run_dir)
+            fd, path = tempfile.mkstemp(suffix='qcow2', prefix=name, dir=self.run_dir)
```
martinpitt, line 636: Wait, `vm-run` (or something else) removes that overlay on shutdown, this is buggy.

martinpitt: Fixed.

jelly (commented):
I used this option quite a bit when playing around with btrfs development, with this change it gives me:

```
[jelle@carbon][~/projects/cockpit/main]%./bots/vm-run -s foo.img fedora-43
^CTraceback (most recent call last):
  File "/home/jelle/projects/cockpit/main/./bots/vm-run", line 78, in <module>
    machine.wait_boot()
    ~~~~~~~~~~~~~~~~~^^
  File "/home/jelle/projects/cockpit-bots/machine/machine_core/ssh_connection.py", line 145, in wait_boot
    if self.wait_execute(timeout_sec=15):
       ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/jelle/projects/cockpit-bots/machine/machine_core/ssh_connection.py", line 100, in wait_execute
    data = sock.recv(10)
KeyboardInterrupt

[jelle@carbon][~/projects/cockpit/main]%./bots/vm-run -S foo.img fedora-43
libvirt: Storage Driver error : Cannot access storage file 'foo.img' (as uid:1000, gid:1000): No such file or directory
  File "/home/jelle/projects/cockpit/main/./bots/vm-run", line 73, in <module>
    machine.add_disk(path=disk, image_type='qcow2')
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jelle/projects/cockpit-bots/machine/machine_core/machine_virtual.py", line 654, in add_disk
    if self._domain.attachDeviceFlags(disk_desc, libvirt.VIR_DOMAIN_AFFECT_LIVE) != 0:
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/libvirt.py", line 721, in attachDeviceFlags
    raise libvirtError('virDomainAttachDeviceFlags() failed')
libvirt.libvirtError: Cannot access storage file 'foo.img' (as uid:1000, gid:1000): No such file or directory
```

machine/machine_core/machine_virtual.py
```diff
@@ -627,27 +627,23 @@ def add_disk(
...
+            subprocess.check_call(["qemu-img", "create", "-q", "-f", "raw", path, size])
+            temporary = True
 
         if not serial:
             serial = f"DISK{index}"
         dev = 'sd' + string.ascii_lowercase[index]
         extra = "<boot order='1'/>" if boot_disk else ""
         disk_desc = TEST_DISK_XML % {
-            'file': image,
+            'file': path,
```
jelly: This might need `os.path.realpath(path)`

jelly: Confirmed, or normalise it in the `if path`

martinpitt: Thanks for spotting! Can reproduce, fixed.

jelly (approved):
Thanks!

PR #9050, opened 2026-05-19
testmap: swap anaconda efi scenario with bios

KKoukiou:
Now all anaconda tests run by default under efi firmware, and we need an extra bios scenario which will run only for some of the tests.


jelly (approved):
Its your party :)

PR #9058, opened 2026-05-20
add `cockpit-ci-images` S3 store on AWS

allisonkarlitskaya:
- [x] https://github.com/cockpit-project/cockpituous/pull/703 for the necessary policy changes

### DIFF
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -6,7 +6,6 @@
 # https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html
 
 import asyncio
-import base64
 import hashlib
 import hmac
 import logging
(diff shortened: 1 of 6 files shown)


lib/s3.py
```diff
@@ -186,22 +185,54 @@ def parse_list(result: ET.Element, *keys: str) -> Iterable[Iterable[str]]:
...
+    }
+    query_string = urllib.parse.urlencode(sorted(query_params.items()), quote_via=urllib.parse.quote)
+
+    canonical_request = f'{method}\n{url.path}\n{query_string}\n{headers_str}\n{headers_list}\nUNSIGNED-PAYLOAD'
+    logger.debug('canonical request: %r', canonical_request)
 
-    h = hmac.HMAC(secret.encode('ascii'), digestmod='sha1')
-    h.update(f'{method}\n\n\n{expires}\n{headers}/{bucket}{url.path}'.encode('ascii'))
-    signature = urllib.parse.quote_plus(base64.b64encode(h.digest()))
+    algorithm = 'AWS4-HMAC-SHA256'
```
martinpitt, line 220: Does that work with Linode as well? What kind of testing did you do with this PR? As you have access to the prod tokens, you could run image-upload/download locally, right?

prometheus-stats
```diff
@@ -232,7 +232,7 @@ merged_failures {merged_failures}
         headers = {"Content-Type": "text/plain"}
         # AWS S3 buckets with BucketOwnerEnforced don't allow ACLs; bucket policy provides public access
         if not opts.s3.hostname.endswith('.amazonaws.com'):
-            headers[s3.ACL] = s3.PUBLIC
+            headers['x-amz-acl'] = 'public-read'
```
martinpitt, line 235: Does that even still work with the bucket policy? With the logs migration, AWS rejected S3 Puts with an ACL for buckets without per-file ACLs. Please test the prometheus-stats upload.

martinpitt (dismissed):
Thanks! LGTM, I just wonder about how much testing went into this.

lib/s3.py
allisonkarlitskaya: Yes.  I tested both of those and I can upload and download without problems, all around.  I did notice that we were failing to include the AWS secret in the testing farm image-refresh workflow though, and pushed an additional commit for that.

prometheus-stats
allisonkarlitskaya: Honestly I have no idea how this works, but this change is a non-change.  I'm just replacing constants that used to be defined in a separate file with the values of the constants.  If this was broken before, it won't change it.

How might I test it?  Or figure out if it's been working on the live infra?

martinpitt: Ah sorry, it's right there -- `if not .. endswith(amazonaws)` :see_no_evil:

PR #9061, opened 2026-05-20
s3: add support for S3 session tokens

allisonkarlitskaya:
Modify our various signing functions to accept `key=` kwargs including a new third tuple field: a session token.


martinpitt (approved):
Works!

PR #9063, opened 2026-05-21
testmap: Set up gating for RHEL backports

jelly:


martinpitt (approved):

lib/testmap.py
```diff
@@ -68,6 +68,12 @@ def contexts(image: str, *scenarios: Iterable[str], repo: str | None = None) ->
             # all skipped
             *contexts('rhel-8-10-distropkg', COCKPIT_SCENARIOS - {'networking'}),
         ],
+        'rhel-10-2': [
```
martinpitt: Typo, the branch name is "10.2", not "10-2"

martinpitt (changes requested):
I triggered the tests, and they immediately fail as it's an unknown branch.

martinpitt:
Discussion summary from matrix: let's keep *only* the rhel-9.8 branch and delete rhel-10.2 (it's currently identical). And change this PR to test both rhel-9-8 and rhel-10-2 on rhel-9.8 branch.

lib/testmap.py
```diff
@@ -68,6 +68,10 @@ def contexts(image: str, *scenarios: Iterable[str], repo: str | None = None) ->
             # all skipped
             *contexts('rhel-8-10-distropkg', COCKPIT_SCENARIOS - {'networking'}),
         ],
+        'rhel-9-8': [
```
martinpitt: Still dot-not-dash 😉

martinpitt (approved):

martinpitt:
I triggered the two tests. They ought to be at least "mostly green" to land this (and if anything fails, they need a follow-up)

Issue #9065, opened 2026-05-21
Plots tests for memory chart fail due to prior packagekitd memory usage


Venefilyn:
testPlots started to fail recently in [our nightly scenario](https://github.com/cockpit-project/cockpit/issues/22824), this can be reproduced by first running `TestUpdates.testUnprivileged` and then `TestPages.testPlots`. 

On the first load of the page, packagekitd eats tons of memory causing a huge spike. The initial MemAvailable range will so small that it falls out of the range of the plateau.

PR #9005, opened 2026-05-11
Image refresh for fedora-44

cockpituous:
Last updated in 02cb3014ee581845269471d5f2ea2de1db91dfea, 7 days ago.
 * [x] image-refresh fedora-44

PR #9066, opened 2026-05-21
naughty: Add `testPlots`

Venefilyn:
Related-to: https://github.com/cockpit-project/bots/issues/9065

### DIFF
--- a/naughty/fedora-44/9065-testplots-mem-mismatch
+++ b/naughty/fedora-44/9065-testplots-mem-mismatch
@@ -1,0 +1,6 @@
+Traceback (most recent call last):
+  File "*verify/check-pages", line *, in testPlots
+    b.wait_js_func("ph_plot_data_plateau", "direct", mem_avail *, mem_avail *, *, "mem")
+*
+testlib.Error: timeout
+wait_js_cond(ph_plot_data_plateau("direct",*,"mem")): Error: condition did not become true


jelly:
It doesn't fail on Fedora 44, our [two week old image went green](https://github.com/cockpit-project/bots/pull/9005) was green.

Venefilyn:
@jelly most recently failed here 8h ago on `fedora-44`. Or is this using a different f44 image?

https://github.com/cockpit-project/cockpit/issues/23315

martinpitt:
No veto as I lost the right to do that :wink: But as a recommendation: human psychology tells us that this essentially means to disable the test forever. So if it's too hard to fix, rather drop/comment out the problematic bit, instead of going through a naughty. As a first step, disable packagekit in that test?

jelly:
> No veto as I lost the right to do that 😉 But as a recommendation: human psychology tells us that this essentially means to disable the test forever. So if it's too hard to fix, rather drop/comment out the problematic bit, instead of going through a naughty. As a first step, disable packagekit in that test?



> @jelly most recently failed here 8h ago on `fedora-44`. Or is this using a different f44 image?
> 
> [cockpit-project/cockpit#23315](https://github.com/cockpit-project/cockpit/issues/23315)

Yes, thats the nightly run but it doesn't fail in pull requests.

jelly:
> No veto as I lost the right to do that 😉 But as a recommendation: human psychology tells us that this essentially means to disable the test forever. So if it's too hard to fix, rather drop/comment out the problematic bit, instead of going through a naughty. As a first step, disable packagekit in that test?

We have https://github.com/cockpit-project/cockpit/pull/23076 just has a really curious failure

mvollmer:
What about running testPlots as a destructive test?

jelly:
Looked again at https://github.com/cockpit-project/cockpit/pull/23076 and this is going green now.

PR #9070, opened 2026-05-22
s3: various cleanups

allisonkarlitskaya:
We need to be a bit more explicit about key handling in order to do the STS stuff and the way this works has been bugging me for a while anyway.  Let's clean up the key stuff, and do a bunch of other drive-bys while we're in there...


martinpitt (approved):
Thanks!

PR #9071, opened 2026-05-22
image-download-rhel: new image-download replacement

allisonkarlitskaya:
This is capable of downloading RHEL images from AWS.

We identify URLs that should be STS-authenticated by using a list (currently 1) of fnmatches on URLs.  This list is only consulted if you don't have a normal static key in ~/.config/cockpit-dev/s3-keys/.

Once we've identified that we want to issue a transient key, the process goes like this:

  - you get a Kerberos ticket with kinit

  - we use gssapi to get a SPNEGO token for auth.redhat.com

  - we post a login attempt using the token and hopefully get back a SAML POST binding HTML page which we can extract the SAML assertion from

  - with the SAML assertion in hand we can use AWS STS to generate a transient token for a particular IAM Role.  If you're in the it-cloud-aws-727920394381-cockpit-ci-images-download rover group this will allow you to download RHEL images from AWS

  - we use the STS token to do the S3 signing operation

This should be more or less transparent to anyone who is in the correct group and logged in with kerberos.  There's no need for us to maintain a list of users and tokens anymore and human developers can delete their old s3-keys.

The new downloader is intended as an eventual replacement for image-download.  It's a rewrite from the ground up and has a couple of other niceties:

  - urllib instead of curl (which was getting complicated with all the extra headers)

  - checksum verification based on filename

  - improved locking vs. the old downloader: no more poll loop or subtle race conditions with deleted lockfiles

  - 🌈 rainbows!

Some features not implemented:

  - `--state` and modification times (which seems to be dead code)
  - `--force`

With S3 keys for EU Linode mirror (light mode):

image

Without any statically-configured S3 keys (dark mode):

image

 - [x] #9070

### DIFF
--- a/lib/ansi.py
+++ b/lib/ansi.py
@@ -1,0 +1,14 @@
+# Copyright (C) 2026 Red Hat, Inc.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import os
+import sys
+
+IS_TTY = os.isatty(sys.stderr.fileno())
+USE_COLOR = 'FORCE_COLOR' in os.environ or (IS_TTY and 'NO_COLOR' not in os.environ)
+
+RED = '\033[31m' if USE_COLOR else ''
+GREEN = '\033[32m' if USE_COLOR else ''
+BLUE = '\033[34m' if USE_COLOR else ''
+RESET = '\033[0m' if USE_COLOR else ''
+CLEAR_LINE = '\033[2K\r' if IS_TTY else ''
--- a/lib/gssapi_saml_sts.py
+++ b/lib/gssapi_saml_sts.py
@@ -1,0 +1,219 @@
+# Copyright (C) 2026 Red Hat, Inc.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""Kerberos → SAML → STS authentication for AWS.
+
+Authenticates via SPNEGO/Negotiate to an IdP, exchanges the SAML assertion
+for temporary AWS credentials via STS AssumeRoleWithSAML, and caches
+credentials to disk.
+"""
+
+import base64
+import dataclasses
+import fnmatch
+import html.parser
+import logging
+import os
+import urllib.error
+import urllib.parse
+import urllib.request
+import xml.etree.ElementTree as ET
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+
+from lib.directories import xdg_cache_home
+from lib.s3 import S3Key
+
+logger = logging.getLogger(__name__)
+
+
+# https://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html
+_STS_URL = 'https://sts.amazonaws.com/'
+_STS_VERSION = '2011-06-15'
+_STS_NS = f'https://sts.amazonaws.com/doc/{_STS_VERSION}/'
+
+
+class AuthError(Exception):
+    """Raised when Kerberos/SAML/STS authentication fails."""
+
+
+def _negotiate_header(idp_url: str) -> str:
+    """Create a Negotiate auth header using Kerberos."""
+    try:
+        import gssapi
+        import gssapi.raw.misc
+    except ImportError as exc:
+        raise AuthError('🐍 no python3-gssapi') from exc
+
+    hostname = urllib.parse.urlparse(idp_url).hostname
+    assert hostname is not None
+    # fixed in python-gssapi 1.9.0: https://github.com/pythongssapi/python-gssapi/pull/338
+    server_name = gssapi.Name(f'HTTP@{hostname}', gssapi.NameType.hostbased_service)  # type: ignore[attr-defined]
+    try:
+        # fixed in python-gssapi 1.9.0: https://github.com/pythongssapi/python-gssapi/pull/338
+        ctx = gssapi.SecurityContext(usage='initiate', name=server_name)  # type: ignore[attr-defined]
+        token = base64.b64encode(ctx.step()).decode()
+    except gssapi.raw.misc.GSSError as exc:
+        statuses = '  '.join(exc.get_all_statuses(exc.min_code, is_maj=False))
+        raise AuthError(f'🐕 {statuses}.  Try kinit.') from exc
+    logger.debug('obtained SPNEGO token for %r', hostname)
+    return f'Negotiate {token}'
+
+
+def _get_saml_assertion(idp_url: str) -> str:
+    """Authenticate via Kerberos and return the base64 SAML assertion."""
+    logger.debug('requesting SAML assertion from %r', idp_url)
+
+    request = urllib.request.Request(idp_url)
+    request.add_header('Authorization', _negotiate_header(idp_url))
+    response = urllib.request.urlopen(request)
+    body = response.read().decode()
+    logger.debug('IdP response status: %r', response.status)
+    logger.debug('IdP response: %d bytes', len(body))
+
+    result: list[str] = []
+
+    class Parser(html.parser.HTMLParser):
+        def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]) -> None:
+            # <input name="SAMLResponse" value="(capture this)">
+            if tag == 'input' and ('name', 'SAMLResponse') in attrs:
+                result.extend(v for k, v in attrs if v and k == 'value')
+
+    Parser().feed(body)
+    if len(result) != 1:
+        logger.debug('response body: %r', body[:500])
+        raise AuthError('🤔 Failed to find SAMLResponse in IdP response')
+
+    return result[0]
+
+
+def aws_sts_assume_role(role_arn: str, provider_arn: str, saml_assertion: str) -> str:
+    """Exchange a SAML assertion for temporary AWS credentials via STS.
+
+    Returns the raw XML response body.
+    """
+    logger.debug('assuming role %r via STS', role_arn)
+
+    params = urllib.parse.urlencode({
+        'Action': 'AssumeRoleWithSAML',
+        'Version': _STS_VERSION,
+        'RoleArn': role_arn,
+        'PrincipalArn': provider_arn,
+        'SAMLAssertion': saml_assertion,
+    }).encode()
+
+    request = urllib.request.Request(_STS_URL, data=params, method='POST')
+    request.add_header('Content-Type', 'application/x-www-form-urlencoded')
+
+    try:
+        response = urllib.request.urlopen(request)
+    except urllib.error.HTTPError as exc:
+        try:
+            tree = ET.parse(exc)
+            message = tree.findtext(f'.//{{{_STS_NS}}}Message', '') or tree.findtext(
+                './/Message', ''
+            )
+        except ET.ParseError:
+            message = ''
+        raise AuthError(
+            f'🐕 STS AssumeRoleWithSAML failed: {message or exc}\n'
+            'Are you in the correct Rover group for cockpit-ci-images access?'
+        ) from exc
+
+    with response:
+        return response.read().decode()
+
+
+def _parse_sts_response(xml: str) -> tuple[S3Key, datetime]:
+    """Parse an STS AssumeRoleWithSAML XML response.
+
+    Returns (S3Key, expiration).  Raises AuthError if the XML is malformed
+    or missing fields.
+    """
+    try:
+        tree = ET.fromstring(xml)
+
+        def text(tag: str) -> str:
+            if val := tree.findtext(f'.//{{{_STS_NS}}}Credentials/{{{_STS_NS}}}{tag}'):
+                return val
+            raise AuthError(f'🤔 STS XML response missing {tag}')
+
+        expiration = datetime.fromisoformat(text('Expiration'))
+        key = (text('AccessKeyId'), text('SecretAccessKey'), text('SessionToken'))
+        return key, expiration
+
+    except (ET.ParseError, ValueError, TypeError) as exc:
+        raise AuthError('🤔 failed to parse STS response') from exc
+
+
+def cache_path(role_name: str) -> Path:
+    """Return the cache file path for STS credentials."""
+    return Path(xdg_cache_home('cockpit-dev', 'sts-credentials', f'{role_name}.xml'))
+
+
+def save_to_cache(role_name: str, xml: str) -> tuple[S3Key, datetime]:
+    """Parse, validate, and cache STS credentials.  Returns (key, expiration)."""
+    key, expiration = _parse_sts_response(xml)
+    path = cache_path(role_name)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with open(os.open(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600), 'w') as fp:
+        fp.write(xml)
+    logger.debug('cached STS credentials to %r (expires %s)', path, expiration)
+    return key, expiration
+
+
+def load_from_cache(role_name: str) -> S3Key | None:
+    """Load cached STS credentials if they exist and aren't expiring within 5 minutes."""
+    try:
+        key, expiration = _parse_sts_response(cache_path(role_name).read_text())
+    except (FileNotFoundError, AuthError) as exc:
+        logger.debug('cached credentials unusable: %r', exc)
+        return None
+    if expiration - datetime.now(timezone.utc) < timedelta(minutes=5):
+        logger.debug('cached credentials expiring soon (%s)', expiration)
+        return None
+    return key
+
+
+@dataclasses.dataclass
+class SAMLTarget:
+    idp_url: str
+    provider_arn: str
+    role_arn: str
+
+
+# Red Hat employee IdP (SAML via Kerberos) → AWS IAM role for cockpit-ci-images
+# Rover group: https://rover.redhat.com/groups/group/it-cloud-aws-727920394381-cockpit-ci-images-download
+TARGETS = {
+    'https://cockpit-ci-images*.s3.*.amazonaws.com/rhel-*': SAMLTarget(
+        idp_url='https://auth.redhat.com/auth/realms/EmployeeIDP/protocol/saml/clients/itaws',
+        provider_arn='arn:aws:iam::727920394381:saml-provider/RedHatInternal',
+        role_arn='arn:aws:iam::727920394381:role/727920394381-cockpit-ci-images-download',
+    ),
+}
+
+
+def _find_target(url: str) -> SAMLTarget | None:
+    for pattern, target in TARGETS.items():
+        if fnmatch.fnmatch(url, pattern):
+            return target
+    return None
+
+
+def try_key(url: str, use_cache: bool = True) -> S3Key | None:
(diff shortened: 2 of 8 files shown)


lib/gssapi_saml_sts.py
```diff
@@ -0,0 +1,219 @@
...
+            logger.debug('SAML/STS credentials unavailable: %s', exc)
+            return None
+        return (creds['access_key'], creds['secret_key'], creds['session_token'])
+
+
+TARGETS = {
+    'https://cockpit-ci-images.s3.us-east-1.amazonaws.com/rhel-*': SAMLTarget(
+        idp_url='https://auth.redhat.com/auth/realms/EmployeeIDP/protocol/saml/clients/itaws',
+        role_arn='arn:aws:iam::727920394381:role/727920394381-cockpit-ci-images-download',
+        provider_arn='arn:aws:iam::727920394381:saml-provider/RedHatInternal',
```
martinpitt, line 190: Can you please add comments/URLs to where these magic values come from?

lib/locking.py
```diff
@@ -0,0 +1,42 @@
...
+    lockpath = path.with_suffix(path.suffix + '.lock')
+    waited = False
+    for _attempt in range(10):
+        lockfile = open(lockpath, 'w')
+        try:
+            fcntl.flock(lockfile, fcntl.LOCK_NB | fcntl.LOCK_EX)
+        except BlockingIOError:
+            if not waited:
+                sys.stderr.write(f'Waiting for concurrent download of {path.name}...\n')
+                waited = True
```
martinpitt: There's no actual waiting going on here AFAICS? Just 10 spinlock rounds?

lib/progressbar.py
```diff
@@ -0,0 +1,155 @@
+# Copyright (C) 2026 Red Hat, Inc.
```
martinpitt, line 1: OMG, can this grow some `pwd.getpwuid(os.getuid()).pw_name != 'martin'`? This is really distracting..

lib/gssapi_saml_sts.py
```diff
@@ -28,192 +27,177 @@
 logger = logging.getLogger(__name__)
 
 
-class AuthError(Exception):
-    """Raised when Kerberos/SAML/STS authentication fails."""
+_STS_URL = 'https://sts.amazonaws.com/'
+_STS_NS = 'https://sts.amazonaws.com/doc/2011-06-15/'
```
martinpitt: doc link please for magic value

lib/gssapi_saml_sts.py
```diff
@@ -28,192 +27,177 @@
...
+def _aws_sts_assume_role(role_arn: str, provider_arn: str, saml_assertion: str) -> str:
+    """Exchange a SAML assertion for temporary AWS credentials via STS.
+
+    Returns the raw XML response body.
+    """
+    logger.debug('assuming role %r via STS', role_arn)
+
+    params = urllib.parse.urlencode({
+        'Action': 'AssumeRoleWithSAML',
+        'Version': '2011-06-15',
```
martinpitt: same magic value, let's factor this into a constant?

saml-login
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python3
```
martinpitt, line 1: This new `saml-login` isn't referenced anywhere yet? Do we actually need this for *uploading* images?

Note: we should eliminate the static `AWS_KEY_LOGS` secret and move to [OIDC/JWT based auth with temporary tokens](https://gitdash.dev/blog/github-actions-oidc-security-identity-federation). For that we can hopefully re-use some machinery that this PR introduces already. But it wouldn't start with a Kerberos token. Let's chat about this today.

martinpitt (commented):
I had a quick look only -- this is rather overwhelming to review in fine detail. It feels like it ought not to be *that* complex, especially as your initial PoC was like 100 lines? Why did that blow up so much, apart from accidental complexity like cute and the webkit script?

lib/progressbar.py
allisonkarlitskaya: `NO_COLOR=1`? :D

no seriously though, I figured this might be controversial.  See the name of the branch ;)

lib/progressbar.py
```diff
@@ -0,0 +1,142 @@
...
+
+def _format_size(size: float) -> str:
+    for unit in ('B', 'kB', 'MB', 'GB'):
+        if size < 1000:
+            return f'{size:.1f} {unit}'
+        size /= 1000
+    return f'{size:.1f} TB'
+
+
+class RateTracker:
```
Venefilyn, line 23: Quick docstring what it does

lib/progressbar.py
Venefilyn, line 1: Maybe rename to something that relates to a progress indicator rather than cute. The rainbow or colors is just extra flair and not related to the progress bar or rate tracker.  Maybe just `progressbar.py`?

lib/progressbar.py
```diff
@@ -0,0 +1,142 @@
...
+    def _format_rate(bytes_per_sec: float) -> str:
+        return f'{_format_size(bytes_per_sec)}/s'
+
+    @staticmethod
+    def _format_duration(seconds: float) -> str:
+        if seconds < 60:
+            return f'{int(seconds)}s'
+        if seconds < 3600:
+            return f'{int(seconds // 60)}m{int(seconds % 60):02d}s'
+        return f'{int(seconds // 3600)}h{int(seconds % 3600 // 60):02d}m'
```
Venefilyn, line 41: This is quite complex to read, I'd simplify it to make it easier

image-download-rhel
```diff
@@ -0,0 +1,248 @@
...
+            logger.debug('%r already downloaded', dest)
+            return True
+
+        sys.stderr.write(f'\nDownloading {GREEN}{name}{RESET}\n\n')
+
+        result = find_store(name, stores, use_cache=use_cache, quiet=quiet)
+        if result is None:
+            sys.stderr.write(f"{RED}couldn't find {name} in any store{RESET}\n")
+            return False
+        url, key, total = result
```
Venefilyn: Is total always known?

lib/progressbar.py
```diff
@@ -0,0 +1,142 @@
...
+                pass
+
+        try:
+            width = max(0, os.get_terminal_size(sys.stderr.fileno()).columns - 60)
+        except OSError:
+            width = 0  # not a tty?  no progress bar!
+
+        bar = ('█' * (self.offset * width // self.total)).ljust(width, '░')
+        if USE_COLOR:
+            bar = ''.join(self._bar_color(i, width) + g for i, g in enumerate(bar))
```
Venefilyn, line 117:
```suggestion
        
        if USE_COLOR:
            bar = ''.join(self._bar_color(i, width) + g for i, g in enumerate(bar))
        else:
            bar = ('█' * (self.offset * width // self.total)).ljust(width, '░')
```

lib/gssapi_saml_sts.py
Venefilyn: OOTL: What is STS?

lib/gssapi_saml_sts.py
```diff
@@ -0,0 +1,214 @@
...
+        raise AuthError('failed to parse STS response') from exc
+
+
+def cache_path(role_name: str) -> Path:
+    """Return the cache file path for STS credentials."""
+    return Path(xdg_cache_home('cockpit-dev', 'sts-credentials', f'{role_name}.xml'))
+
+
+def save_to_cache(role_name: str, xml: str) -> tuple[S3Key, datetime]:
+    """Parse, validate, and cache STS credentials.  Returns (key, expiration)."""
```
Venefilyn, line 155: This doesn't cache the credentials does it? That doesn't seem very good. We can cache the data only but for credentials we should rely on passrings if we can no?

saml-login
```diff
@@ -0,0 +1,106 @@
...
+for use by image-download-rhel.
+"""
+
+import argparse
+import logging
+import sys
+
+import gi  # type: ignore [import-untyped]
+
+from lib.cute import GREEN, RESET
```
Venefilyn: If we are gonna use the colors elsewhere than `cute.py` then lets split it to its own file

Venefilyn (commented):
Quickly went over but feel like it needs more docs in general and a bit concerned about the storing of credentials. How long are the credentials valid for?

lib/locking.py
allisonkarlitskaya: The `fcntl()` is blocking...

lib/gssapi_saml_sts.py
allisonkarlitskaya: AWS Security Token Service: https://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html

saml-login
allisonkarlitskaya: Indeed I already played around with OIDC a bit when trying to implement this stuff and it would fit well into this file.

That's definitely "next step" though.

lib/locking.py
martinpitt: Is it? `fcntl.LOCK_NB` 👀

allisonkarlitskaya: we do one `fcntl.LOCK_NB` to notice that the file is locked by another process just so that we can print the "Waiting..." message.  After that we lock again without `LOCK_NB` and will wait.

image-download-rhel
allisonkarlitskaya: Yes.  We get it from the HEAD that we do to determine the closest mirror.  We also check that the total size is the same on *all* the mirrors and bomb out if it's not.

lib/locking.py
martinpitt: Oh, I see, sorry.

lib/gssapi_saml_sts.py
allisonkarlitskaya: This caches the temporary token that we get issued.  It expires in an hour.  I think this is not the sort of thing that belongs in a keyring...

saml-login
allisonkarlitskaya: Split into `ansi.py` and `cute.py` renamed `progressbar.py`.

lib/progressbar.py
allisonkarlitskaya: Added.

lib/progressbar.py
allisonkarlitskaya: I'm not sure I agree.  It might be kinda complicated to read this from a straight "what does this code do?" sense but it's clear that the intent is to produce a string like "3h10m" or whatever, which is familiar to most humans...

lib/progressbar.py
allisonkarlitskaya: I don't think this makes sense?  The `USE_COLOR` line redefines `bar` based on its previous value... so in your suggested change `bar` would be undefined...

lib/progressbar.py
allisonkarlitskaya: Done

cockpituous:
**Task scheduled:** issue-9071 image-refresh/rhel-8-10

Testing Farm link: https://artifacts.osci.redhat.com/testing-farm/58f156c1-e1ae-40f7-b17b-a75bbd90c7e1

<details>
<summary>Job JSON</summary>

```json
{
  "repo": "cockpit-project/bots",
  "sha": "4901e1caf922b24c8eaae8b1fc2c4739a58cb78a",
  "pull": 9071,
  "slug": "image-refresh-rhel-8-10-4901e1ca-20260608-110242",
  "context": "image-refresh/rhel-8-10",
  "command": [
    "./image-refresh",
    "--verbose",
    "--issue=9071",
    "rhel-8-10"
  ],
  "secrets": [
    "github-token",
    "image-upload"
  ]
}
```
</details>

cockpituous:
**Task scheduled:** issue-9071 image-refresh/arch

Testing Farm link: https://artifacts.osci.redhat.com/testing-farm/18db7dd5-bf04-45b7-bb4d-f0257c576174

<details>
<summary>Job JSON</summary>

```json
{
  "repo": "cockpit-project/bots",
  "sha": "4901e1caf922b24c8eaae8b1fc2c4739a58cb78a",
  "pull": 9071,
  "slug": "image-refresh-arch-4901e1ca-20260608-110246",
  "context": "image-refresh/arch",
  "command": [
    "./image-refresh",
    "--verbose",
    "--issue=9071",
    "arch"
  ],
  "secrets": [
    "github-token",
    "image-upload"
  ]
}
```
</details>

cockpituous:
image-refresh rhel-8-10 started on 2fba6b109fbc. Log: https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/image-refresh-rhel-8-10-4901e1ca-20260608-110242/log.html

cockpituous:
image-refresh arch started on 9f771127b2db. Log: https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/image-refresh-arch-4901e1ca-20260608-110246/log.html

cockpituous:
image-refresh arch succeeded: https://github.com/cockpit-project/bots/commits/image-refresh-arch-20260608-112150

cockpituous:
image-refresh rhel-8-10 succeeded: https://github.com/cockpit-project/bots/commits/image-refresh-rhel-8-10-20260608-113737

lib/gssapi_saml_sts.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+class AuthError(Exception):
+    """Raised when Kerberos/SAML/STS authentication fails."""
+
+
+def _negotiate_header(idp_url: str) -> str:
+    """Create a Negotiate auth header using Kerberos."""
+    import gssapi
+    import gssapi.raw.misc
```
Venefilyn: ~~These imports will cascade up and is never caught anywhere~~

Edit: Ah this is fixed in the last commit

lib/gssapi_saml_sts.py
```diff
@@ -0,0 +1,218 @@
...
+    assert hostname is not None
+    # fixed in python-gssapi 1.9.0: https://github.com/pythongssapi/python-gssapi/pull/338
+    server_name = gssapi.Name(f'HTTP@{hostname}', gssapi.NameType.hostbased_service)  # type: ignore[attr-defined]
+    try:
+        # fixed in python-gssapi 1.9.0: https://github.com/pythongssapi/python-gssapi/pull/338
+        ctx = gssapi.SecurityContext(usage='initiate', name=server_name)  # type: ignore[attr-defined]
+        token = base64.b64encode(ctx.step()).decode()
+    except gssapi.raw.misc.GSSError as exc:
+        statuses = '  '.join(exc.get_all_statuses(exc.min_code, is_maj=False))
+        raise AuthError(f'krb: {statuses}.  Try kinit.') from exc
```
Venefilyn: Huh `raise .. from exc` is a new one to me. TIL

lib/gssapi_saml_sts.py
```diff
@@ -0,0 +1,218 @@
...
+    result: list[str] = []
+
+    class Parser(html.parser.HTMLParser):
+        def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]) -> None:
+            # <input name="SAMLResponse" value="(capture this)">
+            if tag == 'input' and ('name', 'SAMLResponse') in attrs:
+                result.extend(v for k, v in attrs if v and k == 'value')
+
+    Parser().feed(body)
+    if len(result) != 1:
```
Venefilyn, line 83: This by itself doesn't show clearly that result was populated from the `Parser().feed()` call, but it's fine I suppose given the size of the function

lib/gssapi_saml_sts.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    logger.debug('assuming role %r via STS', role_arn)
+
+    params = urllib.parse.urlencode({
+        'Action': 'AssumeRoleWithSAML',
+        'Version': _STS_VERSION,
+        'RoleArn': role_arn,
+        'PrincipalArn': provider_arn,
+        'SAMLAssertion': saml_assertion,
+    }).encode()
```
Venefilyn, line 103: `.urlencode().encode()`? Why twice

lib/gssapi_saml_sts.py
```diff
@@ -0,0 +1,218 @@
...
+
+# Red Hat employee IdP (SAML via Kerberos) → AWS IAM role for cockpit-ci-images
+# Rover group: https://rover.redhat.com/groups/group/it-cloud-aws-727920394381-cockpit-ci-images-download
+TARGETS = {
+    'https://cockpit-ci-images*.s3.*.amazonaws.com/rhel-*': SAMLTarget(
+        idp_url='https://auth.redhat.com/auth/realms/EmployeeIDP/protocol/saml/clients/itaws',
+        provider_arn='arn:aws:iam::727920394381:saml-provider/RedHatInternal',
+        role_arn='arn:aws:iam::727920394381:role/727920394381-cockpit-ci-images-download',
+    ),
+}
```
Venefilyn, line 193: How did you find the `arn`?

lib/ansi.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) 2026 Red Hat, Inc.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import os
+import sys
+
+IS_TTY = os.isatty(sys.stderr.fileno())
+USE_COLOR = 'FORCE_COLOR' in os.environ or (IS_TTY and 'NO_COLOR' not in os.environ)
```
Venefilyn, line 8: We might also just check for `CI` envar and not apply color if that is the case.

lib/progressbar.py
Venefilyn: Aha, yeah you're right. I missed the `enumerate(bar)` so figured it was unnecessary calculations for the first assignment if `USE_COLOR` was set. LGTM then

Venefilyn (dismissed):
Overall LGTM if it works.

lib/gssapi_saml_sts.py
allisonkarlitskaya: I created it... this could probably use a writeup in the cockpituous repo.

lib/gssapi_saml_sts.py
allisonkarlitskaya: Ya I tried a lot to make this suck less... iterator based approaches, some sort of a fancy co-routine situation, raising the result as an exception (ew!), and even almost decided to go with a regexp... there's just no way to do the parsing here in a non-crappy way.  :(

lib/gssapi_saml_sts.py
allisonkarlitskaya: Ya, we have gssapi in the tasks container now but it's not universally deployed in the projects and catching the error here in CI was enough to make me want to fix it for humans who might not have it.

lib/ansi.py
allisonkarlitskaya: I think that's actually wrong.  GitHub actions supports color just fine, for example... and if we did tmate in that environment and ssh'd in, I'd expect TTY to work as well...

lib/gssapi_saml_sts.py
allisonkarlitskaya: We're POSTing the url-encoded body, which expect bytes.  See one line down.  This is all a case of these AWS APIs being realllllly old and not having JSON variants.

Venefilyn (approved):

PR #9078, opened 2026-05-26
lib/stores.py: retire us-east-1.linodeobjects.com

allisonkarlitskaya:
We're slowly moving away from Linode and this is the next step: this mirror is the one that lets anyone download non-RHEL images, but the AWS mirror now fills that role.

The eu-central-1 Linode mirror will remain online for now because of the large number of people who have an S3 key issued there which allows them to download the RHEL images.  We're going to slowly migrate to #9071 for that, but it's not there yet.

Dropping us-east-1 also lets us get rid of our special "public secret" key.

### DIFF
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -57,15 +57,6 @@ def get_key(url: urllib.parse.ParseResult) -> S3Key | None:
             pass
         _, _, hostname = hostname.partition('.')  # strip a leading component
 
-    # Fallback to the "public" secret key if no other keys are configured.
-    # This key will work only for the cockpit-images.us-east-1 bucket.
-    if url.hostname == 'cockpit-images.us-east-1.linodeobjects.com':
-        return (
-            '2RI8J2WQEIC8NI5YS85Z',
-            'V6N8xbuIIDUGXlkHtIXRnje2e9HpIdCGK9nmxaM4',
-            None,
-        )
-
     return None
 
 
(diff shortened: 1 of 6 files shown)


lib/s3.py
```diff
@@ -54,14 +54,6 @@ def get_key(url: urllib.parse.ParseResult) -> tuple[str, str]:
             pass
         _, _, hostname = hostname.partition('.')  # strip a leading component
 
-    # Fallback to the "public" secret key if no other keys are configured.
-    # This key will work only for the cockpit-images.us-east-1 bucket.
-    if url.hostname == 'cockpit-images.us-east-1.linodeobjects.com':
-        return (
-            '2RI8J2WQEIC8NI5YS85Z',
-            'V6N8xbuIIDUGXlkHtIXRnje2e9HpIdCGK9nmxaM4'
```
martinpitt: This will conflict with your cleanup PR, so please choose which one to land first.

martinpitt (dismissed):

allisonkarlitskaya:
Rebased and bonus commit to add the FRA mirror that I spun up.  Playbooks to come in a separate cockpituous PR.

Venefilyn (approved):
Overall LGTM. There are a few mentions of cockpit-images.us-east-1 in bots repo too

martinpitt (approved):
Yay!

PR #9093, opened 2026-05-28
pyproject: Drop npm-update from mypy override

jelly:
This was dropped in 207c73209060cac88adf9499162894c3ef489d35.


allisonkarlitskaya (approved):
thanks!

PR #9102, opened 2026-05-29
image-refresh: factor out some git helpers

allisonkarlitskaya:
Making image-refresh self-contained helped us free ourselves from `task.py` but we also want to bring the other scripts over.  There's good code here and we'd like to share it.  The approach is approximately as follows:

 - GitHub REST API things go in github.py

 - things that involve running the `git` command on a local repository will be collected in a new file called git.py (which we will expand as we go)

An edge case is `git push` which needs to know the remote URL and secrets from the `GitHub` but is a local git command.  We keep it in `git.py` and pass `GitHub` into it to access the remote and the secrets.

This also fixes a bug: the previous code added the secrets via a `git config` command line, which meant that the secret:
 - got written to the disk; and
 - would appear in error messages as the username@; and
 - would have been briefly visible to other users on the system (via `ps`)

The new approach injects it via environment variables as an authorization header.  This should solve all of the above problems.

### DIFF
--- a/lib/git.py
+++ b/lib/git.py
@@ -1,0 +1,61 @@
+# Copyright (C) 2026 Red Hat, Inc.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import logging
+import os
+import re
+import shlex
+import subprocess
+import sys
+from collections.abc import Mapping
+from datetime import datetime, timezone
+from pathlib import Path
+
+from lib.github import GitHub
+
+logger = logging.getLogger(__name__)
+
+
+def _git(
+    *args: str,
+    config: Mapping[str, str] | None = None,
+    dry_run: bool = False,
+) -> str:
+    """Run a git command, logging and returning stdout."""
+    cmd = ["git", *args]
+    logger.debug("+ %s", shlex.join(cmd))
+
+    if dry_run:
+        print('\n** Would', shlex.join(cmd))
+        return ''
+
+    env = {**os.environ, 'GIT_TERMINAL_PROMPT': '0'}
+    if config:
+        env['GIT_CONFIG_COUNT'] = str(len(config))
+        for i, (key, value) in enumerate(config.items()):
+            env[f'GIT_CONFIG_KEY_{i}'] = key
+            env[f'GIT_CONFIG_VALUE_{i}'] = value
+
+    output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, env=env, text=True)
+    sys.stderr.write(output)
+    return output
+
+
+def add(*paths: Path | str) -> None:
+    """Stage files for commit."""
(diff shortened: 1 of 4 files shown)


lib/git.py
```diff
@@ -0,0 +1,75 @@
...
+    env = {**os.environ, 'GIT_TERMINAL_PROMPT': '0'}
+    if config:
+        env['GIT_CONFIG_COUNT'] = str(len(config))
+        for i, (key, value) in enumerate(config.items()):
+            env[f'GIT_CONFIG_KEY_{i}'] = key
+            env[f'GIT_CONFIG_VALUE_{i}'] = value
+    try:
+        output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, env=env, text=True)
+    except subprocess.CalledProcessError as e:
+        logger.debug("! %s", e)
```
martinpitt: This is either redundant (as the `raise` already prints the exception) or useless (as debug messages aren't shown by default). Feels like the whole try/except wrapping could just go?

lib/git.py
```diff
@@ -0,0 +1,75 @@
...
+def commit(message: str, *, allow_empty: bool = False, dry_run: bool = False) -> bool:
+    """Commit staged changes. Returns True if a commit was created."""
+    cmd = ["commit", "-m", message]
+    if allow_empty:
+        cmd.append("--allow-empty")
+
+    if dry_run:
+        print('\n** Would', shlex.join(["git", *cmd]))
+        return True
+
```
martinpitt: DRY, can't this just call _git() with passing `dry_run` along?

lib/git.py
```diff
@@ -0,0 +1,75 @@
...
+    try:
+        _git(*cmd)
+    except subprocess.CalledProcessError:
+        return False
+    return True
+
+
+def push(remote: GitHub, topic: str, *, dry_run: bool = False) -> str:
+    """Configure credentials, generate a branch name, push HEAD. Returns the branch name."""
```
martinpitt: "Configure" vs. "returns" grammar. Also return value should not be in the summary. `</nitpick>`

lib/git.py
allisonkarlitskaya: Yes!  And it reminds me there's another thing I wanted to fix here: I don't like how we deal with empty commits.

lib/git.py
allisonkarlitskaya: You're right.  Fixed it.

lib/git.py
allisonkarlitskaya: Fixed, thanks.

lib/git.py
allisonkarlitskaya: I reworded it but I like it one-line.

martinpitt (approved):
thanks!

PR #9110, opened 2026-06-01
substantial cleanups, drop task.py

allisonkarlitskaya:
This is the final big push to drop task.py.  There are probably mistakes.  We should come up with a plan, but to be honest, "yolo" kinda feels good to me these days... the issues are usually reasonably transparent and easy to fix, and I doubt we'll have another "Friday surprise" like we did with `image-refresh` ;)


task/__init__.py
```diff
@@ -1,55 +1,7 @@
...
-# (at your option) any later version.
-#
-# Cockpit is distributed in the hope that it will be useful, but
-# WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
-# Lesser General Public License for more details.
-#
-# You should have received a copy of the GNU Lesser General Public License
-# along with Cockpit; If not, see <http://www.gnu.org/licenses/>.
+"""Backwards compatibility shim.
```
jelly, line 1: Why drop the Copyright header?

task/__init__.py
```diff
@@ -1,55 +1,7 @@
...
+Older versions of lcov.py do 'from task import github'.
+"""
 
-"""Backwards compatibility shim - import from lib.task instead."""
+from lib import github
```
jelly, line 5: One fallout is probably:

```
[jelle@carbon][~/projects/cockpit/main]%git grep 'import task'
tools/urls-check:from lib import task
```

But hmmm this only runs in Cockpit however and mostly uses it to get the GitHub api for comments

naughty-prune
```diff
@@ -17,72 +17,75 @@
...
+STALE = timedelta(days=21)
 
-def run(unused, verbose=False, **kwargs):
 
-    # Since a month ago
-    now = time.time()
-    since = now - (DAYS * SECONDS_PER_DAY)
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Close stale known issues")
+    parser.add_argument("-n", "--dry-run", action="store_true", help="Dry run to validate this task if supported")
```
jelly, line 35: Do we also want verbose? cockpit-lib-update has that :)

po-refresh
```diff
@@ -1,170 +1,123 @@
...
-    triggers = testmap.tests_for_po_refresh(api.repo)
-    for trigger in triggers:
-        api.post(f"statuses/{head}", {"state": "pending", "context": trigger,
-                                               "description": task.github.NOT_TESTED_DIRECT})
+    # Create a pull request from these changes
+    branch = git.push(github, "po-refresh")
+    pull = github.create_pull_request(
+        branch,
+        "Update translations from Fedora Weblate",
+        labels=['bot', 'no-test', 'release-blocker'],
```
jelly, line 110: bot label is an artifact of the old times? Other workflows don't use it imo

task/__init__.py
jelly, line 1: Dropped license file?

naughty-prune
jelly: Nevermind, this was re-worked during a rebase

po-refresh
allisonkarlitskaya: It has one useful purpose: reducing the number of search results when we go looking for issues.

allisonkarlitskaya: ...and by we, I mean the scripts, like image-trigger

task/__init__.py
allisonkarlitskaya: I mean it's a two-liner now, I don't think it needs a licence...

task/__init__.py
allisonkarlitskaya: Oh.  Interesting.

allisonkarlitskaya: https://github.com/cockpit-project/cockpit/pull/23344

jelly (approved):

PR #9115, opened 2026-06-03
Drop fedora-44* images for anaconda

KKoukiou:
We stopped testing fedora-44 already since the ISO does not change.


jelly (approved):
Thanks!

PR #9137, opened 2026-06-03
github.py: reset extraheader to empty

allisonkarlitskaya:
Before we add our Authorization header make sure we reset the existing value back to empty.  This is a cumulative multi-value and our override will add to the existing value.  Since we stole this technique from GitHub's own checkout action, that's very likely to conflict.

In order to do this we need to move from a dict to a sequence of pairs to allow specifying the value more than once.

### DIFF
--- a/lib/github.py
+++ b/lib/github.py
@@ -175,15 +175,22 @@ class GitHub:
         self.log = Logger(self.cache.directory)
         self.log.write("")
 
-    def config(self) -> Mapping[str, str]:
+    def config(self) -> Sequence[tuple[str, str]]:
         """Git config overrides for authenticating with this remote."""
-        if self.token:
-            basic = base64.b64encode(f'x-access-token:{self.token}'.encode()).decode()
-            return {
-                'credential.helper': '',
-                'http.https://github.com/.extraHeader': f'Authorization: Basic {basic}',
-            }
-        return {}
+
+        # We do this the same way as GitHub's checkout action:
+        # https://github.com/actions/checkout/blob/df4cb1c069e1874edd31b4311f1884172cec0e10/src/git-auth-helper.ts#L55
+        # which means we also need to use an empty value to reset the
+        # multi-value list to empty before adding our own value, just in case
+        # we're running on such a checkout.
+        if not self.token:
+            return ()
+        basic = base64.b64encode(f'x-access-token:{self.token}'.encode()).decode()
+        return (
+            ('credential.helper', ''),
+            ('http.https://github.com/.extraheader', ''),
+            ('http.https://github.com/.extraheader', f'Authorization: Basic {basic}'),
+        )
 
     @functools.cached_property
     def remote(self) -> str:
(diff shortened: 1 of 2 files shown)


lib/github.py
```diff
@@ -175,15 +175,22 @@ def __init__(
...
+        # https://github.com/actions/checkout/blob/df4cb1c069e1874edd31b4311f1884172cec0e10/src/git-auth-helper.ts#L55
+        # which means we also need to use an empty value to reset the
+        # multi-value list to empty before adding our own value, just in case
+        # we're running on such a checkout.
+        if not self.token:
+            return ()
+        basic = base64.b64encode(f'x-access-token:{self.token}'.encode()).decode()
+        return (
+            ('credential.helper', ''),
+            ('http.https://github.com/.extraheader', ''),
```
jelly, line 191: I'm a bit confused by this, is there some special meaning when your env variables are effectively:

`GIT_CONFIG_FOO='$thingfromgitthub' GIT_CONFIG_FOO='' GIT_CONFIG_FOO='our token'`

And the claim is that $thingfromgithub is reset, but I can't find it in the [man page](https://man.archlinux.org/man/extra/git/git-config.1.en#ENVIRONMENT).

jelly: I was honestly confused so had to test this:

```
[jelle@carbon][/tmp/test]%git init
Initialized empty Git repository in /tmp/test/.git/
[jelle@carbon][/tmp/test]% GIT_CONFIG_COUNT=3 \
  GIT_CONFIG_KEY_0=http.extraheader GIT_CONFIG_VALUE_0="wrong-value" \
  GIT_CONFIG_KEY_1=http.extraheader GIT_CONFIG_VALUE_1="" \
  GIT_CONFIG_KEY_2=http.extraheader GIT_CONFIG_VALUE_2="second-value" \
  git config --get-all http.extraheader
```

Am I missing something? :)

allisonkarlitskaya: https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpextraHeader
> Pass an additional HTTP header when communicating with a server. If more than one such entry exists, all of them are added as extra headers. ***To allow overriding the settings inherited from the system config, an empty value will reset the extra headers to the empty list.***

allisonkarlitskaya: to clarify: this isn't visible at the level of `git config get --all`.  That will show you the repo-configured value, followed by the empty value, followed by the override.  The actual "reset on empty entry" logic is implemented in the http code:

https://github.com/git/git/blob/9ac3f193c05c2237e2b14ebaa1149e9fc8a1abe0/http.c#L526

```c
	if (!strcmp("http.extraheader", var)) {
		if (!value) {
			return config_error_nonbool(var);
		} else if (!*value) {
			string_list_clear(&extra_http_headers, 0);
		} else {
			string_list_append(&extra_http_headers, value);
		}
		return 0;
	}
```

ie: during iteration, any encountered empty string clears `extra_http_headers`.

jelly: Aha, I see! Thanks for the clarification.

jelly (approved):

PR #8932, opened 2026-04-17
testmap: move subscriptions to Fedora 44

jelly:

PR #9172, opened 2026-06-09
images: Drop Fedora 42

jelly:
We no longer release to Fedora 42 as it is EOL.


lib/testmap.py
```diff
@@ -190,7 +190,6 @@ def contexts(image: str, *scenarios: Iterable[str], repo: str | None = None) ->
         'main': [
             'centos-10',
             'rhel-10-2',
-            'fedora-42',
             'fedora-43',
         ],
```
Venefilyn, line 195: Should we not add fedora-44 while we're at it?

jelly: Well yes, but that isn't that easy see the [attempt here](https://github.com/cockpit-project/bots/pull/8932) and hopefully [pull request to fix it](https://github.com/cockpit-project/subscription-manager-cockpit/pull/291)

mvollmer: That happens in #8932 I think.

mvollmer (approved):
Less is less!

Venefilyn (approved):

PR #9175, opened 2026-06-10
po-refresh: Fix removing never-tracked .po files

jelly:
Since 23df06f5 git.rm() uses --ignore-unmatch, so it no longer raises on untracked files. The try/except that skipped the commit for those files became a no-op, causing git commit to fail with nothing staged.

Check changes_staged() instead, which works regardless of --ignore-unmatch.


allisonkarlitskaya (approved):
Thanks for find this!  This indeed went back and forth a couple of times: I ended up making the change for the benefit of naughty-prune and it had the negative impact here...

PR #9178, opened 2026-06-10
lib/stores: drop eu-central-1 Linode mirror

allisonkarlitskaya:
This is the last bit of Linode infrastructure that we were using and it's now redundant: public images are openly available again and Red Hat developers are expected to download RHEL images via Kerberos authentication (automatic) or the `saml-login` script.


mvollmer (approved):
AWS works with saml-login for me.

PR #9187, opened 2026-06-11
image-refresh: remove forcepush

allisonkarlitskaya:
We don't need this because:

  - we already have the Closes line that we added ourselves (since we always know the issue number in advance).  Doing the force-push adds a second `Closes:` line.

  - we open the PR as the `cockpituous` user so we don't need the force-push as a workaround for the GitHub anti-recursion rule


mvollmer (approved):
I believe you. :-)

PR #9191, opened 2026-06-15
lib: consolidate S3 signing logic

allisonkarlitskaya:
We have two copies of the S3 signing algorithm: one in lib/ and one in lib/aio/.  Only the non-aio version supports session tokens, but the aio version uses a nice `NamedTuple` for the key.  Merge the best parts of each into `lib/` and have `lib/aio/` used it from there.

There's a subtle functionality change here: files in the s3-keys directory can now contain session keys.  This works the same way as before, but if .split() returns 3 parts, the 3rd part is considered to be the session key.

### DIFF
--- a/lib/s3.py
+++ b/lib/s3.py
@@ -17,6 +17,7 @@ import urllib.parse
 import urllib.request
 import xml.etree.ElementTree as ET
 from collections.abc import Iterable, Mapping, Sequence
+from typing import NamedTuple
 
 from .directories import xdg_config_home
 from .network import host_ssl_context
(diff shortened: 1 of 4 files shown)


lib/s3.py
```diff
@@ -73,46 +81,54 @@ def sign_request(
...
+        **({'x-amz-security-token': key.token} if key.token is not None else {}),
     })
     headers_str = ''.join(f'{k}:{v}\n' for k, v in sorted(headers.items()))
     headers_list = ';'.join(sorted(headers))
 
-    # Extract region from hostname for AWS S3 (e.g., 's3.us-east-1.amazonaws.com' -> 'us-east-1')
-    # AWS requires the actual region; other S3-compatible services accept 'any'
-    region = 'any'
-    if url.hostname.endswith('.amazonaws.com'):
-        # Format: [bucket.]s3.REGION.amazonaws.com - region is third-last component
```
mvollmer, line 96: I guess this is already explained in the "(e.g., ...)" part of the comment above. Otherwise it would be tempting to keep this comment.

mvollmer: (just proving that I am actually reading the code. :-)

mvollmer (approved):
Nice, thanks!

PR #9198, opened 2026-06-16
job-runner: support JSON config

allisonkarlitskaya:
We're going to start generating dynamic configs and it's easier to do this with JSON than TOML.

### DIFF
--- a/lib/aio/jobcontext.py
+++ b/lib/aio/jobcontext.py
@@ -90,8 +90,11 @@ class JobContext(contextlib.AsyncExitStack):
         logger.debug('Loading %s configuration from %s', name, path)
         try:
             with path.open('rb') as file:
-                content = tomllib.load(file)
-        except tomllib.TOMLDecodeError as exc:
+                if path.suffix == '.json':
+                    content = json.load(file)
+                else:
+                    content = tomllib.load(file)
+        except (tomllib.TOMLDecodeError, json.JSONDecodeError) as exc:
             sys.exit(f'{path}: {exc}')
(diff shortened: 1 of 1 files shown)


Venefilyn (approved):

lib/aio/jobcontext.py
```diff
@@ -90,8 +90,11 @@ def load_config(self, path: Path, name: str, *, missing_ok: bool = False) -> Non
         logger.debug('Loading %s configuration from %s', name, path)
         try:
             with path.open('rb') as file:
-                content = tomllib.load(file)
-        except tomllib.TOMLDecodeError as exc:
+                if path.suffix == '.json':
+                    content = json.load(file)
+                else:
+                    content = tomllib.load(file)
+        except (tomllib.TOMLDecodeError, json.JSONDecodeError) as exc:
```
jelly, line 97: With standard lib sure but we could depend on `tomli-w`.

mvollmer (approved):

lib/aio/jobcontext.py
Venefilyn: Unrelated to this PR though since this is just about JSON. But why not use tomllib? Was added in 3.11

allisonkarlitskaya: tomllib is readonly, isn't it?

Venefilyn: So it seems

> [This module does not support writing TOML.](https://docs.python.org/3/library/tomllib.html)

PR #9224, opened 2026-06-24
Add workflow-watch command

allisonkarlitskaya:
This will make it easier to check for failing workflows.  We can add the periodic running of this script to the list of pilot tasks.

 - [x] #9223

PR #9226, opened 2026-06-26
further unit test speedups and cleanups

allisonkarlitskaya:
This is another low-hanging fruit hunt which further improves the single-threaded speed of the unit tests (now down to ~4s from an original ~42s) and makes it possible to parallelize them.  It's also a net decrease in complexity and lines of code (mostly due to removing the json logfile stuff).

Building on #9223 and conflicting #9224 due to the shared first commit.

### DIFF
--- a/lib/github.py
+++ b/lib/github.py
@@ -22,7 +22,6 @@ import json
 import logging
 import os
 import re
-import socket
 import subprocess
 import time
 import urllib.parse
@@ -36,6 +35,8 @@ from lib import cache
 from lib.aio.jsonutil import JsonObject, JsonValue, get_dict, get_dictv, get_int, get_str, typechecked
 from lib.directories import xdg_cache_home, xdg_config_home
 from lib.testmap import is_valid_context
+
+logger = logging.getLogger(__name__)
 
 __all__ = (
     'NOT_TESTED',
@@ -63,20 +64,6 @@ ISSUE_TITLE_IMAGE_REFRESH = "Image refresh for {0}"
 
 _T = TypeVar('_T')
 _DT = TypeVar('_DT')
-
-
-class Logger:
-    def __init__(self, directory: str):
-        hostname = socket.gethostname().split(".")[0]
-        month = time.strftime("%Y%m")
-        self.path = os.path.join(directory, f"{hostname}-{month}.log")
-
-        os.makedirs(directory, exist_ok=True)
-
-    # Yes, we open the file each time
-    def write(self, value: str) -> None:
-        with open(self.path, 'a') as f:
-            f.write(value)
 
 
 class Response(TypedDict):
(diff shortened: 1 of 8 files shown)


lib/github.py
```diff
@@ -150,31 +137,31 @@ def __init__(
...
             except FileNotFoundError:
                 # fall back to GitHub's CLI token
                 try:
                     with open(xdg_config_home("gh/config.yml")) as f:
                         match = re.search(r'oauth_token:\s*(\S+)', f.read())
                     if match:
                         self.token = match.group(1)
+                        logger.debug("github token loaded from %s", f.name)
+                    else:
+                        logger.debug("no oauth_token found in %s", f.name)
```
mvollmer, line 155: Is "oauth token" the same as "github token" in this context? Why does this message say "oauth token"?

mvollmer: Ah, this is saying "no oauth_token assignment in the config file", got it.

test/test_github.py
```diff
@@ -88,21 +87,15 @@ def test_cache(self) -> None:
...
-
-        match = fnmatch.fnmatch(data, expect)
-        if not match:
-            self.fail(f"'{data}' did not match '{expect}'")
+        with self.assertLogs('lib.github', level='DEBUG') as logs:
+            self.api.get("/test/user")
+            self.api.cache.mark(time.time() + 1)
+            self.api.get("/test/user")
+
+        self.assertEqual(logs.output, [
```
mvollmer, line 95: Indentation? Doesn't this need to be under the `with ... as logs:` statement?

mvollmer (changes requested):

lib/github.py
allisonkarlitskaya: Ya this is specifically about the case where (as a final fallback option) we raid the gh cli config and look for an existing token but fail to find it.

test/test_github.py
allisonkarlitskaya: no.  it doesn't.  and maybe even *shouldn't*.  the inside the `with:` is the thing under observation and the outside of it is the inspection of what happened during that observation period.

compare that also with the `assertRaises` for example where it's even more clear why you'd want it that way:

```python
with self.assertRaises(ValueError) as cm:
    do_something()

self.assertEqual(str(cm.exception), "invalid value")
self.assertEqual(cm.exception.args, ("invalid value",))
```

because you can't do the compare until after the exception, but the exception is going to shoot you out of the block.

mvollmer: Aha, thanks!  I was mostly suprised that "logs" is even available outside of the "with". TIL!

mvollmer (approved):

test/test_cache.py
```diff
@@ -43,27 +43,27 @@ def test_read_write(tmp_path: Path) -> None:
 
 
 def test_current(tmp_path: Path) -> None:
-    c = cache.Cache[object](f'{tmp_path}', lag=3)
+    c = cache.Cache[object](f'{tmp_path}', lag=1)
```
jelly, line 46: Isn't this a bit of a micro-optimisation

allisonkarlitskaya: This change is the single biggest improvement in test run times... there's just no need to wait that long... I nearly changed it to 0.1...

PR #9227, opened 2026-06-26
lib/git: allow empty when amending for force-push

allisonkarlitskaya:
...and flush stdout/err before subprocesses for better logging output.

### DIFF
--- a/lib/git.py
+++ b/lib/git.py
@@ -36,6 +36,11 @@ def _git(
             env[f'GIT_CONFIG_KEY_{i}'] = key
             env[f'GIT_CONFIG_VALUE_{i}'] = value
 
+    # stdout is fully buffered when piped (eg. in GitHub Actions), so flush
+    # both streams before spawning git to keep output in execution order
+    sys.stdout.flush()
+    sys.stderr.flush()
+
     return subprocess.check_output(cmd, env=env, text=True)
 
 
--- a/lib/git.py
+++ b/lib/git.py
@@ -120,8 +120,10 @@ def push(remote: GitHub, topic: str, *, dry_run: bool = False) -> str:
     return branch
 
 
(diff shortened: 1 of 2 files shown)


mvollmer (approved):
Thanks!

mvollmer:
I believe that this works, but I don't understand it. So running a subprocess with _its_ stdout connected to a pipe will change the buffering mode of _our_ stdout?

allisonkarlitskaya:
> I believe that this works, but I don't understand it. So running a subprocess with its stdout connected to a pipe will change the buffering mode of our stdout?

No.  In general libc (and Python copying it) have one of three buffering modes.  This is entirely implemented in userspace, inside of the single process:
 - every single print() will write to the fd immediately (unbuffered)
 - print() will buffer writes until it sees a `\n` and then it'll write to the fd ("line buffered")
 - print() will buffer absolutely everything always and only flush the buffer when it's full or atexit when the process is done ("fully buffered")

The default in libc (again, copied by Python) is to use line-buffering if the output in question (stdout, stderr) is connected to a tty.  Otherwise you get fully-buffered.  I guess the idea there is that if you're writing to a pipe or file then interactivity is unimportant and it's better if you can improve performance by doing full block-sized writes every time.

GitHub logging output is a pipe, of course.  Or maybe it's a file?  Either way it's not a tty and "fully buffered" is selected by default.  ie: we never write stdout to the actual fd until the buffer is full.

So given a case where we say like

```python
for i in range(100):
    print(i)
subprocess.run('echo foo')
```

what will happen is that the `print(i)`s will fill the buffer in the Python process and not get printed, the `echo foo` will execute and print to the fd immediately, and then the Python script will continue running and at some point (buffer full or program termination) finally write the `"0\n1\n2\n3\n4\n..."` to stdout *after* the `foo`.  The result of that of course is that the `foo` (or stderr from a failed subcommand) is 100 lines away from where you expect it to be in the log.

PR #9228, opened 2026-06-27
used-tasks-tags: catch 404 on branch lookup

allisonkarlitskaya:
Now that we're using httpx this raises as an exception.  Catch it instead of letting it crash the script.


mvollmer (approved):
Thanks!

PR #9231, opened 2026-06-29
jira: Add meta release label

Venefilyn:
This will help us search Jira for these issues to prevent duplicate
issues as well as filtering out issues we do not care about. In short it
will help find issues related to just what we care about.

### DIFF
--- a/jira-tasks
+++ b/jira-tasks
@@ -46,6 +46,7 @@ class JiraClient:
                      issue_type: str,
                      summary: str, description: str,
                      components: list[str] | None = None,
+                     labels: list[str] | None = None,
                      fix_versions: list[str] | None = None) -> JsonObject:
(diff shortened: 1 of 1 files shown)


jira-tasks
```diff
@@ -65,6 +66,9 @@ class JiraClient:
             "customfield_10638": [{"value": "Automated"}]
         }
 
+        if labels:
+            fields["labels"] = labels
```
jelly, line 70: https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issues#api-rest-api-2-issue-post

Seems good, note the latest is rest api is v3.

jelly (approved):

jira-tasks
tomasmatus: The entire `jira-tasks` script uses v2 api so I think that's fine.

PR #9245, opened 2026-07-03
naughty: Add NetworkManager coredump on Ubuntu 2604

jelly:
Happens in several NetworkManager tests and at random so the naughty matches all crashes:

See:

https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/pull-23428-7a69197a-20260703-064751-ubuntu-2604-networking/log.html


allisonkarlitskaya (changes requested):
Sorry, I kept going in loops looking for the downstream issue link because the Related: here is wrong and I failed to notice it :D

allisonkarlitskaya (approved):
Thanks!

PR #9258, opened 2026-07-08
testmap: drop sub-man branch specification in sub-man-c

tomasmatus:
These are not necessary because as of
https://github.com/cockpit-project/subscription-manager-cockpit/pull/312 sub-man-cockpit is tested against the distro packaged version of sub-man by default.

### DIFF
--- a/lib/testmap.py
+++ b/lib/testmap.py
@@ -209,9 +209,9 @@ REPO_BRANCH_CONTEXT: Mapping[str, Mapping[str, Sequence[str]]] = {
     },
     'cockpit-project/subscription-manager-cockpit': {
         'main': [
-            'centos-9-stream/subscription-manager-1.29',
+            'centos-9-stream',
             'centos-10',
-            'rhel-9-9/subscription-manager-1.29',
+            'rhel-9-9',
             'rhel-10-3',
             'rhel-10-3/devel',
             'fedora-43',


jelly:
> These are not necessary because as of [cockpit-project/subscription-manager-cockpit#312](https://github.com/cockpit-project/subscription-manager-cockpit/pull/312) sub-man-cockpit is tested against the distro packaged version of sub-man by default.

Right, but this drops a few scenarios which were used at least `subscription-manager-1.29`:

```
git clone --quiet https://github.com/candlepin/subscription-manager.git subscription-manager
git -C subscription-manager fetch --quiet --depth=1 origin subscription-manager-1.29
git -C subscription-manager checkout --quiet FETCH_HEAD
checked out subscription-manager/ ref ca7b47b01f9540e273828db90941fe421ce02083
cd subscription-manager && \
fn=$(python3 ./setup.py --fullname); \
python3 ./setup.py sdist && \
mv dist/$fn.tar.gz /work/make-checkout-workdir/dist/subscription-manager.tar.gz
/usr/lib/python3.14/site-packages/setuptools/_distutils/dist.py:289: UserWarning: Unknown distribution option: 'tests_require'
  warnings.warn(msg)
/usr/lib/python3.14/site-packages/setuptools/dist.py:765: SetuptoolsDeprecationWarning: License classifiers are deprecated.
```

So centos-9-stream has:

```
[root@centos-9-stream-127-0-0-2-2201 ~]# rpm -qa | grep subscription-manager
subscription-manager-rhsm-certificates-20220623-1.el9.noarch
python3-subscription-manager-rhsm-1.29.52-1.el9.x86_64
```

So it kinda makes sense to test the  `subscription-manager-1.29` branch to see if subscription-manager-cockpit stays compatible with it however these days there are better ways.

We have [reverse dependency testing](https://artifacts.dev.testing-farm.io/163a95dc-3217-472f-9b2c-26d73a6c0aaf/) in Cockpit for cockpit-files to see if we don't break fsinfo and podman. So in the ideal situation:

* We get a copr repo with subscription-manager-cockpit main builds
* We convince subscription-manager to [setup reverse dependency testing](https://github.com/cockpit-project/cockpit/blob/main/packit.yaml#L50) for `subscription-manager-1.29` and we get rid of the manual `setup.py install` which doesn't really test a packaged version.

Alternatively we rely on CentOS Stream to catch things soon enough, but this seems worth discussing with the candlepin team?

tomasmatus:
> Right, but this drops a few scenarios which were used at least subscription-manager-1.29:

aha, right. The `subscription-manager-1.29` doesn't necessarily mean it is the same as centos branch. I see why this was useful

> We get a copr repo with subscription-manager-cockpit main builds

Should be there already, see [packit.yaml](https://github.com/cockpit-project/subscription-manager-cockpit/blob/main/packit.yaml#L58)

> We convince subscription-manager to [setup reverse dependency testing](https://github.com/cockpit-project/cockpit/blob/main/packit.yaml#L50) for subscription-manager-1.29 and we get rid of the manual setup.py install which doesn't really test a packaged version.

probably something I can make a PR for? Although my main candlepin contact is on PTO right now.

jelly (approved):

PR #9264, opened 2026-07-09
two tests-scan adjustments

allisonkarlitskaya:
...helping us get ready for AWS runners.

### DIFF
--- a/tests-scan
+++ b/tests-scan
@@ -74,16 +74,11 @@ def build_policy(repo: str, requested_contexts: Collection[str]) -> Policy:
     return policy
 
 
-def is_internal_context(context: str) -> bool:
-    for pattern in ["rhel"]:
-        if pattern in context:
-            return True
-    return False
-
-
 def queue_test(entry: QueueEntry, dq: distributed_queue.DistributedQueue) -> None:
     context = entry['job']['context']
-    queue = 'rhel' if is_internal_context(context) else 'public'
+    repo = entry['job']['repo']
+    is_subman = 'subscription-manager' in context or 'subscription-manager' in repo
+    queue = 'rhel' if is_subman and 'rhel' in context else 'public'
     priority = distributed_queue.MAX_PRIORITY if '/devel' in context else distributed_queue.BASELINE_PRIORITY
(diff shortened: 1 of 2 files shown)


tests-scan
```diff
@@ -285,6 +285,7 @@ def cockpit_tasks(api: github.GitHub, contexts: Sequence[str], opts: argparse.Na
                     },
                     "command_subject": command_subject,
                     "slug": slug,
+                    "timeout": 120,
```
jelly, line 280: Kinda long, but lets keep the old timeout alive for now.

tests-scan
```diff
@@ -74,16 +74,11 @@ def build_policy(repo: str, requested_contexts: Collection[str]) -> Policy:
...
-        if pattern in context:
-            return True
-    return False
-
-
 def queue_test(entry: QueueEntry, dq: distributed_queue.DistributedQueue) -> None:
     context = entry['job']['context']
-    queue = 'rhel' if is_internal_context(context) else 'public'
+    repo = entry['job']['repo']
+    is_subman = 'subscription-manager' in context or 'subscription-manager' in repo
```
jelly: We need to discuss this, there is a pull request to get rid of this.

jelly: Now actually.. https://github.com/cockpit-project/subscription-manager-cockpit/pull/333

jelly:
```suggest
    is_subman = 'subscription-manager' in context or 'candlepin/subscription-manager' in repo
```

Our own plugin uses the system packages for subscription-manager now, however `subscription-manager` has a reverse dependency test on `subscription-manager-cockpit` this requires building `subscription-manager` which requires additional packages to be installed.

This is problematic as it uses `setup.py install` will break some day.... So we need to get into the situation we have with podman where they run reverse dependency tests on subscription-manager-cockpit's copr repo.

allisonkarlitskaya:
Now that PSI is offline we really need to land this: rhel jobs no longer have anywhere to run.

tests-scan
allisonkarlitskaya: this is no longer conditionalized based on subman

jelly (approved):

PR #9265, opened 2026-07-09
testingfarm: move environment constraints into fmf plan

jscotka:
Env constraints should be inside FMF plan file instead of hard-coding them in the API request, it is intended to override these values inside plan.
 - arch
 - compose (Fedora-44)
 - virtualization support
 - hardware requirements (memory >= 4 GB, disk >= 60 GiB) . 
  
 Drop these parameters from submit_job() and remove the explicit pipeline timeout, relying on Testing Farm defaults.

### DIFF
--- a/plans/job-runner.fmf
+++ b/plans/job-runner.fmf
@@ -8,7 +8,6 @@ provision:
         disk:
           - size: ">= 60 GiB"
         virtualization:
-            #is-virtualized: true
             is-supported: true
 
 prepare:
--- a/plans/job-runner.fmf
+++ b/plans/job-runner.fmf
@@ -1,4 +1,16 @@
 summary: Run job-runner task remotely
+provision:
+    how: virtual
+    arch: x86_64
+    image: Fedora-44
+    hardware:
+        memory: ">= 4 GB"
+        disk:
+          - size: ">= 60 GiB"
+        virtualization:
+            #is-virtualized: true
+            is-supported: true
+
 prepare:
     how: install
     package:
@@ -16,4 +28,4 @@ execute:
 /job-runner:
     summary: Run job-runner task
     test: ../job-runner json "$JOB_JSON"
-    duration: 2h
+    duration: 4h
--- a/plans/job-runner.fmf
+++ b/plans/job-runner.fmf
@@ -4,8 +4,6 @@ provision:
     image: Fedora-44
     hardware:
         memory: ">= 4 GB"
-        disk:
-          - size: ">= 60 GiB"
         virtualization:
             is-supported: true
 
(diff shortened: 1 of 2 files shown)


plans/job-runner.fmf
```diff
@@ -1,4 +1,16 @@
...
+provision:
+    how: virtual
+    arch: x86_64
+    image: Fedora-44
+    hardware:
+        memory: ">= 4 GB"
+        disk:
+          - size: ">= 60 GiB"
+        virtualization:
+            #is-virtualized: true
```
allisonkarlitskaya: this definitely needs to be true.  it's the most important part...

plans/job-runner.fmf
```diff
@@ -1,4 +1,16 @@
 summary: Run job-runner task remotely
+provision:
+    how: virtual
+    arch: x86_64
+    image: Fedora-44
+    hardware:
+        memory: ">= 4 GB"
+        disk:
+          - size: ">= 60 GiB"
```
allisonkarlitskaya: I'd definitely want @jelly to weigh in on this.  We've had some issues in the past...

plans/job-runner.fmf
jscotka: it should be fine. as you are asking to support virtualization, not that machine itself is virtual and support nested virt

plans/job-runner.fmf
jscotka: these constraints could be dangerous, to not be able to select any machine, but I think it should be fine, in fact I've never used disc size requirements inside TF. But We'll see.

jscotka:
### scheduled from command line
`testing-farm request --git-url https://github.com/cockpit-project/bots --git-ref=tmt_job_runner --plan /job-runner --compose Fedora-44`

seems to be able to run : https://artifacts.osci.redhat.com/testing-farm/616da0f9-4f3b-4b8b-a516-ea1b9a1bd478/work-job-runnerx5xiq0kf/tmt-run.log

https://artifacts.osci.redhat.com/testing-farm/616da0f9-4f3b-4b8b-a516-ea1b9a1bd478/

Various options can be places inside FMF plan file

### some limitations
  - API calls have to select `arch` (CLI probably hides it inside as default if not overriden)
  - also `compose` has to be selected when HTTP calling, otherwise container image will be used.

plans/job-runner.fmf
jelly: We ran into issues with bootc afaik:

images/scripts/lib/bootc.bootstrap:subprocess.check_call(['qemu-img', 'resize', '-f', 'qcow2', args.vmpath, '+10G'])

PR #9274, opened 2026-07-14
s3stream: follow CI logs from the CLI

allisonkarlitskaya:
This lets you get the raw log from a complete or in-progress CI run and includes a `--follow` option which exits when the job completes.

### DIFF
--- a/lib/aio/s3streamer.py
+++ b/lib/aio/s3streamer.py
@@ -212,12 +212,12 @@ class LogStreamer:
 
     def start(self, data: str) -> None:
         # Send the initial data immediately, to get the chunks file written out.
-        self.pending = data.encode()
+        self.pending = data.encode(errors='replace')
         self.send_pending()
         AttachmentsDirectory(self.index, f'{LIB_DIR}/s3-html').scan()
 
     def write(self, data: str) -> None:
-        self.pending += data.encode()
+        self.pending += data.encode(errors='replace')
 
         if len(self.pending) > LogStreamer.SIZE_LIMIT:
             self.send_pending()
--- a/lib/aio/s3streamer.py
+++ b/lib/aio/s3streamer.py
@@ -1,17 +1,92 @@
-# Copyright (C) 2022-2024 Red Hat, Inc.
+# SPDX-FileCopyrightText: 2022-2024 Red Hat, Inc.
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+# S3 Log Streaming Protocol
+# =========================
 #
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU Affero General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
+# This module implements a protocol for streaming log output through S3 (or
+# any object store with similar semantics).  The core problem is that S3
+# objects are immutable once written, so we can't append to a file.  Instead,
+# we write a series of chunk objects and a manifest that tells clients how to
+# reassemble them.  The base filename ("log" in this case) is not special —
+# any name works as long as the writer and clients agree on it.
 #
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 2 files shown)


s3stream
```diff
@@ -0,0 +1,96 @@
...
+            for chunk_size in chunks:
+                chunk_end = chunk_start + chunk_size
+                if bytes_received < chunk_end:
+                    data = fetch(
+                        f"{base_url}.{chunk_start}-{chunk_end}",
+                        bytes_received - chunk_start,
+                    )
+                    bytes_received += len(data)
+                    yield data
+                chunk_start = chunk_end
```
mvollmer, line 61: This assumes that the `fetch` call above always returns the full chunk, no?  I.e., that `bytes_received == chunk_end - chunk_start`.

s3stream
```diff
@@ -0,0 +1,96 @@
...
+        except urllib.error.HTTPError as exc:
+            # S3 returns 403 for missing objects in some bucket configurations
+            if exc.code in (403, 404):
+                break
```
mvollmer, line 65: Does this break the `while` loop or just the `try`?

s3stream
```diff
@@ -0,0 +1,96 @@
...
+            raise
+
+        if not follow:
+            return
+
+        time.sleep(10)
+
+    # Streaming is over (or never started) — fetch the (remaining) final log
```
mvollmer, line 73: So the `break` above gets us here when `.chunks` doesn't exist?

mvollmer (dismissed):
`fetch_log` was difficult to follow, but I have no good ideas for how to make it clearer.

Venefilyn (commented):
Seems fine but you can also provide `log.html` and then it just outputs the HTML, which seems unintended. Would be nice for it to take the `log.html` URL as well and try the log URL itself

That way you can just copy the link from a PR without having to change it

lib/aio/s3streamer.py
```diff
@@ -1,17 +1,90 @@
...
+# any object store with similar semantics).  The core problem is that S3
+# objects are immutable once written, so we can't append to a file.  Instead,
+# we write a series of chunk objects and a manifest that tells clients how to
+# reassemble them.  The base filename ("log" in this case) is not special —
+# any name works as long as the writer and clients agree on it.
+#
+# The entire protocol is bytes-oriented: chunk sizes in the manifest are
+# byte counts, Range request offsets are byte offsets, and chunk filenames
+# encode byte ranges.  Although character offsets are never used for
+# anything, the content is assumed to be UTF-8 text.  The server
```
tomasmatus, line 17: here you promise UTF-8 in the entire protocol but
```py
self.destination.write('log.chunks', json.dumps(chunk_sizes).encode('ascii'))
```

:)

tomasmatus (commented):
Thanks for the docs! It is much clearer to me how it works now

allisonkarlitskaya (commented):
The removal of support for servers without range requests was a mistake because we suggest `python3 -m http.server` for use with our local log driver and.... it doesn't support range requests.

s3stream
allisonkarlitskaya: I reworked this a bit, but the general logic here is: we trust that `urllib` does what we asked it or throws an error.  In the event of an incomplete read, we expect to hear about that...

s3stream
allisonkarlitskaya: `break` doesn't interact with `try` in any way...

s3stream
allisonkarlitskaya: exactly.

lib/aio/s3streamer.py
allisonkarlitskaya: ascii is utf8 :)

s3stream
```diff
@@ -0,0 +1,99 @@
...
+    yield fetch(base_url, bytes_received)
```
tomasmatus, line 74: What happens here if the "(remaining) final log" is an empty string? ie the client already has the full log. That means `bytes_received == size(log)` and the fetch call `fetch(base_url, bytes_received)` would be "off-by-one", no? Because the valid byte ranges are `0 - (size(log) - 1)`. So the server would return 416.

tomasmatus (commented):
Thanks for the docs, it is much clearer how log streaming works. Just some minor nit, otherwise I think it looks fine

PR #9278, opened 2026-07-16
lib/progressbar: show progress on non-tty

allisonkarlitskaya:
This lets people tuning in via s3streamer notice that something is happening.

### DIFF
--- a/lib/progressbar.py
+++ b/lib/progressbar.py
@@ -70,12 +70,32 @@ class ProgressBar(threading.Thread):
         self.offset = 0
         self.status: RateTracker | str | None = None
         self._done = threading.Event()
+        self._percent_shown = 0.0
 
     def track_rate(self) -> None:
         self.status = RateTracker(start_time=time.monotonic(), start_offset=self.offset)
 
     def update(self, n: int) -> None:
         self.offset += n
+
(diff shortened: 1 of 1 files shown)


lib/progressbar.py
```diff
@@ -70,13 +70,26 @@ def __init__(self, total: int) -> None:
...
+        if not IS_TTY and self.total:
+            while self._percent_shown <= self.offset * 100 / self.total:
+                if self._percent_shown == 100:
+                    sys.stderr.write('\n')  # before the summary line
+                elif self._percent_shown % 10 == 0:
+                    sys.stderr.write(f' {self._percent_shown:.0f}% ')
+                else:
+                    sys.stderr.write('.')
+                sys.stderr.flush()
+                self._percent_shown += 2.5
```
tomasmatus, line 98: magic constant `2.5` or is there more to it?

tomasmatus (dismissed):

lib/progressbar.py
allisonkarlitskaya: One dot per 2.5%...

allisonkarlitskaya: Added some comments

tomasmatus (approved):
Thanks!

PR #9279, opened 2026-07-17
Improve the generated SSH command

M4rtinK:
Currently the generated SSH connection command works, but causes two main issues in its current form:
* SSH key ends up being saved persistently on the host for the ephemeral VM
* the user is asked to verify the ephemeral host key is correct, which is useless

So add two more options to the generated SSH command:
* an option no to save the key to the known-hosts file
* an option not to ask for key validation

This should improve the user experience when connecting to machines spawned by bots, as there will be no more need to manually wipe stale keys from past test VMs and being asked to validate ephemeral keys.

Instead, the user should be logged in to the VM immediately every time.

### DIFF
--- a/machine/machine_core/machine.py
+++ b/machine/machine_core/machine.py
@@ -34,7 +34,7 @@ TTY LOGIN
 """
 REMOTE_MESSAGE = """
 SSH ACCESS
-  $ ssh -p {ssh_port} -i bots/machine/identity {ssh_user}@{ssh_address}
+  $ ssh -p {ssh_port} -i bots/machine/identity {ssh_user}@{ssh_address} {key_options}
 
 COCKPIT
   https://{web_address}:{web_port}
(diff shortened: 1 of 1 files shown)


M4rtinK:
Adjusting to keep sane line length. :)

tomasmatus (approved):
Quite useful indeed, the message now looks like this and provides a self-contained info how to access the test VM.
```
SSH ACCESS
  $ ssh -p 2301 -i bots/machine/identity root@127.0.0.2 -o "UserKnownHostsFile=/dev/null" "-o StrictHostKeyChecking=accept-new"
  ```

Note that we already have the same QoL described in [cockpit/test/README.md](https://github.com/cockpit-project/cockpit/blob/main/test/README.md#convenient-test-vm-ssh-access) so I'd recommend adding this snippet to your ssh config for quick and easy access to test VMs if you are hacking on cockpit often.

PR #9305, opened 2026-07-29
image-trigger: Label image-refresh issues

jelly:
Introduce a custom "image-refresh" label for all open image-refresh issues. This allows them to be easily added to the pilot issues board.


tomasmatus (approved):
I see that the label exists, so this should work

PR #9316, opened 2026-08-04
AWS dynamic dispatcher with transient secrets

allisonkarlitskaya:
This is a massive change.  It moves us to being able to dispatch jobs on AWS, but that undersells it substantially.

The three main benefits here (and the resultant complexity) are:
 - dynamic job configuration to move us away from static secrets (although GitHub is still an issue)
 - dynamic runner dispatch to allow ramping up more runners in time of high demand and to avoid runners sitting idle when there's no jobs (which is most of the time)
 - infrastructure-as-code in the bots repository so we can deploy everything in a few seconds.  No more Ansible (and long wait times) for the dispatcher/runners.

The `lib/aws/README.md` file is probably a good place to start reading.


lib/aws/README.md
```diff
@@ -0,0 +1,319 @@
...
+You can also check the [dashboard
+page](https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/dashboard.html) that
+the dispatcher regularly updates when it's running:
+
+### Launch a one-off runner instance for debugging
+
+You can manually launch a runner instance for a given job JSON blob like so:
+
+```
+AWS_PROFILE-cockpit-ci python3 -m lib.aws.launch_runner '{"slug": "abc123", ...}'
```
jelly:
```suggestion
AWS_PROFILE=cockpit-ci python3 -m lib.aws.launch_runner '{"slug": "abc123", ...}'
```

lib/aws/ruff.toml
```diff
@@ -0,0 +1,10 @@
+# The code in lib/aws is meant to work with Python 3.13 and uses the
+# default options for formatting (88 character lines, double quotes).
```
Venefilyn, line 2: Why specifically 88

lib/gssapi_saml_sts.py
```diff
@@ -87,11 +94,12 @@ def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]) -> None
     return result[0]
 
 
-def aws_sts_assume_role(role_arn: str, provider_arn: str, saml_assertion: str) -> str:
+def aws_sts_assume_role(account_id: str, role: str, provider_arn: str, saml_assertion: str, duration: timedelta) -> str:
     """Exchange a SAML assertion for temporary AWS credentials via STS.
 
     Returns the raw XML response body.
     """
+    role_arn = f'arn:aws:iam::{account_id}:role/{role}'
```
Venefilyn, line 104: Can you add a git trailer `See-also` or something for AWS ARN for 2ab800a408cca7780073b9c0ae90aebe158b6fb2?

lib/aio/jobcontext.py
```diff
@@ -128,43 +127,47 @@ def __init__(self, config_file: Path | str | None = None, *, debug: bool = False
...
+                        if ref in self._external_secrets:
+                            resolved[ref] = os.path.expanduser(self._external_secrets[ref])
+                        else:
+                            try:
+                                tmpdir.mkdir(parents=True, exist_ok=True)
+                                resolved[ref] = str(unpack_inline_secret(tmpdir, self._inline_secrets, ref))
+                            except KeyError:
+                                msg = f'container.secrets.{name} references %{{{ref}}} but no value is configured'
+                                raise LookupError(msg) from None
+                all_args.append(re.sub(r'%\{([^}]+)\}', lambda m: resolved[m.group(1)], arg))
```
Venefilyn: Consolidate the two RegEx definitions. Compile it first and run it later. Now one could be changed without the other or get human error'd

lib/aio/job.py
```diff
@@ -173,10 +173,21 @@ async def run_job(job: Job, ctx: JobContext) -> None:
         status = subject.forge.get_status(job.subject.repo, subject.sha, job.context, log.url)
         logger.info('Log: %s', log.url)
 
+        journal = ''
+        invocation_id = os.getenv("INVOCATION_ID")
```
Venefilyn, line 178: I couldn't find `INVOCATION_ID` mentioned anywhere else, where do we get this from? This also needs to be documented IMO

lib/aws/ec2.py
```diff
@@ -0,0 +1,317 @@
...
+                },
+                {
+                    "path": "/etc/cockpit-ci/job.json",
+                    "contents": json_contents(job),
+                    "mode": 0o644,
+                },
+                {
+                    "path": "/usr/local/bin/run-job",
+                    "contents": string_contents(r"""#!/bin/bash
+                        set -euxo pipefail
```
Venefilyn, line 203: Doesn't this technically mean we'll get a huge indentation? Maybe split it out to a different file and load it that way just to make it easier (or just remove the indentation completely)

lib/aws/ec2.py
```diff
@@ -0,0 +1,317 @@
...
+                        maybe_sit() {
+                            # If there are ssh keys configured, let the user inspect it
+                            if [ -s /home/core/.ssh/authorized_keys.d/ignition ]; then
+                                sleep 10m
+                            fi
+                        }
+                        trap maybe_sit ERR
+
+                        # Set SELinux permissive
+                        setenforce 0
```
Venefilyn, line 214: Comment should be why rather than what

lib/aws/ec2.py
```diff
@@ -0,0 +1,317 @@
...
+                        )
+                        mkfs.btrfs -f "$instance_store_device"
+                        mount "$instance_store_device" /var/lib/containers
+
+                        # Download bots and unpack it (we don't need git history)
+                        bots_url="$(</etc/cockpit-ci/bots-url)"
+                        curl -sSLf "${bots_url}" | tar xz --strip-components=1
+
+                        # Download uv and install it into the cwd
+                        curl -sSLf https://astral.sh/uv/install.sh | UV_INSTALL_DIR=. sh
```
Venefilyn, line 232: Not super happy about downloading and calling an unverified bash script like this. If uv gets compromised we'll be infected too if we redeploy while uv is compromised. 

Is there a better way to install uv?

lib/aws/ec2.py
```diff
@@ -0,0 +1,317 @@
...
+
+                        # Required for 'podman run --userns=auto'
+                        tee -a /etc/sub{u,g}id <<< 'containers:2147483647:2147483648'
```
Venefilyn, line 235: Why is it required?

lib/aws/ec2.py
```diff
@@ -0,0 +1,317 @@
...
+                    # blocks IMDS via iptables, causing this to fail
+                    # and delay boot with retries.
+                    "name": "afterburn-sshkeys@core.service",
+                    "mask": True,
+                },
+                {
+                    "name": "run-job.service",
+                    "enabled": True,
+                    "contents": f"""\
+                            [Unit]
```
Venefilyn, line 263: Same here, does this not create the file with huge indentation?

lib/aws/ensure_resource.py
```diff
@@ -0,0 +1,546 @@
...
+    managed_arns.add(arn)
+    return arn
+
+
+def ensure_user(
+    name: str,
+    managed_policies: Sequence[str],
+) -> str:
+    iam = boto3.client("iam")
+    arn = f"arn:aws:iam::{ACCOUNT_ID}:user/{name}"
```
Venefilyn, line 150: We declare this a few times already, can we consolidate it?

lib/aws/infra_definitions.py
```diff
@@ -0,0 +1,506 @@
...
+            },
+        ),
+        managed_policies=[policy_images_download],
+        max_session_duration=REDHAT_SSO_IMAGE_DOWNLOAD_MAX_SESSION,
+    )
+
+    # Users
+    ensure_user(
+        # This is a testing account.  It can do anything that the others can.
+        "cockpit-ci",
```
Venefilyn: Testing as in for us to play around or for CI to run tests with

lib/aws/README.md
```diff
@@ -0,0 +1,319 @@
...
+   Things belong in this file if they're used to setup the infrastructure
+   (`infractl sync`) and also to *use* the infrastructure, for example the
+   names of S3 buckets where images are downloaded.  In general, the constants
+   in this file are really useful with `git grep` to connect infrastructure
+   definitions to actual runtime uses.
+
+### Infrastructure deployment
+
+ - [`authorized_keys`](authorized_keys): an `authorized_keys` file for who is
+   allowed to ssh to the dispatcher
```
Venefilyn, line 18: Do we need kerberos or saml login on top of having an authorized key?

lib/aws/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+with the `/home/lis/src/bots/main` path adjusted to point to your bots
+checkout.
+
+You can test that your admin (or any other profile) access is working by using
+a command like this:
+
+```
+AWS_PROFILE=admin aws sts get-caller-identity
+```
```
Venefilyn, line 139: I'm assuming this is `awscli2`?

lib/aws/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+Finally, if you want to login to the AWS web console, you need to use this URL:
+https://auth.redhat.com/auth/realms/EmployeeIDP/protocol/saml/clients/itaws
+
+In general, the web console is good for poking around or doing experiments, but
```
Venefilyn: but..

lib/aws/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+You can test that your admin (or any other profile) access is working by using
+a command like this:
+
+```
+AWS_PROFILE=admin aws sts get-caller-identity
+```
+
+Finally, if you want to login to the AWS web console, you need to use this URL:
+https://auth.redhat.com/auth/realms/EmployeeIDP/protocol/saml/clients/itaws
```
Venefilyn, line 142: Tried to take a look. No clue what is going on - so many things there.

lib/aws/README.md
```diff
@@ -0,0 +1,319 @@
...
+redeploy the entire infrastructure from scratch (which is harmless and fast).
+
+
+### Make changes to the CI infrastructure
+
+You can make the required changes by modifying `lib/aws/infra_definitions.py` and running
+
+```
+AWS_PROFILE=admin python3 -m lib.aws.infractl sync --bots-ref main
+```
```
Venefilyn, line 219: Would be nice if we could run `infractl` instead of `python3 -m lib.aws.infractl`. Could we add a repo top-level invocation script?

```python
#! /usr/bin/python3 -sP

if __name__ == '__main__':
    sys.exit(callhere)
```

lib/html/dashboard/dashboard.js
```diff
@@ -0,0 +1,228 @@
...
+ *   logs_visible?: boolean,
+ * }} Job
+ *
+ * @typedef {{
+ *   jobs: Record<string, Job>,
+ *   instances: Record<string, Instance>,
+ * }} DashboardData
+ */
+
+import { css, html, LitElement } from "https://cdn.jsdelivr.net/gh/lit/dist@3/all/lit-all.min.js";
```
Venefilyn, line 23: IMO hardcode the value and import it through `dashboard.html` instead, that way we can add integrity as well and use ESM.

```html
<script type="importmap">
  {
    "imports": {
      "lit": "https://cdn.jsdelivr.net/npm/lit@3.3.3/+esm"
    },
    "integrity": {
      "https://cdn.jsdelivr.net/npm/lit@3.3.3/+esm": "sha384-YynveZF4bP2Bqr6B3IosyScL+dLGEx74GFZ88Zpu/VjO6ah3rBikPBow+4YA74j7"
    }
  }
</script>
```

then in `dashboard.js` you can then `import x from lit` instead.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/importmap

Venefilyn (changes requested):
Thanks for splitting the changes up into multiple commits!

---

Typo in commit message 2ab800a408cca7780073b9c0ae90aebe158b6fb2

> Move that to to the

---

Personal peeve: In 6d8f7e78d4def39720c2f513271ee2888e5eee2c I see git trailer
`Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>`

Change to `Assisted-by: AI` or something to remove advertisement :p

lib/aws/README.md
```diff
@@ -0,0 +1,319 @@
...
+group](https://rover.redhat.com/groups/edit/basic/cockpit-infra-accounts).
+Contact one of the owners of that group if you want to be added.
+
+You can add those credentials to `~/.aws/credentials` in a format similar to this:
+
+```
+[cockpit-ci-infractl]
+aws_access_key_id = AKIA...
+aws_secret_access_key = secret...
+```
```
Venefilyn, line 112: Mention this should be the `infractl` user explicitly. Or that the profile itself should be the user

allisonkarlitskaya: How do you mean?

lib/aws/infractl.py
```diff
@@ -0,0 +1,336 @@
...
+                {p.name: p.read_text() for p in args.secrets_dir.iterdir()}
+                if args.secrets_dir
+                else {}
+            )
+
+            print("\n# Deployment")
+            print(f"  - bots ref: {args.bots_ref}")
+            print(f"  - bots sha: {sha}")
+            print(f"  - bots url: {bots_url}")
+            print(f"  - secrets: {list(secrets)}")
```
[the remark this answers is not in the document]
allisonkarlitskaya: Go home, you're drunk.

(list(secrets) on a dict will take only the key names)

saml-login
```diff
@@ -11,53 +11,99 @@
...
+from lib.s3 import S3Key
 
 logger = logging.getLogger(__name__)
 
 
+def print_credentials(key: S3Key, output: Literal['awscreds', 'env']) -> None:
+    match output:
+        case 'env':
+            print(f'export AWS_ACCESS_KEY_ID={shlex.quote(key.access)}')
+            print(f'export AWS_SECRET_ACCESS_KEY={shlex.quote(key.secret)}')
```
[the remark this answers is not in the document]
allisonkarlitskaya: THIS IS LITERALLY THE ENTIRE POINT OF THIS SCRIPT

lib/aws/ruff.toml
allisonkarlitskaya: That's the default and the Python standard (per PEP8).

allisonkarlitskaya: Sorry, it's the default but *not* the PEP8 recommendation: https://docs.astral.sh/ruff/rules/line-too-long/

Either way, it's the default used by both Black and Ruff...

lib/gssapi_saml_sts.py
allisonkarlitskaya: You mean like, a link to what defines role ARNs?  I figure if we wanted to get into that game we'd end up with quite a lot of these comments around...

lib/aio/jobcontext.py
allisonkarlitskaya: I don't believe in precompiled regexps (since they reduce readability and Python caches them internally anyway) but I agree that this code is way too two-pass and the duplication is definitely a smell.  I cleaned it up to do a single pass with a memoizing function.  Thanks!

lib/aws/ec2.py
allisonkarlitskaya: yes; we'll definitely get the indentation but i don't think it matters much at all.  the only line that it matters for is the sh-bang and it's not indented.

I could maybe see splitting this into a separate file for better syntax highlighting or something but I actually prefer having everything here inline in one place.  I also prefer optimizing the indentation for human readability rather than saving a few bytes in a language where whitespace is ignored...

lib/aws/README.md
allisonkarlitskaya: Yes.  We could definitely document that better, thanks.

allisonkarlitskaya: I mentioned installing `awscli2` but didn't document that it was for the `aws` CLI (although it might be guessed from the name).  I bridged the gap in both places.

lib/aws/README.md
allisonkarlitskaya: thanks.  fixed.

lib/aws/README.md
allisonkarlitskaya: ya.  it's a bit much....

lib/aws/README.md
allisonkarlitskaya: Ya.  I've gone back and forth on this.  I think this might be OK except that it's currently very AWS specific.  That's fine when you invoke it like `lib.aws.infractl` but when you invoke it directly by a script in the root maybe it should have a more descriptive name...

...that's sort of where I stopped thinking about it and left it like it is.  Ideas?

lib/html/dashboard/dashboard.js
allisonkarlitskaya: I had it like this and I actually changed it.  I'm not really a web person so I'm not super bright on this stuff...

lib/gssapi_saml_sts.py
Venefilyn: No just like a git trailer in the commit message with link to AWS docs about ARN stuffs

lib/aws/README.md
Venefilyn: You have the docs say

> You can add those credentials to `~/.aws/credentials` in a format similar to this:

But without seeing the file header `[cockpit-ci-infractl]` I wasn't sure what Bitwarden entry to add. Didn't know which ones _we_ should use vs CI

allisonkarlitskaya:
The job-runner fail is a cockituous issue: https://github.com/cockpit-project/cockpituous/pull/713

Venefilyn (dismissed):

allisonkarlitskaya:
@Venefilyn thanks for the review.  @bookwar wanted more people to look at this so let's leave it open for more people to chime in.

lib/aws/account.py
```diff
@@ -0,0 +1,102 @@
...
+# Things in this section are things that are to some extent outside of
+# our control and may need to change.  The account number is obvious but
+# we may also need to change S3 bucket names since they are globally
+# unique and first-come, first-serve.
+ACCOUNT_ID = '727920394381'
+
+# S3 BUCKETS
+# Bucket names are globally unique and first-come, first-serve.
+LOGS_BUCKET = 'cockpit-ci-logs'
+CI_IMAGES_BUCKETS: Mapping[str, BucketLocationConstraintType | Literal['us-east-1']] = {
```
mvollmer, line 33: Why is the `| Literal['us-east-1']` here?  Bug in BucketLocationConstraintType?

mvollmer (dismissed):
I love it! :-)

pyproject.toml
```diff
@@ -2,11 +2,15 @@
...
+[tool.pyright]
+stubPath = "lib/stubs"
+
 [tool.mypy]
 python_version = "3.14"
 strict = true
 follow_imports = 'silent'  # https://github.com/python-lsp/pylsp-mypy/issues/81
 scripts_are_modules = true  # allow checking all scripts in one invocation
+mypy_path = 'lib/stubs'
 warn_return_any = false
```
jelly, line 14: The concern with this commit is that it will get out of sync over a while. Pika is still an active project, can't we instead donate this to a pika-stubs repository they maintain or contribute the trivial typing.

https://github.com/pika/pika/

lib/aws/account.py
```diff
@@ -0,0 +1,102 @@
...
+
+# This file defines constants for things like policy, role, bucket, etc. names
+# along with some related constants like our AWS account number and SSO URLs.
+# When adding new resources, ensure that 'cockpit-ci' appears in the name
+# (ideally at the start) to make it easier to find resources that we've
+# deployed.
+
+# It should be possible to find the resources with a command like:
+#
+#     aws resource-explorer-2 search --query-string "cockpit"
```
jelly, line 21: And with what profile, it turns out that the aws cli is not very user friendly and:

```
[default]
region = us-east-1
```

is just ignored..

saml-login
```diff
@@ -11,53 +11,99 @@ for use by image-download.
 """
 
 import argparse
+import json
 import logging
+import shlex
 import sys
+from datetime import timedelta
+from typing import Literal
```
jelly, line 19: I tested this and it does not work:

```
[jelle@carbon][~]%AWS_PROFILE=admin aws sts get-caller-identity

aws: [ERROR]: Expecting value: line 1 column 1 (char 0)
```

With

```
[admin]
credential_process = /home/lis/src/bots/main/saml-login --output=awscreds 727920394381-admin
```

spawns SSO login but after closing the browser it doesn't pass credentials correctly

lib/html/dashboard/dashboard.html
```diff
@@ -0,0 +1,49 @@
...
+                --fg: #c0caf5;
+                --border: #3b4261;
+                --dim: #565f89;
+                --cyan: #7dcfff;
+                --green: #9ece6a;
+                --yellow: #e0af68;
+                --red: #f7768e;
+                --blue: #7aa2f7;
+                --instance-bg: #24283b;
+            }
```
jelly, line 32: The dark mode contract isn't great, its very hard to read the greyed out text.

lib/html/dashboard/dashboard.js
```diff
@@ -0,0 +1,228 @@
+// @ts-check
```
jelly, line 1: This dashboard seemingly so simple I would prefer if we didn't introduce more frameworks then currently we have:

* PatternFly/React for Cockpit
* log.html which is vanilla JS + PatternFly

This introduces a third option, lit with biome/tsc.

lib/html/dashboard/dashboard.js
```diff
@@ -0,0 +1,228 @@
...
+    disconnectedCallback() {
+        super.disconnectedCallback();
+        clearInterval(this._fetchInterval);
+        clearInterval(this._tickInterval);
+    }
+
+    /** @returns {Promise<void>} */
+    async _fetch() {
+        try {
+            const resp = await fetch("summary.json");
```
jelly: It seems that we get an ETag from the server so we could optimise this but a [conditional request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Conditional_requests) based on either last modified or ETag.

jelly (changes requested):
Partially reviewed this huge PR, requesting changes as before landing I want to discuss the dashboard (and make sure all commands in the readme / code work)

Another blocker before landing is that several people have tried to deploy the new infra, I believe @tomasmatus has but it would be good if two others also did it.

lib/aws/account.py
allisonkarlitskaya: There's some confusion about ~/.aws/ vs ~/.config/aws/ that might be contributing to that...

saml-login
allisonkarlitskaya: the version on `main` doesn't currently support the output of credentials.  you need to use the ec3-thingy version for now...

if you have ec3-thingy checked out in your `main` directory then maybe it's another bug.  does it output nothing at all when you run it directly?

lib/html/dashboard/dashboard.html
allisonkarlitskaya: blame it on the l-l-l-l-l-l-lm

lib/html/dashboard/dashboard.js
allisonkarlitskaya: log.html isn't exactly "vanilla JS".  it starts by introducing its own hacked up version of the lit html´´` template....

also, as mentioned during the meeting: I have an outstanding PR for doing the log.html work too.

lib/html/dashboard/dashboard.js
allisonkarlitskaya: true!

lib/html/dashboard/dashboard.js
jelly: > log.html isn't exactly "vanilla JS". it starts by introducing its own hacked up version of the lit html´´` template....

Yes, and honestly it sucks and does  a terrible job.

lib/html/dashboard/dashboard.html
jelly: A human made this pull request :)

saml-login
jelly: Issue was that I don't have access to the admin account, but using the poweruser account works so that sounds worth documenting.

lib/html/dashboard/dashboard.js
mvollmer: log.html used to use Mustache until quite recently.

jelly: Indeed, if we consolidate everything to lit that sounds more reasonable then having 3 different JavaScript stacks.

allisonkarlitskaya:
In this round:

  - the cockpit-ci-infractl IAM user can now write all the dispatcher SSM parameters
  - SSM secrets moved to a /secrets/ subpath, separating them from config parameters (since it was getting a bit messy in there)
  - `dispatcher set max-jobs`, `dispatcher set max-awaiting-logs`, and (particularly useful) `dispatcher set secrets` added to allow updating those things without requiring admin access for a full `sync`
  - `dispatcher console` command added to infractl
  - infractl dispatch refactored from match/case to `set_defaults(func=...)` per subparser as recommended in https://docs.python.org/3/library/argparse.html#subcommands (search "particularly effective way of handling subcommands")
  - `README.md` updated to document the secrets subpath and new dispatcher set commands

lib/html/dashboard/dashboard.js
allisonkarlitskaya: done, thanks for the suggestion.

allisonkarlitskaya:
This round: added the etag thing suggested from @jelly

lib/aws/README.md
```diff
@@ -0,0 +1,332 @@
...
+package can also be very helpful (for the `aws` CLI tool).
+
+If you don't already have a `~/.aws/config`, here's a reasonable starting point:
+
+```
+[default]
+region = us-east-1
+```
+
+The credentials can be found in the `cockpit-infra-accounts` Bitwarden group.
```
mvollmer, line 100: This is the "edit" link to the group, which doesn't work for me. This should be the normal "read" link, https://rover.redhat.com/groups/group/cockpit-infra-accounts

mvollmer (changes requested):

jelly:
```
[jelle@carbon][~/projects/cockpit-bots]%AWS_PROFILE=cockpit-ci python3 -m lib.aws.launch_runner '{"repo": "cockpit-project/subscription-manager-cockpit", "slug": "jelle-subman-c-rhel-10", "env": {"TEST_OS": "centos-10" }}'
INFO:botocore.credentials:Found credentials in shared credentials file: ~/.aws/credentials
  File "<frozen runpy>", line 203, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/jelle/projects/cockpit-bots/lib/aws/launch_runner.py", line 183, in <module>
    main()
    ~~~~^^
  File "/home/jelle/projects/cockpit-bots/lib/aws/launch_runner.py", line 149, in main
    instance_id = prepare_and_launch(
        ec2,
    ...<7 lines>...
        ssh_keys=ssh_keys,
    )
  File "/home/jelle/projects/cockpit-bots/lib/aws/dispatcher.py", line 101, in prepare_and_launch
    job_config=job_runner_config(
               ~~~~~~~~~~~~~~~~~^
        slug,
        ^^^^^
    ...<4 lines>...
        credential_duration=credential_duration,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ),
    ^
  File "/home/jelle/projects/cockpit-bots/lib/aws/jobconfig.py", line 148, in job_runner_config
    "token": params["secrets/github-token"],
             ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'secrets/github-token'
```

jelly:
> ```
> [jelle@carbon][~/projects/cockpit-bots]%AWS_PROFILE=cockpit-ci python3 -m lib.aws.launch_runner '{"repo": "cockpit-project/subscription-manager-cockpit", "slug": "jelle-subman-c-rhel-10", "env": {"TEST_OS": "centos-10" }}'
> INFO:botocore.credentials:Found credentials in shared credentials file: ~/.aws/credentials
> Traceback (most recent call last):
>   File "<frozen runpy>", line 203, in _run_module_as_main
>   File "<frozen runpy>", line 88, in _run_code
>   File "/home/jelle/projects/cockpit-bots/lib/aws/launch_runner.py", line 183, in <module>
>     main()
>     ~~~~^^
>   File "/home/jelle/projects/cockpit-bots/lib/aws/launch_runner.py", line 149, in main
>     instance_id = prepare_and_launch(
>         ec2,
>     ...<7 lines>...
>         ssh_keys=ssh_keys,
>     )
>   File "/home/jelle/projects/cockpit-bots/lib/aws/dispatcher.py", line 101, in prepare_and_launch
>     job_config=job_runner_config(
>                ~~~~~~~~~~~~~~~~~^
>         slug,
>         ^^^^^
>     ...<4 lines>...
>         credential_duration=credential_duration,
>         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     ),
>     ^
>   File "/home/jelle/projects/cockpit-bots/lib/aws/jobconfig.py", line 148, in job_runner_config
>     "token": params["secrets/github-token"],
>              ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
> KeyError: 'secrets/github-token'
> ```

Resolved by uploading new secrets.

lib/aws/account.py
allisonkarlitskaya: It's because of an oddity in the AWS API: us-east-1 was the original location and used to be the only place you could create an S3 bucket, so the official way to create a bucket there is to provide no location constraint.  Our type annotations actually caught this bug because if you say "us-east-1" is fails (so we have a check for that elsewhere).  We might rather write `| None` here with the understanding that `None` means `us-east-1` but it's confusing and also inconsistent with other APIs which *do* accept `us-east-1` as a region.

allisonkarlitskaya:
> Resolved by uploading new secrets.

Thanks for testing that and making sure it works for non-admins.

lib/aws/README.md
allisonkarlitskaya: Fixed, thanks.

allisonkarlitskaya:
I pushed the latest set of changes as a fixup to make it easier to interdiff.  We'll have to autosquash before merging.

lib/stubs/pika/adapters/asyncio_connection.pyi
```diff
@@ -0,0 +1,18 @@
+from collections.abc import Callable
```
jelly, line 1: What about https://pypi.org/project/types-pika-ts/ ?

lib/aio/amqp.py
```diff
@@ -0,0 +1,161 @@
...
+        def memfd(name: str) -> str:
+            fd = os.memfd_create(name)
+            stack.callback(os.close, fd)
+            os.write(fd, credentials[name].encode())
+            return f'/proc/self/fd/{fd}'
+
+        context = ssl.create_default_context(cafile=memfd('secrets/ca.pem'))
+        context.load_cert_chain(
+            keyfile=memfd('secrets/amqp-client.key'),
+            certfile=memfd('secrets/amqp-client.pem'),
```
jelly, line 43: Sneaky :-)

lib/aio/amqp.py
```diff
@@ -0,0 +1,161 @@
...
+
+    return pika.ConnectionParameters(
+        host,
+        int(port),
+        ssl_options=pika.SSLOptions(context, server_hostname=host),
+        credentials=pika.credentials.ExternalCredentials(),
+    )
+
+
+class Queue:
```
jelly, line 55: This is potentially confusing? There is a builtin [Queue](https://docs.python.org/3/library/queue.html#module-queue)

lib/aio/amqp.py
```diff
@@ -0,0 +1,161 @@
...
+            self.close(reason)
+
+        def on_channel_opened(channel: Channel) -> None:
+            logger.debug('AMQP channel opened')
+            self._channel = channel
+            channel.add_on_close_callback(on_channel_closed)
+            channel.basic_qos(prefetch_count=1, global_qos=True)
+            for queue in self._queues:
+                logger.debug('declaring queue %r', queue)
+                channel.queue_declare(queue, durable=True, arguments={"x-max-priority": 9})
```
jelly, line 81: That's `MAX_PRIORITY` no? (lib/distributed_queue.py)

jelly: This should be temporary, we want the `/devel` scenario to be queued "higher" (currently it is `MAX_PRIORITY`)

lib/aws/README.md
```diff
@@ -0,0 +1,334 @@
...
+regulated by [a separate Rover
+group](https://rover.redhat.com/groups/group/it-cloud-aws-727920394381-admin).
+Contact the admins of that group if you need access.
+
+The easiest way to get this working is by configuring an `admin` profile in
+`~/.aws/credentials`, something like so:
+
+```
+[admin]
+credential_process = /home/lis/src/bots/main/saml-login --output=awscreds 727920394381-admin
```
mvollmer, line 127: I would like to use this also for the low privileged "arn:aws:iam::727920394381:user/cockpit-ci" user so that I don't have to store the creds permanently on my laptop.

saml-login
```diff
@@ -11,53 +11,99 @@ for use by image-download.
...
+                'SecretAccessKey': key.secret,
+                'SessionToken': key.token,
+            }))
+
+
 def main() -> None:
     parser = argparse.ArgumentParser(description='Log in to AWS via SAML using a web browser')
+    # fmt: off
     parser.add_argument('-d', '--debug', action='store_true', help='Enable debug logging')
+    parser.add_argument('role', nargs='?', default=REDHAT_SSO_IMAGE_DOWNLOAD_ROLE,
```
mvollmer, line 63: There is help message somewhere "Are you in the correct Rover group for cockpit-ci-images access?"  I think this message should mention the correct Rover group for the role, if possible.

jelly:
Some follow ups:

* Document how we can test changes for the CoreOS configuration
* Diagram with the workflow
* Naming of the Queue variable shadowing Python internals
* Adjusting the help message for `saml-login` -> Are you in the correct Rover group for cockpit-ci-images access?" I think this message should mention the correct Rover group for the role
* Discuss the btrfs container usage, its non-standard so I would like to know if we have a good reason to not use overlayfs? (reduce all possible moving parts)
* Systemd unit is not de-dentend but weirdly indentend on the dispatcher (same for the shell script)
* We should not make the dispatcher dependent on $random sha on Github and use proper GitHub releases for it
* No alerting at all when the dispatcher is broken, this didn't exists in our previous setup but we should do better

lib/aws/ec2.py
jelly: This is for running containers unprivileged, alternatively we can do this via usermod

```
usermod --add-subuids 100000-165535 --add-subgids 100000-165535 containers
```

lib/aws/account.py
```diff
@@ -0,0 +1,102 @@
...
+    'cockpit-ci-images': 'us-east-1',
+    'cockpit-ci-images-fra': 'eu-central-1',
+}
+
+# TAGS
+# Read corporate policy before changing anything here!
+# https://source.redhat.com/departments/products_and_global_engineering/red_hat_public_cloud_services/public_cloud_services_wiki/resource_tagging_policy
+# https://source.redhat.com/departments/products_and_global_engineering/red_hat_public_cloud_services/public_cloud_services_wiki/resource_tagging_names
+MANDATORY_TAGS: Mapping[str, str] = {
+    'app-code': 'ARR-001',
```
jelly: It turns out we should not use this but our own CMDB, I have submitted an application for that.

lib/aws/account.py
mvollmer: Thanks!

lib/aws/README.md
mvollmer: This can be a follow up if we agree it would be a useful feature.

lib/aio/amqp.py
jelly: Ok turns out, that `run-queue` effectively gets messages from the `webhook`.

lib/aws/jobconfig.py
```diff
@@ -0,0 +1,182 @@
...
+    if "image-download" in secrets:
+        yield (
+            "image-download",
+            str(assume_role(sts, IMAGE_DOWNLOAD_ROLE, duration=duration)),
+        )
+
+    if "github-token" in secrets:
+        yield "github-token", params["secrets/github-token"]
+
+    if "fedora-wiki" in secrets:
```
jelly, line 78: This is wrong, it should be `fedora-wiki.json` as it is in ci-secrets.

lib/aws/infra_definitions.py
```diff
@@ -0,0 +1,535 @@
...
+            ),
+            allow(
+                "autoscaling:SetDesiredCapacity",
+                f"arn:aws:autoscaling:{CI_RUNNER_REGION}:{ACCOUNT_ID}:autoScalingGroup:*:autoScalingGroupName/{DISPATCHER_ASG}",
+            ),
+            allow(
+                "ssm:PutParameter",
+                [
+                    f"arn:aws:ssm:{CI_RUNNER_REGION}:{ACCOUNT_ID}:parameter/cockpit-ci/dispatcher-url",
+                    f"arn:aws:ssm:{CI_RUNNER_REGION}:{ACCOUNT_ID}:parameter{DISPATCHER_PARAMS}/*",
```
jelly, line 280: Missing `ampq-server`

jelly:
All of the rebasing and squashing makes it impossible to determine what is running right now:

```
aws ssm get-parameter --name /cockpit-ci/dispatcher/runner-url --query Parameter.Value --output text
https://github.com/cockpit-project/bots/archive/210b1b6afb090d32d1a02ba441ca3e39193475b3.tar.gz
```

[allisonkarlitskaya](https://github.com/allisonkarlitskaya) [force-pushed](https://github.com/cockpit-project/bots/compare/31ca448f41d846b0c3eb72c253fc64474e5b2167..210b1b6afb090d32d1a02ba441ca3e39193475b3) the ec3-thingy branch 2 times, most recently from [31ca448](https://github.com/cockpit-project/bots/commit/31ca448f41d846b0c3eb72c253fc64474e5b2167) to [210b1b6](https://github.com/cockpit-project/bots/commit/210b1b6afb090d32d1a02ba441ca3e39193475b3) [last week](https://github.com/cockpit-project/bots/pull/9316#event-29315125289)

So we are currently running an older version of this pull request.

lib/aws/launch_runner.py
```diff
@@ -0,0 +1,183 @@
...
+    ec2.terminate_instances(InstanceIds=[instance_id])
+    watch_instance(ec2, instance_id)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    # fmt: off
+    parser.add_argument("--debug", action="store_true", help="Enable debug logging")
+    parser.add_argument("--ami", help="FCOS AMI ID (default: latest)")
+    parser.add_argument("--instance-type", default="m8id.4xlarge")
```
jelly, line 100: This should not be hardcoded, you sold us on a deploy script with all AWS definitions in once place, so that should be there too.

lib/aws/dispatcher.py
```diff
@@ -0,0 +1,420 @@
...
+            try:
+                job.launched_instance = await loop.run_in_executor(
+                    None,
+                    lambda: prepare_and_launch(
+                        ec2,
+                        sts,
+                        job=job_json,
+                        params=self.params,
+                        bots_url=self.params["runner-url"],
+                        instance_type="m8id.4xlarge",
```
jelly, line 222: Same comment as `launch_runner` make this a constant so we can easily change it (and find it)

jelly (approved):

mvollmer (approved):
Thanks!

PR #9335, opened 2026-08-12
images: drop ubuntu-2404 and cleanup leftovers from ubuntu-2204

tomasmatus:


images/scripts/debian.setup
```diff
@@ -115,13 +115,6 @@ if [ "$IMAGE" = "ubuntu-2604" ] ; then
...
 
-if [ "$IMAGE" = "ubuntu-2404" ]; then
-    # introduced in util-linux 2.40.1-5
-    TEST_PACKAGES="${TEST_PACKAGES/lastlog2/}"
-    # renamed at some point
-    TEST_PACKAGES="${TEST_PACKAGES/sos/sosreport}"
-fi
-
 # apt-ftparchive moved to apt-utils in ubuntu-2510
 if [ "${ID}" = "ubuntu" ] && [ "${VERSION_ID}" != "22.04" ]; then
```
jelly, line 119: another 22.04 leftover hihi, the comment seems wrong?

jelly (approved):

images/scripts/debian.setup
tomasmatus: aha! Hope there isn't more... Because I only grepped for "ubuntu-2404" and didnt realize we might also work based on `${VERSION_ID}`

PR #9340, opened 2026-08-13
logs.html → {js,html} split + cleanup + lit + tsc + biome

allisonkarlitskaya:
This simplifies the code substantially, fixes several bugs, and makes this code a *lot* easier to work on going forward.


allisonkarlitskaya:
Predicted feedback from @Venefilyn: use an import map and integrity :)

allisonkarlitskaya:
@Venefilyn should we also give the PF CSS the integrity treatment?  In both cases this means we'd need to pin to a specific version...

Venefilyn:
> Predicted feedback from @Venefilyn: use an import map and integrity :)

Yes!

> @Venefilyn should we also give the PF CSS the integrity treatment? In both cases this means we'd need to pin to a specific version...

We don't need integrity treatment for CSS, but we'd need to pin the version

lib/html/s3streamer/log.js
```diff
@@ -0,0 +1,671 @@
...
+                    <input class="pf-v6-c-switch__input" type="checkbox" role="switch"
+                           .checked=${has_failed ? this.show_only_failed : false}
+                           @change=${(/** @type {Event} */ e) => {
+                               this.show_only_failed = /** @type {HTMLInputElement} */ (e.target).checked;
+                           }}>
+                    <span class="pf-v6-c-switch__toggle"></span>
+                    <span class="pf-v6-c-switch__label" aria-hidden="true">Show only failed</span>
+                </label>
+            </div>`;
+    }
```
Venefilyn, line 512: This doesn't have animation when switching, maybe rerendered each time? Only happens on Firefox

https://github.com/user-attachments/assets/6e6da837-02f0-419e-91d3-16c4f3c0c6e9

lib/html/types/lit.d.ts
Venefilyn, line 1: Why do we need `lit.d.ts`?

HACKING.md
```diff
@@ -69,20 +69,32 @@ pushing PRs that will fail on trivial errors:
...
-* Make changes in `log.html` and see changes refresh live in the browser
+python -m http.server -d ./lib/html/s3streamer
+```
+
+If you need to test the streaming code itself, it's also possible to run CI
+jobs locally using `job-runner run cockpit-project/starter-kit` or something
+similar.  In the default configuration, `job-runner` will write to the local
+disk instead of uploading to S3, but the s3streamer algorithm is still used and
+live updates should work in the same way.  In that case you should run the
+webserver like this:
```
Venefilyn, line 91: I would like to make it possible to use logs from an existing run. That way you can compare with live code. I could do that with current main by just modifying `log.html`

lib/html/s3streamer/log.js
```diff
@@ -0,0 +1,671 @@
...
+                        </div>
+                    </div>
+                </div>
+            </li>`;
+    }
+}
+
+customElements.define("test-entry", TestEntry);
+
+class LogViewer extends LitElement {
```
Venefilyn, line 329: This is the main component but in the middle of the file, could we move to the bottom of the file

lib/html/s3streamer/log.js
```diff
@@ -0,0 +1,671 @@
...
+    static properties = {
+        href: { type: String },
+        content: { state: true },
+        raw: { state: true },
+        show_only_failed: { state: true },
+    };
```
Venefilyn: Would love comments

Venefilyn (commented):
Overall missing comments throughout. 

I kinda like having raw logs available like we have in main atm, it is just a lot faster to scroll through etc. but seems fine here so sure lets go with it - means it's faster to switch between

Nitpicks about Lit in general. Performance wise it is noticeably slower, I don't know how Lit works in regards to re-rendering and performance optimizations or how Web Components handle performance when re-rendering. Maybe there is something we're missing?

I checked performance recording for main vs this PR and there is a lot more happening. When going from full list to smaller (only failures) it is going from 20ms to 50ms. Going from only failures to full list goes from 130ms to 190ms. 

---

These `test/verify/check-selinux TestSelinuxEnforcing.test` run green 3 times, just retried as the test is touched. So looks weird to show that - we shouldn't show it IMO and keep it to how we had it before. See how we [have it live atm](https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/pull-23617-ee41aa55-20260813-123950-fedora-44-bots%239331/log.html) 
 
Image

---

I'm missing the reason for the retries, they're missing completely

Image

HACKING.md
allisonkarlitskaya: I tried to follow your instructions and it failed with a CORS message in the debugger...

lib/html/tsconfig.json
```diff
@@ -3,6 +3,7 @@
         "checkJs": true,
         "strict": true,
         "noEmit": true,
+        "incremental": true,
```
jelly: speed what up? And do we want this in cockpit?

Venefilyn:
Re-request review when my previous review is addressed and I'll take a look :)

PR #9386, opened 2026-09-01
.github/workflows: disable codeql scanning

allisonkarlitskaya:
This produces almost entirely false positives.


mvollmer (commented):
I  would like a (brief) discussion in the team about this. I personally don't remember getting any value from codeql that we wouldn't also get from test/common/static-code.

I believe that is is not sufficiently good for actually detecting vulnerabilities. The various AI scanners that bombard us now are doing a much better job.

allisonkarlitskaya:
I've (nearly) never seen it report something useful.  Particularly now that I ask Claude to take a look at my PRs before submitting, the false-positive rate is now like 100%...

allisonkarlitskaya:
More silliness: https://github.com/cockpit-project/bots/pull/9394#discussion_r3922163784

jelly (approved):

tomasmatus (approved):

PR #9388, opened 2026-09-01
infractl: add 'dispatcher set amqp-server'

allisonkarlitskaya:
This allows live switching of the amqp-server parameter used by the dispatcher.

This is supported via new update_amqp_server() function in `infra_definitions.py` which is what all of the existing `set` verbs have.

While we're at it, update `sync_ssm()` to use the `update_*` functions for everything instead of duplicating that logic.


jelly (approved):
Thanks!

PR #9394, opened 2026-09-03
Move `test/verify/check-bots-api` from the `cockpit` repository

allisonkarlitskaya:


test/test_machines_api.py
```diff
@@ -0,0 +1,120 @@
...
+from pathlib import Path
+
+import pytest
+
+from lib.constants import BOTS_DIR, MACHINE_DIR
+from machine import testvm
+
+
+@pytest.fixture
+def test_os() -> str:
```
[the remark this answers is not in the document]
allisonkarlitskaya: ... the return value of `pytest.skip` is literally `NoReturn`.  `ruff` and `mypy` are both perfectly happy with that, as am I.

lib/testmap.py
```diff
@@ -17,7 +17,7 @@
 
 import fnmatch
 import itertools
-import os.path
+import os
```
jelly, line 21: hihi ruff?

lib/testmap.py
```diff
@@ -336,9 +336,6 @@ def projects() -> Iterable[str]:
 
 
 def get_default_branch(repo: str) -> str:
-    branches = tests_for_project(repo)
```
jelly, line 338: This feels very opportunistic and breaking things for no reason. What unknown 3rd party project do we break with this?

.github/workflows/machines-api.yml
```diff
@@ -0,0 +1,60 @@
...
+            libvirt-daemon-system \
+            python3-libvirt \
+            python3-pytest \
+            python3-pytest-asyncio \
+            python3-pytest-cov \
+            qemu-kvm
+
+      # virtqemud spawns QEMU processes as a non-root user which needs /dev/kvm access
+      - name: Enable KVM access for QEMU
+        run: sudo chmod 666 /dev/kvm
```
jelly, line 51: I saw a lot of other projects use:

```
      - name: Enable KVM group perms
        run: |
            echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
            sudo udevadm control --reload-rules
            sudo udevadm trigger --name-match=kvm
```

allisonkarlitskaya: I guess this could be interesting?  I'd prefer to keep it simple, personally.

lib/testmap.py
allisonkarlitskaya: `tests_for_project()` only works when called on our own testmap, which no longer contains any project with "master" on it, so this is well and truly dead code.  The only possible way you could hit this at all is via the undocumented/internal `COCKPIT_TESTMAP_INJECT` envvar which is only used from the cockpituous repository with the main branch.

lib/testmap.py
allisonkarlitskaya: srsly.  I actually had my editor showing me about this as a warning, so now I'm wondering where that warning was coming from...

allisonkarlitskaya: pyright

lib/testmap.py
allisonkarlitskaya: The main reason for doing this is that it enables me to write 'main' as the default in a couple of places.

lib/testmap.py
```diff
@@ -336,9 +336,6 @@ def projects() -> Iterable[str]:
...
-    if 'master' in branches:
-        return 'master'
```
jelly, line 341: What about the 3rd party projects, anaconda webui uses main, but there might be others?

lib/testmap.py
```diff
@@ -252,11 +253,10 @@ def product(image: str, *scenarios: Iterable[str], repo: str | None = None) -> S
         *product(TEST_OS_DEFAULT, ['firefox'], COCKPIT_SCENARIOS, repo='cockpit-project/cockpit'),
         *product('ubuntu-stable', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit'),
         *product('debian-trixie', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit'),
-        *product('rhel-9-8', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit'),
-        "rhel-8-10@cockpit-project/cockpit/rhel-8",
+        *product('rhel-9-8', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit/rhel-9.8'),
+        *product('rhel-8-10', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit/rhel-8'),
         "rhel-8-10@candlepin/subscription-manager/subscription-manager-1.28",
         "rhel-9-8@candlepin/subscription-manager/subscription-manager-1.29",
-        "rhel-10-2@cockpit-project/subscription-manager-cockpit",
```
jelly, line 259: This commit does  two things.. cleans up `IMAGE_REFRESH_TRIGGERS` by dropping subscription-manager-cockpit as it uses the candlepin container on itself now and something else.

lib/testmap.py
```diff
@@ -252,11 +253,10 @@ def product(image: str, *scenarios: Iterable[str], repo: str | None = None) -> S
         *product(TEST_OS_DEFAULT, ['firefox'], COCKPIT_SCENARIOS, repo='cockpit-project/cockpit'),
         *product('ubuntu-stable', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit'),
         *product('debian-trixie', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit'),
-        *product('rhel-9-8', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit'),
-        "rhel-8-10@cockpit-project/cockpit/rhel-8",
+        *product('rhel-9-8', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit/rhel-9.8'),
+        *product('rhel-8-10', COCKPIT_SCENARIOS, repo='cockpit-project/cockpit/rhel-8'),
```
jelly, line 257: I wondered if rhel-8 did not support scenarios but it does.

.github/workflows/machines-api.yml
```diff
@@ -0,0 +1,60 @@
...
+              python3 -m test.affected_images |
+              tee -a "$GITHUB_OUTPUT"
+
+  machines-api:
+    needs: query-affected
+    if: needs.query-affected.outputs.affected_images != '[]'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      pull-requests: none
```
jelly, line 29: is that better then `permissions: {}`? Just curious

jelly (commented):
One nitpick, otherwise good

.github/workflows/machines-api.yml
allisonkarlitskaya: vibes

lib/testmap.py
allisonkarlitskaya: There's really not.  See the above comment: tests_for_project() only works for things that are actually in the testmap and nothing in our testmap has a master branch anymore.

jelly (approved):

PR #9395, opened 2026-09-03
Fedora update automation

jelly:


jelly:
https://github.com/cockpit-project/bots/pull/9396

PR #9397, opened 2026-09-03
cockpit-container-update: Script to update Containerfile automatically

jelly:
Automate the updating of the Fedora containers we use in CI (cockpit-tasks) and cockpit-ws to the latest stable Fedora release.

This script can be used in our GitHub workflows like `cockpit-lib-update` and gets rid of manual often forgotten tasks.

---

Depends on a tasks container update https://github.com/cockpit-project/cockpituous/pull/715

Tested on both cockpituous and cockpit main. Would either run weekly, every two weeks or monthly. Whatever is enough to make us not forget.

```
cockpit-bots/cockpit-container-update --dry-run --containerfile tasks/container/Containerfile
```

### DIFF
--- a/cockpit-container-update
+++ b/cockpit-container-update
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+
+# Copyright (C) 2026 Red Hat, Inc.
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# Update the cockpit-ws and cockpit-tasks container to the latest stable Fedora
+# automatically.
+
+import argparse
+import pathlib
+import re
+
+import fedora_distro_aliases
+
+from lib import git
+from lib.github import GitHub
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Update Containerfile to latest stable Fedora")
+    parser.add_argument("-n", "--dry-run", action="store_true", help="Dry run")
+    parser.add_argument('containerfile', type=pathlib.Path)
+    args = parser.parse_args()
+
+    aliases = fedora_distro_aliases.get_distro_aliases()
+    latest_stable = aliases["fedora-latest-stable"][0].major_version
+    print(f"Latest stable Fedora: {latest_stable}")
+
+    content = args.container_file.read_text()
+    match = re.search(r"^(FROM\s+.*fedora:)(\d+)", content, re.MULTILINE)
+    if not match:
+        raise RuntimeError(f"Could not find Fedora version in {args.containerfile}")
+
+    current = int(match.group(2))
+    print(f"Current Containerfile uses: fedora:{current}")
+
+    if current == latest_stable:
+        print("Already up to date")
+        return
+
+    content = re.sub(r"(FROM\s+.*fedora:)\d+", rf"\g<1>{latest_stable}", content)
+    args.container_file.write_text(content)
+
+    title = f"{args.containerfile.parent.name}: Update container base image to Fedora {latest_stable}"
+    body = f"Fedora {latest_stable} is now the latest stable release."
+
+    git.add(args.containerfile)
+    git.commit(f"{title}\n\n{body}")
+
+    github = GitHub()
+    branch = git.push(github, "cockpit-container-update", dry_run=args.dry_run)
+    github.create_pull_request(branch, title=title, body=body, dry_run=args.dry_run)
+    git.amend_and_forcepush(github, branch, dry_run=args.dry_run)
+
+
+if __name__ == '__main__':
+    main()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,7 @@ module = [
     'boto3',
     'botocore',
     'botocore.*',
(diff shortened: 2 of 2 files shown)


cockpit-container-update
```diff
@@ -0,0 +1,62 @@
...
+        raise RuntimeError(f"Could not find Fedora version in {args.containerfile}")
+
+    current = int(match.group(2))
+    print(f"Current Containerfile uses: fedora:{current}")
+
+    if current == latest_stable:
+        print("Already up to date")
+        return
+
+    content = re.sub(r"(FROM\s+.*fedora:)\d+", rf"\g<1>{latest_stable}", content)
```
allisonkarlitskaya, line 41: I've got `\g<9>9` problems but this ain't one.

cockpit-container-update
```diff
@@ -0,0 +1,62 @@
...
+
+import fedora_distro_aliases
+
+from lib import git
+from lib.github import GitHub
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Update Containerfile to latest stable Fedora")
+    parser.add_argument("-c", "--containerfile", help="Path to Containerfile", required=True)
```
allisonkarlitskaya: seems like this is a positional... maybe a multi-value positional, even.

allisonkarlitskaya: also, `type=Path`?

allisonkarlitskaya (commented):
Seems fine, but I guess I'd like to see it being used first?  Do you have PRs for what the workflow looks like?

jelly:
> Seems fine, but I guess I'd like to see it being used first? Do you have PRs for what the workflow looks like?

Not yet, would basically be the same as `cockpit-lib-update`

cockpit-container-update
jelly: Fair, can do.

jelly:
> Seems fine, but I guess I'd like to see it being used first? Do you have PRs for what the workflow looks like?

Needs a new tasks container, but see https://github.com/cockpit-project/cockpit/pull/23679

Would also live in cockpituous to update cockpit-tasks to a new fedora

allisonkarlitskaya:
> Would also live in cockpituous to update cockpit-tasks to a new fedora

Can you clarify what you mean by this?  We keep a second copy there, or we check out bots there for a workflow hosted in that repository?

jelly:
> > Would also live in cockpituous to update cockpit-tasks to a new fedora
> 
> Can you clarify what you mean by this? We keep a second copy there, or we check out bots there for a workflow hosted in that repository?

Yes, checkout bots there and run it weekly on a workflow

jelly:
@allisonkarlitskaya I worked on this again, so see my own forks [test run](https://github.com/jelly/cockpit/actions/runs/35359076579/job/105645554335) of the [Cockpit workflow PR](https://github.com/cockpit-project/cockpit/pull/23679). And it successfully made [a new PR](https://github.com/jelly/cockpit/pull/245) to update the container!

pyproject.toml
```diff
@@ -19,6 +19,7 @@ module = [
     'boto3',
     'botocore',
     'botocore.*',
+    'fedora_distro_aliases',
```
allisonkarlitskaya: https://github.com/cockpit-project/cockpituous/pull/715 means we can drop the ignore, right?

cockpit-container-update
```diff
@@ -0,0 +1,57 @@
...
+    title = f"{args.containerfile.parent.name}: Update container base image to Fedora {latest_stable}"
+    body = f"Fedora {latest_stable} is now the latest stable release."
+
+    git.add(args.containerfile)
+    git.commit(f"{title}\n\n{body}")
+
+    github = GitHub()
+    branch = git.push(github, "cockpit-container-update", dry_run=args.dry_run)
+    github.create_pull_request(branch, title=title, body=body, dry_run=args.dry_run)
+    git.amend_and_forcepush(github, branch, dry_run=args.dry_run)
```
allisonkarlitskaya: This is because the PR gets opened by the github-actions user, presumably.

cockpit-container-update
```diff
@@ -0,0 +1,57 @@
...
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Update Containerfile to latest stable Fedora")
+    parser.add_argument("-n", "--dry-run", action="store_true", help="Dry run")
+    parser.add_argument('containerfile', type=pathlib.Path)
+    args = parser.parse_args()
+
+    aliases = fedora_distro_aliases.get_distro_aliases()
+    latest_stable = aliases["fedora-latest-stable"][0].major_version
```
allisonkarlitskaya: So the idea here is that the package will be regularly updated by Fedora?  And the check is offline?  So we're not getting the latest Fedora version but rather the latest version that our tasks container knows about, yes?

pyproject.toml
jelly: Not when you work locally and dont have it installed.

PR #9398, opened 2026-09-03
workflows: Auto-add new issues to GH project

Venefilyn:
Adds a GitHub Project workflow which auto-adds issues and adds bot
specific PRs.

- One workflow as we're gonna be copying this around a lot, so we
   may as well make it a bit easier.  Keeps things as DRY as they can
   be, given the circumstances.
- Use the `gh-project` environment (there has never been a global
  `GH_PROJECTS` secret and we don't want to add one)
- Use `pull_request_target`: we don't get access to the environment (and
  secret) otherwise.  This is generally considered bad practice but it's
  pretty safe here since we never check out the PR's code.
- Add a comment about the 3 other approaches we tried/considered

### DIFF
--- a/.github/workflows/gh-project.yml
+++ b/.github/workflows/gh-project.yml
@@ -1,0 +1,39 @@
+# GitHub Projects lets you import from multiple repositories on creation, but the
+# built-in "Auto-add to project" workflow only supports a single repository (without
+# paying for more). This workflow replicates that functionality across all of our
+# repositories.
+#
+# We also considered:
+#   - a single workflow in a single repository (eg: bots) that checks all of
+#     the repositories periodically: this wouldn't be realtime, and it would
+#     be a lot of wasted action runs, 99% of which would do nothing at all
+#   - driving a small Python helper from our webhook: this would be a nice
+#     option, but most of our "GitHub administrativia" tasks are done via GitHub
+#     workflows, so let's keep with the trend
+
+name: Add issues and bot PRs to Cockpit GitHub Project
+
+on:
+  issues:
+    types:
+      - opened
+  pull_request_target:
+    types:
+      - labeled
+
+jobs:
+  add-to-project:
+    if: >-
+      github.event_name == 'issues' ||
+      (
+        github.event_name == 'pull_request_target' &&
+        contains(github.event.pull_request.labels.*.name, 'bot')
+      )
+    name: Add to project
+    environment: gh-project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v2
+        with:
+          project-url: https://github.com/orgs/cockpit-project/projects/10
+          github-token: ${{ secrets.GH_PROJECT }}


.github/workflows/gh-project-add-prs.yml
```diff
@@ -0,0 +1,20 @@
...
+  pull_request:
+    types:
+      # We only need to add to a project
+      - labeled
+
+jobs:
+  # Needs repo and project scopes in a PAT token in the org
+  # https://github.com/actions/add-to-project#creating-a-pat-and-adding-it-to-your-repository
+  add-to-project:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'bot') }}
```
allisonkarlitskaya: Why just bot?  I thought the idea was to add everything and use labels to filter...

allisonkarlitskaya (changes requested):
Can we do this the other way around as an automation on the project side?  I've heard something about limits on free accounts but if that's the case it would be nice to see that spelled out explicitly.

jelly:
> Can we do this the other way around as an automation on the project side? I've heard something about limits on free accounts but if that's the case it would be nice to see that spelled out explicitly.

So this landed in `cockpit` already but another concern might be having to setup a token manually per project for this.

Venefilyn:
> Can we do this the other way around as an automation on the project side? I've heard something about limits on free accounts but if that's the case it would be nice to see that spelled out explicitly.

Project have limitations on the amount of project-workflows it can have. It also doesn't handle multiple repos in a workflow so you'd need to duplicate it. AFAICT this is why GH Project that has all issues was never done in Cockpit before as it didn't look feasible. But with each project adding themselves that works fine

> So this landed in `cockpit` already but another concern might be having to setup a token manually per project for this.

We can setup a secret globally in the org can't we? I don't have perms for it though
https://github.com/organizations/cockpit-project/settings/secrets/actions

.github/workflows/gh-project-add-prs.yml
Venefilyn: Both myself and Aleksandra want to avoid PRs in the GH Project. I added this to the [project README](https://github.com/orgs/cockpit-project/projects/10/?pane=info):

> ## Stance on PRs in the board
> 
> This project should primarily be a way to track work and help us realize how things are going. PRs aren't banned from being added to this project but should only be done for PRs that are very menial or bots related tasks (such as dependency updates or cockpit container bumps). 
> 
> If we were to track PRs from code contributions that we have written we run into a few issues. 
> 
> - We could split the PR into multiple PRs if it becomes too big, yet they all track towards a main goal and need all to be merged before the goal is complete - having an issue here is a must.
> - We could close a PR as there was no work done to it in a while or we just don't have time for it - without an issue we loose tracking of it
> - PRs could be closed in favor of other PRs to solve the same task, if it was not tracked by an issue we'd lose the information surrounding that as well.

allisonkarlitskaya (approved):

jelly (approved):

PR #9405, opened 2026-09-07
various: drop the 'rhel' queue

allisonkarlitskaya:
We no longer send jobs here and nothing reads from it.

Also drop the redhat_network() function because we no longer need it to determine if we should read from the rhel queue or not.


jelly (approved):
Thanks!

PR #9406, opened 2026-09-07
workflows: Auto-add to project from all repos

Venefilyn:
Needs to use API directly instead of GH Actions


Venefilyn:
Closing in favor of https://github.com/cockpit-project/bots/pull/9398

See reasons in said PR

PR #9413, opened 2026-09-09
.github: Introduce dependabot to update workflow actions

jelly:
Our workflows use an old version of actions/checkout and we manually up date them. In all our other projects we use dependabot to keep this up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    open-pull-requests-limit: 3
+    labels:
+      - "bot"
+    schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    open-pull-requests-limit: 3
+    labels:
+      - "no-test"
```
jelly: We can probably drop `no-test`.

allisonkarlitskaya: Do you want to do this and push a new version?

jelly: done

Venefilyn (approved):
LGTM

PR #9416, opened 2026-09-10
job-runner: add timestamps to log output

allisonkarlitskaya:
This prints the UTC time at the start and end of a test run.

### DIFF
--- a/checkout-and-run
+++ b/checkout-and-run
@@ -10,6 +10,7 @@ import subprocess
 import sys
 import time
 from collections.abc import Sequence
+from datetime import datetime, timezone
 
 
 def log(*parts: object) -> None:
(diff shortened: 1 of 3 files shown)


allisonkarlitskaya:
The timestamps are always in UTC with ISO format followed by a "▶" (like `2026-09-10 16:10:22 ▶`) which should make them easy to find and parse.

checkout-and-run
```diff
@@ -82,7 +83,7 @@ def checkout_and_run(repository: str, revision: str | None, rebase: str | None,
 
     for cmd in commands:
         try:
-            log('\n+', shlex.join(cmd))
+            log('\n+ {datetime.now(timezone.utc):%T %F} ▶', shlex.join(cmd))
```
jelly, line 86: Ack, so this logs the total time of `test/run` and substract TESTS PASSED and you get 'setup time'

jelly:
Can we have a test run with `launch_runner` with this change?

Issue #9432, opened 2026-09-17
cockpit-bridge does not recover from user D-Bus connection failure


mvollmer:
As of this writing, the cockpit-machines CI tests on debian-trixie expose a bug in systemd-ctypes that causes cockpit-bridge to crash semi-rarely, but often enough to make test runs with "retries of affected tests" impossible to get green.

This shows up in the journal with a backtrace like:

```
Exception in callback _Transport._read_ready()
handle: <Handle _Transport._read_ready()>
Traceback (most recent call last):
  File "/usr/lib/python3.13/asyncio/events.py", line 89, in _run
    self._context.run(self._callback, *self._args)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/transports.py", line 110, in _read_ready
    self._protocol.data_received(data)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/protocol.py", line 192, in data_received
    result = self.consume_one_frame(self.buffer)
  File "/usr/lib/python3/dist-packages/cockpit/protocol.py", line 142, in consume_one_frame
    self.frame_received(data[start:end])
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/protocol.py", line 96, in frame_received
    self.control_received(data)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/protocol.py", line 106, in control_received
    self.channel_control_received(channel, command, message)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/router.py", line 223, in channel_control_received
    endpoint.do_channel_control(channel, command, message)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cockpit/channel.py", line 166, in do_channel_control
    self.do_control(command, message)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
[...]

PR #9433, opened 2026-09-17
naughty: Add pattern for #9432 to debian-trixie

mvollmer:
The real problem here is not the refused connection, but the fact that cockpit-bridge crashes because of it under specific conditions.  We can't detect the crash from the browser log, unfortunately, so we hook onto the reported connection error.


mvollmer:
Proof that it works: https://cockpit-ci-logs.s3.us-east-1.amazonaws.com/pull-2661-511db09b-20260917-063035-debian-trixie-bots%239433/log.html

tomasmatus (approved):

PR #9435, opened 2026-09-17
testmap: Setup cockpit-files backport branch CI

jelly:


tomasmatus (commented):
Looks good, we'll also need this for cockpit-machines. Would you mind doing it here while you're at it? :)

tomasmatus (commented):
... and cockpit too for rhel-9.9 and 10.3

jelly:
> Looks good, we'll also need this for cockpit-machines. Would you mind doing it here while you're at it? :)

I'll leave that up @mvollmer now just making sure files is ready.

tomasmatus (approved):

PR #9453, opened 2026-09-21
images: drop ubuntu-stable image

### LINKED ISSUE #9447: Drop Ubuntu-stable image
* [x] Drop the ubuntu-stable image from the bots repository
* [x] Drop the image from the testmap
* [ ] Remove ubuntu-stable workarounds in our repositories

jscotka:
Closes #9447


images/scripts/debian.setup
```diff
@@ -160,7 +160,7 @@ EOF
 
 fi
 
-# Attention: Enable proposed for ubuntu-stable after 24.10 release
+# Attention: Enable proposed for ubuntu images after a new release
```
Venefilyn: Wonder if we can delete it altogether. @mvollmer or @jelly do you know what the previous comment meant? It's from 2 years ago by Martin

Venefilyn (commented):
Overall LGTM! Everything that needs changing is done, wondering about one comment though from 2 years ago and what it actually means

images/scripts/debian.setup
jelly: [proposed updates](https://help.ubuntu.com/community/UbuntuUpdates) are like updates-testing for Ubuntu.

Do you propose deleting the whole comment? The commit message in 809886b74 explains it pretty well. Once we want to move ubuntu-stable to 26.10, disable proposed updates so we don't test stuff which is highly likely to break.

jscotka: I've also tried to understand what this does mean and If this is the place what we should/could reword.

Venefilyn: @jelly I saw that commit and I can't really understand it either

> Disable -proposed while this is still in the development phase, as that
are by definition the broken packages.

To me this just seems like we can remove the comment no? As we no longer use "oracular"

jelly: There are two Ubuntu releases LTS (.4) and the .10 release -stable used to track .10 releases. As we wanted to test them early in beta phase we disabled the equivalent of `updates-testing` for Ubuntu to not run into issues with broken updates.

jscotka: so I've removed this line completely, seems as good solution?

Venefilyn (approved):
Thanks!

PR #9454, opened 2026-09-21
Support running CI on embargoed GHSA forks

allisonkarlitskaya:
This is some necessary precondition work but it's not the whole story.  In theory this would support issuing private jobs manually with launch_runner, but we probably want to get some more automation than that.

It's also not currently working: job-runner crashes for some reason that I need to dig into.

The biggest open question here is where to put the logs.  There are two options:
 - in public but with an unguessable and clearly embargoed URL like `embargoed-vUBWKocQ9lNxS8EmB_1hcQ`: this is the easy option because it means we can directly post those URLs into statuses on the GHSA PR
 - in a private bucket or gated by a security policy
   - we would need some sort of a proxy or viewer app to view these because something needs to gain access to the correct IAM role and do the S3 signing


allisonkarlitskaya:
Worth noting: GHSA temporary private forks seem to lack support for commit statuses, so we're going to need to figure that out.  This is fine right now because `launch_runner` already sets `post=False`, so it just works.

allisonkarlitskaya:
It might make sense to put `ci-run` into `lib/aws` and run it like `python3 -m lib.aws.ci_run` or so.  It kinda fits in with the rest of those scripts using boto3 and requiring an AWS credentials setup (with the most direct comparison being `launch_runner`) but it also does some higher-level GitHub-style stuff, so it's a bit of a weird case....

lib/aws/dispatcher.py
```diff
@@ -320,6 +321,22 @@ async def scan_instances(self, ec2: "EC2Client") -> None:
...
             )
 
+        # In the name of belt-and-suspenders when it comes to burning cash, we
+        # want to notice and terminate-on-overdue "embargoed" instances (done
+        # above) but we don't otherwise want to make any note of them for
+        # bookkeeping (done below): this means that they don't count towards
+        # our total instance count, don't get their logs tracked, and don't
+        # show up on the dashboard.
+        self.instances = {
+            iid: inst for iid, inst in instances.items() if inst.slug != EMBARGOED_SLUG
```
Venefilyn: I'd mark `iid` as unused `_iid`/`_`

lib/aws/launch_runner.py
```diff
@@ -136,6 +139,10 @@ def main() -> None:
 
     job = typechecked(json.loads(args.job_json), dict)
 
+    if args.embargo:
+        job["slug"] = f"{EMBARGOED_SLUG}-{secrets.token_urlsafe(16)}"
+        print(f"Log URL: {LOGS_URL}{job['slug']}/log.html")
```
Venefilyn: If we set the slug here with embargoed slug already set, we don't need to do the same in ec2.py no?

Also isn't `secrets.token_urlsafe` related to the auth token?

Venefilyn (commented):
Didn't look it through fully yet. Too much brain needed

allisonkarlitskaya:
Direction from @bookwar (from the stand up): we should focus on landing "option 0" instead of looking into :heavy_check_mark: :x: sort of functionality.  We can add the other stuff later if we want.

lib/aws/dispatcher.py
allisonkarlitskaya: `iid` is the key, which is deconstructed and used back in the constructed dictionary, unmodified.  I actually looked for `.filter_by_value()` or so but it doesn't exist in Python, unfortunately.

Or did I misunderstand?

lib/aws/ec2.py
```diff
@@ -295,8 +297,8 @@ def launch_instance(
         {"Key": k, "Value": v}
         for k, v in {
             **TAGS,
-            "Name": f"{RUNNER_NAME_PREFIX}{slug}",
-            RUNNER_INSTANCE_SLUG_TAG: slug,
+            "Name": f"{RUNNER_NAME_PREFIX}{EMBARGOED_SLUG if embargoed else slug}",
+            RUNNER_INSTANCE_SLUG_TAG: EMBARGOED_SLUG if embargoed else slug,
```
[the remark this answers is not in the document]
allisonkarlitskaya: It's the slug (ie: the `image-refresh-fedora-coreos-0dc08c89-20260902-072103` part of log urls) passed in from outside, usually from the job JSON.  `tests-scan` chooses it, but it's theoretically kinda possible to imagine a situation in which it might be guessable so the idea here is to replace it with something unguessable.

lib/aws/launch_runner.py
allisonkarlitskaya: So there's two things going on here: the slug here is what was always used for determining the subdirectory that we write to on the logs server.  That's the main purpose of the slug.

In `ec2.py` we also use the slug (as a sort of side effect) to name the jobs.  That has two purposes: we can see from the Name tag of running ec2 instances what the log URL is (which is a purely human-friendly feature, but it is also kinda useful in the output of `infractl costs`) and we can (via the `cockpit-ci-slug` tag) find out the logs URL for a running instance: that's how the dashboard can link running instances to their logs.

Both of those things are *bad* for embargoed CI runs.  We don't want people poking around in the dashboard, the AWS account web UI, or the costs output (which is a fairly widely shared permission on our particular account) seeing actual log urls.  So we just write the word "embargoed" there instead.  That's also a signal to the dispatcher not to show it in the dashboard.

So two layers:

 1. Use randomized/unguessable slugs to start with because slugs might be guessable (that's the part here with using `secrets.token_urlsafe()`).
 2. Don't establish a link between the running instance and the slug (because it could leak the log URL, and we're trying hard to avoid that).

And `secrets.token_urlsafe()` is just a convenient way in the Python stdlib to get "secure randomness" in a particular form (ie: urlsafe base64) and a number of bits.  It has nothing to do with authentication: it's just pure randomness.

